PR #1, opened 2017-08-16
A basic Flask application

nottrobin:
Just as a starting point, create a basic hello world Flask application.

QA
--

``` bash
FLASK_APP=hello.py flask run
```

Browse to http://127.0.0.1:5000/ and check you see "Hello, World!".


nottrobin:
I've decided this was a bit too trivial a step for its own PR. I'll submit one with slightly more work in a bit.

PR #10, opened 2017-08-25
Add Sass and vanilla-framework

nottrobin:
- vanilla-framework
- basic markup structure to layout template
- package.json to include vanilla-framework and node-sass
- document how to run the site in the readme

Fixes #8 

QA
--

Follow the instructions in the README and check you can get the site running and that it looks styled like a vanilla site.

PR #11, opened 2017-08-29
Snap details page

### LINKED ISSUE #3: Create snap page template and URL endpoint
Create a snap page template with dummy data, and a route & view so that requests for snap details URLs will display the dummy page.

### LINKED ISSUE #4: Fetch data from API for snap details page
Wire up the back-end to the API to retrieve live data for each snap.

nottrobin:
Fixes #3
Fixes #4

- A snap details template
- A default view to process all URLs matching `/<something>`
- Only look for public snaps in the stable channel
- Cause a 404 with a custom error message if snap not found
- Use vanilla markup syntax ready for merging with #10

## Implementation

Could the relevant people please comment on these implementation details to let me know if they seem sensible.

### Data passed to the template

I am only passing through certain data to the template, according to what @matthewpaulthomas listed in https://github.com/CanonicalLtd/snappy-design-squad/issues/69#issuecomment-320259725 - I am providing all the data with ticks next to them, except the geo location data which isn't yet available.

The data that's passed through can easily be updated in `app.py`, but each datum should always be explicitly passed through, rather than transparently returning everything from the API.

@matthewpaulthomas please comment on this.

### Caching and timeout

The application will maintain a cache of its API requests for 60 seconds. Any subsequent requests from the same application instance within 60 seconds of the first request will therefore be served from the cache, rather than querying the API directly. This should mean that within each minute, the application should only be call a maximum of the number of times corresponding with how many app units are running. 4 units = 4 times a minute.

Each API request will timeout after 2 seconds. This is to prevent the application delaying significantly to respond to the user.

On any error (timeouts, connection errors or HTTP statuses above 499), the application will return a response from the cache if available, along with a notification that there was an API error.

@cprov would you or someone else from the Store side mind commenting on my implementation here? Let me know if it seems sensible to you?

## QA instructions

``` bash
python3 -m venv env3
source env3/bin/activate
pip install -r requirements.txt
FLASK_APP=app.py FLASK_DEBUG=true flask run
```

Now browse to:

- http://127.0.0.1:5000/canonical-livepatch/
- http://127.0.0.1:5000/documentation-builder/
- http://127.0.0.1:5000/non-existent-snap/

And check each page shows what is expected.

### DIFF
--- a/app.py
+++ b/app.py
@@ -6,7 +6,10 @@ import humanize
 
 
 app = Flask(__name__)
-cached_request = CachedSession(expire_after=60)
+cached_request = CachedSession(
+    expire_after=60,
+    old_data_on_error=True
+)
 snap_details_url = (
     "https://api.snapcraft.io/api/v1/snaps/details/{snap_name}"
     "?channel=stable"
@@ -17,8 +20,9 @@ snap_details_url = (
 def page_not_found(error):
     return render_template('404.html', description=error.description), 404
 
+
 @app.route('/<snap_name>/')
-def page_not_found(snap_name):
+def snap_details(snap_name):
     query_headers = {
         'X-Ubuntu-Series': '16',
     }
@@ -68,4 +72,3 @@ def page_not_found(snap_name):
         'snap-details.html',
         **context
     )
-
--- a/app.py
+++ b/app.py
@@ -1,3 +1,9 @@
+"""
+A Flask application for snapcraft.io.
+
+The web frontend for the snap store.
+"""
+
 import flask
 import requests
 import requests_cache
@@ -19,6 +25,11 @@ snap_details_url = (
 
 @app.errorhandler(404)
 def page_not_found(error):
+    """
+    For 404 pages, display the 404.html template,
+    passing through the error description.
+    """
+
     return flask.render_template(
         '404.html', description=error.description
     ), 404
@@ -26,6 +37,14 @@ def page_not_found(error):
 
 @app.route('/<snap_name>/')
 def snap_details(snap_name):
+    """
+    A view to display the snap details page for specific snaps.
+
+    This queries the snapcraft API (api.snapcraft.io) and passes
+    some of the data through to the snap-details.html template,
+    with appropriate sanitation.
+    """
+
     query_headers = {
         'X-Ubuntu-Series': '16',
     }
--- a/app.py
+++ b/app.py
@@ -110,6 +110,7 @@ def snap_details(snap_name):
 
         # Context info
         'api_error': response.old_data_from_error,
+        'is_linux': 'Linux' in flask.request.headers['User-Agent']
     }
 
     return flask.render_template(
--- a/app.py
+++ b/app.py
@@ -28,8 +28,14 @@ uncached_session.mount(
     'https://api.snapcraft.io',
     HTTPAdapter(max_retries=retries)
 )
-cached_session = requests_cache.CachedSession(expire_after=60)
+
+# The cache expires after 5 seconds
+cached_session = requests_cache.CachedSession(expire_after=5)
+
+# Requests should timeout after 2 seconds in total
 request_timeout = 2
+
+# Request only stable snaps
 snap_details_url = (
     "https://api.snapcraft.io/api/v1/snaps/details/{snap_name}"
     "?channel=stable"
@@ -60,6 +66,7 @@ def snap_details(snap_name):
 
     query_headers = {
         'X-Ubuntu-Series': '16',
+        'X-Ubuntu-Architecture': 'amd64',
     }
 
     response = _get_from_cache(
@@ -80,27 +87,29 @@ def snap_details(snap_name):
     snap_data = json.loads(response.text)
 
     context = {
-        # Always available
+        # Data direct from API
         'name': snap_data['title'],
-        'api_error': response.old_data_from_error,
+        'icon_url': snap_data['icon_url'],
         'version': snap_data['version'],
         'revision': snap_data['revision'],
-        'filesize': humanize.naturalsize(snap_data['binary_filesize']),
         'download_url': snap_data['download_url'],
         'publisher': snap_data['publisher'],
         'screenshot_urls': snap_data['screenshot_urls'],
         'prices': snap_data['prices'],
         'support_url': snap_data.get('support_url'),
+        'summary': snap_data['summary'],
+        'description': snap_data['description'],
+
+        # Transformed API data
+        'filesize': humanize.naturalsize(snap_data['binary_filesize']),
         'last_updated': (
             humanize.naturaldate(
                 parser.parse(snap_data.get('last_updated'))
             )
         ),
 
-        # May be available
-        'summary': snap_data.get('summary'),
-        'description': snap_data.get('description'),
-        'icon_url': snap_data.get('icon_url'),
+        # Context info
+        'api_error': response.old_data_from_error,
     }
 
     return flask.render_template(
--- a/app.py
+++ b/app.py
@@ -1,7 +1,71 @@
-from flask import Flask, render_template
+from flask import abort, Flask, render_template
+from requests_cache import CachedSession
+from dateutil import parser
+import json
+import humanize
+
+
 app = Flask(__name__)
+cached_request = CachedSession(expire_after=60)
+snap_details_url = (
+    "https://api.snapcraft.io/api/v1/snaps/details/{snap_name}"
+    "?channel=stable"
+)
 
 
 @app.errorhandler(404)
 def page_not_found(error):
     return render_template('404.html', description=error.description), 404
+
+@app.route('/<snap_name>/')
+def page_not_found(snap_name):
+    query_headers = {
+        'X-Ubuntu-Series': '16',
+    }
+
+    response = cached_request.get(
+        snap_details_url.format(snap_name=snap_name),
+        headers=query_headers,
+        timeout=2
+    )
+
+    if response.status_code >= 400:
+        message = (
+            'Failed to get snap details for {snap_name}'.format(**locals())
+        )
+
+        if response.status_code == 404:
+            message = 'Snap not found: {snap_name}'.format(**locals())
+
+        abort(response.status_code, message)
+
+    snap_data = json.loads(response.text)
+
+    context = {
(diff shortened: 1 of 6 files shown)


matthewpaulthomas:
> - A default view to process all URLs matching /<something>
>
>…
> Could the relevant people please comment on these implementation details to let me know if they seem sensible.

I do not think it is sensible for package pages to use the top-level snapcraft.io/______ URL namespace, because it assumes that all pages we will want to add to snapcraft.io in future will either live on a subdomain, have an URL path that hasn’t been reserved as a snap name, or have an URL path where we can force the snap to be renamed. A big counterexample to my opinion is Facebook, where `/BillGates`, `/Canonical-125818784107695`, and `/zuck` sit happily alongside `/careers`, `/marketplace`, and `/recipes`. But I expect Facebook can force renames more easily than a package management system could.

> I am only passing through certain data to the template, according to what @matthewpaulthomas listed in [CanonicalLtd/snappy-design-squad#69](https://github.com/CanonicalLtd/snappy-design-squad/issues/69#issuecomment-320259725)
>…
> @matthewpaulthomas please comment on this.

At that time I identified “channel map” as something that would not be needed. But probably we will need it, to let people choose to install a non-default track or branch.

(I could not test the app because `python3 -m venv env3` returned an error saying that I need to install `python3-venv`, which isn’t available.)

nottrobin:
> I do not think it is sensible for package pages to use the top-level `snapcraft.io/______` URL namespace.

This boat has sailed - the decision was made Warsaw. I do not think it's too problematic because it is perfectly fine for us to reserve as many names as we like in the snap store that then can't be registered as snaps where we need to. And we have done this for most things we can think we might need - category/ies, images, assets, css etc.

And on the whole, I do agree with Mark that if snapcraft.io is the home of snaps, then `snapcraft.io/<snap-name>` is simplest possible URL at which to find details of specific snaps - and in this case that ease outweighs the inconvenience for implementation.

> At that time I identified “channel map” as something that would not be needed. But probably we will need it, to let people choose to install a non-default track or branch.

Here is a sample of the data returned from an API call:

``` json
{
  "aliases": null,
  "anon_download_url": "https://api.snapcraft.io/api/v1/snaps/download/b96UJ4vttpNhpbaCWctVzfduQcPwQ5wn_22.snap",
  "architecture": [
    "amd64"
  ],
  "binary_filesize": 4558848,
  "channel": "stable",
  "confinement": "strict",
  "content": "application",
  "date_published": "2016-08-31T19:49:17.576703Z",
  "deltas": [],
  "description": "Canonical Livepatch Client",
  "developer_id": "canonical",
  "developer_name": "Shared snappy store account",
  "download_sha3_384": "cc4b79fdd8d0b46466eec7b661d70423723748f7eb2b27e96b75990ee0a74304a024ff820340475bbc4660c0e224bd0b",
  "download_sha512": "b6bf3e70e0d8776345cadc444c63ad579fd96aed27dfa0805edbd18b8c8c731e04770d1cb9a1c7a1fd94c52d61a2fcdf90ee16b8fea391b7095fc143351ed9a5",
  "download_url": "https://api.snapcraft.io/api/v1/snaps/download/b96UJ4vttpNhpbaCWctVzfduQcPwQ5wn_22.snap",
  "epoch": "0",
  "icon_url": "https://dashboard.snapcraft.io/site_media/appmedia/2016/08/pictogram-upgrade-orange-hex.svg.png",
  "last_updated": "2017-03-24T17:06:30.316236+00:00",
  "name": "canonical-livepatch.canonical",
  "origin": "canonical",
  "package_name": "canonical-livepatch",
  "prices": {},
  "private": false,
  "publisher": "Canonical",
  "ratings_average": 0,
  "release": [
    "16"
  ],
  "revision": 22,
  "screenshot_urls": [],
  "snap_id": "b96UJ4vttpNhpbaCWctVzfduQcPwQ5wn",
  "summary": "Canonical Livepatch Client",
  "support_url": "mailto:snappy-canonical-storeaccount@canonical.com",
  "title": "canonical-livepatch",
  "version": "7"
}
```

I think all it tells you is the channel of the current snap. I'm not even sure right now if there's a way to find out which channels are available for a given snap.

@matthewpaulthomas can you clarify what you're expecting for the channel map?

matthewpaulthomas:
> This boat has sailed - the decision was made Warsaw.

I trust that you have this particular issue handled. But in general, “the decision was made Warsaw” doesn’t mean it will work out. URLs should change less often than most other parts of a UI, but it’s still highly risky to design them independently from the rest of the UI, especially for a UGC-based app.

> @matthewpaulthomas can you clarify what you're expecting for the channel map?

Tracks and risk levels are useful only if people can install them. People can install them using a GUI only if the GUI displays them. So I am assuming that the end-user snap page will need to list them, and [I showed this as a menu in my wireframes](https://github.com/CanonicalLtd/snappy-design-squad/issues/69#issuecomment-321767579).

Now, I’m not clear on exactly how much of the installation flow will be handled by the Web site, vs. by your local client (such as Gnome Software or KDE Discover). It may be that the end-user page doesn’t have an “Install” button at all, just a “View in Store” or similar button that opens your local client. Either way, though, your local client will _also_ need a way to learn the tracks and channels, and that probably should involve the same API.

app.py
```diff
@@ -1,7 +1,141 @@
...
+from dateutil import parser
+
+
+app = flask.Flask(__name__)
+uncached_session = requests.Session()
+cached_session = requests_cache.CachedSession(expire_after=60)
+request_timeout = 2
+snap_details_url = (
+    "https://api.snapcraft.io/api/v1/snaps/details/{snap_name}"
+    "?channel=stable"
```
cprov: FWIW `?channel=stable` is the default scope and because the way `channel_maps_list` field is exposed (all channels), it's unlikely that the store-front would ever need to switch scope.

app.py
```diff
@@ -1,7 +1,141 @@
...
+    """
+    A view to display the snap details page for specific snaps.
+
+    This queries the snapcraft API (api.snapcraft.io) and passes
+    some of the data through to the snap-details.html template,
+    with appropriate sanitation.
+    """
+
+    query_headers = {
+        'X-Ubuntu-Series': '16',
```
cprov: Snap revisions are architecture-specific (the vast majority), so it the API. The `X-Ubuntu-Architecture: <arch>` defaults to 'amd64' and is required for arch-specific fields like `channel_maps_list`.
It will certainly have UI implications.

app.py
```diff
@@ -1,7 +1,141 @@
...
+    }
+
+    response = _get_from_cache(
```
cprov: We discussed this briefly in our call today. I would be more concerned about retrying failed requests than caching results in the context. Sure, the load will be not neglectable on the store services but we are in a better position to cache it properly, specially when authentication comes to play. Also, considering the possible topology of these services, both deployed in PS4.5, request round-trips won't be expensive.
OTOH, retrying 5xx request has proven to be relevant to avoid network glitches to prevent external request to be fulfilled.

app.py
```diff
@@ -1,7 +1,141 @@
...
+
+    if response.status_code >= 400:
+        message = (
+            'Failed to get snap details for {snap_name}'.format(**locals())
+        )
+
+        if response.status_code == 404:
+            message = 'Snap not found: {snap_name}'.format(**locals())
+
+        flask.abort(response.status_code, message)
```
cprov: ditto, 5xx should be retried a given number of times before giving up and 4xx logged/collected as an incident (sentry or similar).

app.py
```diff
@@ -1,7 +1,141 @@
...
+        'last_updated': (
+            humanize.naturaldate(
+                parser.parse(snap_data.get('last_updated'))
+            )
+        ),
+
+        # May be available
+        'summary': snap_data.get('summary'),
+        'description': snap_data.get('description'),
+        'icon_url': snap_data.get('icon_url'),
```
cprov: These fields are always present (i.e available in the response payload), they just happen to be eventually empty. No need for `.get()`s.

app.py
```diff
@@ -1,7 +1,141 @@
...
+    }
+
+    return flask.render_template(
+        'snap-details.html',
+        **context
+    )
+
+
+def _get_from_cache(url, headers):
```
cprov: As mentioned above, I think local caching (even with short TTL like this) is an unnecessary complexity at this point. The store backend should handle the additional load impact and if necessary cache the context appropriately.

app.py
nottrobin: I was aware it's the default, but I'd still rather be explicit in this URL. It informs the reader of this code immediately (without having to have knowledge of the API) that we're only requesting stable channels, and illustrated how to filter queries in general.

app.py
nottrobin: Ah, so by default we're only going to see snaps that exist for `amd64`? In that case I'd rather provide the `X-Ubuntu-Architecture` header, again to make it clear. I'll do that. Thanks.

app.py
nottrobin: I like the idea of retrying requests - that certainly makes sense. But I'd rather retry for a set amount of time (2 seconds, probably) rather than a set number of times, as I want to avoid this application hanging on a delayed API response.

I still have concerns about not providing some redundancy in this layer. For these public API requests, it seems to me to be perfectly valid to cache the response, and I think the public facing details pages should be especially resilient.

If it doesn't make sense to cache the authenticated API responses, then that's fine - I think an authenticated user would be less surprised to find that an API call was required in their requests - and it would be slightly less impactful if those pages broke.

I'd be happy to set up a different cache which honours caching headers (probably https://pypi.python.org/pypi/CacheControl) so that your application can choose how the response is cached if that would help. But for public pages at least, I'd still like to return cached data rather than a 500 page if we have an API error. We will still display a warning notification to explain that the API response failed.

nottrobin: Discussion about caching moved to https://github.com/canonical-websites/snapcraft-flask/issues/20

nottrobin:
@cprov I've made the following tweaks:

- Try each API request 5 times, with a total timeout of 2 seconds
- Explicitly request `X-Ubuntu-Architecture`
- Explicitly expect all API response items to exist (rather than using `.get(`)
- Moved discussion of caching to https://github.com/canonical-websites/snapcraft-flask/issues/20

Thanks for your help. I'm going to merge this for now. We can make further changes in future PRs.

PR #23, opened 2017-09-04
Snapcraft header

### LINKED ISSUE #15: Implement header with snapcraft logo for snap details page
Designs: https://github.com/CanonicalLtd/snapcraft-design/issues/54#issuecomment-326014685

bartaz:
Vanilla styling for header

Fixes #15

## QA
- Update local branch
- Run `FLASK_APP=app.py FLASK_DEBUG=true flask run`
- Go to some snap page: http://localhost:5000/canonical-livepatch/
- Header should be vanilla navigation light pattern with snapcraft.io logo
 
screen shot 2017-09-05 at 09 14 38


static/sass/header.scss
```diff
@@ -0,0 +1,28 @@
+// based on p-navigation--light pattern from Vanilla
+@mixin navigation-x-light {
+  .p-navigation--x-light {
```
bartaz: Vanilla doesn't have pattern for white background heading, so adding it here.
Based on built in patterns not to write too much custom styles.

static/sass/header.scss
```diff
@@ -0,0 +1,28 @@
...
+
+    // remove double bottom border on open nav
+    &:target .p-navigation__nav .p-navigation__link:last-child {
+      border-bottom: 0;
+    }
+  }
+}
+
+@mixin snapcraft-logo {
+  .p-navigation__snapcraft-logo {
```
bartaz: Some custom styling to give snapcraft.io same size and spacing as on http://snapcraft.io

anthonydillon:
These changes are good. My only suggestion would be to keep the same class names as vanilla and just modify them locally like you have. These means of any patterns or pushed up to vanilla you can just remove the custom styling.

templates/_layout.html
```diff
@@ -14,8 +14,8 @@
       <div class="row">
         <div class="p-navigation__banner">
           <div class="p-navigation__logo">
-            <a href="/">
-              <img src="https://assets.ubuntu.com/v1/d45097a4-snapcraft.io-logotype.svg" style="max-width: 150px; display: block" alt="Snapcraft.io">
+            <a href="https://snapcraft.io">
+              <img class="u-float-left" src="https://assets.ubuntu.com/v1/d45097a4-snapcraft.io-logotype.svg" width="148" alt="Snapcraft.io" />
```
bartaz: Using `u-float-left` seemed the only way in Vanilla to force `display: block` on img without adding custom styles.
But feels a bit hacky to me...

bartaz:
@nottrobin I removed all the custom styling and used only Vanilla patterns.

templates/_layout.html
nottrobin: Yep, I agree. This is the right solution for now. 

I filed [an issue](https://github.com/vanilla-framework/vanilla-framework/issues/1313) about this in Vanilla - please add to it if you feel I missed anything.

PR #28, opened 2017-09-06
Sanitise and linkify markup

nottrobin:
**Proposal**. I don't know if we want this feature, but here's an example of how we might handle markup and URLs in snap descriptions.

This will:

- Remove any dangerous tags (e.g. `<script>`) from descriptions
- Allow safe markup tags (e.g. `<strong>`) in descriptions, and render them properly
- Add `<a rel="nofollow">` tags to URLs found in text
- Add `rel="nofollow"` to existing link tags

QA
--

`./run`, then visit <http://0.0.0.0:8022/dwarf-fortress/> and see the links are clickable! I'm not sure if there's an example of a snap with markup in its description, but if one exists we could try it out as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,14 +1,17 @@
 {
   "scripts": {
     "test": "sass-lint static/**/*.scss --verbose --no-exit",
-    "build": "node-sass --include-path node_modules static/sass --output static/css",
-    "watch": "node-sass --include-path node_modules --source-map true --watch static/sass --output static/css",
+    "build": "node-sass --include-path node_modules static/sass --output static/css && postcss --use autoprefixer --replace static/css/*.css",
+    "watch": "watch -p 'static/sass/*.scss' -c 'yarn run build'",
     "clean": "rm -rf node_modules yarn-error.log static/css *.log *.sqlite _site/ build/ .jekyll-metadata"
   },
   "author": "Canonical webteam",
   "devDependencies": {
+    "autoprefixer": "^7.1.3",
+    "postcss-cli": "^4.1.0",
     "node-sass": "^4.5.3",
     "sass-lint": "^1.10.2",
-    "vanilla-framework": "1.5.1"
+    "vanilla-framework": "1.5.1",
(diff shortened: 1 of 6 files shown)


bartaz:
Making links clickable would be a nice touch, but of course we don't control where they link.
For sure they should be marked as external links to make users aware they are going out of snapcraft.

As for the other styling - I don't think anyone will put HTML into their snapcraft.yaml descriptions.
Markdown would be a better choice for that, but again - probably just a limited scope of it (to avoid people adding headings, images, etc) and that would need to be clearly communicated and documented in snapcraft.yaml docs.

nottrobin:
@bartaz okay I've updated the logic so it now does the simpler treatment of just sanitising markup tags so the user will simply see the markup, while also adding links around any URLs.

Please have another look?

nottrobin:
@bartaz I've now added the `p-link--external` classes to external links in descriptions.

To support this I've also added autoprefixing to the CSS builder. Please check `./run build`, `./run watch` and `./run serve --watch` work as expected.

package.json
```diff
@@ -1,14 +1,17 @@
...
+    "postcss-cli": "^4.1.0",
     "node-sass": "^4.5.3",
     "sass-lint": "^1.10.2",
-    "vanilla-framework": "1.5.1"
+    "vanilla-framework": "1.5.1",
+    "watch-cli": "^0.2.2"
   }
-}
\ No newline at end of file
+}
```
bartaz: I think we keep having diffs like this.
I suppose my editor is adding new lines at the end of file by default, while yours is removing them ;)

nottrobin: I don't think my editor does anything with them as I'm usually using vim, but `package.json` often gets generated/updated by yeoman - with `yo canonical-webteam:run`. I'm guessing that removes newlines.

bartaz: OK. Maybe my editor (Atom) adds them to edited files. As I've seen added newlines in some files that I touched.

PR #29, opened 2017-09-06
Rename title and add package_name

nottrobin:
I noticed that we were using the `title` for both the "heading" and in the snap URL (`snap://{{ name }}`). This was my mistake - because `title` and `package_name` happened to be the same for `canonical-livepatch` I thought they were always the same.

I'm now passing through `snap_title` and `package_name` separately, and using them where appropriate.

QA
--

`./run` and visit <http://0.0.0.0:8022/dwarf-fortress/>. Check the "install" link works.

### DIFF
--- a/templates/snap-details.html
+++ b/templates/snap-details.html
@@ -1,24 +1,24 @@
 {% extends "_layout.html" %}
 
-{% block title %}{{ name }} snap details{% endblock %}
+{% block title %}{{ snap_title }} snap details{% endblock %}
 
 {% block content %}
   <div class="p-strip is-shallow is-bordered">
     <div class="row">
       <div class="col-2">
         {% if icon_url %}
-          <img src="{{ icon_url }}" alt="{{ name }} snap" />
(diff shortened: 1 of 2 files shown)


templates/snap-details.html
```diff
@@ -1,23 +1,23 @@
 {% extends "_layout.html" %}
 
-{% block title %}{{ name }} snap details{% endblock %}
+{% block title %}{{ snap_title }} snap details{% endblock %}
```
bartaz: Not related much to this PR, but maybe it would be good to define how the title should look like (as this is going to be a page linked, indexed by google etc).

As part of snapcraft.io maybe it should have similar title format, something like: `{{ snap_title }} - Snap details - Snaps are universal Linux packages`.
Would be good to confirm this at some point - but not really blocking this PR on that.

nottrobin: Yeah I think you should open an issue about that. Or if you think you know what wording would be good just submit a PR with the suggestion.

PR #34, opened 2017-09-07
Added social icons to the footer

### LINKED ISSUE #33: Add social icons to footer
Add social icons (like on snapcraft.io) to footer.

Create it as a pattern that could be proposed to Vanilla.

bartaz:
Fixes #33 

Added social icons to footer.
Prepared styles for it as a vanilla pattern that could be proposed.

### QA

- ./run
- Go to any snap details page (http://localhost:8022/canonical-livepatch/)
- Footer should contain social icons

screen shot 2017-09-07 at 14 22 59

### DIFF
--- a/static/sass/_patterns_social.scss
+++ b/static/sass/_patterns_social.scss
@@ -1,18 +1,10 @@
 @mixin p-social {
-  .p-social__heading {
-    color: $color-mid-dark;
-    font-size: .875rem;
-    text-transform: uppercase;
+  .p-social-list {
+    @extend .p-inline-list;
+    margin-top: 1rem;
   }
 
-  .p-social__heading + * {
-    margin-top: .5rem;
-  }
-
-  .p-social__icon--twitter,
-  .p-social__icon--google,
-  .p-social__icon--facebook,
-  .p-social__icon--youtube {
+  %p-social-list-icon {
     background-image: url('#{$assets-path}6bfc3d6a-build.snapcraft.social.optimize.svg');
     background-repeat: no-repeat;
     background-size: 160px 80px; // 2x4 icons 40x40px each (in 2 rows for regular and hover state)
@@ -21,24 +13,29 @@
     overflow: hidden;
     text-indent: -99999px;
     width: 40px;
+
+    &:hover {
+      background-position-y: -40px;
+    }
   }
 
-  .p-social__icon--google {
+  .p-social-list__icon--twitter {
+    @extend %p-social-list-icon;
+    background-position: 0 0;
+  }
+
+  .p-social-list__icon--google {
+    @extend %p-social-list-icon;
     background-position: -40px 0;
   }
 
-  .p-social__icon--facebook {
+  .p-social-list__icon--facebook {
+    @extend %p-social-list-icon;
     background-position: -80px 0;
   }
 
-  .p-social__icon--youtube {
+  .p-social-list__icon--youtube {
+    @extend %p-social-list-icon;
     background-position: -120px 0;
-  }
-
-  .p-social__icon--twitter:hover,
-  .p-social__icon--google:hover,
-  .p-social__icon--facebook:hover,
-  .p-social__icon--youtube:hover {
-    background-position-y: -40px;
   }
 }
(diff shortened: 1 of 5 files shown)


static/sass/_patterns_social.scss
```diff
@@ -0,0 +1,44 @@
+@mixin p-social {
+  .p-social__heading {
+    color: $color-mid-dark;
+    font-size: .875rem;
+    text-transform: uppercase;
+  }
+
+  .p-social__heading + * {
+    margin-top: .5rem;
```
bartaz: Default Vanilla typography adds a lot of margin (2rem) after the heading. So we need to override it so it doesn't look to far away from the icons...

static/sass/_patterns_social.scss
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  .p-social__icon--twitter,
+  .p-social__icon--google,
+  .p-social__icon--facebook,
+  .p-social__icon--youtube {
```
bartaz: Layout of social icons can be done with `p-list--inline` pattern, so styles are mostly icons.

In theory they could be added to vanilla icons, but social icons differ a bit because they are links with hover state, so they are not just static icons like the others.

nottrobin: Could you call this block `.p-social__icon` and then, inside each of the modified versions, add `@extends .p-social__icon`?

static/sass/_patterns_social.scss
```diff
@@ -0,0 +1,44 @@
...
+
+  .p-social__icon--youtube {
+    background-position: -120px 0;
+  }
+
+  .p-social__icon--twitter:hover,
+  .p-social__icon--google:hover,
+  .p-social__icon--facebook:hover,
+  .p-social__icon--youtube:hover {
+    background-position-y: -40px;
```
nottrobin: You should add this as `&:hover {}` inside the above `.p-social__icon` block.

static/sass/_patterns_social.scss
```diff
@@ -0,0 +1,44 @@
+@mixin p-social {
+  .p-social__heading {
```
nottrobin: Please use the undocumented `p-muted-heading` for this

static/sass/_patterns_social.scss
nottrobin: We should avoid using `*`. I think you should address this by something like:

``` sass
.p-social-list {
  @extends .p-inline-list;
  margin-top: 1rem;  // I think 0.5rem is a bit too close
}
```

static/sass/_patterns_social.scss
nottrobin: Maybe you should also move the `background-image` bit into the modifiers, as it doesn't make much sense without the `background-position`

static/sass/_patterns_social.scss
bartaz: I found it (by searching for "uppercase" in codebase), but it still needed different bottom spacing and color.

But if we introduce `.p-social-list` as you suggested below it, we could define spacing for `.p-social-list` and ignore color I guess...

nottrobin: yeah I'd rather use the existing standard headings as used on www.canonical.com, then if we're not happy with the colour we should lobby to get p-muted-heading modified.

static/sass/_patterns_social.scss
```diff
@@ -0,0 +1,41 @@
+@mixin p-social {
+  .p-social-list {
+    @extend .p-inline-list;
```
bartaz: sass-lint warns here about extending .class instead of %pattern.

static/sass/_patterns_social.scss
```diff
@@ -0,0 +1,41 @@
...
+    margin-top: 1rem;
```
bartaz: If the only reason to introduce `p-social-list` is changing the top margin maybe it would be nice to have means to control margin from vanilla itself?

nottrobin: I do agree it feels kinda superfluous. Maybe it would be neater to implement our own utility class....

PR #51, opened 2017-09-15
Table key value pattern proposal

bartaz:
Added new key-value table pattern.

## QA
- ./run
- Go to any snap page
- Snap details table should have new style

screen shot 2017-09-19 at 11 22 45

## NOTE

As discussed, implemented a design with borders between the rows (but no border at the bottom of the table).


nottrobin:
Fixes #51

PR #53, opened 2017-09-18
Expose geodata to the template

### LINKED ISSUE #52: Expose geodata from API to view

nottrobin:
This makes use of the public geodata API that @cprov built.

At present I'm requesting a month of historical data about the percentage of users of a specific snap in each country.

The API returns the data broken down into days, so I am averaging all the days of data returned over the last month to return a single figure. This should hopefully help to avoid anomalous activity that might have happened e.g. in the last 24 hours.

Fixes #52 

QA
--

``` bash
./run
```

Go to a snap page, e.g. <http://0.0.0.0:8022/lxd/>, and view source. You should see the geodata precentages pretty clearly in the source code.

### DIFF
--- a/app.py
+++ b/app.py
@@ -8,7 +8,6 @@ import flask
 import requests
 import requests_cache
 import datetime
-import json
 import humanize
 import re
 import bleach
@@ -84,6 +83,16 @@ def snap_details(snap_name):
     )
     details = details_response.json()
 
+    if details_response.status_code >= 400:
+        message = (
+            'Failed to get snap details for {snap_name}'.format(**locals())
+        )
+
+        if details_response.status_code == 404:
+            message = 'Snap not found: {snap_name}'.format(**locals())
+
+        flask.abort(details_response.status_code, message)
+
     metrics_query_json = [
         {
             "metric_name": "installed_base_by_country_percent",
@@ -107,16 +116,6 @@ def snap_details(snap_name):
         percentages = [p for p in percentages_with_nulls if p is not None]
         average_percentage = sum(percentages) / len(percentages)
         user_percentage_by_country[country_code] = average_percentage
-
-    if details_response.status_code >= 400:
-        message = (
-            'Failed to get snap details for {snap_name}'.format(**locals())
-        )
-
-        if details_response.status_code == 404:
-            message = 'Snap not found: {snap_name}'.format(**locals())
-
-        flask.abort(details_response.status_code, message)
 
     description = details['description'].strip()
     paragraphs = re.compile(r'[\n\r]{2,}').split(description)
--- a/app.py
+++ b/app.py
@@ -13,7 +13,7 @@ import humanize
 import re
 import bleach
 import urllib
-from dateutil import parser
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -66,29 +75,50 @@ def snap_details(snap_name):
     some of the data through to the snap-details.html template,
     with appropriate sanitation.
     """
+    today = datetime.date.today()
+    month_ago = today - relativedelta.relativedelta(months=1)
```
cprov: What about using UTC timestamps ? ```today = datetime.datime.utcnow().date()```
Also, I am not sure how you intend to display 1 month of installed_base data. Perhaps the delta between day 1 & day 30 ? The simplest thing the occurs to me, would be to show installed_base 'yesterday'.

app.py
```diff
@@ -66,29 +75,50 @@ def snap_details(snap_name):
...
 
-    if response.status_code >= 400:
+    user_percentage_by_country = {}
+
+    for country_percentages in geodata:
+        country_code = country_percentages['name']
+        percentages_with_nulls = country_percentages['values']
+        percentages = [p for p in percentages_with_nulls if p is not None]
+        average_percentage = sum(percentages) / len(percentages)
+        user_percentage_by_country[country_code] = average_percentage
```
cprov: What is the rationale for averaging last 30d of installed_base ? I am afraid we are diluting meaning unnecessarily.

nottrobin: To start with we'd like to implement a simple solution so we can ship it. So rather than offering a map that can be filtered by date, we'll simply show an up-to-date snapshot - so we're only going to show a single data point per country.

In discussion with @matthewpaulthomas we decided that simply showing the *mos recent* day of data might be too subject to anomalous fluctuations (people's computers were switched on), so it's better to take an average over a larger period. We chose a month, but if you think a different period makes sense then let me know.

app.py
nottrobin: Good point, I'll use `utcnow`.

app.py
```diff
@@ -43,6 +42,15 @@
     "https://api.snapcraft.io/api/v1/snaps/details/{snap_name}"
     "?channel=stable"
 )
+details_query_headers = {
+    'X-Ubuntu-Series': '16',
+    'X-Ubuntu-Architecture': 'amd64',
```
sparkiegeek: I think you need to look at the installs across all the architectures. Even if we're not intending on exposing the different architectures in the page, we should at least count them in the graph (i.e. loop over all of them)

nottrobin: The details API is not the same as the metrics API. The metrics query isn't filtered by architecture. The details API just gets data about a specific revision, not metrics. Also, AFAIK (check with @cprov) the details API will always choose an architecture - if I don't specify it it will default to `amd64`.

app.py
nottrobin: I discussed the question of the this smoothing over a period of time with @pmahnke and he agreed that, for the time being, taking the average over the last month seems like it should probably produce fresh-enough data. So I'm going to merge it as it is currently implemented. However, we can continue to discuss and refine this implementation.

PR #63, opened 2017-09-25
Redirects for docs and community

nottrobin:
Redirect /community to the homepage, as it was in the old
[snapcraft.io-static-pages repo](https://github.com/canonical-websites/snapcraft.io-static-pages/).

Redirect /docs to docs.snapcraft.io. This is in preparation for when
docs.snapcraft.io becomes the home of snapcraft docs. It won't effect
anything right now, because [kubernetes takes care of](https://github.com/canonical-webteam/k8s-deployments/blob/master/staging.snapcraft.io/snapcraft.io-ingress.yaml#L27) routing /docs to
the docs application, before it hits this application.

QA
--

Run the server in one terminal window:

``` bash
$ ./run
```

In a different terminal window, check `/community/` redirects to the 
homepage, and `/docs/some/page` redirects to `docs.snapcraft.io`:

``` bash
$ curl -sI 127.0.0.1:8022/community/ | grep Location
Location: http://127.0.0.1:8022/
$ curl -sI 127.0.0.1:8022/docs/some/page/ | grep Location
Location: https://docs.snapcraft.io/some/page/
```

### DIFF
--- a/app.py
+++ b/app.py
@@ -53,6 +53,8 @@ metrics_query_headers = {
 }
 
 
+# Error handlers
+# ===
 @app.errorhandler(404)
 def page_not_found(error):
     """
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -65,6 +67,20 @@ def page_not_found(error):
     ), 404
 
 
+# Redirects
+# ===
+@app.route('/docs/<path:path>')
```
bartaz: Should it also catch `/docs/` without path?
There is a link to /docs in header navigation that currently is in 404.

But I guess it's currently handled on kubernates.

nottrobin: Very true. I'll fix this.

nottrobin: Fixed.

PR #71, opened 2017-09-27
Snap details map

bartaz:
Adds a map of snap usage to snap details page.

## QA
- ./run
- Go to snap page
- Map should appear

Or try demo:
http://snapcraft.io-pr-71.run.demo.haus/lxd/
http://snapcraft.io-pr-71.run.demo.haus/gnome-calculator/
http://snapcraft.io-pr-71.run.demo.haus/canonical-livepatch/
http://snapcraft.io-pr-71.run.demo.haus/core/

screen shot 2017-10-12 at 15 13 40


anthonydillon:
The map does not scroll on small screens.

bartaz:
@anthonydillon Yes, map needs to either scale or be scrollable on smaller screens.

bartaz:
@nottrobin @carlaberkers @spencerbygraves @matthewpaulthomas 

Map is ready to be reviewed on demo server. Couple of examples:
http://snapcraft.io-pr-71.run.demo.haus/lxd/
http://snapcraft.io-pr-71.run.demo.haus/gnome-calculator/
http://snapcraft.io-pr-71.run.demo.haus/canonical-livepatch/
http://snapcraft.io-pr-71.run.demo.haus/core/

The numbers from API tend to be rather small, so colours are quite pale in most of the cases.
So we may need to consider changing a bit the scale to make the map more interesting (?).

Also it probably would be useful to explain somehow (a legend?) what do these colours represent. It's not very clear what we are actually showing on this map. And we don't show any numbers so it's hard to interpret the values.

.eslintrc.js
```diff
@@ -0,0 +1,33 @@
+module.exports = {
```
nottrobin: :heart: - please bring this up in the next dev catch-up if I forget? We should sync up on JS linting standards.

package.json
```diff
@@ -1,18 +1,24 @@
 {
   "scripts": {
     "test": "sass-lint static/**/*.scss --verbose --no-exit",
-    "build": "node-sass --include-path node_modules static/sass --output static/css && postcss --use autoprefixer --replace static/css/styles.css && postcss static/css/styles.css --use cssnano > static/css/styles.min.css",
+    "build": "yarn run build-js && yarn run build-css",
+    "build-css": "node-sass --include-path node_modules static/sass --output static/css && postcss --use autoprefixer --replace static/css/styles.css && postcss static/css/styles.css --use cssnano > static/css/styles.min.css",
+    "build-js": "cp node_modules/d3/build/d3.min.js static/js/modules && cp node_modules/topojson-client/dist/topojson-client.min.js static/js/modules",
     "watch": "watch -p 'static/sass/*.scss' -c 'yarn run build'",
```
nottrobin: We should probably `watch` this to just `yarn run build-css`, as the JS doesn't need copying every time a Sass file changes.

package.json
```diff
@@ -1,18 +1,24 @@
...
     "node-sass": "^4.5.3",
     "postcss-cli": "^4.1.0",
     "sass-lint": "^1.10.2",
     "vanilla-framework": "1.5.2",
-    "watch-cli": "^0.2.2",
-    "cssnano": "^3.10.0"
+    "watch-cli": "^0.2.2"
+  },
+  "dependencies": {
+    "d3": "^4.10.2",
```
nottrobin: As these still won't get packaged up with the site directly (they get copied to the static folder and packaged from there) they're really just as much devDependencies as the rest of the stuff. But in other projects we simply use `dependencies`, as there's not really a material difference for our sites (they are never included as modules in other projects, which is when the distinction matters). Could you just put everything under `dependencies`?

static/js/snap-details-map.js
```diff
@@ -0,0 +1,104 @@
+/* global d3, topojson */
+
+!(function(window, document, d3, topojson) {
+  function renderMap(el, snapData) {
+    var width = 988; // 990 - 1px borders
```
nottrobin: Is this from design advice? It would seem more robust if JavaScript were to calculate this width based on the width of a specific element - what if Vanilla decides pages are now 1200px wide? or 800px?

I get that you want the map to always be the same width and then scroll, but it would still be nice if there was a way for that maxWidth to be relative to something. Can you think of any solution?

matthewpaulthomas:
Cool!

> Also it probably would be useful to explain somehow (a legend?) what do these colours represent. It's not very clear what we are actually showing on this map. And we don't show any numbers so it's hard to interpret the values.

What we are answering is questions of two types:
a. Is this snap significantly more popular in territory X than territory Y?
b. Is this snap significantly more popular in territory X than this other snap is?

Neither of those require numbers, which is fortunate both because publishers would be unlikely to want that level of detail, and because the number thresholds will change over time to keep the maps interesting. The latter reason is also why we can’t show a legend.

> The numbers from API tend to be rather small, so colours are quite pale in most of the cases.

The grey looks a bit too pale regardless. If I look at the screen at an angle much different from 90°, grey territories disappear altogether.

> So we may need to consider changing a bit the scale to make the map more interesting (?).

I’d be able to specify a more interesting scale if I had access to data for the current distribution of daily active device counts for all snap-territory combinations.

> I get that you want the map to always be the same width and then scroll

There’s a tension here. On one hand, ~95% of the time the map will be acting only as an illustration. That would suggest (a) it should not scroll by default, and (b) it should be a lot smaller than it is. (To put it another way, it shouldn’t be taking up more screen space than the description of most snaps; it’s less important.)

On the other hand, presenting territories distinctly is an incomplete feature if people can’t reliably see and click on (for example) Slovenia or Costa Rica.

I think the way to satisfy both those constraints is for the map container to be smaller, for the map to be scaled by default to exactly fit its container, and for it to be zoomable.

A few issues with the tooltips:
- They are the same width regardless of name, which looks wonky for most territories. They should be just wide enough for the color swatch and name.
- The color swatches are missing.
- Grey territories currently have no tooltip at all.

nottrobin:
@matthewpaulthomas you can zoom in on the whole page, and therefore on the map too. I just tried it out. Might that be sufficient? Should we go ahead and make it always scale to the width of the container then?

carlaberkers:
Would it be possible to make it full width? By having it in a box in the content area of the page it looks like important content.

If it is more illustrative, to give a feel of popularity, maybe it can be bigger and maybe it can be a background to the summary at the top. That way snap details look distinct when we load the page. Would be nice to see a mock of what that would look like.q

For this initial version, it might also help if the border was less dark in contrast to the map.

spencerbygraves:
Looking good, thanks.

I agree the grey is too light, we need to finalise those colours. The border is the correct colour though.

The design had zoom controls, can these be added?

The spacing for tooltips can be seen in the vanilla specs - https://github.com/ubuntudesign/vanilla-design/tree/master/Tooltips

@carlaberkers I tried a design with the map in the background of the header, but it seemed related to the design of the app. It may also be difficult to hover with the other elements over the top.

bartaz:
@nottrobin  The code related review comments were addressed, so feel free to give it a second look.

nottrobin:
@bartaz code & QA looks good.

We really just need clarification as to whether this is okay as a first pass to be improved, or whether we it has issues that need to be fixed before we release the first version.

carlaberkers:
Review with @bartaz @spencerbygraves @lyubomir-popov @matthewpaulthomas @nottrobin @carlaberkers

Fix before landing:
- change map to Google projection
- zooming the page is enough, no extra zoom
- Spencer & Lyubomir review and propose colours
- keep colour in the tooltip for now

Nice to have:
- add a tooltip to grey countries

Potential longer-term improvements:
- Full-width display
- Adding panning
- Adding zoom

bartaz:
@spencerbygraves @lyubomir-popov @matthewpaulthomas @nottrobin @carlaberkers

To give you an idea of scale of the map with new projection:

screen shot 2017-10-05 at 13 32 00

That's how it renders by default (without any clipping), so we may consider clipping a bit of the map from top/bottom (poor Antarctica again) to make the map less overwhelming.

spencerbygraves:
@bartaz well, I don't think that will work. Lyubomir will discuss other projections with you. Thanks

matthewpaulthomas:
Obligatory _The West Wing_: [“Are you saying the map is wrong?” “Oh dear yes.”](https://www.youtube.com/watch?v=eLqC3FNNOaI)

lyubomir-popov:
Hi guys I'll look through the different projections and see if I can find something more suitable ( https://bl.ocks.org/mbostock/29cddc0006f8b98eff12e60dd08f59a7 )

In the mean time, the suggestion bellow aims to address the following:
1) Set a fixed height and fit whatever projection we choose in that - so we get full control over how much of the screen height the map takes. 
2) Get a panoramic, full width look by stretch the background.
3) Integrate the title into the coloured background, to address the fact that a title outside the map area makes it too disjointed, as Carla mentioned.
 
screen shot 2017-10-06 at 09 35 36

lyubomir-popov:
Also wondering whether a fuzzier visualisation style might be more suitable in a case like this where showing exact numbers is not allowed. I realise this is way out of scope, so not really suggesting it as a possibility at this stage.
screen shot 2017-10-06 at 11 20 00
(http://www.digitalearthlab.com/wp-content/uploads/2012/08/Distribution.png)

bartaz:
@lyubomir-popov I found "Natural Earth" projection: https://github.com/d3/d3-geo/blob/master/README.md#geoNaturalEarth1
Said to be designed to be "appealing to the eye for small-scale maps of the whole world." which is kind of our use case I think.


It's not really a rectangle, but it is panoramic so we should be able to easily display it in a rectangle area.

lyubomir-popov:
I like it, I was just looking at this one, robinson and wagner6 and trying to spot any differences.

Only one minor thing I'm not clear about - what happens at the top left/right edges when you remove the parallels and meridians? Is it possible to avoid visible clipping by nudging the projection so that the Russian parts of Alaska move out of view at the left edge? If not, we can leave as is, or apply a subtle fade if needed.

bartaz:
@lyubomir-popov Yes, I found a way to rotate the projection a bit. This is more or less how it may look with current settings:

screen shot 2017-10-06 at 13 50 59

We could probably clip it a bit from the top (to remove empty space) and maybe bottom too.

lyubomir-popov:
I've updated the mockup to reflect Spenser's comments - title is now outside the blue area, height of the map is set to 500px to leave some space above/below on all screen sizes, including mobile.

screen shot 2017-10-09 at 09 52 00

lyubomir-popov:
As for colours, I've picked the closest one we have in the secondary colours palette - #48929B. The background in the mockup below is 5%, and the rest should be in increments matching the bands - I think it was 10% increments? @bartaz do you need the actual hex values, or can you use a sass function like lighten or fade-out to obtain them from the base shade?
screen shot 2017-10-09 at 10 35 02
(Note - this projection is wrong, I just couldn't find a clean svg from the d3 examples, sketch has some glitches when opening the NaturalEarth one). This mockup is just to preview the new colour.

bartaz:
@lyubomir-popov We are using sass so I think I will be able to use functions to calculate colour values.

As for the steps - we have 5 step scale currently, so I suppose the colour steps will be by 20%?
Meaning #48929B is the darkest, and 4 others would be 20% lighter (as in `lighten(#48929B, 20%)`?

matthewpaulthomas:
Omitting Antarctica is unlikely to anger anyone, but chopping off southern Chile? Really? The Google projection is specifically designed to be fast to calculate, an issue that doesn’t even apply here. So if that projection is so tall that we need to crop it, the obvious answer is not to crop it, it’s to use a different projection.

lyubomir-popov:
@bartaz sounds good, only modification I see is to replace the stretched background with a horizontal rule (fixed width, inside a row) (due to the grey background colour of the footer which wasn't reflected in the mockups up until now. Attaching a screenshot:
screen shot 2017-10-10 at 10 56 10

@matthewpaulthomas sorry I should have mentioned when I uploaded the screenshot with the dark blue countries, it was just the outdated map that was in the file. It is not a suggestion to clip Chile.

(I'm having a bit of trouble getting the actual NaturalEarth projection into Sketch due to the file size of the svg I think (some paths appear glitchy and corrupt). I'll screenshot it and update the mockup once Bartek updates the calculator example.)

bartaz:
@lyubomir-popov 

So, to clarify scope of changes to implement:
- projection should be changed to Natural Earth
- map background colour stays white
- instead of border around the map we only have a top rule between map and the heading
- colour scale is based from #48929B and lighter shades of it

lyubomir-popov:
@bartaz yes to all 4 points.

carlaberkers:
Looks great @bartaz! Are we clipping Antartica?

bartaz:
@lyubomir-popov @matthewpaulthomas 

Map demo is updated with new design, colours and tooltips for all countries:
http://snapcraft.io-pr-71.run.demo.haus/gnome-calculator/
http://snapcraft.io-pr-71.run.demo.haus/lxd/

Are we fine with leaving it like that or do we want some more updates?
Is it fine to leave Antarctica as it is or do we want to clip it more (or remove as a whole)?

lyubomir-popov:
@bartaz I'd prefer Antarctica clipped as in the screenshot below.
screen shot 2017-10-11 at 15 33 15

nottrobin:
@bartaz I added a few commits to simplify the python code a bit and fix the external icon CSS through autoprefixing. Just check they look okay to you.

Apart from that, this is all still :+1: from me.

lyubomir-popov:
@bartaz sorry one more thing (lowest priority), could we use .p-tooltip--top-center? I think it will be slightly less likely that the mouse pointer will overlap the tooltip if it appears above. 

Ideally a tooltip should follow the cursor, and originate right above it instead of the country centre, but I'm assuming that is trickier.

bartaz:
@lyubomir-popov @mpt Map demo updated to latest version:
http://snapcraft.io-pr-71.run.demo.haus/lxd/
http://snapcraft.io-pr-71.run.demo.haus/gnome-calculator/
http://snapcraft.io-pr-71.run.demo.haus/canonical-livepatch/
http://snapcraft.io-pr-71.run.demo.haus/core/

Because we started to show all the countries that we don't have data for (usage is 0) with the same colour as countries where usage is close to 0 (0-20%) it made the map look quite flat with USA having a bigger usage and distinct colour and rest of the world being flat 0.

We could probably make countries with little (but any) users more distinct from 0 if we introduce another colour step. Meaning that countries with no data (no users) will be lighter then countries between 0-20%. I'm not sure if we want reveal this information - but it's likely to make the map a bit more interesting and useful.

bartaz:
Examples for same snap (same data):

Current 'flat' map with countries with 0 users and 0-20 users in same colour:
screen shot 2017-10-12 at 15 33 13

Possible map with countries with 0 users with lighter colour than 0-20:

screen shot 2017-10-12 at 15 32 19

lyubomir-popov:
@bartek would it be straightforward to:
1. Switch to 10% bands instead of 20%, to get more variation
2. Switch to one of the GnBu colorbrewer choropleth palettes available in d3? (http://bl.ocks.org/mhkeller/10504471)
( I've been reading about colours in maps and charts as part of the mockup of the graph-heavy "Measure" owner facing page (I'll upload the finished mockups tomorrow), and one of the recommendations for maps/charts is to stick to certain palettes that fit certain requirements like colour-blind-friendly etc. )

I realise this is a change of the original request, so if not straightforward please ignore for now.

Preview - 
screen shot 2017-10-12 at 15 58 44

bartaz:
@lyubomir-popov We could switch to different color scale to make steps more distinct.

But the problem on our map is not that colours are too similar it's that they are exactly the same.
Countries that we don't have any data for are in the same scale step as countries that have very little users. And it turns out that **most** of the countries fall into this lowest scale step.

If we introduce 6th scale step (for countries with 0 users) this will make the countries with some (but very little) users at least visible on the map making the visualisation more interesting (and a little bit different between different snaps).

bartaz:
@lyubomir-popov Some examples with new d3 green-blue 10 step scale:

screen shot 2017-10-13 at 11 26 55
screen shot 2017-10-13 at 11 26 05
screen shot 2017-10-13 at 11 25 54
screen shot 2017-10-13 at 11 25 35

It looks better (colours are more distinct), but sometimes US feels like having colour out of scale (because of significant value difference).
Not sure if it's an issue we can avoid.

lyubomir-popov:
@bartaz I think it is better. As to the US appearing very distinct, is it a consequence of the fact this is an absolute range across all snaps? The colour scheme is quite gradual, so my guess is the numbers in the US are just that much higher in the case of the speed test example.

I personally think it is ok to leave as is. If we have to delve further, one option would be to alter the way we sample the gradient - but to make that call we need to know where the data is concentrated - something like a scatter plot of all data points across all snaps might help to see where the majority of them are.

Below an example with 10 bands of what that might look like:

screen shot 2017-10-13 at 12 36 11

So to get to the dark blue, a country would need to have 500 times more users than a country in the first band, whereas on the current scale, dark blue means 10 times more than the first band.

lyubomir-popov:
@bartaz just noticed that the tooltip colour doesn't match. It does change to a darker shade compared to other countries, but not to the actual shade of the contry you hover on. screen shot 2017-10-13 at 13 43 56

bartaz:
@lyubomir-popov ah yes, I updated CSS for tooltips but not JS that generates them.

lyubomir-popov:
@bartaz One more glitch, not map related - the selected element is inherigint a margin-top that causes a white gap under the footer. zeroing the margin-top fixes it.
screen shot 2017-10-13 at 14 01 41

bartaz:
@lyubomir-popov Can you report the margin bug separately at: https://github.com/canonical-websites/snapcraft.io/issues ?

lyubomir-popov:
@bartaz done https://github.com/canonical-websites/snapcraft.io/issues/94

bartaz:
@lyubomir-popov Demo was just updated with fixed colours in tooltips.

matthewpaulthomas:
@bartaz Following discussion at the retrospective on Friday, please could you un-merge this? We need to solve the problem that (for example) [core](https://staging.snapcraft.io/core/) looks as if it is less popular than [gnome-calculator](https://staging.snapcraft.io/gnome-calculator/), which is impossible.

bartaz:
@matthewpaulthomas Yes, map is removed in #100

carlaberkers:
@matthewpaulthomas @bartaz @nottrobin @lyubomir-popov
What is our plan for solving the problem?

PR #76, opened 2017-09-29
Show license information

### LINKED ISSUE #77: Show license info in snap details pages

nottrobin:
I've just put it at the bottom of the key-value table for snap info at the moment.

However, this will need a UX review to check it's okay.

Fixes #77

QA
--

`./run`, go to e.g.: http://0.0.0.0:8022/dwarf-fortress/.

It's probably also worth trying a few different snaps just in case one of them has a long license, to see if that looks okay.

If it doesn't look okay, maybe I should comment out the template change to get this PR merged, and then let a frontender take over.

### DIFF
--- a/app.py
+++ b/app.py
@@ -204,6 +204,7 @@ def snap_details(snap_name):
         'icon_url': details['icon_url'],
         'version': details['version'],
         'revision': details['revision'],
+        'license': details['license'],
         'publisher': details['publisher'],
         'screenshot_urls': details['screenshot_urls'],
         'prices': details['prices'],
--- a/templates/snap-details.html
+++ b/templates/snap-details.html
@@ -55,6 +55,7 @@
         <table class="p-table-key-value">
           <tr><th width="100">Version</th><td>{{ version }}</td></tr>
           <tr><th>Size</th><td>{{ filesize }}</td></tr>
+          <tr><th>License</th><td>{{ license }}</td></tr>
         </table>
       </div>
     </div>


bartaz:
Design-wise looks quite similar to what @spencerbygraves proposed: https://github.com/CanonicalLtd/snapcraft-design/issues/54#issuecomment-324969516

The only difference is the lack of text "Open source". But I'm not sure it's needed or if we have this distinction in API.

nottrobin:
That's a good point. @spencerbygraves do you think we need the "open source"? If we do then it'll delay things as I'll need to ask the store team a few things. We need an exhaustive list of the open source license options to test against before adding the "open source" wording - unless the license thing is a free text field somewhere, in which case adding "open source" won't be possible.

nottrobin:
@bartaz did you QA it? Could you add QA +1?

nottrobin:
@matthewpaulthomas could we get your opinion on the "open source" issue when you get a sec?

matthewpaulthomas:
I copied the “Open source” phrase from [the equivalent display in Ubuntu Software Center](https://wiki.ubuntu.com/SoftwareCenter#license), where we had the reverse problem: [we didn’t have the information needed to make it more precise](https://launchpad.net/bugs/435183).

If we can be more precise with snaps, then great! What is the distribution of lengths of this field for current snaps?

nottrobin:
Merging this as it appears to have @matthewpaulthomas's approval. We discussed verbally that we could merge and see if the field length turns out to be a problem down the line.

PR #78, opened 2017-09-29
Add X-Commit-ID and X-Hostname headers to all requests

nottrobin:
These should pick up on the COMMIT_ID environment variable and the hostname of the running context respectively.

The COMMIT_ID variable should exist in the built docker image. The `hostname` should be the name of the pod when the image is running in Kubernetes.

If the COMMIT_ID environment variable isn't found, the HTTP header will still be there, but will be set to None.

This PR also add updates from run script v2.1.1: https://github.com/canonical-webteam/generator-canonical-webteam/pull/43. Feel free to review that too if you wish.

QA
--

``` bash
./run
```

Check the site works.

Now check the built image has the commit id

``` bash
$ docker build --build-arg commit_id=`git rev-parse HEAD` --tag snapcraft .  # Build the image
$ docker run --rm --name throwaway -ti -d -p 80:80 snapcraft  # Run the image in the background
$ curl -sI localhost | grep 'X-'  # Check the headers of the response for "commit"
X-Commit-ID: ab279404423effe21454323ab35
X-Hostname: 70cb5292e2eb
$ docker kill throwaway  # Kill the background image
```


nottrobin:
Note to self: Add the k8s pod header before meeting

PR #98, opened 2017-10-18
New snapcraft.io home page

### LINKED ISSUE #96: Implement new home page
TODO: 

- [x] update languages icons
- [x] logos responsive styles code more like Vanilla pattern
- [x] copy text review and updates
- [x] decide on 'Get started with GitHub' button text (comments in copy doc)
- [ ] footer design

bartaz:
New snapcraft.io home page

Fixes #96

Demo:
http://snapcraft.io-pr-98.run.demo.haus/index2


templates/index.html
```diff
@@ -1,692 +1,417 @@
...
+    for(var e,l='article aside footer header nav section time'.split(' ');e=l.pop();document.createElement(e));
+    </script>
+
+    <meta name="description" content="Introducing snaps and commands to install and remove them on Ubuntu 16.04 LTS.">
+
+    <link rel="apple-touch-icon-precomposed" sizes="144x144" href="https://assets.ubuntu.com/v1/3361409d-apple-touch-icon-144x144-precomposed.png" />
+    <link rel="apple-touch-icon-precomposed" sizes="114x114" href="https://assets.ubuntu.com/v1/5fe4d3c8-apple-touch-icon-114x114-precomposed.png" />
+    <link rel="apple-touch-icon-precomposed" sizes="72x72" href="https://assets.ubuntu.com/v1/09460d9a-apple-touch-icon-72x72-precomposed.png" />
+    <link rel="apple-touch-icon-precomposed" href="https://assets.ubuntu.com/v1/cc66da65-apple-touch-icon-precomposed.png" />
+
```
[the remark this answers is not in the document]
bartaz: Good catch, thanks. They were just copied from old snapcraft.io page (where they shouldn't probably be in the first place for the same reasons).

templates/index.html
```diff
@@ -1,692 +1,417 @@
...
+    </div>
 
-                <p>Typically, configuration is stored in one of these, along with
-                system-wide data for the snap.</p>
+    <div class="p-strip is-shallow is-bordered u-no-padding--bottom">
+      <div class="row u-vertically-center">
+        <div class="col-6">
+          <h2>Auto-build from GitHub</h2>
+          <p class="p-heading--five">Automate your build process and publish software straight from your GitHub repos with build.snapcraft.io</p>
+        <p><a class="p-button--neutral" href="https://build.snapcraft.io/auth/authenticate"><i class="p-icon--github"></i> Get started with GitHub</a></p>
```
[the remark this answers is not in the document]
bartaz: Moved this discussion to a copy doc, as we also have "Get started with GitHub" in top banner, so we should probably change both of them.

static/sass/_snapcraft_custom-typography.scss
```diff
@@ -0,0 +1,43 @@
+@mixin snapcraft-custom-typography {
```
anthonydillon: Custom typography?

static/sass/_patterns_testimonial.scss
```diff
@@ -0,0 +1,73 @@
+@mixin vf-p-testimonial {
```
anthonydillon: Can you not use the Vanilla pull-quote style? https://docs.vanillaframework.io/en/patterns/pull-quote

static/sass/_snapcraft_custom.scss
```diff
@@ -0,0 +1,12 @@
+@mixin snapcraft-custom {
+  .snapcraft-hero {
+    background-image: url('#{$assets-path}19da7e00-snapcraft-hero-background.png');
```
anthonydillon: Please use the strip-image pattern: https://docs.vanillaframework.io/en/patterns/strip#strip-image

static/sass/_snapcraft_custom.scss
```diff
@@ -0,0 +1,12 @@
...
+    background-position: 50% 0;
+    background-repeat: no-repeat;
+    background-size: 1440px 1440px;
+  }
+
+  .u-width-auto {
```
anthonydillon: This doesn't do anything where used. The buttons already have `width: auto`

static/sass/styles.scss
```diff
@@ -1,6 +1,11 @@
 // vanilla patterns
 @import 'vanilla-framework/scss/vanilla';
-@include vanilla-base;
+
+$color-brand: #00302f; // TODO: this is a temprary brand colour
+$color-accent: $color-brand; // temporary for strip background
+$color-accent-background: $color-accent;
```
anthonydillon: I believe this is default styling: https://github.com/vanilla-framework/vanilla-framework/blob/develop/scss/_settings_colors.scss#L23

templates/index.html
```diff
@@ -1,692 +1,425 @@
...
+
+    <div class="p-strip--accent is-deep snapcraft-hero">
+      <h1 class="u-off-screen">Snapcraft</h1>
+      <div class="row">
+        <div class="col-12">
+          <p class="p-heading--two">Package, distribute, and update any app for Linux and IoT</p>
+          <p class="p-heading--five">Snaps are containerised software packages that are simple to create and install. They auto-update and are safe to run and because they bundle their dependencies, they work on all major Linux systems without modification.</p>
+          <a href="https://tutorials.ubuntu.com/tutorial/create-your-first-snap" class="u-width-auto p-button--positive is-inline--right">Build your first snap</a>
+          <span class="is-inline--right">or</span>
+          <a href="https://build.snapcraft.io" class="u-width-auto p-button--neutral">Get started with GitHub</a>
```
anthonydillon: Would you consider putting the button section in a `p` tag, for example:
```html
<p>
    <a href="https://tutorials.ubuntu.com/tutorial/create-your-first-snap" class="p-button--positive is-inline--right">Build your first snap</a> or <a href="https://build.snapcraft.io" class="p-button--neutral">Get started with GitHub</a>
</p>
```

templates/index.html
```diff
@@ -1,692 +1,425 @@
...
-                system-wide data for the snap.</p>
+    <div class="p-strip is-shallow is-bordered u-no-padding--bottom">
+      <div class="row u-vertically-center">
+        <div class="col-6">
+          <h2>Auto-build from GitHub</h2>
+          <p class="p-heading--five">Automate your build process and publish software straight from your GitHub repos.</p>
+        <p><a class="p-button--neutral" href="https://build.snapcraft.io/auth/authenticate"><i class="p-icon--github"></i> Get started with GitHub</a></p>
+        </div>
+        <div class="col-6">
+          <img class="u-float-right" src="https://assets.ubuntu.com/v1/006db1f4-build-snapcraft-heroku.png" />
```
anthonydillon: Instead of using `u-no-padding--bottom` please use the `image-position` pattern: https://docs.vanillaframework.io/en/utilities/image-position

templates/index.html
```diff
@@ -1,692 +1,425 @@
...
+      <div class="row">
+        <div class="col-12">
+          <h3>Featured snaps</h3>
+        </div>
+      </div>
+      <div class="row">
+        <div class="col-3">
+          <div class="p-heading-icon">
+            <div class="p-heading-icon__header">
+              <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/2ca6d139-atom.png" alt="icon" />
```
anthonydillon: Could you provide a better alt text or leave it empty.

templates/index.html
```diff
@@ -1,692 +1,425 @@
...
-            <a href="https://www.facebook.com/snapcraftio" class="social-list__item--facebook">Share on Facebook</a>
-          </li>
-          <li class="social-list__item">
-            <a href="https://www.youtube.com/snapcraftio" class="social-list__item--youtube">Share on YouTube</a>
-          </li>
-        </ul>
+    <footer class="p-footer p-strip--light">
+      <div class="row">
+        <div class="col-9">
+          <p>&copy; 2017 Canonical Ltd. Ubuntu and Canonical are registered trademarks of Canonical Ltd.</p>
```
anthonydillon: Try and update the year with a template variable as we are nearing the end of year :)

templates/index.html
```diff
@@ -1,692 +1,425 @@
 <!DOCTYPE html>
 <html>
```
anthonydillon: Please add `lang="en"`

templates/index.html
```diff
@@ -1,692 +1,425 @@
...
-            </div>
+        <div class="col-6">
+          <p class="p-heading--five">Snaps are available for any Linux OS running snapd, the service that runs and manage snaps.</p>
+          <p><a href="https://docs.snapcraft.io/core/install">Enable snaps on your OS</a></p>
+        </div>
+      </div>
+      <div class="row">
+        <div class="col-12 p-logo-links">
+          <a class="p-logo-link" href="https://docs.snapcraft.io/core/install-ubuntu">
+            <img src="https://assets.ubuntu.com/v1/d6c13911-ubuntu.png" />
```
anthonydillon: Please ensure images have alt tags.

templates/index.html
```diff
@@ -1,692 +1,425 @@
...
+            <li class="p-navigation__link" role="menuitem"><a href="https://build.snapcraft.io">Build</a></li>
+            <li class="p-navigation__link" role="menuitem"><a href="/docs">Docs</a></li>
+            <li class="p-navigation__link" role="menuitem"><a class="p-link--external" href="https://tutorials.ubuntu.com">Tutorials</a></li>
+            <li class="p-navigation__link" role="menuitem"><a class="p-link--external" href="https://forum.snapcraft.io/categories">Forum</a></li>
+          </ul>
+        </nav>
+      </div>
+    </header>
+
+    <div class="p-strip--accent is-deep snapcraft-hero">
```
anthonydillon: Please add `id="main-content"` here please. So the Jump to main content link above has a target.

templates/index.html
anthonydillon: Please add `alt` attr

static/sass/_snapcraft_custom.scss
bartaz: I think what @lyubomir-popov used it for is mobile buttons - to prevent them from getting full width on mobile.

static/sass/_snapcraft_custom-typography.scss
bartaz: These currently contain updates to line height for headings (multiline) used on home page and the max-width limiting based on font size (proposed and discussed on vanilla group before).

static/sass/_patterns_testimonial.scss
bartaz: Its font size is way too big for displaying longer testimonials next to each other.

static/sass/_snapcraft_custom.scss
lyubomir-popov: @anthonydillon wasn't aware of the strip image, thanks for pointing it out. It replicates most of what's needed here, apart from:
```background-size: {exact image dimensions; background-position: center 0;}```

Unlike ```background-size:cover```,  the above give you full control over how background image and content in the foreground overlap.

If you look at this element of the background in the strip example, it moves as you resize, and eventually goes underneath the text.
image

image

Specifying exact dimensions and an "anchoring" point ensures the image remains exactly as designed, which is very handy when you have predictable length text due to max-width.

You can see that when you resize the screen. The diagonal runs through the dot in the last paragraph.

Another example (from maas) is this diagonal, which I can ensure will run precisely across the play button in the foreground:
image

Comment (lyubomir-popov) [reply static/sass/_snapcraft_custom.scss]: @anthonydillon As to the width: auto, I was trying to prevent this from happening but I should have at least put it in a media query:
image
Full width buttons look nice on mobile (up to a certain breakpoint, and I'd argue 767px is far too wide) if you just have one. But when you have 2 as in this case, it looks too disjointed (especially when we have or in between the two).

File: static/sass/_patterns_testimonial.scss (hunk 4 above)
Comment (anthonydillon) [reply static/sass/_patterns_testimonial.scss]: Thought so, might be good to propose a modification to the testimonial pattern in Vanilla

File: static/sass/_snapcraft_custom-typography.scss (hunk 3 above)
Comment (anthonydillon) [reply static/sass/_snapcraft_custom-typography.scss]: Ok thanks, we need to fix the line-heights in Vanilla.

File: static/sass/_snapcraft_custom.scss (hunk 6 above)
Comment (anthonydillon) [reply static/sass/_snapcraft_custom.scss]: Add the background-size styling to the element too. It's very bespoke styling, therefore, the same as using a bespoke class.

Comment (anthonydillon) [reply static/sass/_snapcraft_custom.scss]: Regarding the `width: auto` I didnt check small screens. Thanks.

**Comment by lyubomir-popov** [comment]
@bartaz @nottrobin @carlaberkers minutes from the review meeting:

1. Quotes: 
• Check whether titles are available
• Add light grey background
• Move quotes above "Autobuild with github" section to avoid 2x 3 column layouts next to each other

"What is Snapcraft" section: 
• try h5 headings instead of paragraph styling works well in check list
• cross browswer testing

• Change p to h5 where possible
• custom styles to apply to tags as well, not just p-heading--*  classes

File: templates/index.html (hunk 8 above)
Comment (lyubomir-popov) [reply templates/index.html]: @anthonydillon I removed the wrapping p around the buttons as it was causing a duplicated margin-top: p+p (1.5rem) + *+* (2rem) on the two anchors inside. I know u-no-margin--top can address that, but then on mobile it leads to the two buttons (or the text "or" in between) to get too close to each other like this:
image

Curious though, since a button is meant to look and behave like a single line block, while a paragraph is something that naturally wraps, is there a reason it needs to be inside a p?

File: static/sass/_snapcraft_custom.scss (hunk 6 above)
Comment (lyubomir-popov) [reply static/sass/_snapcraft_custom.scss]: Just to confirm about full width buttons all the way up to 767px - @anthonydillon 
image 


Was it designed to be like this, or a missing media query between $breakpoint-small and $breakpoint-medium ?

Text + padding seems to make a lot more sense to me, especially in the context of left-aligned preceding copy:

image

Comment (anthonydillon) [reply static/sass/_snapcraft_custom.scss]: Yes it was designed this way mainly due to a lack of a breakpoint lower then 768px. You could go full width with the buttons and hide the `or` on smaller screens?

File: templates/index.html (hunk 8 above)
Comment (anthonydillon) [reply templates/index.html]: The p is for proper spacing but I see you would get margin top on the second button. It was just a suggestion. Feel free to leave it as it is.

File: static/sass/_snapcraft_custom.scss (hunk 6 above)
Comment (lyubomir-popov) [reply static/sass/_snapcraft_custom.scss]: @anthonydillon just had a word with @kwm14 about full width buttons, as there is a compromise involved in both scenarios - the simplest solution seems to be a design decision to not have full width buttons unless we can have an upper limit on their width, at a much smaller width than 767px. 

So just dropping the ```width: 100%``` in vanilla and removing the overrides here once vanilla is updated. I'll make a vanilla proposal for the button patterns.

File: templates/index.html [hunk 15]
```diff
@@ -1,692 +1,417 @@
...
-
-<ul class="footer__links">
-    <li class="footer__list-item"><a href="http://www.ubuntu.com/legal">Legal information</a><span class="footer__dot"> &#183;</span></li>
-    <li class="footer__list-item"><a href="https://github.com/ubuntudesign/snapcraft.io/issues/new">Report a bug on this site</a><span class="footer__dot"> &#183;</span></li>
-    <li class="footer__list-item"><a href="http://status.snapcraft.io/">Service status</a></li>
-</ul>
-
-<p class="footer__text">Powered by <a href="https://www.ubuntu.com/kubernetes">the Canonical Distribution of Kubernetes</a></p>
-
-<p class="note six-col">Yocto Project and all related marks and logos are trademarks of The Linux Foundation. This website is not, in any way, endorsed by the Yocto Project or The Linux Foundation. Some artwork by the <a href="http://www.gnome.org">GNOME Project</a>, licensed <a href="http://creativecommons.org/licenses/by-sa/3.0/">CC-BY-SA</a>.</p>
```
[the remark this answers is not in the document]
bartaz: Added.

static/sass/styles.scss
bartaz: Yes it is. It didn't work for me because I defined values after importing vanilla.

Moving assign before import fixes this and makes it possible to assign `$color-brand` only and leave `$color-accent-background` as default.

bartaz:
@anthonydillon Thanks for your comments. I think they were addressed. Feel free for second review when you have a moment.

static/sass/_snapcraft_custom.scss
nottrobin: @anthonydillon are we still using set breakpoints in Vanilla? That's mildly disappointing.

anthonydillon: @nottrobin Yes, here are the [available breakpoints](https://github.com/vanilla-framework/vanilla-framework/blob/develop/scss/_settings_breakpoints.scss). What would you like to see instead?

templates/index.html
```diff
@@ -1,692 +1,427 @@
...
-                will also integrate all the snap daemons (services) with your
-                init system so that they are automatically started and stopped
-                as needed.</p>
+          </div>
+        </div>
+        <div class="col-3">
+          <div class="p-heading-icon">
+            <div class="p-heading-icon__header">
+              <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/4eca1f4e-discord.png" alt="" />
+              <h5 class="u-no-margin"><a href="/atom">Atom</a></h5>
```
tbille: Here it should be Discord

Lukewh: Could we also get some x2 images for hdpi?

anthonydillon:
Thanks @bartaz, can this merge once the conflicts are resolved?

anthonydillon:
From @bartaz 
> It's on hold for now, there is some design work for footer planned and A/B testing plan we need to prepare next iteration before it can be merged.

bartaz:
Not to postpone merging this too much we decided to move this to separate file and URL, so it can be merged without changing current home page.

bartaz:
@matthewpaulthomas As your initial comments here were addressed already, could you approve it?

We'd like to merge it as it is and improve on it later without exposing it as a home page before it's ready.

templates/index2.html
```diff
@@ -0,0 +1,442 @@
...
+      </div>
+    </div>
+
+    <div class="p-strip is-deep is-bordered">
+      <div class="row">
+        <div class="col-6">
+          <h2>The Snap Store</h2>
+        </div>
+        <div class="col-6">
+          <p class="p-heading--five">The Snap Store is enabled by default on tens of millions Linux systems. It allows developers to release free or paid apps for multiple architectures, on multiple release channels from daily builds to stable releases.</p>
```
matthewpaulthomas: Sorry I missed this last time: “tens of millions _of_ Linux systems”.

templates/index2.html
```diff
@@ -0,0 +1,442 @@
...
+            </div>
+          </div>
+        </div>
+        <div class="col-3">
+          <div class="p-heading-icon">
+            <div class="p-heading-icon__header">
+              <a class="p-heading-icon__img" href="/go">
+                <img src="https://assets.ubuntu.com/v1/b3e93ee1-golang.png" alt="" />
+              </a>
+              <h5 class="u-no-margin"><a href="/go">Golang</a></h5>
```
matthewpaulthomas: In my previous review I said “As with the language logos above, each snap icon here should be part of the same link as the text.”

This is an improvement, in that the icons are links. Unfortunately they’re separate links, with no text of their own, which is an accessibility problem: a screenreader will resort to saying something like “Graphic link: B 3 E 9 3 E E 1 dash golang dot P N G”. This could be avoided by making the icon part of the same link as the text, as you’ve done with the language logos.

bartaz: This vanilla pattern doesn't really allow that by default. Will try to work around that later.

I haven't heard that screen reader would try to read a URL of an image, I would rather expect it to read URL where link leads to ("/go" in this case). But I guess adding alt="Golang" to the image could fix that.
But that still would mean two similar links would be read by screenreader.

PR #141, opened 2017-11-13
Simple HTML for search results using Vanilla media object pattern

bartaz:
Added basic Vanilla based HTML for search results.

### QA

- ./run
- Go to search https://localhost:8004/search
- Search for some snaps
- Results should show up as Vanilla Media object pattern

Or try demo:
http://snapcraft.io-pr-141.run.demo.haus/search?q=gnome&limit=10&offset=0

screen shot 2017-11-13 at 14 54 55

### DIFF
--- a/templates/search.html
+++ b/templates/search.html
@@ -1,55 +1,69 @@
 {% extends "_layout.html" %}
 
-{% block title %}{{ snap_title }} — Linux software in the Snap Store{% endblock %}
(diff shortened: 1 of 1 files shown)


bartaz:
@Lukewh Yep, descriptions themselves tend to be long sometimes, so I guess they should be trimmed for search results anyway.

PR #156, opened 2017-11-28
Add get feature snaps function

tbille:
# Summary

Add function to get the featured snaps

# QA 

When the discover page is added the return of this function needs to be added to the template.
This pr is in prerparation for the discover page, so not possible to QA for now. 
You can print the return of the function on the homepage to see that you get a result.

But basically I am just calling this endpoint:
```
curl "https://api.snapcraft.io/api/v1/snaps/search?confinement=strict,classic&q=&section=featured"
```

### DIFF
--- a/app.py
+++ b/app.py
@@ -69,6 +69,28 @@ search_query_headers = {
 }
 
 
+def get_searched_snaps(search_results):
+    return (
+        search_results['_embedded']['clickindex:package']
+        if search_results['_embedded']
+        else []
+    )
+
+
+def get_featured_snaps():
+    url = (
+        "https://api.snapcraft.io/api/v1/snaps/search"
+        "?confinement=strict,classic&q=&section=featured"
+    )
+
+    featured_response = _get_from_cache(
+        url,
+        headers=search_query_headers
+    )
+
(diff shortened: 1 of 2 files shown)


tbille:
# QA

* `./run`
* [http://0.0.0.0:8004/](http://0.0.0.0:8004/)
* Open developper tools (CTRL-MAJ-i)
* Inspect code, check that at the bottom of the code you have a comment with some json data (the featured snaps)

Lukewh:
for the example could you wrap it in a script tag:

```<script>console.log(//jsoninthecomment);</script>```

to make it easier to see. Can't tell in the html comment whether the json is valid.

bartaz:
@tbille It seems that it doesn't work on demo:
http://snapcraft.io-pr-156.run.demo.haus/discover gives 404

tbille:
@Lukewh I won't add the log, because it's a python object, it would be long and useless to convert this object to JSON.
@bartaz the featured snaps have the same structure as the snaps displayed in the search page.
+ demo not working because my first PR didn't had the discover page merged. It seems that the demo server doesn't update when we push new commits (or modified commits)

PR #166, opened 2017-12-05
Discover featured snaps

### LINKED ISSUE #162: Build featured snaps [/discover - phase 2]
### Task
@bartaz to build 'featured snaps' section using @tbille's endpoint for #197 

### Stakeholders
@greggless @lyubomir-popov 

(moved from https://github.com/CanonicalLtd/snapcraft-design/issues/198)

bartaz:
Adds featured snaps from store API to discover page.
Also updates featured snaps on home page to make them clickable as a whole block.

Fixes #162 

### QA

- ./run
- Go to http://localhost:8004/discover
- Featured snaps should be shown

Or: http://snapcraft.io-pr-166.run.demo.haus/discover/

screen shot 2017-12-05 at 13 16 41

### DIFF
--- a/templates/discover.html
+++ b/templates/discover.html
@@ -25,7 +25,36 @@
     </div>
   </section>
 
-  <!--
-    {{ featured_snaps }}
-  -->
+  <section class="p-strip is-deep is-bordered">
+    <div class="row">
+      <div class="col-12">
+        <h2>Featured snaps</h2>
+      </div>
+    </div>
+    <div class="row">
+      {% for snap in featured_snaps %}
+        <div class="col-3">
+          <div class="p-heading-icon">
+            <div class="p-heading-icon__header">
+              <a class="p-heading-icon__img" href="/{{ snap.package_name }}">
+                {% if snap.icon_url %}
+                  <img src="{{ snap.icon_url }}" alt="">
+                {% else %}
+                  <img src="https://assets.ubuntu.com/v1/610689b1-default-package-icon.svg" alt="">
+                {% endif %}
+              </a>
+              <h5 class="u-no-margin"><a href="/{{ snap.package_name }}">{{ snap.title }}</a></h5>
+            </div>
+          </div>
+        </div>
+
+        {# end the row after every 4th snap #}
+        {% if (loop.index % 4 == 0) and not loop.last %}
+          </div>
+          <div class="row">
+        {% endif %}
+
+      {% endfor %}
+    </div>
+  </section>
 {% endblock %}
--- a/templates/discover.html
+++ b/templates/discover.html
@@ -39,7 +39,7 @@
               {% if snap.icon_url %}
                 <img class="p-heading-icon__img" src="{{ snap.icon_url }}" alt="">
               {% else %}
-                <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/610689b1-default-package-icon.svg" alt="">
+                <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/6fbb3483-snapcraft-default-snap-icon.svg" alt="">
               {% endif %}
               <h5 class="u-no-margin">{{ snap.title }}</h5>
             </a>
--- a/templates/discover.html
+++ b/templates/discover.html
@@ -35,16 +35,14 @@
       {% for snap in featured_snaps %}
         <div class="col-3">
           <div class="p-heading-icon">
-            <div class="p-heading-icon__header">
-              <a class="p-heading-icon__img" href="/{{ snap.package_name }}">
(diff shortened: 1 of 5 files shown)


templates/discover.html
```diff
@@ -25,7 +25,34 @@ <h2>Discover snaps for any Linux device</h2>
...
+              {% else %}
+                <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/610689b1-default-package-icon.svg" alt="">
+              {% endif %}
+              <h5 class="u-no-margin">{{ snap.title }}</h5>
+            </a>
+          </div>
+        </div>
+
+        {# end the row after every 4th snap #}
+        {% if (loop.index % 4 == 0) and not loop.last %}
```
bartaz: Because of how Vanilla grid works we need to split the snaps into rows.
This means we need this tricky bit inside to loop to detect when the row should end and 'inject' a new row in between the snaps.

templates/index2.html
```diff
@@ -292,84 +292,68 @@ <h3>Featured snaps</h3>
       <div class="row">
         <div class="col-3">
           <div class="p-heading-icon">
-            <div class="p-heading-icon__header">
-              <a class="p-heading-icon__img" href="/atom">
-                <img src="https://assets.ubuntu.com/v1/694b2e80-atom.png" alt="" />
-              </a>
-              <h5 class="u-no-margin"><a href="/atom">Atom</a></h5>
-            </div>
+            <a class="p-heading-icon__header" href="/atom">
```
nottrobin: Shouldn't this be generated from the same feed as the `/discover` page?

nottrobin:
Is this based on a design or is it just something you came up with?

If it's your design, how would you feel about adding the description as well, to give it a little more flavour?

spencerbygraves:
@lyubomir-popov is working on the design.

templates/discover.html
lyubomir-popov: @bartaz there's actually a very handy column() mixin - https://github.com/vanilla-framework/vanilla-framework/blob/develop/scss/grid/shelves/mixins/_column.scss  that is accessible in vanilla - it allows you to clear first column of the next row, so you don't have to generate additional .row markup. This will clear 
`        >:nth-child(3n+1) {
          @include column(6, $grid-columns, true);
        }
`
I've used it to fix a hard-coded 50% instead of 6-col here so it should be permissable: https://github.com/vanilla-framework/vanilla-framework/blob/develop/scss/_patterns_lists.scss#L218

(apologies if I'm saying something you know and avoid doing because of other reasons)

lyubomir-popov:
@spencerbygraves I think this was scheduled to go out without a design, and implement the design as an enhancement - this is because development is ahead of UX. My design will be reviewed tomorrow, but I don't think this should stop this from going live - at least if we follow the current workflow which is build with vanilla, design later.

bartaz:
@nottrobin It was designed already for home page (Featured snaps section): http://snapcraft.io-pr-166.run.demo.haus/index2

So it's basically reusing same pattern here.

templates/index2.html
bartaz: Home page was designed with hardcoded list of featured snaps.

Yes, it seems it could make sense to reuse the same feed on home page as well.

templates/discover.html
bartaz: @lyubomir-popov 
I generally avoid adding CSS as we are aiming to use Vanilla as it is. Especially reusing Vanilla internal implementation (such as column() mixin and $grid-columns doesn't seem good from maintenance point of view, as these can change in Vanilla at any time).

templates/index2.html
nottrobin: @bartaz please could you do that? I don't think we should have two different lists of "featured" snaps on the same site.

bartaz:
Review meeting decisions:
- update default logo to: https://github.com/CanonicalLtd/snapcraft-design/blob/master/assets/logos/snap-default-icon.svg
- put same API featured snaps on home page

templates/discover.html
lyubomir-popov: @bartaz Ah I see. Makes sense. Would it be possible to have the snaps display 2 per row on mobile & tablet resolutions with the above markup?

bartaz:
@nottrobin Home page featured snaps are fetched from API, and default icon is updated.

Could you re-review it?

templates/discover.html
bartaz: @lyubomir-popov No. It uses grid, so on mobile it just goes down to one column.

As you are redesigning it anyway for next phase, I would propose leaving it as it is now and address these issues when implementing new design. At that point we could use similar flexbox solution to one we used for logos on home page.

PR #172, opened 2017-12-06
Average map data over a week

### LINKED ISSUE #167: Reduce the smoothing of map data to a week (from a month)
We've agreed with @cprov to reduce the smoothing of the map data to weekly - as in the information shown on the map will be the result of averaging the data over the last week.

nottrobin:
The map was previously showing usage averaged over the preceding month.
@cprov was concerned that this might not show fresh enough data for new
snaps, so I'm changing it to show an average over the last week.

Fixes #167

QA
--

``` bash
./run
```

go to e.g. <http://0.0.0.0:8004/lxd/> and check the map still works.

Try to find cases of fairly new snaps where the data looks different than on live. Or if you can't find any, just trust that it's working ;)

### DIFF
--- a/app.py
+++ b/app.py
@@ -249,7 +249,7 @@ def snap_details(snap_name):
     with appropriate sanitation.
     """
     today = datetime.datetime.utcnow().date()
-    month_ago = today - relativedelta.relativedelta(months=1)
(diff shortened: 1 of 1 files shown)


matthewpaulthomas:
Seems reasonable to me.

PR #173, opened 2017-12-08
Moves new home page to root URL and fixes small grid issues

bartaz:
Moves new snapcraft home page to / root URL and fixes small issues with the new grid gutter (in footer and testimonials).

### QA

Go to:
http://snapcraft.io-pr-173.run.demo.haus/

New home page should be displayed, there should be no layout issues (for example social icons in footer should be in one line).


app.py
```diff
@@ -189,13 +189,8 @@ def get_pages_details(links):
...
 @app.route('/')
 def homepage():
-    return flask.render_template('index.html', now=datetime.datetime.utcnow())
-
-
-@app.route('/index2')
-def new_homepage():
     return flask.render_template(
-        'index2.html',
+        'index2.html',  # old home page is in index.html in case we need it
```
nottrobin: Please remove `index2.html` and put this at `index.html` and remove this comment. We have version control if we need to retrieve anything.

PR #176, opened 2017-12-08
Measure page

Lukewh:
Initial measure page


Lukewh:
## QA

- `./run serve`
- Visit http://0.0.0.0:8004/snaps/dwarf-fortress/measure
- Or visit http://snapcraft.io-pr-176.run.demo.haus/snaps/dwarf-fortress/measure

templates/publisher/measure.html
```diff
@@ -0,0 +1,102 @@
...
+      <div class="row">
+        <div class="u-clearfix">
+          <h3 class="u-float--left">Installs</h3>
+          <strong class="p-heading--three u-float--right u-no-margin--top">{{ installs_total }}</strong>
+        </div>
+        <div id="installs_metrics" class="snapcraft-metrics__graph">
+        </div>
+      </div>
+    </section>
+    <!-- <div class="row">
```
bartaz: Maybe hide it with flask template comment? `{# #}`

Lukewh: Done! Thanks @bartaz

spencerbygraves:
Looking good :+1

package.json
```diff
@@ -3,16 +3,18 @@
     "test": "sass-lint static/**/*.scss --verbose --no-exit",
     "build": "yarn run build-js && yarn run build-css",
     "build-css": "node-sass --include-path node_modules static/sass --output static/css && postcss --use autoprefixer --replace 'static/css/**/*.css' && postcss --use cssnano --dir static/minified 'static/css/**/*.css'",
-    "build-js": "cp node_modules/d3/build/d3.min.js static/js/modules && cp node_modules/d3-geo/build/d3-geo.min.js static/js/modules && cp node_modules/topojson-client/dist/topojson-client.min.js static/js/modules",
+    "build-js": "cp node_modules/d3/build/d3.min.js static/js/modules && cp node_modules/d3-geo/build/d3-geo.min.js static/js/modules && cp node_modules/topojson-client/dist/topojson-client.min.js static/js/modules && cp node_modules/billboard.js/dist/billboard.min.js static/js/modules && cp node_modules/moment/min/moment.min.js static/js/modules",
```
bartaz: I think we are getting to the point when we could use something more flexible than that. Listing all the scripts manually in package.json will get painful really soon.

static/js/snap-metrics.js
```diff
@@ -0,0 +1,251 @@
...
+      var yAxis = el.querySelector('.bb-axis-y');
+
+      ticks = yAxis.querySelectorAll('.tick');
+      cullYAxis(ticks);
+    }
+
+    /**
+     * Generate the tooltip.
+     * @param {Object} data The point data.
+     * @returns {HTMLElement}
```
bartaz: It doesn't seem to return an element but a HTML string.

package.json
Lukewh: I though about rolling something in to this pr, but thought we could discuss it this week and maybe we could do a cheeky pr at some point.

static/js/snap-metrics.js
Lukewh: Good spot, I'm a LIAR!

Lukewh: dun

nottrobin:
What are we doing with the "30 days" and "All tracks and branches"? Having disabled select boxes is confusing. Could we turn this into a version that makes sense in the absence of other choices?

I would say the options are either comment out those select boxes or turn them into plain text.

static/js/snap-metrics.js
```diff
@@ -0,0 +1,285 @@
+/* global d3, bb, moment */
+
+(function (window, d3, bb, moment) {
+  /**
+   * Debounce
+   * @param {Function} func Function to run.
+   * @param {Number} wait Time to wait between tries.
+   * @param {Boolean} immediate Immediately call func.
+   */
+  function debounce(func, wait, immediate) {
```
bartaz: This seems like something that could be useful in other places as well.
So the question is how to best share it across different modules...

Probably not worth to block this PR on this decision, but something to think of sooner than later.

static/js/snap-metrics.js
```diff
@@ -0,0 +1,285 @@
...
+    /**
+     * Generate the tooltip.
+     * @param {Object} data The point data.
+     * @returns {String} A string of HTML.
+     */
+    function snapcraftGraphTooltip(data) {
+      var contents = ['<div class="p-tooltip p-tooltip--top-center">'];
+      contents.push('<span class="p-tooltip__message" role="tooltip">');
+      contents.push('<span class="snapcraft-graph-tooltip__title">' + moment(data[0].x).format('YYYY-MM-DD') + '</span>');
+      for (var i = 0, ii = data.length; i < ii; i += 1) {
```
bartaz: Why not `data.forEach`?

static/js/snap-metrics.js
```diff
@@ -0,0 +1,285 @@
...
+        x: e.x,
+        y: e.y
+      };
+    };
+
+    window.addEventListener('resize', resize);
+    window.addEventListener('mousemove', mouseMove);
+  }
+
+  window.renderMetrics = renderMetrics;
```
bartaz: We also need to think of how to do JS moving forward (meaning having JS functionality and data from the template). Exposing functions directly in global scope will not scale very well.

We could at least namespace it in `snapcraft.renderMetrics` or something. But also this feels so not like JS these days...

Thoughts?

bartaz:
Tooltip above last bar seems to be bit misaligned:

screen shot 2017-12-13 at 11 07 13

lyubomir-popov:
@Lukewh image
Can we hide the box shadow above the top edge? Or use a background-image gradient instead so we just get the part below the tabstrip?

static/js/snap-metrics.js
Lukewh: @bartaz I agree to all points. My only extra thought is that snapcraft.io will, for now, not have a lot of JS that is required for all pages and should be quite localised, so shouldn't be a massive issue.... For now.

I quite like `snapcraft.*` where we perhaps namespace `snapcraft.metrics.*`. But like you say, this doesn't feel hipster enough.

While we're thinking about this we should consider minification... Adding some cheeky ES[latest] and transpilation will sort this right out though AND we'll feel like the hippest hipsters on JavaScript Lane in Hipsterville.

static/js/snap-metrics.js
Lukewh: I don't like the capitalisation.... Changed.

static/js/snap-metrics.js
Lukewh: Yeah I think a lib of some kind will definitely be required moving forwards.

Lukewh:
@nottrobin I've hidden the dropdowns for now to get this landed, I can revisit next iteration while dealing with the new graphs.

static/js/snap-metrics.js
```diff
@@ -0,0 +1,293 @@
...
+      if (callNow) func.apply(context, args);
+    };
+  }
+
+  /**
+   * Render all metrics
+   * @param {Object} metrics An object of metrics from the API.
+   */
+  function renderMetrics(metrics) {
+    if (!d3 || ! bb) {
```
bartaz: Nit picking but space between `! bb` seems unnecessary.

static/js/snap-metrics.js
```diff
@@ -0,0 +1,293 @@
...
+     * Cull Y Axis. The built in culling does not provide enough control.
+     *  - hide labels that are not every X_TICK_FREQUENCY ticks.
+     *  - remove month abreviation from label for sequential dates that have 
+     *    the same month.
+     * @param {NodeList} ticks X axis tick elements.
+     */
+    function cullXAxis(ticks) {
+      var tick, totalTicks, text, monthCache;
+
+      for (tick = 0, totalTicks = ticks.length; tick < totalTicks; tick += 1) {
```
bartaz: Unless you need a `tick` for anything else then iterator, this seems like it could be a `forEach` as well...

static/js/snap-metrics.js
```diff
@@ -0,0 +1,293 @@
...
+        text = ticks[tick].querySelector('text');
+
+        if (tick % X_TICK_FREQUENCY !== 0) {
```
bartaz: Ah, here you use `tick` directly.
So I don't mind leaving it as it is, but `forEach` will remove some unnecessary bloat.

static/js/snap-metrics.js
```diff
@@ -0,0 +1,293 @@
...
+
+    /**
+     * Cull Y Axis. The built in culling does not provide enough control.
+     *  - hide labels that are not every Y_TICK_FREQUENCY ticks.
+     * @param {NodeList} ticks Y axis tick elements.
+     */
+    function cullYAxis(ticks) {
+      var tick, totalTicks, text;
+
+      for (tick = 0, totalTicks = ticks.length; tick < totalTicks; tick += 1) {
```
bartaz: See above about `forEach`.

static/js/snap-metrics.js
Lukewh: They're also SVG elements, not sure forEach works with them as they're classed as a NodeList?

PR #185, opened 2017-12-13
Fixes copyright notice in footer of directory page layout.

bartaz:
Updates footer copyright notice on directory pages (discover, search, snap details) so it contains full copyright of Canonical and Ubuntu.

screen shot 2017-12-13 at 13 32 11

### QA

- ./run serve
- go to http://localhost:8004/discover http://localhost:8004/lxd
- footer should contain "© 2017 Canonical Ltd. Ubuntu and Canonical are registered trademarks of Canonical Ltd."

### DIFF
--- a/app.py
+++ b/app.py
@@ -92,6 +92,17 @@ def get_featured_snaps():
     return get_searched_snaps(featured_response.json())
 
 
+# Context processors
+# ===
+@app.context_processor
+def inject_datetime():
+    """
+    Adds current date to all the templates, so we can display
+    current year in the footer copyright disclaimer.
+    """
+    return dict(now=datetime.datetime.utcnow())
+
+
 # Error handlers
 # ===
 @app.errorhandler(404)
@@ -191,8 +202,7 @@ def get_pages_details(links):
 def homepage():
     return flask.render_template(
         'index.html',
-        featured_snaps=get_featured_snaps(),
-        now=datetime.datetime.utcnow()
+        featured_snaps=get_featured_snaps()
     )
 
 
--- a/app.py
+++ b/app.py
@@ -92,17 +92,6 @@ def get_featured_snaps():
     return get_searched_snaps(featured_response.json())
 
 
-# Context processors
-# ===
-@app.context_processor
-def inject_datetime():
-    """
-    Adds current date to all the templates, so we can display
-    current year in the footer copyright disclaimer.
-    """
-    return dict(now=datetime.datetime.utcnow())
-
-
 # Error handlers
 # ===
 @app.errorhandler(404)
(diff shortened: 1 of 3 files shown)


app.py
```diff
@@ -234,7 +237,8 @@ def search_snap():
 
     return flask.render_template(
         'search.html',
-        **context
+        **context,
+        now=datetime.datetime.utcnow()
```
bartaz: @nottrobin @tbille Is it possible in flask to make sure something (like `now` here) is always added to context by default, without adding it separately in every view?

tbille: If we were only using `context` in every function we could create a function that creates this object, and that adds each time this var.
Otherwise use JS :P

bartaz: OK, found it myself. It's called `@app.context_processor` ;)

app.py
```diff
@@ -92,6 +92,17 @@ def get_featured_snaps():
     return get_searched_snaps(featured_response.json())
 
 
+# Context processors
+# ===
+@app.context_processor
+def inject_user():
```
tbille: change function name

app.py
nottrobin: @bartaz what do you need the time for? Is this to calculate the year for the copyright notice?

Could I suggest we don't make that dynamic? The copyright *starts* from the year mentioned, it doesn't need to be the latest year:

> The general rule is that the year to include in a copyright notice is the year of first publication of the work. First publication is when the work is made available to the public without restriction.

(From http://www.copyrightlaws.com/copyright-notice-year/)

I think if we're going to use a year in the copyright notice (and I'd love to talk about removing the year, but feat that discussion will delay this unnecessarily), I think the most valid year to use is the current year at the point that the copyright notice is written - *not* the current year. So make i t 2017 now, and it should stay 2017 until we make a new website, or actively decide to update it for another reason.

nottrobin: See also: https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

> The date on the notice establishes how far back the claim is made. This means if you update the date, you are no longer claiming the copyright for the original date

templates/_layout.html
```diff
@@ -41,7 +41,7 @@
       <footer class="p-footer u-no-margin--top">
         <div class="row">
           <div class="col-6">
-            <p>Directory &copy; 2017 Canonical Ltd.</p>
+            <p>&copy; {{ now.year }} Canonical Ltd. Ubuntu and Canonical are registered trademarks of Canonical Ltd.</p>
```
nottrobin: Please change this to 2017, as per my previous reply

bartaz:
@nottrobin OK. Done

PR #187, opened 2017-12-13
Say language or framework, not just language (covers ROS, Electron, etc)

evandandrea:
screen shot 2017-12-13 at 14 31 13

Suggestion from @jamiebennett. Say "language or framework" instead of "language" as we have guides for Electron, MOOS, ROS, and ROS 2.

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -168,8 +168,8 @@
     <div class="p-strip is-deep is-bordered">
       <div class="row">
         <div class="col-12">
-          <h2>Snapcraft integrates with your language</h2>
-          <p class="p-heading--five">See how Snapcraft works with your language to make packaging and installing your software easy.</p>
+          <h2>Snapcraft integrates with your tools</h2>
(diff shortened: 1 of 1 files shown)


bartaz:
@Greggless @matthewpaulthomas Could you have a look if that's the wording we want to use here on the home page?

matthewpaulthomas:
Adding “or framework” makes sense to me. However, it does make it more of a problem that the body text repeats almost all of the heading. I suggest changing the heading to something like “Snapcraft integrates with your tools” or even “Snapcraft integrates with your stuff”, leaving the body text to explain what kind of stuff we’re referring to.

(I imagine that the list of languages and frameworks might eventually be joined by CI systems, and/or IDEs, integrating in different ways with each.)

Greggless:
@bartaz +1 to MPT's comment. 

I vote for:
> ## Snapcraft integrates with your tools
> See how Snapcraft works with your language or framework to make packaging and installing your software easy.

bartaz:
@Greggless Updated as per your suggestion.

Could you +1 this so we can merge?

Greggless:
Done 🚀

PR #191, opened 2017-12-19
Rollup js

Lukewh:
This pull request implements basic rollupjs transpilation for metrics.

The reasoning behind this approach is to namespace and transpile only parts that are 'app-like'. Metrics is a good example of this as it's only required on one page and is fairly self contained, but will have multiple 'views' (graphs) that can share quite a bit of code.


package.json
```diff
@@ -3,7 +3,8 @@
     "test": "sass-lint static/**/*.scss --verbose --no-exit",
     "build": "yarn run build-js && yarn run build-css",
     "build-css": "node-sass --include-path node_modules static/sass --output static/css && postcss --use autoprefixer --replace 'static/css/**/*.css' && postcss --use cssnano --dir static/minified 'static/css/**/*.css'",
-    "build-js": "cp node_modules/d3/build/d3.min.js static/js/modules && cp node_modules/d3-geo/build/d3-geo.min.js static/js/modules && cp node_modules/topojson-client/dist/topojson-client.min.js static/js/modules && cp node_modules/billboard.js/dist/billboard.min.js static/js/modules && cp node_modules/moment/min/moment.min.js static/js/modules",
+    "build-js": "cp node_modules/d3/build/d3.min.js static/js/modules && cp node_modules/d3-geo/build/d3-geo.min.js static/js/modules && cp node_modules/topojson-client/dist/topojson-client.min.js static/js/modules && cp node_modules/billboard.js/dist/billboard.min.js static/js/modules && cp node_modules/moment/min/moment.min.js static/js/modules && yarn run build-metrics",
```
bartaz: Maybe I would split it for now into separate tasks for: "copy third party" "build metrics" and one that combines them all.

static/js/libs/mouse.js
```diff
@@ -0,0 +1,18 @@
+class Mouse {
+  constructor() {
+    this.position = {x: 0, y: 0};
+
+    window.addEventListener('mousemove', this.updatePosition.bind(this));
```
bartaz: Not sure if it part of this PR, but do we need a global mouse object that always listens to mouse move on whole window? Instead of some event handling inside particular module that needs this?

static/js/metrics/axis.js
```diff
@@ -0,0 +1,92 @@
+var X_TICK_FREQUENCY = 7;
```
bartaz: Not sure if rewriting this needs to be part of this PR or later improvements, but these could be `const`.

static/js/metrics/metrics.js
```diff
@@ -0,0 +1,31 @@
+import installsMetrics from './graphs/installs';
+
+/**
+ * Render all metrics
+ * @param {Object} metrics An object of metrics from the API.
+ */
+function renderMetrics(metrics) {
+  if (!d3 || ! bb) {
```
bartaz: This space between `!` and `bb` is haunting me 👻

static/js/metrics/axis.js
Lukewh: Yeah I think these need moving in to the config too, will address in later PR

static/js/libs/mouse.js
Lukewh: All the graphs will use the mouse to position the tooltips, one event handler is more efficient than multiple - iirc... I could be wrong?

package.json
Lukewh: Done

Lukewh:
Thanks @bartaz

PR #193, opened 2017-12-19
Adds a channel badge to snap details page.

### LINKED ISSUE #145: Add channels info to snap details page

bartaz:
Fixes #145

Adds channel badge to snap details page.

### QA

- ./run
- go to snap details page of a snap in beta or candidate
- badge saying 'Beta' or 'Preview' should appear

Some examples (these may stop working if new revisions are released to channels by publishers):
Example stable snap (no badge): http://snapcraft.io-pr-193.run.demo.haus/phpstorm/
Example 'beta' snap: http://snapcraft.io-pr-193.run.demo.haus/webstorm/
Example 'preview' snap: http://snapcraft.io-pr-193.run.demo.haus/brave/

screen shot 2017-12-19 at 14 01 35
screen shot 2017-12-19 at 14 00 37

### DIFF
--- a/app.py
+++ b/app.py
@@ -47,7 +47,7 @@ request_timeout = 2
 # Request only stable snaps
 snap_details_url = (
     "https://api.snapcraft.io/api/v1/snaps/details/{snap_name}"
-    "?channel=stable"
+    "?channel="
 )
 details_query_headers = {
     'X-Ubuntu-Series': '16',
@@ -338,6 +338,7 @@ def snap_details(snap_name):
         'icon_url': details['icon_url'],
         'version': details['version'],
         'revision': details['revision'],
+        'channel': details['channel'],
         'license': details['license'],
         'publisher': details['publisher'],
         'screenshot_urls': details['screenshot_urls'],
--- a/static/sass/_snapcraft_channel.scss
+++ b/static/sass/_snapcraft_channel.scss
@@ -1,0 +1,33 @@
+@mixin snapcraft-channel {
+
+  %p-snap-channel {
+    position: relative;
+
+    &::after {
+      background: $color-snapcraft-light;
+      border-radius: 2px;
+      font-size: 1rem;
+      left: 100%;
+      margin-left: $sp-x-small;
+      padding: $sp-xx-small $sp-x-small;
+      position: absolute;
+      top: 0;
+    }
+  }
+
+  .p-snap-channel--beta {
+    @extend %p-snap-channel;
+
+    &::after {
+      content: 'Beta';
+    }
+  }
+
+  .p-snap-channel--candidate {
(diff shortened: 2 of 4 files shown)


app.py
```diff
@@ -47,7 +47,7 @@
 # Request only stable snaps
 snap_details_url = (
     "https://api.snapcraft.io/api/v1/snaps/details/{snap_name}"
-    "?channel=stable"
+    "?channel="
```
bartaz: @nottrobin @tbille 

I added this `?channel=` for testing purposes, so that we can show snap details page for snaps that are not in stable.

But I don't think that works as intended by us. It simply shows the latest revision of the snap, so if the latest revision is in stable, it will be in stable, if in beta it will be beta, etc.
I believe our intended behaviour is to show stable snap details (if they are available) even if they are not latest revision. If there are no revisions in stable, we should try to show beta if available, etc.
As per spec: https://docs.google.com/document/d/1p0FOzR9boI35RmsCKPOn0HNRY_kbFAS-zXqpPLh2Ot4/edit#heading=h.9m3zk1dpntv

I don't know how to do that with current details API, as it only allows to ask for a specific channel (defaults to stable) or ask for latest revision from any channel (using `channel=`).

Is that something we can implement on our side, or does it need to be changed in API?

cprov: @bartaz, so far we continue to search/browse only over snaps currently released to stable, eventually that restriction will be lifted on the API (see https://forum.snapcraft.io/t/showing-edge-and-beta-in-search-results/4004). Access the snap-details API with a channel only makes sense if users provide the channel, then the 'yes or no' answer is fine (`snap refresh foo --edge`).

What I think we need in the public snap page (fwiw, in publisher page as well) is the `channel_maps_list` field from the snap-details API, which is what `snap info foo` uses:

```
curl -s -H 'X-Ubuntu-Series: 16' 'https://api.snapcraft.io/api/v1/snaps/details/surl?fields=channel_maps_list' | jq
{
  "channel_maps_list": [
    {
      "architecture": "amd64",
      "map": [
        {
          "binary_filesize": 9560064,
          "channel": "stable",
          "confinement": "strict",
          "epoch": "0",
          "info": "released",
          "revision": 96,
          "version": "0.4"
        },
        {
          "channel": "candidate",
          "info": null
        },
        {
          "channel": "beta",
          "info": null
        },
        {
          "channel": "edge",
          "info": null
        }
      ],
      "track": "latest"
    }
  ],
  "package_name": "surl"
}
```
Since it's not a default field, I think you have to start to specify all fields you need from the details API explicitly.
Also not this field is architecture-specific atm, it will default to `amd64` in the absence of `X-Ubuntu-Architecture` header.

Lukewh:
This is on hold until further discussion and agreement.

PR #195, opened 2017-12-19
Add promoted snaps endpoint

tbille:
# Summary

- Add promoted snaps endpoint

# QA 

- Checkout to this branch
- `./run`
- Go to http://0.0.0.0:8004/snaps/
- See promoted snaps ;)

# Problem...
I don't know how to center a title with vanilla...

### DIFF
--- a/app.py
+++ b/app.py
@@ -80,6 +80,15 @@ search_query_headers = {
     'Accept': 'application/hal+json'
 }
 
+promoted_query_url = (
+    "https://api.snapcraft.io/api/v1/snaps/search"
+    "?promoted=true"
+    "&confinement=strict,classic"
+)
+promoted_query_headers = {
+    'X-Ubuntu-Series': '16'
+}
+
 
 def redirect_to_login():
     return flask.redirect(''.join([
@@ -108,6 +117,15 @@ def get_featured_snaps():
     )
 
     return get_searched_snaps(featured_response.json())
+
+
+def get_promoted_snaps():
+    promoted_response = _get_from_cache(
+        promoted_query_url,
+        headers=promoted_query_headers
+    )
+
+    return get_searched_snaps(promoted_response.json())
 
 
 # Error handlers
@@ -294,6 +312,14 @@ def discover():
     return flask.render_template(
         'discover.html',
         featured_snaps=get_featured_snaps()
+    )
+
+
+@app.route('/snaps/')
+def snaps():
+    return flask.render_template(
+        'promoted.html',
+        snaps=get_promoted_snaps()
     )
 
 
--- a/templates/promoted.html
+++ b/templates/promoted.html
@@ -1,0 +1,33 @@
+{% extends "_layout.html" %}
+
+{% block title %}
+  Promoted Snaps — Linux software in the Snap Store
+{% endblock %}
+
+{% block content %}
+  <section class="p-strip is-shallow">
+      <div class="row">
+        {% for snap in snaps %}
+          <div class="col-3">
+            <div class="p-heading-icon">
+              <a class="p-heading-icon__header" href="/{{ snap.package_name }}">
+                {% if snap.icon_url %}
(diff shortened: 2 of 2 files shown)


templates/promoted.html
```diff
@@ -0,0 +1,38 @@
+{% extends "_layout.html" %}
+
+{% block title %}
+  Promoted Snaps — Linux software in the Snap Store
+{% endblock %}
+
+{% block content %}
+  <section class="p-strip is-shallow">
+      <div class="col-12">
```
bartaz: There should be a `<div class="row">` around that col.

templates/promoted.html
```diff
@@ -0,0 +1,38 @@
...
+          <h3>Featured snaps</h3>
```
bartaz: These are not featured snaps. Do we need a title here?

app.py
```diff
@@ -204,6 +222,14 @@ def discover():
     )
 
 
+@app.route('/promoted/')
```
bartaz: Is there meant to be this separate page for those, or should 'promoted' snaps be displayed on some other existing page (search, discover?)? @Greggless?

templates/promoted.html
Greggless: We won't need a title

app.py
Greggless: On phase 2 of /discover (https://github.com/CanonicalLtd/snapcraft-design/issues/197) we link to a page using `see more snaps`. This is the core of that page, a listing of all 'promoted snaps. 

It needs some design love, mainly including a search in the hero. @tbille put this together quickly last night (thanks!)

templates/promoted.html
```diff
@@ -0,0 +1,40 @@
...
+
+{% block title %}
+  Promoted Snaps — Linux software in the Snap Store
+{% endblock %}
+
+{% block content %}
+  <section class="p-strip is-shallow">
+      <div class="row">
+          <div class="col-12">
+              <h1>Promoted snaps</h1>
```
bartaz: I think @Greggless mentioned we don't need title at all on this page.

templates/promoted.html
```diff
@@ -0,0 +1,40 @@
+{% extends "_layout.html" %}
+
+{% block title %}
+  Promoted Snaps — Linux software in the Snap Store
```
bartaz: I'm not sure - is 'promoted snaps' official way how they are called? Is that the title we want for that page?

nottrobin:
@tbille is this linked to an issue? I'm trying to track down where the decision to call this "promoted" came from, to work out who I should be bring up this point to:

I raised an issues with that wording, that "promoted" is semantically very similar to "featured" and will almost certainly confuse people both internally and externally. I'd prefer an adjective like "approved" or something, but maybe for the URL endpoint we could use something entirely neutral like simply "/snaps/".

Greggless:
I believe it had been specified in the [sections and promoted](https://docs.google.com/document/d/1J5a0W92RbuKeGNrYmaTbzXqbE0NpQ8e0xnJG6D4MLVA/edit) documentation but I don't know its true origin.

I agree with @nottrobin  that these shouldn't be called 'promoted' snaps, and have no objection to using 'approved'.

Having the URL as `https://snapcraft.io/snaps` makes sense to me. It'd look something like [this](https://docs.google.com/document/d/138XzmkXhjuS7pVSR0geEpgQbUWOrsHLFsh9hrr6zm_E/edit) until we decide to implement ~~categories~~ sections, filters and sorting as specified in **Phase 3** of the [phased rollout of /discover](https://docs.google.com/a/canonical.com/spreadsheets/d/1Mn9WfpsHA5KfrM_c4VuBpKD44khIph0Q_FPS70cY45M/edit?usp=sharing)

app.py
nottrobin: This should be changed to `/snaps/`.

Greggless:
# MERGEDAT  🚀 :shipit:

We will use this endpoint when we come to complete phase 2 of /discover plan

PR #197, opened 2017-12-20
active devices, eslint and code refactor

Lukewh:
## Summary

- Added active devices graph to measure page
- Added eslint to yarn test
- Refactored some of the es6 code

implements design from https://github.com/CanonicalLtd/snapcraft-design/issues/116

## QA

- Checkout the branch
- `./run test` - make sure the tests work
- `./run serve`
- http://0.0.0.0:8004/snaps/dwarf-fortress/measure or http://snapcraft.io-pr-197.run.demo.haus/snaps/dwarf-fortress/measure
- Check both graphs work as expected


Greggless:
## What to do when version = 0
### The issue
With the current implementation, 0 results are displayed as a `version row` on the `tooltip`, meaning if there are 10 versions in the timeframe of the graph, and 7 have 0 results, we'll still display them in the `tooltip`, when a user may only care about versions with active devices.

Is it bad for us to assume a user will understand if the version is not in the `tooltip`, that it does, in fact, have no active devices?

### The proposal 💍 
- The vertical position of the `tooltip` should remain constant
- Each later `version row` should be added to the bottom of the existing 'version rows'
- The tooltip should extend vertically downwards as the number of `version rows` increase

@Lukewh pointed out that the position of the tooltip is determined by the height of the mouse, this means that as the `tooltip` will extend, and the mouse may cover the data points... 

@matthewpaulthomas @lyubomir-popov it'd be great of have your noggins on this one

kwm14:
## Visual review

- [ ] Y-axis should have a maximum number relative to total as seen on **Installs** chart
- [ ] X-axis add 1rem padding indent to both left and right side of date scale
- [ ] Add 40px padding to bottom of chart

anthonydillon:
Considering this is the first pass and some of it is dummy code to be removed.

carlaberkers:
Impressed Luke, great first start!
- we should look at the positioning of the label in the area on the daily installs graph
- especially on the daily installs graph there should be a hover on the day. 
at the moment it looks like you are hovering an area that spans multiple days but actually you are seeing info about a specific day.

Lukewh:
@Greggless:
I've added 'No data' for days that have no data and removed any point from a tooltip that is 0.

@kwm14 
I've updated the space below the graph. I can't find a reasonable way to round up to the nearest visible labeled tick, the spacing and tick count is generated from within the graphing library (so sometimes it does display correctly). I tried a few different approaches and all of them created other odd behaviours. Would you consider this a blocker?

lyubomir-popov:
A couple of comments:

1) @Lukewh would it be possible to include an <hr> underneath the graph heading on all graphs? 
image
2) Can the is-shallow class be applied to all graph containers? Currently the active devices graph is too close to the bottomof the page:
image
3) (Scope creep) Since we won't have dual axis graphs at this stage, could we make the graphs span the 12th column (which was originally reserved for a second axis on graphs that have one).

static/js/metrics/graphs/activeDevices.js
```diff
@@ -0,0 +1,162 @@
...
+import { snapcraftGraphTooltip, positionTooltip } from '../tooltips';
+import { COLORS } from '../config';
+
+function showGraph(el) {
+  formatAxis(el);
+  el.style.opacity = 1;
+}
+
+function findStart(data) {
+  for (let i = 0, ii = data.length; i < ii; i += 1) {
```
tbille: `for (let i = 0; i < data.length; i += 1) {`

Lukewh: the ii caches the `.length` so it doesn't need to be checked each iteration.

Lukewh:
@lyubomir-popov can you take another look at http://snapcraft.io-pr-197.run.demo.haus/snaps/dwarf-fortress/measure when you get chance please?

> 1. Would it be possible to include an hr underneath the graph heading on all graphs?

Added

> 2. Can the is-shallow class be applied to all graph containers? Currently the active devices graph is too close to the bottomof the page:

Should be done

> 3. (Scope creep) Since we won't have dual axis graphs at this stage, could we make the graphs span the 12th column (which was originally reserved for a second axis on graphs that have one).

Should be done

> 4. I managed to resize the screen to a position where labels on the active devices graph are at the edge, as opposed to in the "center of mass" of the respective area (see screenshot above). Is there any room for adjustment there? And can we keep the labels visible at smaller resolutions?

I've had to implement a rudimentary function for positioning these, I've modified it a bit so hopefully it should work a little better, let me know what you think.

lyubomir-popov:
@Lukewh 

1 & 2 look good, 3 is still an issue but not a blocker.

One more thing - strong needs to be inside a .p-heading--* for the font-weight to cascade properly. Otherwise the headinc class overwrites it.
image

Lukewh:
@lyubomir-popov updated!

Greggless:
**Daily active devices graph** looks great :+1: from UX

templates/publisher/measure.html
```diff
@@ -73,14 +77,18 @@ <h3 class="u-float--left">Daily active devices</h3>
         <div class="col-6">
           <div class="u-clearfix">
             <h3 class="u-float--left">Territories</h3>
-            <strong class="p-heading--three u-float--right u-no-margin--top">74</strong>
+            <div class="p-heading--three u-float--right u-no-margin--top">
```
lyubomir-popov: @Lukewh I don't see this when I reload the demo, so might be wrong, but when I move the p-heading-three class onto the div, the max width limits it and the number moves. What I meant was, wrap the strong in a <p> -
 <p class="p-heading--three"><strong>{{number}}</strong></p>

image

Please ignore this if you've found another way to get it to look correctly.

lyubomir-popov:
@Lukewh I've given it a plus one, please see my comment on the commit (I can't see the update in the demo yet)

templates/publisher/measure.html
Lukewh: I can't reproduce this @lyubomir-popov I'm going to merge and if we notice it while I'm working on the other graphs we can revisit!

Lukewh:
Thanks all!

PR #202, opened 2017-12-20
Measure territories

Lukewh:
Depends on https://github.com/canonical-websites/snapcraft.io/pull/197 landing first.

## Summary
- Added territories graph to publisher metrics
- Converted public territories graph to es6 module and added a build step for public js
- Moved snap-metrics build to publisher build, so we now have dist/public.js and dist/publisher.js.

## QA
- Checkout the branch
- ./run serve
- http://0.0.0.0:8004/snaps/dwarf-fortress/measure or http://snapcraft.io-pr-202.run.demo.haus/snaps/dwarf-fortress/measure
- Check all graphs work correctly
- Ensure the map show's "x daily active users" in tooltips
- Go to http://0.0.0.0:8004/dwarf-fortress or http://snapcraft.io-pr-202.run.demo.haus/dwarf-fortress
- Ensure the map still works as before and the map does not show "x daily active users" in tooltips


Greggless:
## UX review
1. No separator line underneath graph title
2. Colour block on screen tip should be center aligned as in [the design](https://user-images.githubusercontent.com/2741678/31540497-be0f5126-b003-11e7-9d2c-d18392f29917.png)
3. Countries are labelled slighly oddly _I.E._ `Congo, Democratic Republic Of The` Is this correct?
come on lets do the congo

4. USA is a darker hue, with a fewer daily active users than other countries, is this a bug or just dummy data?
### Reference
- [Snap details - owner facing - visual design issue #1](https://github.com/CanonicalLtd/snapcraft-design/issues/67)
- [Snap details - owner facing - visual design issue #2](https://github.com/CanonicalLtd/snapcraft-design/issues/116)
- [Snap details - owner facing - measure design](https://user-images.githubusercontent.com/2741678/31540497-be0f5126-b003-11e7-9d2c-d18392f29917.png)
- [Publisher service](https://docs.google.com/document/d/1PQzECKRrVZ-y3W7DcdkfeZipBVr5Xl2mpMj5Vqhux2E/edit#heading=h.2j98marqeb32)

lyubomir-popov:
@Lukewh The number next to territories in the screenshot above (2428) seems way too high. Is it the number of total users? It sits next to territories, so I thought it would have an upper limit the number of countries/territories in the world?

All numbers should be bold - 
image

How can I make the owner facing tooltip appear? I se the public facing one - 
image

Greggless:
+1 Total numbers of territories the snap is used in

Lukewh:
Thanks @Greggless @lyubomir-popov I'll update the total to be 'active territories'

> 1. There should be a separator line underneath graph title

There should be. Will add.

> 2. Colour block on screen tip should be center aligned as in the design

I completely missed that, thanks.

> 3. Countries are labelled slighly oddly I.E. Congo, Democratic Republic Of The Is this correct?

This list comes from https://pypi.python.org/pypi/pycountry/

> 4. USA is a darker hue, with a fewer daily active users than other countries, is this a bug or just dummy data?

Yeah the dummy data is complete lies and nothing matches up (because it's all generated separately).

Lukewh:
@lyubomir-popov I've updated the header total to be the total number of territories that have active users, I've also ensured that on a country that has 0 active users the tooltip displays '0 daily active users' instead of showing the default, public tooltip.

lyubomir-popov:
@Lukewh can territories have the same is-shallow class? all graphs should be equally spaced. It is currently too close to the footer.
image

Greggless:
@Lukewh Is there any way to see the country's colour reflect the data? Would be good to get a real feel for it. Other than that it's a 👍 :shipit: from me

static/js/public/snap-details/map.js
```diff
@@ -0,0 +1,121 @@
+/* global d3, topojson */
+
+export default function renderMap(el, snapData, isPublic) {
```
bartaz: I'm not a big fan of bool params hanging in the end, as it makes it hard to deduce their meaning from the calls:
`snapcraft.public.map('#js-snap-map', {{ countries|tojson }}, true);` - what is `true`?

Also in this case we are interested in the opposite of `isPublic` - even in the code below we are checking `if (!isPublic)`, so it seems that `isAuthenticated` or `isPublisher` would make more sense.

So a simpler suggestion is to change to `isPublisher` that defaults to false/undefined and has to be set to true if user can see more data in the tooltip.

On the other hand - if the user is not supposed to see the user counts they should not be in data, so we don't really need `isPublic` or `isPublisher`, because it's the data itself that will define what to show in the tooltip.

static/js/public/snap-details/map.js
```diff
@@ -0,0 +1,121 @@
...
+            countrySnapData.percentage_of_users
+          )}`);
+
+          tooltip
+            .style('top', pos[1] + 'px')
+            .style('left', pos[0] + 'px')
+            .style('display', 'block');
+
+          let content = ['<span class="u-no-margin--top">', countrySnapData.name];
+          if (!isPublic) {
```
bartaz: As per comment above, we could possibly not need this flag and check if `number_of_users` exists in data and if not show simplified (public) tooltip, otherwise show the data for publisher in the tooltip.
If the data for public users contains `number_of_users` it's a bug, as we shouldn't expose it in any form.

bartaz:
Map seems very close to the footer
screen shot 2018-01-04 at 13 50 06

static/js/public/snap-details/map.js
Lukewh: Yeah I had it like that before (using the data) however @lyubomir-popov brought up that not showing the 'x daily active users' made him think it was the public map, so I added this bool to ensure that '0 daily active users' is displayed... I also had it that way round initially, being a `isPublisher` but I swapped it for a reason I don't remember.... Will switch it back though as I agree.

Lukewh:
@bartaz addressed your comments, can you take another look please?

Lukewh:
@greggless the data for the heat is currently from the 'public' data but the numbers shown are made up (just to have something in the tooltips to work with). I'm going to look at getting real data in to the graphs soon - at that point it'll be much clearer!

static/js/public/snap-details/map.js
bartaz: @Lukewh Thanks for update.

Still, I think that existence of `number_of_users` (even if it's zero) can trigger showing publisher tooltip while nonexistence of it (when it's undefined, not 0) will show public tooltip.

Lukewh: Yeah I thought this after doing the last commit. Will update again, thanks @bartaz 💟

Lukewh: pushed @bartaz

PR #212, opened 2018-01-03
Adds social media meta tags for nicer sharing.

bartaz:
Fixes https://github.com/canonicalltd/snappy-design-squad/issues/313

Adds Open Graph meta tags to snapcraft.io pages.

### QA

- go to snap detail page (http://snapcraft.io-pr-212.run.demo.haus/brave/)
- view source
- additional meta tags should be there
- try sharing demo page on facebook - nice preview should appear
  
screen shot 2018-01-03 at 15 18 48
screen shot 2018-01-03 at 15 14 32

### DIFF
--- a/templates/_layout.html
+++ b/templates/_layout.html
@@ -5,6 +5,14 @@
     <meta charset="UTF-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    {% block meta %}
+      <meta property="og:title" content="{{ self.title() }}"/>
+      <meta property="og:site_name" content="Snapcraft"/>
+      <meta property="og:type" content="website"/>
+      <meta property="og:description" content="Snaps are containerised software packages that are simple to create and install. They auto-update and are safe to run. And because they bundle their dependencies, they work on all major Linux systems without modification."/>
+      <meta property="twitter:site" content="@snapcraftio" />
+    {% endblock %}
+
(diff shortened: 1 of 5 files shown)


carlaberkers:
great! 

is it possible to have some more padding around the icon?
if yes, @lyubomir-popov what do you think?

Lukewh:
Do we care about Twitter? They use different tags `<meta name="twitter:....">` I believe

lyubomir-popov:
@carlaberkers @bartaz Do we control this layout? 
If we do, I would leave the same amount the text card is padded by. Additionally, I would make the icon no bigger than 48px in either direction and centered.

Lukewh:
@carlaberkers the only way to add more padding is to add more padding to the image files - facebook etc. just take an image url and display it how they want.

Lukewh:
@lyubomir-popov https://github.com/canonical-websites/snapcraft.io/pull/212#issuecomment-355033403 applies to your question too.

Facebook just take raw url and text and display it how they like

bartaz:
@carlaberkers @lyubomir-popov We don't control the layout, only provide the content (image, text). So to add more padding we would need a special padded version of the image just for Facebook - which is hard to expect for snap icons.

@Lukewh As far as I understand Twitter fallbacks to Open Graph tags if specific twitter ones are not provided: https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup. But if that wouldn't work we could duplicate them for twitter.

PR #219, opened 2018-01-08
Add market endpoint

tbille:
# Summary

Add market endpoint with basic html page

# QA

- `./run`
- If you are logged in: http://0.0.0.0:8004/logout
- http://0.0.0.0:8004/account/snaps/INSERT A SNAP YOU HAVE ACCESS TO/market
- Should redirect to login
- There is a bug in the openid library we use so the next url is /account
- http://0.0.0.0:8004/account/snaps/INSERT A SNAP YOU HAVE ACCESS TO/market
- You should see a JSON with the metadata of you snap

### DIFF
--- a/app.py
+++ b/app.py
@@ -89,6 +89,29 @@ promoted_query_headers = {
     'X-Ubuntu-Series': '16'
 }
 
+metadata_query_url = (
+    "https://dashboard.snapcraft.io/dev/api"
+    "/snaps/{snap_id}/metadata"
+)
+
+
+def get_metadata_snap(snap_id):
+    authorization = authentication.get_authorization_header(
+        flask.session['macaroon_root'],
+        flask.session['macaroon_discharge']
+    )
+
+    headers = {
+        'Authorization': authorization
+    }
+
+    metadata_response = _get_from_cache(
+        metadata_query_url.format(snap_id=snap_id),
+        headers=headers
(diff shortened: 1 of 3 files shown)


app.py
```diff
@@ -738,3 +761,32 @@ def publisher_snap(snap_name):
         'publisher/measure.html',
         **context
     )
+
+
+@app.route('/account/snaps/<snap_name>/market', methods=['GET'])
```
Lukewh: final slash?

authentication.py
```diff
@@ -55,7 +55,7 @@ def request_macaroon():
     response = requests.request(
         url='https://dashboard.snapcraft.io/dev/api/acl/',
         method='POST',
-        json={'permissions': ['package_access']},
+        json={'permissions': ['package_access', 'package_upload']},
```
Lukewh: should we add this list as a param of request_macaroon at some point in the future?

tbille: yes I think so but for now, as we said we can keep it like this and find some good practice on how to use permissions with macaroons...
Let's not mix everything in this PR. I created an issue for it: https://github.com/canonical-websites/snapcraft.io/issues/220

Lukewh: xox

Lukewh:
go5it @tbille

PR #221, opened 2018-01-08
Add POST endpoint to update snap metadata

tbille:
# Summary

Update metadata for a snap

# QA 

- `./run`
- http://0.0.0.0:8004/account/snaps/SNAP YOU CAN MODIFY/market/
- Update data
- Apply
- Make sure that the data as been modified on the page

### DIFF
--- a/app.py
+++ b/app.py
@@ -111,10 +111,11 @@ def get_authorization_header():
     }
 
 
-def get_snap_metadata(snap_id):
+def snap_metadata(snap_id, json=None):
     metadata_response = _get_from_cache(
         metadata_query_url.format(snap_id=snap_id),
-        headers=get_authorization_header()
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -784,10 +785,41 @@ def market_snap(snap_name):
...
+    ]
+
+    body_json = {
+        key: flask.request.form[key]
+        for key in whitelist if key in flask.request.form
+    }
+
+    metadata_snap(flask.request.form['snap_id'], body_json)
+
+    return flask.render_template(
```
bartaz: This may look a bit weird when posting to a form returns back to the same page. But let's see how it works and improve from there.

tbille: yep for now I added this because I didn't really where to redirect...

PR #223, opened 2018-01-08
Add realease enpoint for publishers

tbille:
# Summary

Added backend endpoint to access the release page with a basic template

# QA 

- `./run`
- `http://0.0.0.0:8004/account/snaps/SNAP YOU OWN/release/`
- See data about the releases

### DIFF
--- a/app.py
+++ b/app.py
@@ -94,23 +94,39 @@ metadata_query_url = (
     "/snaps/{snap_id}/metadata"
 )
 
+status_query_url = (
+    "https://dashboard.snapcraft.io/dev/api"
+    "/snaps/{snap_id}/status"
+)
 
-def get_metadata_snap(snap_id):
+
+def get_authorization_header():
     authorization = authentication.get_authorization_header(
         flask.session['macaroon_root'],
         flask.session['macaroon_discharge']
     )
 
-    headers = {
+    return {
         'Authorization': authorization
     }
 
+
+def get_snap_metadata(snap_id):
     metadata_response = _get_from_cache(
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -94,25 +94,41 @@
...
         metadata_query_url.format(snap_id=snap_id),
-        headers=headers
+        headers=get_authorization_header()
     )
 
     return metadata_response.json()
 
 
+def get_status_snap(snap_id):
+    status_response = _get_from_cache(
```
Lukewh: does `_get_from_cache` actually get it from a cache? We should be careful with some of this stuff as it will need to be live and up to date! If not, it seems badly named!

WillMoggridge: It is using [requests-cache](https://pypi.python.org/pypi/requests-cache) which says is ignores cache headers. So it is a good idea to be careful with this. Though it only seem to cache content for 5 seconds currently (https://github.com/tbille/snapcraft.io/blob/4046872cde64cd60c7dff1b545cd97c628c7c4a8/app.py#L52).

Caching may need to be revisited soon to allow more flexibility.

app.py
```diff
@@ -94,25 +94,41 @@
...
     metadata_response = _get_from_cache(
         metadata_query_url.format(snap_id=snap_id),
-        headers=headers
+        headers=get_authorization_header()
     )
 
     return metadata_response.json()
 
 
+def get_status_snap(snap_id):
```
WillMoggridge: Can these functions be named more like `get_snap_status`?

PR #225, opened 2018-01-09
Measure staging data

Lukewh:
- http://snapcraft.io-pr-225.run.demo.haus/login
- Open up a snap with data


app.py
```diff
@@ -38,7 +38,7 @@
     status_forcelist=[500, 502, 503, 504]
 )
 uncached_session.mount(
-    'https://api.snapcraft.io',
+    'https://api.stagin.snapcraft.io',
```
WillMoggridge: Looks like a typo for `staging` here.

tbille: this code should be rebased with the last version of s.io 
The URLs are handled in a different way now

Lukewh: This is not the MR you're looking for

PR #227, opened 2018-01-10
Improve colors map

tbille:
# Summary

Based on the mail discussion. The colors are now generated: the country which has the biggest percentage of users has the darkest color, and the rest of thecountries color adapts from the biggest country.
screenshot-2018-1-10 atom linux software in the snap store

# QA

- `./run`
- http://0.0.0.0:8004/atom/
- Admire the beautiful use of colors

### DIFF
--- a/app.py
+++ b/app.py
@@ -479,27 +479,67 @@ def snap_details(snap_name):
 
     # Normalise geodata from API
     users_by_country = {}
+    max_users = 0.0
     for country_percentages in geodata:
         country_code = country_percentages['name']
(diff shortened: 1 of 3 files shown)


tbille:
@matthewpaulthomas Here instead of having 10 colors (one every 10%) the colors are generated between a certain range. This means that we could have a color per country between this range. Is ok like this?

tbille:
@Lukewh Bug with tooltip fixed

matthewpaulthomas:
Wot, my thumbs-up emoji wasn’t enough?

PR #230, opened 2018-01-11
Small updates to account page layout

bartaz:
Fixes #230

Small updates to account page layout.
We don't have much more to show there ATM.

screen shot 2018-01-11 at 14 37 45

### DIFF
--- a/app.py
+++ b/app.py
@@ -309,6 +309,7 @@ def get_account():
     user_snaps = response.json()
     return flask.render_template(
         'account.html',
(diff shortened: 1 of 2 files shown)


Lukewh:
@bartaz should the col wit the icon be a 1-col? That gap seems pretty wide?

bartaz:
@Lukewh With 1 col the seem visually too close. And Market page shows icon in 2 cols as well, so I used that as an example.

But hopefully this will get updated when properly designed.

PR #233, opened 2018-01-11
Measure changes from data

Lukewh:
## QA

- Goto http://snapcraft.io-pr-233.run.demo.haus/login.
- Visit a snap with data ask around if you don't have access to one.
 /account/snaps/<snap-name>/measure/

This is not perfect and still requires work, but I think it would be good to get it landed as-is before I start really ~breaking~ fixing things.


static/js/publisher/metrics/metrics.js
```diff
@@ -13,32 +13,41 @@ function renderMetrics(metrics) {
...
+
+  activeDevices.buckets = activeDevices.buckets.map(bucket => {
+    return moment(bucket);
   });
+  activeDevices.buckets.unshift('x');
 
-  activeDevicesMetrics(days, activeDevices);
+  activeDevicesMetrics(
+    activeDevices.buckets,
+    activeDevices.series);
```
tbille: new line before `)`

app.py
```diff
@@ -580,152 +589,133 @@ def publisher_snap(snap_name):
...
+                "metric_name": "installed_base_by_country",
+                "snap_id": details['snap_id'],
+                "start": start.strftime('%Y-%m-%d'),
+                "end": end.strftime('%Y-%m-%d')
+            }
+        ]
+    }
+
+    authed_metrics_headers = publisher_metrics_query_headers.copy()
+    auth_header = authentication.get_authorization_header(
```
tbille: I wrote a function to do this:
`get_authorization_header()['Autorization']`

static/js/publisher/metrics/colorScale.js
```diff
@@ -0,0 +1,23 @@
+export default function(steps) {
+  const start = [ 255, 232, 200 ];
+  const end = [ 226, 74, 51 ];
```
tbille: If you wanted the same colors as the map:
```
maxColor = rgb(8, 64, 129)
minColor = rgb(29, 245, 223)
```

Lukewh: this is for the activeDevices, but good to know!

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -44,50 +45,58 @@ function getHighest(data) {
...
     let _points = points.slice(0);
     const version = _points.shift();
-    labelPosition(el, version, _points, index === 0, index === activeDevices.length - 1);
+    labelPosition(
+      el,
+      version,
+      _points,
+      activeDevicesMetrics.data(),
+      index === 0,
+      index === activeDevices.length - 1);
```
tbille: new line before `)`

static/js/publisher/metrics/metrics.js
```diff
@@ -13,32 +13,42 @@ function renderMetrics(metrics) {
...
 
-  installsMetrics(days, installs);
+  installs.buckets = installs.buckets.map(bucket => {
+    return moment(bucket);
+  });
+  installs.buckets.unshift('x');
+
+  installsMetrics(
+    installs.buckets,
+    installs.values);
```
tbille: )

app.py
```diff
@@ -194,6 +199,36 @@ def get_snap_id(snap_name):
     return details_response.json()['snap_id']
 
 
+def calculate_color(thisCountry, maxCountry, maxColor, minColor):
+    countryFactor = float(thisCountry)/maxCountry
+    colorRange = maxColor - minColor
+    return int(colorRange*countryFactor+minColor)
+
+
+def calculate_colors(countries, max_users):
```
bartaz: There is so much different stuff in app.py already. We should split it up into smaller managable pieces at some point.

tbille: I agree with you, but not in this PR I am planning on doing a refactor of the file next week

app.py
```diff
@@ -658,16 +668,20 @@ def _get_from_cache(url, headers, json=None, method=None):
 
 # Publisher views
 # ===
-@app.route('/snaps/<snap_name>/measure')
-def publisher_snap_measure(snap_name):
+@app.route('/account/snaps/<snap_name>/measure/')
+def publisher_snap(snap_name):
```
bartaz: We are not very consistent in naming these handlers.
As this is a measure page not 'snap page', name could be more specific.

On the other hand for market page we have separate get_ and post_ handlers which makes naming even more inconsistent...

app.py
```diff
@@ -194,6 +199,36 @@ def get_snap_id(snap_name):
...
+
+
+def calculate_colors(countries, max_users):
+    for country_code in countries:
+        countries[country_code]['color_rgb'] = [
+            calculate_color(
+                countries[country_code]['percentage_of_users'],
+                max_users,
+                8,
+                229
```
bartaz: What are these magic numbers, are these colour definitions?

app.py
Lukewh: This specific case was my bad, have fixed it. But yeah we need to figure out how we're going to make the monolith of a file more manageable.

app.py
Lukewh: I believe so - @tbille?

In this PR I just moved the code so it could be used twice.

app.py
```diff
@@ -686,155 +700,152 @@ def publisher_snap_measure(snap_name):
...
         'snap_title': details['title'],
         'package_name': details['package_name'],
-        'metric_period': metric_period_int,
+        'metric_period': metric_period,
 
         # Metrics data
-        'installs_total': sum(installs_values),
-        'installs_metrics': installs_metrics,
-        'active_devices_total': active_devices_total,
+        'installs_total': "{:,}".format(installs_total),
```
bartaz: What is "{:,}" doing?

app.py
tbille: yes here we define the max and min colors
Same here I don't like that we do this in the backend. We need to have a discussion about what should be send to the frontend...

static/js/publisher/metrics/colorScale.js
bartaz: It's not ideal that we hardcode colours in the code.
If we can't move these into CSS because of functionality maybe we can later have some file were we define colour variables/constants or something...

app.py
Lukewh: converts 1000 to 1,000 or 10000 to 10,000 @bartaz

static/js/publisher/metrics/colorScale.js
Lukewh: I have a config file already, will move them there

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -28,66 +29,48 @@ function findEnd(data) {
...
 
   let label = document.querySelector(`[data-version="${labelText}"]`);
   if (!label) {
     label = document.createElementNS('http://www.w3.org/2000/svg', 'text');
+    label.setAttribute('class', 'snapcraft-metrics__active-devices-label');
     label.setAttribute('data-version', labelText);
     label.setAttribute('stroke', '#000');
-    label.setAttribute('x', highestLeft + leftPadding);
-    label.setAttribute('y', bbox.y + (bbox.height / 2));
     const labelSpan = document.createElementNS('http://www.w3.org/2000/svg', 'tspan');
```
bartaz: Oh boy, it's a pain that we need to be that low level...
But I guess we need to manually create SVG parts here, right?

Lukewh: Yeah we could add *ANOTHER* library, but it seemed pointless for such a small part of the project.

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -28,66 +29,48 @@ function findEnd(data) {
...
+  label.setAttribute('y', bbox.y + (bbox.height / 2));
 };
 
-const positionLabels = function(el, activeDevices) {
+const positionLabels = function(el, activeDevices, activeDevicesMetrics) {
   activeDevices.forEach((points, index) => {
     let _points = points.slice(0);
     const version = _points.shift();
-    labelPosition(el, version, _points, index === 0, index === activeDevices.length - 1);
+    labelPosition(
```
bartaz: It seems that `labelPosition` was changed and now only takes 3 parameters, not 6.

Lukewh: It's good to have a buffer just incase.... I'll remove them.

package.json
```diff
@@ -5,10 +5,9 @@
...
     "build": "yarn run build-js && yarn run build-css",
     "build-css": "node-sass --include-path node_modules static/sass --output static/css && postcss --use autoprefixer --replace 'static/css/**/*.css' && postcss --use cssnano --dir static/minified 'static/css/**/*.css'",
-    "build-js": "yarn run copy-3rd-party-js && yarn run build-publisher && yarn run build-public",
+    "build-js": "yarn run copy-3rd-party-js && yarn run build-js-transpile",
     "copy-3rd-party-js": "cp node_modules/d3/build/d3.min.js static/js/modules && cp node_modules/d3-geo/build/d3-geo.min.js static/js/modules && cp node_modules/topojson-client/dist/topojson-client.min.js static/js/modules && cp node_modules/billboard.js/dist/billboard.min.js static/js/modules && cp node_modules/moment/min/moment.min.js static/js/modules",
-    "build-publisher": "node_modules/.bin/rollup static/js/publisher/publisher.js --o static/js/dist/publisher.js -f iife -n snapcraft.publisher",
-    "build-public": "node_modules/.bin/rollup static/js/public/public.js --o static/js/dist/public.js -f iife -n snapcraft.public",
+    "build-js-transpile": "node_modules/.bin/rollup -c",
```
bartaz: I believe from scripts in package.json simply `"rollout -c"` should work, as it should have installed modules in path for this project.

Lukewh: Yeah it should, but I've had issues with build environments before where they don't properly link though I notice sass-lint is called 'normally' so I'll update. Thanks @bartaz

Lukewh:
Thanks all

PR #236, opened 2018-01-12
Fixup templates

Lukewh:
## QA

- Visit http://0.0.0.0:8004/account or http://snapcraft.io-pr-236.run.demo.haus/account
- Click on a snap - you should go to the market page
- Click on the measure tab - you should go to the measure page
- Click on the market tab - you should return to the market page
- Click on My snaps - you should return to the account page

Celebrate! 🍾

### DIFF
--- a/app.py
+++ b/app.py
@@ -838,7 +838,7 @@ def publisher_snap_measure(snap_name):
 
     context = {
         # Data direct from details API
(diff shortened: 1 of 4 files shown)


Lukewh:
Thanks @tbille 🎊

PR #238, opened 2018-01-16
Attach a GA event to snap Install button

caldav:
As discussed during the GA mentoring session last month, this fires a GA event when the install button is triggered on snap details pages.

## QA

1. Open a snap page such as  http://snapcraft.io-pr-238.run.demo.haus/scummvm/
2. Click on the **Install** button
3. Ensure the event has been properly logged by going to the real-time/events view of the "0) Device - DevSites" GA property
4. Click on the most recent "Snap details" link in the "Event Category" col, ensure the "Event label" col contains the name of the snap you tried to install


templates/snap-details.html
```diff
@@ -28,7 +28,7 @@ <h1>{{ snap_title }}</h1>
 
             <p class="p-media-object__content--large">
               {% if is_linux %}
-                <a class="p-button--positive" href="snap://{{ package_name }}">Install</a>
+                <a class="p-button--positive" href="snap://{{ package_name }}" onclick="ga('gtm1.send', {hitType: 'event', eventCategory: 'Snap details', eventAction: 'Clicked Install button', eventLabel: 'Install {{ package_name }}'})">Install</a>
```
bartaz: When we have more places where we would send custom events it would be good to abstract it somehow to avoid inline click events, but it's fine for now.

PR #239, opened 2018-01-16
Add ability to dev to update URL from env var.

tbille:
# Summary

Now developpers can use other urls (for example stagings urls) by adding them in the .env file.

# QA

- `./run`
- http://0.0.0.0:8004/search?q=no
- http://0.0.0.0:8004/account
- http://0.0.0.0:8004/logout
- http://0.0.0.0:8004/discover
- All those pages should work as usual
- Now stop the run script
- In the .`.env` file add those lines:
```
SNAPCRAFT_IO_API=https://api.staging.snapcraft.io/api/v1/
DASHBOARD_API=https://dashboard.staging.snapcraft.io/dev/api/
SEARCH_API=https://search.apps.staging.ubuntu.com/api/v1/
LOGIN_URL=https://login.staging.ubuntu.com
```
- `./run`
- http://0.0.0.0:8004/search?q=no
- http://0.0.0.0:8004/account
- http://0.0.0.0:8004/logout
- http://0.0.0.0:8004/discover
- You should have access to all those pages, but the data displayed should be the data from staging.


app.py
```diff
@@ -32,6 +32,23 @@
 app.secret_key = os.environ['SECRET_KEY']
 app.wtf_csrf_secret_key = os.environ['WTF_CSRF_SECRET_KEY']
 
+SNAPCRAFT_IO_API = os.environ.get(
```
WillMoggridge: I do have a preference to using `os.getenv()`, which is specifically created for this use.

app.py
```diff
@@ -59,54 +76,75 @@
 request_timeout = 2
 
 # Request only stable snaps
-snap_details_url = (
-    "https://api.snapcraft.io/api/v1/snaps/details/{snap_name}"
-    "?channel=stable"
-)
+snap_details_url = ''.join([
```
WillMoggridge: There are a mix of lowercase constant variables and uppercase. I think we should go with just uppercase.
Also there are two sections of URL variables separated by the request session block. Can these URL blocks be put next to each other?

tbille:
@WillMoggridge 

> Also there are two sections of URL variables separated by the request session block. Can these URL blocks be put next to each other?

No because the block of request needs the URLs from the env files and the block of URLs definitions are strongly linked with the functions right under.
This is one of the reason this file needs to be refactored...

PR #241, opened 2018-01-17
UI for editing icon on market page

bartaz:
Fixes #248

Adds a UI for editing icon on snap market page.

NOTE: there is currently no backend API done for saving the icon, so it wont actually change when saved

UPDATE: icon editing code is there but is disabled so we can merge it without making functionality available yet

### QA

- ./run
- go to market page of a snap you own
- click on an icon to edit it
- choose different image to replace icon with
- icon should update on the page

### DIFF
--- a/templates/publisher/market.html
+++ b/templates/publisher/market.html
@@ -9,7 +9,7 @@
     <section class="p-strip is-shallow">
       <div class="row">
         <p><a href="/account">My snaps&nbsp;&rsaquo;</a></p>
-        <h1 class="u-no-margin--top">{{ snap_title }}</h1>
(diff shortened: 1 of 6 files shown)


templates/publisher/market.html
```diff
@@ -29,6 +29,8 @@ <h1 class="u-no-margin--top">{{ snap_title }}</h1>
       <form method="POST">
```
tbille: can you add this to the form
` enctype="multipart/form-data"`

bartaz:
Not merging for now, because we don't clarification on how to save the icon with API.

bartaz:
UPDATE: icon editing code is there but is disabled so we can merge it without making functionality available yet

PR #243, opened 2018-01-17
Refactor redundant API calls

tbille:
# REFACTOTO
## Summary

Step by step refactor: Refactor redundant code

## QA

- `./run`
- http://0.0.0.0:8004/account/snaps/speed-test/market/
- http://0.0.0.0:8004/speed-test/
- http://0.0.0.0:8004/search?q=web&limit=2&offset=2
- OK

### DIFF
--- a/app.py
+++ b/app.py
@@ -163,6 +163,25 @@ def get_authorization_header():
     }
 
 
+def get_snap_details(snap_name):
+    details_response = _get_from_cache(
+        SNAP_DETAILS_URL.format(snap_name=snap_name),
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -661,8 +656,6 @@ def external(attrs, new=False):
         'countries': country_data,
 
         # Context info
-        'details_api_error': details_response.old_data_from_error,
-        'metrics_api_error': metrics_response.old_data_from_error,
```
tbille: I removed those lines because:
1. we don't actually use them
2. We handle errors in another way (see details function for example)

PR #244, opened 2018-01-17
Refactor metrics

tbille:
# Refectoto

## Summary

Refactor metrics functions

## QA

- `./run`
- http://0.0.0.0:8004/account/snaps/speed-test/measure/
- http://0.0.0.0:8004/speed-test/
- The map is still displayed

### DIFF
--- a/app.py
+++ b/app.py
@@ -252,6 +252,16 @@ def normalize_searched_snaps(search_results):
     )
 
 
+def get_public_metrics(snap_name, json):
+    metrics_response = _get_from_cache(
+        SNAP_METRICS_URL.format(snap_name=snap_name),
+        headers=METRICS_QUERY_HEADERS,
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -557,53 +581,18 @@ def snap_details(snap_name):
...
-            'name': country.name,
-            'code': country.alpha_2,
-            'percentage_of_users': percentage_of_users,
-            'color_rgb': color_rgb
-        }
+    users_by_country = normalize_metrics(
+        metrics_response[0]['series']
+    )
+    country_data = build_country_info(
+        users_by_country,
```
bartaz: Not sure about our python code style, but these calls seem unnecessarily split into multiple lines as they would easily fit into one line as they don't take many arguments.

PR #251, opened 2018-01-22
Use flask-openid stateless version

tbille:
# Summary
The library used (Flask-OpenId) is not stateless. We are going to use the fork that fixes this problem: 
https://github.com/nward/flask-openid

This should be removed if the PR is merged in the future:
https://github.com/mitsuhiko/flask-openid/pull/50

# QA
- `./run`
- http://0.0.0.0:8004/account
- Make sure you can still login and access your account page

### DIFF
--- a/app.py
+++ b/app.py
@@ -65,7 +65,8 @@ uncached_session.mount(
 )
 
 oid = OpenID(
-    app,
+    app=app,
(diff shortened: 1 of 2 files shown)


Lukewh:
QA good too 👍

PR #253, opened 2018-01-23
Access private snaps on market and measure pages

tbille:
# Summary

Access private snaps for publisher

# QA

- `./run`
- http://0.0.0.0:8004/account/snaps/PRIVATE_SNAP/measure/
- http://0.0.0.0:8004/account/snaps/PRIVATE_SNAP/market/
- http://0.0.0.0:8004/account/snaps/SNAP/measure/
- http://0.0.0.0:8004/account/snaps/SNAP/market/
- Access a non uploaded snap:
- http://0.0.0.0:8004/account all non uploaded snaps shouldn't have a link to the market/measure page
- http://0.0.0.0:8004/account/snaps/NOT_UPLOADED_SNAP/measure/
- Should display page with an error message
- http://0.0.0.0:8004/account/snaps/NOT_UPLOADED_SNAP/market/
- Should display page with error message, and all fields in readonly mode
- If you force HTML to send a request, the request shouldn't be proccessed

### DIFF
--- a/templates/publisher/market.html
+++ b/templates/publisher/market.html
@@ -12,6 +12,7 @@
         <h1 class="u-no-margin--top">{{ title }}</h1>
       </div>
     </section>
+
     <section class="metrics__nav-wrap p-strip is-shallow u-no-padding--top u-no-padding--bottom">
       <nav class="p-tabs">
         <ul class="p-tabs__list" role="tablist">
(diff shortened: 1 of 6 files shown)


tbille:
Endpoint to get the details for all snaps (public, private, ...) https://dashboard.snapcraft.io/docs/api/snap.html#obtaining-information-about-a-snap

bartaz:
### QA

1. accessing snap that is public but in edge channel - doesn't show additional details (version, etc)
screen shot 2018-02-01 at 16 53 55

2. accessing 'snaps' that were just registered names (no snap uploaded) is possible showing empty form - in dashboard they are not editable at all...

screen shot 2018-02-01 at 16 54 13

tbille:
@bartaz 
1. We removed Version and size because this depends of the architecture and the channel (you might have to hard refresh your page, or relaunch the run script)
2. This is a pb from the API we need to see this with the snap store team I guess...

cprov:
@tbille and @Lukewh planned API changes deployed in production.

Summing up, `account-info` now provides `snaps['16']['<name>']['uploaded']` flag indicating wether there is data to be presented on `snap-info` or not, also `snaps['16']['<name>']['icon_url']` when it is set (null otherwise).

The other relevant change is that reviewers and store_viewers (read-only) have permission to access the APIs, i.e. they will be able to access https://snapcraft.io/account/snaps/<name>/{market,measure} assuming they know the snap name (for now)

bartaz:
@tbille I think some divs are not closed properly or something, because layout got messed up a bit:

screen shot 2018-02-07 at 17 04 09

templates/publisher/market.html
```diff
@@ -118,23 +130,6 @@ <h1 class="u-no-margin--top">{{ title }}</h1>
...
-          </div>
-        </div>
-        <div class="row">
-          <div class="col-2">
-            <label>Size: </label>
-          </div>
-          <div class="col-8">
-            {{ details.filesize }}
-          </div>
         </div>
```
bartaz: You forgot to remove this closing </div>

templates/publisher/market.html
```diff
@@ -11,7 +11,19 @@
         <p><a href="/account">My snaps&nbsp;&rsaquo;</a></p>
         <h1 class="u-no-margin--top">{{ title }}</h1>
       </div>
+      {% if not uploaded %}
```
bartaz: Not a big thing, but I think it would look better to have error notification below tab navigation, because it says about content of particular tab.

Same goes to same notification on measure page.

On the other hand maybe there should be no tabs or no content at all if snap is not uploaded? Maybe it should be 404? But that's probably a question to design how these pages should look for not uploaded snaps.

tbille: Done

tbille:
New views:

# Measure
53a156b5-e3d4-4338-b677-3c36d73162ab

# Market
screenshot-2018-2-8 market details for

PR #256, opened 2018-01-23
 Post screenshots and icons for market page

tbille:
# Summary

Post screenshots and icons for market page

/!\ To avoid conflicts with dashboard for now I need reupload every screenshots/icons.... This will need to be optimized

# QA

- `./run`
- http://0.0.0.0:8004/account/snaps/SNAP/market/
- Update the icon

### DIFF
--- a/app.py
+++ b/app.py
@@ -961,12 +961,19 @@ def get_market_snap(snap_name):
         get_snap_details(snap_name)
     )
 
+    context = {
+        "snap_id": snap_id,
+        "snap_name": snap_name,
+        "title": metadata['title'],
+        "summary": metadata['summary'],
+        "description": metadata['description'],
+        "license": metadata['license'],
+        "details": details
+    }
+
     return flask.render_template(
         'publisher/market.html',
-        snap_id=snap_id,
-        snap_name=snap_name,
-        **metadata,
-        details=details
+        **context
     )
 
 
@@ -1046,14 +1053,21 @@ def post_market_snap(snap_name):
                 get_snap_details(snap_name)
             )
 
+            context = {
+                "snap_id": flask.request.form['snap_id'],
+                "snap_name": snap_name,
+                "title": metadata['title'],
+                "summary": metadata['summary'],
+                "description": metadata['description'],
+                "license": metadata['license'],
+                "details": details,
+                "screenshots": screenshots_response,
+                "error_list": error_list
+            }
+
             return flask.render_template(
                 'publisher/market.html',
-                snap_id=flask.request.form['snap_id'],
-                snap_name=snap_name,
-                **metadata,
-                details=details,
-                screenshots=screenshots_response,
-                error_list=error_list
+                **context
             )
 
     flask.flash("Changes applied successfully.", 'positive')
--- a/app.py
+++ b/app.py
@@ -5,6 +5,7 @@ The web frontend for the snap store.
 """
 
 import authentication
+import hashlib
 import flask
 import requests
 import requests_cache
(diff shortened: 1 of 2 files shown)


bartaz:
Not a problem with this PR, but we need to start showing errors (which we don't right now I think), because for some images (too big I guess) it doesn't work and doesn't show any info why.

app.py
```diff
@@ -939,38 +982,81 @@ def snap_release(snap_name):
...
+
         body_json = {
             key: flask.request.form[key]
             for key in whitelist if key in flask.request.form
         }
 
         metadata = snap_metadata(flask.request.form['snap_id'], body_json)
-
         if 'error_list' in metadata:
+            error_list.append(
```
bartaz: I'm not sure if that's correct.
It seem to append an array of errors INTO an array, making error_list contain another array instead of list of errors, like:

`[[{'message': 'Image must be exactly 256x256 pixels in size.', 'extra': {'name': 'icon'}, 'code': 'invalid-field'}]]`

app.py
```diff
@@ -939,38 +982,81 @@ def snap_release(snap_name):
...
+        if icon is not None:
+            info.append(build_image_info(icon, 'icon'))
+            files.append(icon)
+
+        screenshots = flask.request.files.getlist('screenshots')
+        for screenshot in screenshots:
+            info.append(build_image_info(screenshot, 'screenshot'))
+            files.append(screenshot)
+
+        if not files:
```
nottrobin: Maybe it would be better to consistently use the word "screenshot" in related variable names? 'cos we're probably going to be dealing with a few different types of data in this method. E.g. `screenshot_files`, `screenshots_json`. Or, maybe it would make sense to encapsulate all the screenshot work into its own function (`_get_screenshots` or something) - in that case the shorter contextual variable names would be fine.

app.py
```diff
@@ -166,6 +168,17 @@ def get_authorization_header():
     }
 
 
+def transform_api_data(details):
```
nottrobin: If this function is only used inside this file, could we prefix it with an underscore: `_tranform_api_data`?

app.py
```diff
@@ -185,11 +198,40 @@ def get_snap_details(snap_name):
     return details_response.json()
 
 
-def snap_screenshots(snap_id, json=None):
+def build_image_info(image, image_type):
+    hasher = hashlib.sha256(image.read())
```
nottrobin: Would you mind adding brief docstrings to all the functions you create here?

app.py
```diff
@@ -939,38 +982,81 @@ def snap_release(snap_name):
...
+            files.append(screenshot)
+
+        if not files:
+            # API requires a multipart request, but we have no files to push
+            # https://github.com/requests/requests/issues/1081
+            files = {'info': ('', dumps(info))}
+        else:
+            images_json = {'info': dumps(info)}
+
+        screenshots_list = snap_screenshots(
```
nottrobin: Personally I'd call the variable simply `screenshots` - I think it's pretty implicitly obvious that if there are multiple screenshots it's going to be a list. It makes things like `for screenshot in screenshots:` a bit more poetic.

app.py
```diff
@@ -939,38 +982,81 @@ def snap_release(snap_name):
...
+                get_snap_details(snap_name)
+            )
+
             return flask.render_template(
                 'publisher/market.html',
                 snap_id=flask.request.form['snap_id'],
                 snap_name=snap_name,
-                metadata=flask.request.form,
-                error_list=metadata['error_list']
+                **metadata,
```
nottrobin: I'm not a fan of implicitly passing attributes through directly from the API. I think the back-end should offer the front-end with certainty about what data it's providing, and raise and handle any errors itself if data is missing. In other words the back-end should be a solid reliable layer in and of itself If you implicitly pass data through then subtle changes in the API data can much more easily cause errors on the frontend because at some point it started relying on data it was never supposed to use.

So I'd rather we grab and explicitly pass through any metadata that the front-end needs. And only what it needs.

PR #260, opened 2018-01-24
Adds flash notifications when market page form is saved or reverted

bartaz:
First part of https://github.com/CanonicalLtd/snapcraft-design/issues/249

Adds flash notification message when market page form is saved or reverted.
It currently doesn't check if there were any changes in the form. Just shows a message based on which button was used.

Also, follow up PR will be added later to remove the notification after some delay and when changes are made in the form.

### QA
- ./run
- go to market page of a snap you own
- click Apply or Revert
- proper message should appear
- edit form, message should be removed
- click Apply or Revert
- message should appear
- wait 20 seconds, message should be removed

screen shot 2018-01-24 at 16 39 32
screen shot 2018-01-24 at 16 39 23

### DIFF
--- a/templates/publisher/market.html
+++ b/templates/publisher/market.html
@@ -26,7 +26,7 @@
     </section>
 
     <div class="p-strip is-shallow">
-      <form method="POST" enctype='multipart/form-data'>
+      <form id="market-form" method="POST" enctype='multipart/form-data'>
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
         <input type="hidden" name="snap_id" value="{{ snap_id }}" />
         <input type="file" name="icon" id="snap_icon" hidden class="hidden" accept="image/*"/>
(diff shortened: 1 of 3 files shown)


templates/publisher/market.html
```diff
@@ -33,11 +33,28 @@ <h1 class="u-no-margin--top">{{ title }}</h1>
 
         <div class="row">
           <div class="col-12 u-align--right">
-              <a href="/account/snaps/{{ snap_name }}/market" class="p-button--neutral">Revert</a>
-              <input type="submit" class="p-button--positive" value="Apply" />
+              <input type="submit" class="p-button--neutral" name="submit_revert" value="Revert" />
```
bartaz: We don't really keep any state on client side to revert to, so to cancel changes we need to 'post' but ignore the data and just show flash message that changes were reverted.

Lukewh:
This good to merge @bartaz ?

bartaz:
@Lukewh @tbille If you can have a second look, because in the meantime I added a script to remove the notification on delay or form change to this PR as well.

PR #264, opened 2018-01-25
Measure territories and remove installs

Lukewh:
## Changes

- Installs graph removed
- Measure territories grey for lowest shade
- Territories changed to weekly metric

## QA

- Pull the branch
- `./run`
- Click on a snap with metrics data
- The top 'Installs' graph should not longer be there
- The 'Territories' map should have grey as the lowest colour scale and the tooltips should show exact numbers

### DIFF
--- a/app.py
+++ b/app.py
@@ -804,12 +804,12 @@ def build_country_info(users_by_country, display_number_users=False):
         country_info = users_by_country.get(country.alpha_2)
         number_of_users = 0
         percentage_of_users = 0
-        color_rgb = [229, 245, 223]
+        color_rgb = [247, 247, 247]
         if country_info is not None:
             if display_number_users:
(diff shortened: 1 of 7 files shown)


app.py
```diff
@@ -840,62 +840,47 @@ def publisher_snap_measure(snap_name):
     metric_bucket = ''.join([i for i in metric_period if not i.isdigit()])
     metric_period_int = int(metric_period[:-1])
 
+    installed_base_metric = flask.request.args.get(
+        'active-devices',
+        default='version',
+        type=str)
```
tbille: )

Lukewh: done

tbille:
Same problem on this PR I can't see the map on my screen

Lukewh:
@tbille the map problem we'll have to fix in a future branch I think

PR #268, opened 2018-01-29
Create cache module

tbille:
# Refactoto S01E03

## Summary

Create module for cache, this is an helper file for the futures api modules we will have in the project.
There is no modification in the logic of the functions, just code moved.

## QA 
- `./run`
- http://0.0.0.0:8004/
- http://0.0.0.0:8004/atom
- http://0.0.0.0:8004/store
- http://0.0.0.0:8004/account
- http://0.0.0.0:8004/account/snaps/SNAP/market
- http://0.0.0.0:8004/account/snaps/SNAP/metrics
- ...
- Try access pages that make API calls and see if it still works


src/cache.py
```diff
@@ -0,0 +1,73 @@
...
+)
+
+# The cache expires after 5 seconds
+cached_session = requests_cache.CachedSession(expire_after=5)
+
+# Requests should timeout after 2 seconds in total
+request_timeout = 2
+
+
+def get_from_cache(url, headers, json=None, method=None):
```
nottrobin: how about simply `def get(...):`? So you can do:

``` python
import cache

cache.get(url, headers)
```

bartaz:
Not sure if that's necessarily for this PR, but now that we have `src` folder, shouldn't `app.py` be there as well?

nottrobin:
Oh yeah I agree with @bartaz - if there's a `src/` directory, I would expect it to contain all the source code. But I don't think that's what we want here - `app.py` should be at the top level. So if you really want to keep `cache.py` in its own folder, I would put it in `lib/` or something. *but* personally I'd just keep it at the top level for now, and only put it in a subfolder if we get to the point where we have so many python files that it's getting confusing - and at that point we could hopefully start to create subfolders with meaningful names (like "models" or something).

tbille:
Maybe the folder should be called `modules` instead. Because this file will also have the modules `public` and `publisher`.

PR #269, opened 2018-01-29
Routing fixes

nottrobin:
- Tidy code slightly
- Fix redirection error
- Correct URL paths for URLs without slashes

QA
--

``` bash
./run
```

Then navigate to `/account` and `/account/`, with and without being already logged in. Check everything still works, and that `/account/` is now *not* a 404.

### DIFF
--- a/app.py
+++ b/app.py
@@ -34,6 +34,7 @@ app = flask.Flask(__name__)
 app.wsgi_app = ProxyFix(app.wsgi_app)
 app.secret_key = os.environ['SECRET_KEY']
 app.wtf_csrf_secret_key = os.environ['WTF_CSRF_SECRET_KEY']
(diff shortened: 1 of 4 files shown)


app.py
```diff
@@ -155,6 +156,20 @@
...
 
 
+@app.before_request
+def clear_trailing():
+    """
+    Remove trailing slashes from all routes
+    We like our URLs without slashes
+    """
+
+    from flask import redirect, request
```
tbille: like we did for the rest of the app, use like this:
```
flask.request ...
flask.redirect ...
```

nottrobin:
@tbille I've fixed the tests and added a whole load more tests. Now almost every endpoint is tested - except the ones that require authentication to reach, which is much more tricky to test.

PR #276, opened 2018-01-31
Refactor all public views on its own module

tbille:
# Refactoto S01E04

## Summary 
Refactor all public views on its own module

## QA

- `./run`
- Test all pubic and publishers pages and see if they still work :)

### DIFF
--- a/app.py
+++ b/app.py
@@ -4,26 +4,24 @@ A Flask application for snapcraft.io.
 The web frontend for the snap store.
 """
 
-import bleach
 import datetime
 import flask
 import hashlib
 import humanize
 import modules.cache as cache
 import modules.authentication as authentication
+import modules.public.views as public_views
+import modules.public.api as public_api
 import os
-import pycountry
-import re
 import socket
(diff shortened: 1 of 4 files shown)


app.py
```diff
@@ -502,75 +356,16 @@ def create_redirect():
...
-        links_result['self'] = convert_limit_offset_to_size_page(
-            links['self']['href']
-        )
-
-    return links_result
-
-
 # Normal views
 # ===
 @app.route('/')
```
Lukewh: If we're splitting the methods into separate files I feel we should put the route definitions there too... I find it weird to declare the route in one file but implement the behaviour in another. Maybe it's just me?

tbille: After a discussion we had with @nottrobin we decided to keep the routes in the "entrance" file to keep the backend simple. So that if a new developer arrives on the project he can open the `app.py` file and follow the path of the issue in a simple way

Lukewh: I don't like it, but I'm okay with it.

PR #277, opened 2018-02-01
Refactor all publisher views on its own module

tbille:
# Refactoto S01E05

## Summary 
Refactor all publisher views on its own module

## QA

- `./run`
- Test all public and publishers pages and see if they still work :)

### DIFF
--- a/app.py
+++ b/app.py
@@ -4,23 +4,16 @@ A Flask application for snapcraft.io.
 The web frontend for the snap store.
 """
 
-import datetime
 import flask
-import hashlib
-import humanize
-import modules.cache as cache
 import modules.authentication as authentication
 import modules.public.views as public_views
-import modules.public.api as public_api
+import modules.publisher.views as publisher_views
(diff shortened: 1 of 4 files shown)


app.py
```diff
@@ -390,239 +188,31 @@ def snap_details(snap_name):
 
 # Publisher views
 # ===
-@app.route('/account/snaps/<snap_name>/measure')
+@app.route('/account')
```
bartaz: Would it be possible in flask to 'delegate' part of the routing to separate module?

I mean - with this PR even that all the publisher views are in separate module it's `app.py` that defines all URLs for them.
Would it be possible to say: everything in `/account` is handled by publisher module, and then the publisher module will define specific publisher URLs?

tbille: As discussed, for now we are going to keep the route in app to keep the app simple. If we feel that we'll need to move the routes on their own modules we will create a new Refactoto Season 😄

PR #278, opened 2018-02-01
Refactor app.py and account endpoint

tbille:
# Refactoto S01E06 - FINAL

## Summary 
* Reorganise app.py
* Refactor account endpoint

## QA

- `./run`
- Test all public and publishers pages and see if they still work :)
- http://0.0.0.0:8004/account still accessible


bartaz:
@tbille Small thing, but in meta tags for snap details page we use URL with / in the end: https://github.com/canonical-websites/snapcraft.io/blob/master/templates/snap-details.html#L7

I guess if we redirect to urls without / we should change that to remove unnecessary redirect.

I'm not sure if there are other places that we have links to URLs with slashes that could be updated...

PR #283, opened 2018-02-05
Market page - adding screenshots (step 1)

bartaz:
Adds basic screenshots UI to market page form.

- Currently it's only possible to add screenshots.
- Because of current backend implementation added screenshots **override** screenshots added previously when saved
- Not possible to move or delete screenshots yet
- No validation on number of screenshots yet

## QA

- ./run
- go to market page of snap you own
- existing screenshots should appear (if you have any)
- add some screenshots
- screenshots should show up
- click "Apply" to save - screenshots should be uploaded and saved in dashboard

screen shot 2018-02-05 at 14 22 00

screen shot 2018-02-05 at 14 22 53

### DIFF
--- a/static/js/publisher/market/market.js
+++ b/static/js/publisher/market/market.js
@@ -13,12 +13,33 @@ function initSnapIconEdit(iconElId, iconInputId) {
 }
 
 function initSnapScreenshotsEdit(screenshotsToolbarElId, screenshotsWrapperElId, data) {
-  const state = {};
-  state.screenshots = data.map((url) => { return { url }; });
-
+  // DOM elements
   const screenshotsToolbarEl = document.getElementById(screenshotsToolbarElId);
   const screenshotsWrapper = document.getElementById(screenshotsWrapperElId);
 
+  // simple state handling (and serializing as JSON in hidden input)
+  const state = {};
+  const stateInput = document.createElement('input');
+  stateInput.type = "hidden";
+  stateInput.name = "state";
+
+  screenshotsToolbarEl.parentNode.appendChild(stateInput);
+
+  const setState = function(nextState) {
+    for (let key in nextState) {
+      if (nextState.hasOwnProperty(key)) {
+        state[key] = nextState[key];
+      }
+    }
+
+    stateInput.value = JSON.stringify(state);
+  };
+
+  setState({
+    screenshots: data.map((url) => { return { url }; })
+  });
+
+  // templates
   const screenshotTpl = (screenshot) => `
     <div class="col-2">
       <img src="${screenshot.url}" alt="" />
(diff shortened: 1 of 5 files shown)


static/js/publisher/market/market.js
```diff
@@ -12,6 +12,91 @@ function initSnapIconEdit(iconElId, iconInputId) {
...
+  const render = () => {
+    renderScreenshots(state.screenshots);
+  };
+
+  render();
+
+  const onScreenshotsChange = function() {
+    const fileList = this.files;
+
+    for (var i = 0; i < fileList.length; i++) {
```
tbille: `for(const file in fileList)`

bartaz: Tried that, but it doesn't work. FileList is not really an array, just and array-like object.

Lukewh:
Removed the code +1 as changes are still being made

static/js/publisher/market/market.js
```diff
@@ -12,6 +12,91 @@ function initSnapIconEdit(iconElId, iconInputId) {
...
+  const renderScreenshots = (screenshots) => {
+    if (screenshots.length) {
+      screenshotsWrapper.innerHTML = screenshots.map(screenshotTpl).join("");
+    } else {
+      screenshotsWrapper.innerHTML = emptyTpl();
+    }
+  };
+
+  const render = () => {
+    renderScreenshots(state.images.filter(i => i.type === "screenshot"));
```
tbille: remane `i` to `image`

static/js/publisher/market/market.js
```diff
@@ -12,6 +12,91 @@ function initSnapIconEdit(iconElId, iconInputId) {
...
+  const onScreenshotsChange = function() {
+    const fileList = this.files;
+
+    for (let i = 0; i < fileList.length; i++) {
+      const file = fileList[i];
+      setState({
+        images: state.images.concat([{ file, url: URL.createObjectURL(file), name: file.name, type: "screenshot" }])
+      });
+
+      render();
```
tbille: Don't you want to render after the `for loop`?

static/js/publisher/market/market.js
```diff
@@ -12,6 +12,91 @@ function initSnapIconEdit(iconElId, iconInputId) {
...
+
+    stateInput.value = JSON.stringify(state);
+  };
+
+  setState(initialState);
+
+  // templates
+  const screenshotTpl = (screenshot) => `
+    <div class="col-2">
+      <img src="${screenshot.url}" alt="" />
```
tbille: I don't think I am sending you the value but we should add the name in the `alt`:

`alt="${screenshot.filename}"`

static/js/publisher/market/market.js
bartaz: Right, good catch.

static/js/publisher/market/market.js
bartaz: Yep, right now I only have urls from `details`. Maybe screenshots API gives more data.

templates/publisher/market.html
```diff
@@ -178,5 +196,18 @@ <h1 class="u-no-margin--top">{{ title }}</h1>
 <script>
```
tbille: If snap is not uploaded I have an error because the object icon_url is undefined.

One solution could be to add `{% if uploaded %}`  for the script tag?

bartaz: Fixed

PR #299, opened 2018-02-13
Add terms and conditions page

tbille:
# Summary

Add terms and conditions page

# QA
- With a new user
- `./run`
- http://0.0.0.0:8004/account
- http://0.0.0.0:8004/account/agreement
- Accept terms and conditions
- Should be redirected to /account page

# Screenshot
screenshot-2018-2-12 developer program agreement linux software in the snap store

### DIFF
--- a/app.py
+++ b/app.py
@@ -211,6 +211,18 @@ def get_account():
     return publisher_views.get_account()
 
 
+@app.route('/account/agreement')
+@login_required
+def get_agreement():
+    return publisher_views.get_agreement()
+
+
+@app.route('/account/agreement', methods=['POST'])
+@login_required
+def post_agreement():
+    return publisher_views.post_agreement()
+
+
 @app.route('/account/snaps/<snap_name>/measure')
 @login_required
 def publisher_snap_measure(snap_name):
--- a/modules/publisher/api.py
+++ b/modules/publisher/api.py
@@ -22,6 +22,11 @@ ACCOUNT_URL = ''.join([
     'account',
 ])
 
+AGREEMENT_URL = ''.join([
+    DASHBOARD_API,
+    'agreement/'
+])
+
 METADATA_QUERY_URL = ''.join([
     DASHBOARD_API,
     'snaps/{snap_id}/metadata',
@@ -104,6 +109,32 @@ def get_account():
         }
 
     return response.json()
+
+
+def get_agreement():
+    headers = get_authorization_header()
+
+    agreement_response = cache.get(
+        AGREEMENT_URL,
+        headers
+    )
+
+    return agreement_response.json()
+
+
+def post_agreement(agreed):
+    headers = get_authorization_header()
+
+    json = {
+        'latest_tos_accepted': agreed
+    }
+    agreement_response = cache.get(
+        AGREEMENT_URL,
+        headers,
+        json
+    )
+
+    return agreement_response.json()
 
 
 def get_publisher_metrics(json):
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -13,6 +13,15 @@ def get_account():
     if 'redirect' in account:
         return account['redirect']
 
+    error_list = []
+    if 'error_list' in account:
+        for error in account['error_list']:
+            if error['code'] == 'user-not-ready':
+                if 'has not signed agreement' in error['message']:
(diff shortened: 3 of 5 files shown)


lyubomir-popov:
Not sure about the box, usually you see a window with the full terms and conditions. Here it feels like we're trying to hide them behind a click. Not a blocker, just don't think it's ideal. Adding design +1 as I don't think that's a crucial thing and can be improved later.

tbille:
@lyubomir-popov For now I took the exact same page as we have on dashboard.

lyubomir-popov:
@tbille I assumed that's the case, and have added the design +1 label. My comment was just to indicate that there are enhancements I'd like to propose if / when an opportunity arises.

matthewpaulthomas:
I have no opinion on whether it’s best to improve the design in this PR, or leave it for later. But if it’s best to leave it for later, then it would be quicker not to add “Design needed” in the first place.

If it’s best to improve the design in this PR, then:
- [ ] As Lyubo said, show the terms inline.
- [ ] Decide on one overall heading for the terms, rather than two.
- [ ] Remove the punctuation from the end of the checkbox label.
- [ ] Follow our standard commit button layout: “Sign up” rightmost, “Cancel” left of it.

cprov:
@tbille buid.s.io actually depends on the existing (bad) API behavior, see https://github.com/canonical-websites/build.snapcraft.io/blob/master/src/common/actions/auth-store.js#L246

Instead of changing this in a backward-incompatible way (there is not API versioning support atm), we should probably check the error `message` if code is `user-not-ready`. Moreover, we should really be solving both "problems" at once (accepting the agreement and setting username if necessary).

bartaz:
I got an error when doing QA:

My steps:

- go to http://snapcraft.io-pr-299.run.demo.haus/account
- create new account
- validate email
- accept and log in
- when I got redirected from login to http://snapcraft.io-pr-299.run.demo.haus/account I got a KeyError:

screen shot 2018-02-15 at 09 40 05

tbille:
@bartaz this is going to be fixed in this PR https://github.com/canonical-websites/snapcraft.io/pull/304

tbille:
@matthewpaulthomas 
> As Lyubo said, show the terms inline.

- Do you mean diplaying all the terms and conditions in the page instead of redirecting to ubuntu.com? Because this will mean that we will need to update this page all the time the conditions are changed on ubuntu.com...

> Decide on one overall heading for the terms, rather than two.

- I am open to suggestions :smile: 

@cprov I am now doing the same control as we have in b.s.io

@bartaz Error fixed and merge #304

lyubomir-popov:
Would it be possible to add a couple of rems of padding underneath? On pages taller than the window the buttons touch the footer. Also as @matthewpaulthomas said, the buttons need to be right aligned, green one right-most, cancel to the left. 

image

tbille:
@lyubomir-popov can you hard refresh the page because it seems it is due to the cache

lyubomir-popov:
I did, still see it as above. Ideally this should be like that:

screen shot 2018-02-21 at 14 57 26

Lukewh:
@lyubomir-popov @matthewpaulthomas if you check the latest demo are we ok to land this?

lyubomir-popov:
@tbille not sure if it's cache (tried disabling when devtools is open, didn't help) but I still see the u-align-right and col-4 on the same div, leading to the 4-col div not having a margin-left. Could we at least have this sort of markup so the utility class doesn't interfere with the grid classes?
```
<div class="col-4"><div class="u-align--right">
            <a class="p-button--neutral" href="/">Cancel</a>
            <button class="p-button--positive" disabled="disabled" type="submit">Continue</button>
          </div>
</div>
```

that makes it look like this (still too much white space above, but at leaast aligned on the right):

<div class="col-4"><div class="u-align--right">
            <a class="p-button--neutral" href="/">Cancel</a>
            <button class="p-button--positive" disabled="disabled" type="submit">Continue</button>
          </div>
</div>

Lukewh:
Sorry @lyubomir-popov should be sorted now, if you could try again thanks.

matthewpaulthomas:
Good to see that the rest of the layout fixes have survived. The buttons still aren’t exactly in the right place, but that’s not Budapestworthy.

PR #300, opened 2018-02-13
Create username if user doesn't have one on snap store

tbille:
# Summary
- Handle errors when a user is new:
  - Create a snap store username

# QA

- `./run`
- Create new account on login.ubuntu.com
- http://0.0.0.0:8004/account
- you should be redirected to the create username page
- http://0.0.0.0:8004/account/username
- Create username
- http://0.0.0.0:8004/account
- You should land on account page, Namespace should be the username choosen

# Screenshots

## New username
screenshot-2018-3-15 choose username linux software in the snap store 1

### DIFF
--- a/app.py
+++ b/app.py
@@ -247,6 +247,18 @@ def post_agreement():
     return publisher_views.post_agreement()
 
 
+@app.route('/account/username')
+@login_required
+def get_account_name():
+    return publisher_views.get_account_name()
+
+
+@app.route('/account/username', methods=['POST'])
+@login_required
+def post_account_name():
+    return publisher_views.post_account_name()
+
+
 @app.route('/account/snaps/<snap_name>/measure')
 @login_required
 def publisher_snap_measure(snap_name):
--- a/modules/authentication.py
+++ b/modules/authentication.py
@@ -71,7 +71,13 @@ def request_macaroon():
     response = requests.request(
         url=url,
         method='POST',
-        json={'permissions': ['package_access', 'package_upload']},
+        json={
+            'permissions': [
+                'package_access',
+                'package_upload',
+                'edit_account'
+            ]
+        },
         headers={
             'Accept': 'application/json, application/hal+json',
             'Content-Type': 'application/json',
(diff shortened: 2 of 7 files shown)


Greggless:
Great, now just some help text :)
> This name will appear on any snaps you create

matthewpaulthomas:
- [ ] If it’s possible to combine this page with the “Developer Programme Agreement” page, that would save a click and a page load.
  - [ ] If not, then at most only one of the pages should have a “Sign up” button. The other should have “Continue” or similar, so that we don’t imply people are signing up twice.
- [ ] `<input name="username" id="username" type="text" value="{{ username }}">` suggests that there are no restrictions on valid user names. 32769 characters long, emoji, spaces, tabs, line-break characters — are they all fine? That seems … implausible.
- [ ] There’s no hint as to what this name will be used for. Will it be shown on end-user snap pages, for example? That’s important if I was about to call myself, say, LadyKiller1983.
- [ ] How does this interact with the user name shown during Ubuntu One sign-up?

Lukewh:
Updated screenshot:
screenshot from 2018-02-23 16-25-26

matthewpaulthomas:
According to @tbille there is no longer any way to review this, for the same reason that there are no tests for it. Making it testable might be more important than anything design-related.

Based on the screenshot, though, it looks like only the second point from my previous review has been addressed: “have ‘Continue’ or similar”. There’s an attempt at addressing my fourth point — “hint as to what this name will be used for” — but it’s not clear what “appear as the owner” means. Where? In “snap info”?

An example of helpful text might be: (disclaimer: I don’t know whether this is accurate!)
> Lower-case letters, numbers, and dashes only. This ID may appear next to the publisher name on public pages, and on the command line.

Lukewh:
I've made some updates to the design, I'm not going to figure out how to test it.

lyubomir-popov:
@Lukewh just one thing, can we have an hr above the buttons? people were complaining about buttons being too detached; an hr will guide your eyes from the left to the right

Issue #291, opened 2018-02-08
The 'past year' option on the 'measure' should not be shown.


thomir:
On the 'measure' page of the new publisher dashboard. there 'past year' drop-down option probably should not be shown, as we don't have close to one year's worth of data. 

We have data from 2017-10-31.

I can think of a few strategies we could employ here:

1) We could change this to say "For all time", which would show data from 2017-10-31 until we have one year's worth of data, at which point we could change it back to say "Past year".

2) We could remove this option altogether, until we have 6 months of data, at which point we could add the '6 months' option. When we have one year of data we could add the 'one year' option.

Lukewh:
Hey @thomir I've put in a PR to address this. I've removed the 1 year option. Adding a 'For all time' and then switching it back could cause some friction with users, so I think your option 2 is a far better solution - add specific periods when it makes sense.

thomir:
Works for me - can you let me know when it's live please, and I'll do some further testing.

Cheers,

Issue #289, opened 2018-02-08
country map on 'measure' page sometimes does not show up.


thomir:
When viewing the 'measure' page on the new publisher dashboard, sometimes the world map fails to render.

It only happens when the "By OS" option in "Weekly Active devices" is selected, and only for some snaps.

For example, viewing the core snap, this happens:

![image](https://user-images.githubusercontent.com/372783/35949869-942ab738-0cd8-11e8-89c1-f0241fa2a498.png)

On the developer console there's a javascript error:

```
activeDevices.js:48 Uncaught TypeError: Cannot read property 'getBBox' of null
    at labelPosition (activeDevices.js:48)
    at activeDevices.forEach (activeDevices.js:81)
    at Array.forEach (<anonymous>)
    at positionLabels (activeDevices.js:78)
    at activeDevices (activeDevices.js:155)
    at Object.renderMetrics [as metrics] (metrics.js:33)
    at measure:184
```

Selecting "By Version" from the drop-down box causes the map to render again.

Issue #290, opened 2018-02-08
'measure' page OS graph is hard to interpret


thomir:
On the 'measure' page of the new publisher dashboard, the 'By OS' graph is hard to interpret, especially for snaps with a large install base. 

For example, here's the core snap (with numbers removed):

![image](https://user-images.githubusercontent.com/372783/35949994-258ec1ec-0cd9-11e8-8fef-ad58fc0fcbe1.png)

The colours in the graph are too similar to each other, and it becomes impossible to differentiate one OS from another.

Additionally, the tooltip entries appear to be sorted by colour, but I think they should be sorted in the same order as the stacking order on the graph (i.e.- from largest to smallest).

Finally, on our internal metrics we've started grouping non-Ubuntu operating systems by family (essentially stripping the '/<version>' part of the OS name). This has made these graphs much easier to interpret. You may want to consider doing the same.

thomir:
Hi @Lukewh 

I think there's been a regression in the publisher dashboard - now whenever I attempt to load the `measure` page for core (or any other snap I have viewer rights on) I get an `HTTP 500` response.

Are these new systems hooked up to sentry? I couldn't see them in the snap-store sentry, but maybe you have a separate instance somewhere?

Cheers,

lyubomir-popov:
@Lukewh 
How about using the [sequential colorbrewer scales](http://colorbrewer2.org/#type=sequential&scheme=BuGn&n=9) - you get up to 9 colours per OS, 6 distinct scales.

> Additionally, the tooltip entries appear to be sorted by colour, but I think they should be sorted in the same order as the stacking order on the graph (i.e.- from largest to smallest).

The order should definitely match, the question is what makes more sense - sorted by OS version or by relative size. 

1) Sorting OSs by share and then versions within OS by version number seems most straightforward to me - share of OS version is clearly indicated by the surface area, and colours will be banded together by hue. I think the fact that the biggest ones aren't on top is less of an issue than the next alternative bellow.

2) Not grouping versions by OS will allow vertically sorting by size, but will inevitably lead to confusion as with lighter shades you can't always tell which scale a colour belongs to, take for example the lighter shades of these two - 

![image](https://user-images.githubusercontent.com/2741678/36255566-7378b0e8-1247-11e8-92b7-1401ff2df9f5.png)

Lukewh:
@thomir we'll look in to that regression thanks for finding it.

thomir:
Hi,

[...]

PR #303, opened 2018-02-14
Weekly active devices updates

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/291 and https://github.com/canonical-websites/snapcraft.io/issues/290 (to some extent, colours still need work)
should also fix https://github.com/canonical-websites/snapcraft.io/issues/289 (seemes to be JS error, not a map specific error)

# Work done

- Defended against an error when a label isn't present for some reason.
- Removed the 'Year' option for Weekly active devices.
- Combined non-Ubuntu OS versions (will probably need to be revised in the future).
- Sorted Weekly active device data in an attempt to make the data easier to read.
- Ensured axis labels are visible.
- Sorted tooltip data.

# QA

- Pull this branch
- `./run` and visit http://0.0.0.0:8004/account or http://snapcraft.io-pr-303.run.demo.haus/account
- Click through to the market page for a snap (with data)
- Ensure everything looks as expected and the tooltip data is sorted appropriately
- +1 this PR.

# Screenshot

screenshot from 2018-02-14 12-15-04


static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -147,12 +205,23 @@ export default function activeDevices(days, activeDevices) {
...
+      columns: [days].concat(deviceData)
+    },
+    onrendered: () => {
+      // Because we're modifying the axis with CSS the mask crops the labels
+      // We don't need the masks, so here we remove them each render.
+      Array.from(
+        el.querySelectorAll('clipPath')
+      ).forEach(mask => {
+        if (mask.id.indexOf('axis') !== -1) {
+          mask.remove();
```
bartaz: I guess it would be good to define browsers we support. `remove` is not available in IE at all (only in Edge). So are many other new features (arrow functions, const/let, etc) that we use and don't transpile.

Especially that we develop and test only in latest browsers where all of this works, and it will just throw badly when just one feature doesn't work.

Not sure if it's something to fix in this PR, but something we may want to address now that we add more and more JS features.

templates/publisher/measure.html
```diff
@@ -120,7 +119,10 @@ <h3 class="u-float--left">Retention</h3>
   snapcraft.publisher.selector('.metrics-period', 'period');
   snapcraft.publisher.selector('.active-devices', 'active-devices');
   snapcraft.publisher.metrics({
-    activeDevices: {{ active_devices|safe }},
+    activeDevices: {
+      metrics: {{ active_devices|safe }},
```
bartaz: I think there is a `json` or (`tojson`?) filter that may be better to use here instead of `safe`. 
Unless it's HTML that is rendered here.

bartaz:
Not sure if it was desired or not, but it seems bit strange to show label when there is no data:

screen shot 2018-02-15 at 15 10 34

static/js/publisher/metrics/graphs/activeDevices.js
Lukewh: Yeah good point, I've been thinking we should add some transpilation soon, even just to figure out how we want to do it moving forward, and it should help with things like this.

templates/publisher/measure.html
Lukewh: I tried tojson and it didn't seem to work - flask route converts to json before sending it to the template, which I guess is the reason?

bartaz: Ok.

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -86,14 +91,64 @@ const positionLabels = function(el, activeDevices) {
   });
 };
 
-export default function activeDevices(days, activeDevices) {
+export default function activeDevices(days, activeDevices, type) {
   const el = document.getElementById('active_devices');
 
+  activeDevices.sort((a, b) => {
+    const a1 = a.slice(1);
```
bartaz: Why are first items removed from these arrays?
I guess the structure of the data explains it, but it's not clear from the code. Would be useful to leave some comment in code.

PR #312, opened 2018-02-16
snap details: Use common_name if available in map regions

### LINKED ISSUE #310: Politically incorrect region name "Taiwan, Province of China"
The problematic section

The region name "Taiwan, Province of China" is politically incorrect and offending to Taiwanese users, please use the neutral name "Taiwan" instead.

If the website rely on the iso-codes package, please use the "common name" field when available instead as with this [patch](https://launchpadlibrarian.net/53161408/use_common_name_instead_when_it_exists.patch) did with the [Ubuntu LoCo Team Portal](http://loco.ubuntu.com/) site.

REF: [Bug #614236 “Wrong Country Name. (Taiwan)” : Bugs : LoCo Team Portal](https://bugs.launchpad.net/loco-team-portal/+bug/614236)

Thanks in advance!

Lin-Buo-Ren:
Common names are more politically neutral, so use those if possible.

Fixes #310.

Signed-off-by: 林博仁(Buo-Ren Lin) <Buo.Ren.Lin@gmail.com>

### DIFF
--- a/modules/public/views.py
+++ b/modules/public/views.py
@@ -83,8 +83,14 @@ def build_country_info(users_by_country, display_number_users=False):
             percentage_of_users = country_info['percentage_of_users'] or 0
             color_rgb = country_info['color_rgb'] or [247, 247, 247]
 
+        # Use common_name if available to be less political offending (#310)
+        try:
+            country_name = country.common_name
+        except AttributeError:
+            country_name = country.name
+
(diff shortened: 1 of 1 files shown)


Lin-Buo-Ren:
default

The patch is verified to fix the issue without noticing regressions (yet), please review.

NOTE: I don't really speak Python, please really do review.

PR #316, opened 2018-02-16
Basic sentry integration.

cprov:
The production/staging story involves deploying Sentry in k8s (sentry + redis + postgres), but then it's just the `SENTRY_DSN` we use in those environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Once the containers are setup, you can visit <http://127.0.0.1:8004> in your bro
 
 ### Use staging APIs
 
-To use staging APIs locally you can add the following lines to an`.env.local` file:
+To use staging APIs locally you can add the following lines to an `.env.local` file:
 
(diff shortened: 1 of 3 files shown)


WillMoggridge:
In the situation of us not setting the environment variable, will the site run without errors?

I believe we need to do some investigating to get Sentry set up for our stack and putting the Sentry client behind a feature flag seems like a good move.

nottrobin:
@WillMoggridge Do you have any reason to believe it won't run successfully if you don't declare `SENTRY_DSN`? It seems fine to me.

@cprov This looks like a great service, but are you saying that to actually use Sentry we need to deploy redis & postgres alongside the snapcraft.io app in Kubernetes? Because we can't have stateful services in Kubernetes. I don't think we should merge this until we know for sure how we would deploy it into production, as that is where it's actually useful.

PR #318, opened 2018-02-19
Further account updates

Lukewh:
# Done

This pull request satisfies the remaining 'Phase 1' items in https://docs.google.com/document/d/1L6Dj77LGUJnbngyzceZcDrkf8iygx2HNAc471xjRaQE/edit#heading=h.q0rl1q8l4z1o:

1. User details (editable) - Links off to SSO (login.ubuntu.com)
2. Snap list - Registered names vs actual snaps

This should also satisfy https://github.com/CanonicalLtd/snappy-design-squad/issues/358 in it's current state (@Greggless to confirm).

# QA

- Pull this branch
- `./run`
- View your account page:
  - 'Username' should have an 'Edit' button that links to https://login.ubuntu.com
  - Spacing of icon, snap name and 'private' flag should be better
  - The 'Status' column should indicate whether a snap is a registered name.

### DIFF
--- a/app.py
+++ b/app.py
@@ -39,6 +39,12 @@ LOGIN_URL = os.getenv(
 )
 
 
+# Make LOGIN_URL available in all templates
+@app.context_processor
+def inject_login():
+    return dict(LOGIN_URL=LOGIN_URL)
+
+
 def login_required(func):
     """
     Decorator that checks if a user is logged in, and redirects
--- a/templates/account.html
+++ b/templates/account.html
@@ -1,69 +1,83 @@
 {% extends "_layout.html" %}
 
 {% block title %}
-    Account page — Linux software in the Snap Store
+Account page — Linux software in the Snap Store
 {% endblock %}
 
 {% block content %}
-
   <section class="p-strip is-shallow">
-      <div class="row">
-          <div class="col-2">
-              {% if user.image %}
-                  <img class="p-media-object__image--large" src="{{ user.image }}" alt="{{ user.nickname }} account" />
-              {% else %}
-                  <img class="p-media-object__image--large" src="https://assets.ubuntu.com/v1/6fbb3483-snapcraft-default-snap-icon.svg" alt="" />
-              {% endif %}
-          </div>
-          <div class="col-9">
-              <h1 class="u-no-margin">{{ user.nickname }}</h1>
-              <table class="p-table-key-value">
-                  <tr><th width="100">Full name</th><td>{{ user.fullname }}</td></tr>
-                  <tr><th>Username</th><td>{{ user.nickname }}</td></tr>
-                  <tr><th>Email</th><td>{{ user.email }}</td></tr>
-                  <tr><th>Namespace</th><td>{{ namespace }}</td></tr>
-              </table>
-          </div>
+    <div class="row">
+      <div class="col-2">
+        {% if user.image %}
+          <img class="p-media-object__image--large" src="{{ user.image }}" alt="{{ user.nickname }} account" />
+        {% else %}
+          <img class="p-media-object__image--large" src="https://assets.ubuntu.com/v1/6fbb3483-snapcraft-default-snap-icon.svg" alt="" />
+        {% endif %}
       </div>
+      <div class="col-9">
+        <h1 class="u-no-margin">{{ user.nickname }}</h1>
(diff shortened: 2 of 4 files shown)


app.py
```diff
@@ -197,7 +197,7 @@ def snap_details(snap_name):
 @app.route('/account')
 @login_required
 def get_account():
-    return publisher_views.get_account()
+    return publisher_views.get_account(LOGIN_URL)
```
tbille: could you add instead a function like this (see [doc](http://flask.pocoo.org/docs/0.12/templating/#context-processors)):

```python
@app.context_processor
def inject_login():
  return dict(LOGIN_URL=LOGIN_URL)
```

I am pretty sure we will use this URL in other context (login button for example)

templates/account.html
```diff
@@ -19,7 +19,7 @@
               <h1 class="u-no-margin">{{ user.nickname }}</h1>
               <table class="p-table-key-value">
                   <tr><th width="100">Full name</th><td>{{ user.fullname }}</td></tr>
-                  <tr><th>Username</th><td>{{ user.nickname }}</td></tr>
+                  <tr><th>Username</th><td>{{ user.nickname }} <a href="{{ login_url }}" class="p-link--external">Edit</a></td></tr>
```
tbille: This line starts to be a little bit long:

```html
<tr>
  <th>Username</th>
  <td>
    {{ user.nickname }} <a href="{{ login_url }}" class="p-link--external">Edit</a>
  </td>
</tr>
```

templates/account.html
```diff
@@ -43,22 +43,32 @@ <h3>My snaps</h3>
...
-                                    <img src="https://assets.ubuntu.com/v1/6fbb3483-snapcraft-default-snap-icon.svg" width="32" height="32" />&nbsp;
-                                {% endif %}
-                                {% if user_snaps[snap].uploaded %}
-                                    <a href="/account/snaps/{{snap}}/market" class="u-no-margin--top">{{snap}}</a>
+                            <td class="u-vertically-center p-snap-list">
+                              {% if user_snaps[snap].uploaded %}
+                                <a href="/account/snaps/{{snap}}/market" class="u-no-margin--top u-vertically-center">
+                              {% endif %}
+                              {% if user_snaps[snap].icon_url %}
+                                  <img src="{{ user_snaps[snap].icon_url }}" width="32" height="32" class="p-snap-list__image" />
```
tbille: Different indentations in this file...

templates/account.html
```diff
@@ -43,22 +43,32 @@ <h3>My snaps</h3>
...
+                              {% if user_snaps[snap].uploaded %}
+                                </a>
+                              {% endif %}
+                              {% if user_snaps[snap].private %}
+                                  <small class="u-no-margin--top">Private</small>
+                              {% endif %}
+                            </td>
+                            <td>
+                                {% if not user_snaps[snap].uploaded %}
+                                    Name registered
```
tbille: Don't we want a message like: `Snap not uploaded`?

templates/account.html
tbille: You added a scss file, maybe those lines ` width="32" height="32"` could be added in it

templates/account.html
tbille: And question, do we want to leave s.io ? Because we can't come back from the SSO. Maybe we could open on a new tab?

templates/account.html
cprov: Additionally to a better `status` terminology, I would also order the list by their status, then name.  Snaps 'uploaded' would be listed first.
Later on we can introduce new (default) order, possibly by last upload/change or installed_base.

Lukewh: @cprov new issue created here: https://github.com/canonical-websites/snapcraft.io/issues/323 /cc @tbille

app.py
Lukewh: done

templates/account.html
Lukewh: We should avoid opening new tabs etc. where we can, especially if we own the site linked to. New tabs are annoying and anyone that wants to can using normal methods.

Fixed the line length though.

templates/account.html
Lukewh: done

Greggless:
I'm due to start work on this next iteration, so we can do merge what we have now and then give it another pass next week

PR #319, opened 2018-02-19
Testing publisher pages (with authentication).

cprov:
Tentative proposal for testing publisher pages using response doubles.

Minor fix included for login redirects (previously pointing to non-evaluated route, e.g `.../<snap_name>/market` and ending up on `/account` because of the redirect in `verify_response()` ... kinda crazy)

### DIFF
--- a/app.py
+++ b/app.py
@@ -54,7 +54,7 @@ def login_required(func):
 
 
 def redirect_to_login():
-    return flask.redirect('login?next=' + flask.request.url_rule.rule)
+    return flask.redirect('login?next=' + flask.request.path)
 
 
 # Error handlers
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ pymacaroons==0.12.0
 python-dateutil==2.6.1
 requests==2.18.4
 requests-cache==0.4.13
+responses==0.8.1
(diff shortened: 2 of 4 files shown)


cprov:
@nottrobin it is a trend, each layer of tests will catch a group of problems for a specific cost.

1) basic unittests (`assert foo(x) = y`) will catch local programming errors but will not help with integration logic;
2) unittests with doubles (what I proposed) will also catch integration problems but won't cover page rendering, API provider changes/regressions or the internet quirks;
3) browser testing cover page rendering, API provider and the 'internet', but requires functional credentials and known dataset.
4) Behavior testing additionally to static features/pages, would also cover UX patterns and workflows  

In the given order, costs increase while access (where you can run and how) decreases. Considering we have ZERO tests (effectively) and many problems, I've decided to attack this in the successive-approximation (sort of) mode, it's okay if you want to invest some time now into browser (non-intrusive) tests

nottrobin:
> what I proposed will also catch integration problems but won't cover page rendering

You don't actually test integration do you? What you do is mock responses and then test that the code handles those mocked responses as expected - this is effectively a unit test that tests the internal logic of the view. Which is definitely helpful, but doesn't prove that it will successfully communicate with the actual API.

I was asking if there was a way to authenticate with the API for a test account with known data in a purely scripted fashion - as in we provide secret credentials for that test account through environment variables and then have our automated tests use those to run scripted tests that use the actual API.

But if you think the only way to achieve that is through browser tests, then we can do that too.

Issue #321, opened 2018-02-20
'measure' Implement tooltip hover indication


Lukewh:
As per the [spec](https://docs.google.com/document/d/1PQzECKRrVZ-y3W7DcdkfeZipBVr5Xl2mpMj5Vqhux2E/edit#heading=h.rtb5soqn673v) - embolden tooltip info on hover.

https://github.com/CanonicalLtd/snapcraft-design/issues/278 for discussion.

PR #322, opened 2018-02-20
Measure: Improved label positions

Lukewh:
# Done

- Better vertical alignment of active device labels

This partly fixes https://github.com/CanonicalLtd/snapcraft-design/issues/278
Follow up issue:
- https://github.com/canonical-websites/snapcraft.io/issues/321

# QA

- Pull this branch
- `./run`
- Visit a snap measure page http://0.0.0.0:8004/account/snaps/SNAP_NAME/measure
- Have a look at the labels - do the work?

### DIFF
--- a/static/js/publisher/metrics/graphs/activeDevices.js
+++ b/static/js/publisher/metrics/graphs/activeDevices.js
@@ -11,86 +11,6 @@ function showGraph(el) {
   el.style.opacity = 1;
 }
 
-const findCenter = function(points) {
-  const maxValue = Math.max.apply(Math, points);
-
-  let sections = [];
-  let cache = [];
-  points.forEach((point, index) => {
-    if (point / maxValue < 0.5 && cache.length !== 0) {
-      sections.push(cache.slice(0));
-      cache = [];
-    }
-    if (point / maxValue >= 0.5) {
-      cache.push(index);
-    }
-
-    if (index === points.length - 1 && cache.length > 0) {
-      sections.push(cache);
-    }
-  });
-
-  let longestSection = { length: 0 };
-  sections.forEach(section => {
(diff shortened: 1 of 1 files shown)


Lukewh:
@tbille you need to approve in the code tab otherwise I will be forever blocked.

tbille:
As discussed here is the screenshot: 
screenshot-2018-2-21 publisher metrics for toto

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -58,6 +59,11 @@ const labelPosition = function(graphEl, labelText, points) {
     return;
   }
 
+  const activePoints = points.slice(range[0], range[range.length - 1]);
+  const topActiveItem = Math.max.apply(Math, activePoints);
+  const topItem = Math.max.apply(Math, points);
+  let y = bbox.y + (((bbox.height / topItem) * topActiveItem) / 2);
```
matthewpaulthomas: This is only ever going to work for any version that is more than twice as popular as all newer versions put together. Which, for most snaps, means only the newest version.

For any version X where newer versions combined are more than half as popular, they will ascend to more than half the height of X’s area, so they’ll obscure or completely X’s halfway-down label.

Lukewh:
This is a crazy hard problem to solve when not using d3 directly. I've removed the labels for now as I believe buggy/ broken functionality is worse than not having functionality. Hopefully adding a hover indicator will improve the legibility (to be done in a separate PR).

matthewpaulthomas:
Verified that the area labels are gone. Reported #331 on restoring them once they can be positioned properly.

Lukewh:
Thanks all

PR #324, opened 2018-02-20
Change search API URL

tbille:
search.apps.ubuntu.com is deprecated; the endpoint we use is now provided by api.snapcraft.io.

(same modification that is in b.s.io https://github.com/canonical-websites/build.snapcraft.io/pull/1061 )


Lukewh:
QA's good, if anything is seems faster!

PR #332, opened 2018-02-22
public sentry

Lukewh:
# Done

- Added Raven from the sentry CDN
- Checks for `SENTRY_PUBLIC_DSN` in `.env.local`

# QA

- Pull the branch
- Setup an account on https://sentry.io and get a public DSN
- Add the public dsn to your `.env.local` (or create one):`SENTRY_PUBLIC_DSN=<public_dsn>`
- Change the name of a variable in the JS
- `./run`
- Visit the page with the bad JS
- Check your sentry dashboard, you should see the error immediately
- Rejoice

# Screenshot

screenshot from 2018-02-22 17-20-58

### DIFF
--- a/app.py
+++ b/app.py
@@ -29,11 +29,15 @@ from modules.macaroon import (
 
 
 app = flask.Flask(__name__)
-sentry = Sentry(app)
 app.wsgi_app = ProxyFix(app.wsgi_app)
 app.secret_key = os.environ['SECRET_KEY']
 app.wtf_csrf_secret_key = os.environ['WTF_CSRF_SECRET_KEY']
 app.url_map.strict_slashes = False
+app.sentry_public_dsn = os.getenv(
+    'SENTRY_PUBLIC_DSN',
+    ''
+)
+sentry = Sentry(app)
 
 open_id = OpenID(
     app=app,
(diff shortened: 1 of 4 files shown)


app.py
```diff
@@ -41,8 +45,11 @@
 
 # Make LOGIN_URL available in all templates
 @app.context_processor
-def inject_login():
-    return dict(LOGIN_URL=LOGIN_URL)
+def inject_template_variables():
+    return {
+        'LOGIN_URL': LOGIN_URL,
+        'sentry_public_dsn': app.sentry_public_dsn
```
tbille: could you UPPERCASE this var

app.py
```diff
@@ -23,6 +23,10 @@
 app.secret_key = os.environ['SECRET_KEY']
 app.wtf_csrf_secret_key = os.environ['WTF_CSRF_SECRET_KEY']
 app.url_map.strict_slashes = False
+app.sentry_public_dsn = os.getenv(
```
tbille: can you rebase on master so that you can modify the already existing var?

Lukewh: I rebased, but nothing changed!?

PR #363, opened 2018-03-07
Move back to using the SENTRY_PUBLIC_DSN env var

nottrobin:
We want to use the environment variable so that we don't have to worry
about people cloning and hosting this site elsewhere, and polluting our
sentry logs.

We now `strip()` the variable contents, so newlines or extra spaces
shouldn't break anything.


nottrobin:
@Lukewh I don't trust you. @tbille is also checking it :p

PR #364, opened 2018-03-07
Sentry whitelist

Lukewh:
# Done

Added snapcraft.io and run.demo.haus as white listed urls for error reporting.

# QA

- Wait for the demo to start.
- Go to a measure page.
- See an error in the page - d4 is not defined.
- See an error in sentry - d4 is not defined.
- Check the count of NS_ERROR_UNKNOWN_PROTOCOL errors in sentry
- Go to http://snapcraft.io-pr-364.run.demo.haus/slack
- Click the install button (on an OS that doesn't support snap urls)
- Make sure the NS_ERROR_UNKNOWN_PROTOCOL error count doesn't increase.

### DIFF
--- a/templates/publisher/market.html
+++ b/templates/publisher/market.html
@@ -206,7 +206,9 @@
 <script src="https://cdn.ravenjs.com/3.22.3/raven.min.js" crossorigin="anonymous"></script>
 <script src="/static/js/dist/publisher.js"></script>
 <script>
-  Raven.config('{{ SENTRY_PUBLIC_DSN }}').install();
+  Raven.config('{{ SENTRY_PUBLIC_DSN }}', {
+    whitelistUrls: ['snapcraft.io']
+  }).install();
   Raven.context(function () {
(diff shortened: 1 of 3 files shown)


static/js/publisher/metrics/metrics.js
```diff
@@ -12,6 +12,10 @@ function renderMetrics(metrics) {
     return false;
   }
 
+  if (d4) {
```
tbille: Awesome code

Lukewh: just getting ready for when d4 is released and replaces d3

nottrobin: ? I'm so confused. Surely we're not merging that?

Lukewh: Of course not... But we need to create an error to ensure our errors are still coming through.... Once QA is done I'll remove this code

Lukewh: More details on ^ and because we white list hosts, I've white listed the demo url for QAing

PR #368, opened 2018-03-07
Defend against snapcraft 'not being defined'

Lukewh:


Greggless:
⚠️ `Dumb designer question` ⚠️

> What is this? @Lukewh

bartaz:
It's an error from sentry.
Means that JavaScript didn't load properly and didn't run (so for example map was not shown on details page).

This PR means we will stop getting this error, but still JavaScript will not run.
We will just not know about it, because it will silently fail for user.

I'm not sure if it's good approach. Even though we cannot have any good way to debug it (maybe user blocks JavaScript or it fails to load for some other reason) - we should try to investigate why it maybe happening.

tbille:
This is a hack, @bartaz found out where the problem comes from: we need to transpile our code. PR is on its way!

Greggless:
Thanks for explaining! 

# :electron:

Lukewh:
PR is here: https://github.com/canonical-websites/snapcraft.io/pull/370

PR #370, opened 2018-03-08
Add babel to rollup

Lukewh:
To fix the continuing alerts about snapcraft not being defined.

# QA

@bartaz can you try and reproduce the error.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,10 @@
+{
+  "presets": [
+    ["env", {
+      "targets": {
+        "browsers": ["last 2 versions"]
+      },
+      "modules": false
+    }]
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -16,6 +16,8 @@
   "author": "Canonical webteam",
   "dependencies": {
     "autoprefixer": "^6.3.1",
+    "babel-core": "^6.26.0",
+    "babel-preset-env": "^1.6.1",
     "billboard.js": "^1.1.1",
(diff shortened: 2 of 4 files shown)


bartaz:
QA +1 - JavaScript parses correctly on IE now, doesn't throw `snapcraft is undefined` anymore.

package.json
```diff
@@ -30,5 +30,11 @@
     "topojson-client": "^3.0.0",
     "vanilla-framework": "^1.6.5",
     "watch-cli": "^0.2.2"
+  },
+  "devDependencies": {
```
bartaz: I'd move it all to `dependencies`.
It's not a project that is going to be installed as npm module, so difference between dependencies and devDependencies is not needed and in long term may only be confusing (should something be dev or not).

.babelrc
```diff
@@ -0,0 +1,11 @@
+{
+  "presets": [
+    ["env", {
+      "targets": {
+        "browsers": ["last 2 versions"]
+      },
+      "modules": false
+    }]
+  ],
+  "plugins": ["external-helpers"]
```
bartaz: Are we using `external-helpers`? From what I understand here (https://babeljs.io/docs/plugins/external-helpers/) is they they have to be explicitly built and included separately. It doesn't seem like we do it. Also doesn't seem like we would need to do it in a short-term (as our codebase is not too big enough to benefit from such optimisations).

Unless rollout does something with that automatically?

PR #373, opened 2018-03-09
Remove flask dependency api layer

tbille:
# Refactoto S02E01 - The first Endpoint

## Summary

* Remove flask dependency from api public layer.
* First draft for a better error handler in the api layer
* Work on public snap access

## QA

- `./run`
- `./run test`
- 0.0.0.0:8004/toto
- Should have a snap
- 0.0.0.0:8004/qsdfsqdfqsdf
- Should return 404 page

### DIFF
--- a/modules/public/api.py
+++ b/modules/public/api.py
@@ -1,4 +1,3 @@
-import flask
 import modules.cache as cache
 import os
 
@@ -54,6 +53,14 @@ PROMOTED_QUERY_HEADERS = {
 }
 
 
+class InvalidResponseContent(Exception):
+    pass
+
+
+class ApiErrorResponse(Exception):
+    pass
+
+
 def normalize_searched_snaps(search_results):
     return (
         search_results['_embedded']['clickindex:package']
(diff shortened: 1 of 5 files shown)


modules/public/api.py
```diff
@@ -99,17 +106,27 @@ def get_snap_details(snap_name):
...
-
-    return details_response.json()
+    try:
+        details = details_response.json()
+    except ValueError as decode_error:
+        error_message = ''.join([
+            "JSON decoding failed: ",
+            str(decode_error),
+        ])
+        raise InvalidResponseContent(error_message)
```
cprov: Nitpicking on style, the string-joining looks strange: 
```
error_message = 'JSON decoding failed: {}'.format(decode_error)
```

modules/public/api.py
```diff
@@ -99,17 +106,27 @@ def get_snap_details(snap_name):
...
+
+    if details_response.status_code != 200:
```
cprov: The store API uses other 2xx status and they should be treated as success.

modules/public/api.py
```diff
@@ -99,17 +106,27 @@ def get_snap_details(snap_name):
...
+        if 'error_list' in details:
+            api_error_exception = ApiErrorResponse("Error list")
```
cprov: You might get some expiration from https://git.launchpad.net/snapdevicegw/tree/snapdevicegw/exceptions.py and have a chain of exceptions that would treat the API errors internally, `APIError(error_response)`

modules/public/views.py
```diff
@@ -230,7 +230,16 @@ def snap_details(snap_name):
...
+    try:
+        details = api.get_snap_details(snap_name)
+    except api.InvalidResponseContent as invalid_response_content:
+        message = str(invalid_response_content)
+        flask.abort(500, message)
+    except api.ApiErrorResponse as api_error_exception:
+        if api_error_exception.errors:
+            flask.abort(api_error_exception.status, api_error_exception.errors)
+        else:
+            flask.abort(api_error_exception.status, ["Unknown error"])
```
cprov: It doesn't seem to be a good idea to blindly 'leak' snap-details error status to you own application, nor leaking bad-json responses to ugly 500s to the app users. Wouldn't it be possible to tolerate errors and carry on with `details = {...}` ?

PR #374, opened 2018-03-09
Updated 404 page

Lukewh:
# Done

- Removed logo from the 404 page

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/asdlkgjwjnbnrjnb

OR

- Visit http://snapcraft.io-pr-374.run.demo.haus/asdlkgjwjnbnrjnb

# Screenshot

screenshot from 2018-03-13 13-25-27


templates/404.html
```diff
@@ -4,9 +4,9 @@
 
 {% block content %}
 <div class="p-strip">
-  <div class="row">
-    <div class="col-6">
-      <img src="/static/confused-hex.svg" />
```
bartaz: We should remove the `confused-hex.svg` file from project if it's not used anymore.

lyubomir-popov:
@Lukewh Just consulted with Tom, we shouldn't use the bird for things like that, although it looks nice (sorry!). Having a creative 404 page is a good idea, but we don't have visuals that work with the message at the moment. So can we remove it and have the copy flush left?

Lukewh:
@lyubomir-popov sounds good, was just an idea Greg had on the sprint (as the current 404 isn't "on brand" in any way). I'll update this PR and ping you

lyubomir-popov:
@Lukewh Looks good. One minor niggle, more of a ux thng so maybe something for @Greggless - at the moment there's a bit of repetition - 
"Page not found"
"Snap not found: {snap-name}"

Could these be combined into one? would help if the snap name doesn't look like fine print.

Also shouldn't we be providing a helpful suggestion - maybe a search input?

lyubomir-popov:
(Adding design plus one as the above comments are not a blocker from visual point of view.)

Greggless:
We should definitely design a more useful 404. 

This'll do (I hate saying that) for now, but I'll open an issue for something more useful

PR #375, opened 2018-03-09
Market page screenshots fixes

### LINKED ISSUE #360: Screenshots - limiting number of screenshots to 5 not always works
There are couple of related issues around limiting number of screenshots to 5 when you try to add/delete them at once:

### Choosing more then 5

You can choose more then 5:
- Click add button
- Choose many images (more then 5)
- All images as added to the list and will be uploaded

### You can't add new images if you mark other for deleting

- Go to market page
- Have 5 screenshots (max limit)
- Mark some of them to be deleted
- You cannot add new ones (even that after deleting you would have less then 5)

bartaz:
Fixes #360 

- changed adding screenshots, so user can choose only one file at once (to prevent going over the 5 files limit)
- fixed rendering more then 5 screenshots (when some are marked for deletion and some are being added)
- adds some unit tests for screenshots functionality

### QA
- ./run
- go to market page of some snap you own
- add some screenshots
- delete some screenshots
- add all 5 screenshots (Apply)
- mark old screenshots for delete and add new screenshots

screen shot 2018-03-09 at 11 17 21


static/js/publisher/market/screenshots.js
```diff
@@ -0,0 +1,249 @@
...
+        );
+      }
+    }
+  });
+}
+
+export {
+  initSnapScreenshotsEdit,
+  // for testing
+  templates,
```
Lukewh: If we need to export things that aren't 'public' api's for a module, should we namespace them? It should provide a red flag to anyone who wants to import that function to maybe move it somewhere more generic.

Something like
```
const __tests = {
  templates: templates,
  serializeState: serializeState
};

export {
  initSnapScreenshotsEdit,
  __tests
}
```

This might be overkill - just a thought.

bartaz: Yes, I was thinking about something like this. Didn't want to over-engineer it at this stage.
In theory you should only need to test public functions (private should be covered if they are used internally), but in this case it's just easier to unit test some helper function then whole thing operating on DOM.

I would leave it for now. Hopefully when we move to React and per field editing it will be better structured and easier to test.

PR #376, opened 2018-03-09
Handle errors public api calls

tbille:
# Refactor s02e02 - Public protector

## Summary

-  Handle errors for homepage and store page for featured snaps
-  Handle errors for snap store
-  Handle errors for metrics page
- Handle errors for search without illegal characters
- Handle timeouts

## QA

- `./run`
- http://0.0.0.0:8004/ 
- http://0.0.0.0:8004/toto
- http://0.0.0.0:8004/qsdfsqdf
- http://0.0.0.0:8004/store
- http://0.0.0.0:8004/search?q=%3A%3A 
All should work as usual

### DIFF
--- a/modules/cache.py
+++ b/modules/cache.py
@@ -1,4 +1,5 @@
 import requests
+from modules.exceptions import ApiErrorResponse
 
 
 def get(
@@ -20,12 +21,21 @@ def get(
     if method is None:
         method = "POST" if json else "GET"
 
-    return requests.request(
-        method=method,
-        url=url,
-        headers=headers,
-        json=json,
-        files=files,
-        data=data,
-        timeout=2
-    )
+    try:
+        return requests.request(
+            method=method,
+            url=url,
+            headers=headers,
+            json=json,
+            files=files,
+            data=data,
+            timeout=2
+        )
+    except requests.exceptions.Timeout as exception_response:
+        api_error_exception = ApiErrorResponse("Timeout")
+        api_error_exception.errors = [{
+            'code': 'timeout-server',
+            'message': 'API timeout'
+        }]
+        api_error_exception.status = 500
+        raise api_error_exception
--- a/modules/cache.py
+++ b/modules/cache.py
@@ -1,5 +1,8 @@
 import requests
-from modules.exceptions import ApiTimeoutError
+from modules.exceptions import (
+    ApiTimeoutError,
+    ApiConnectionError
+)
 
 
 def get(
@@ -31,8 +34,13 @@ def get(
             data=data,
             timeout=2
         )
-    except requests.exceptions.Timeout as exception_response:
+    except requests.exceptions.Timeout:
         api_error_exception = ApiTimeoutError(
             'The request to {} took longer than 2 seconds'.format(url),
+        )
+        raise api_error_exception
+    except requests.exceptions.ConnectionError:
+        api_error_exception = ApiConnectionError(
+            'Failed to establish connection to {}.'.format(url)
         )
         raise api_error_exception
--- a/modules/cache.py
+++ b/modules/cache.py
@@ -32,10 +32,8 @@ def get(
             timeout=2
         )
     except requests.exceptions.Timeout as exception_response:
-        api_error_exception = ApiErrorResponse("Timeout")
-        api_error_exception.errors = [{
-            'code': 'timeout-server',
-            'message': 'API timeout'
-        }]
-        api_error_exception.status = 500
+        api_error_exception = ApiErrorResponse(
+            'The request to {} took longer than 2 seconds'.format(url)
+        )
+        api_error_exception.status_code = 504
         raise api_error_exception
--- a/modules/cache.py
+++ b/modules/cache.py
@@ -1,5 +1,5 @@
 import requests
-from modules.exceptions import ApiErrorResponse
+from modules.exceptions import ApiResponseError
 
 
 def get(
@@ -32,7 +32,7 @@ def get(
             timeout=2
         )
     except requests.exceptions.Timeout as exception_response:
-        api_error_exception = ApiErrorResponse(
+        api_error_exception = ApiResponseError(
             'The request to {} took longer than 2 seconds'.format(url)
         )
         api_error_exception.status_code = 504
--- a/modules/cache.py
+++ b/modules/cache.py
@@ -1,5 +1,5 @@
 import requests
-from modules.exceptions import ApiResponseError
+from modules.exceptions import ApiTimeoutError
 
 
 def get(
@@ -32,8 +32,7 @@ def get(
             timeout=2
         )
     except requests.exceptions.Timeout as exception_response:
-        api_error_exception = ApiResponseError(
-            'The request to {} took longer than 2 seconds'.format(url)
+        api_error_exception = ApiTimeoutError(
+            'The request to {} took longer than 2 seconds'.format(url),
         )
-        api_error_exception.status_code = 504
         raise api_error_exception
--- a/modules/exceptions.py
+++ b/modules/exceptions.py
@@ -6,6 +6,14 @@ class ApiError(Exception):
     pass
 
 
+class ApiConnectionError(ApiError):
+    """
+    Communication with the API failed
+    """
+
+    pass
+
+
 class ApiTimeoutError(ApiError):
     """
     Communication with the API timed out
--- a/modules/exceptions.py
+++ b/modules/exceptions.py
@@ -1,6 +1,2 @@
-class InvalidResponseContent(Exception):
-    pass
-
-
-class ApiErrorResponse(Exception):
+class ApiResponseError(Exception):
     pass
--- a/modules/exceptions.py
+++ b/modules/exceptions.py
@@ -1,0 +1,6 @@
+class InvalidResponseContent(Exception):
+    pass
+
+
+class ApiErrorResponse(Exception):
+    pass
--- a/modules/exceptions.py
+++ b/modules/exceptions.py
@@ -1,2 +1,43 @@
-class ApiResponseError(Exception):
+class ApiError(Exception):
+    """
+    Base exception for any errors in the API layer
+    """
+
     pass
+
+
+class ApiTimeoutError(ApiError):
+    """
+    Communication with the API timed out
+    """
+
+    pass
+
+
+class ApiResponseDecodeError(ApiError):
+    """
+    We failed to properly decode the response from the API
+    """
+
+    pass
+
+
+class ApiResponseError(ApiError):
+    """
+    The API responded with an error
+    """
+
+    def __init__(self, message, status_code):
+        self.status_code = status_code
+        return super().__init__(message)
+
+
+class ApiResponseErrorList(ApiResponseError):
+    """
+    The API responded with a list of errors,
+    which are included in self.errors
+    """
+
+    def __init__(self, message, status_code, errors):
+        self.errors = errors
+        return super().__init__(message, status_code)
--- a/modules/public/api.py
+++ b/modules/public/api.py
@@ -125,4 +125,4 @@ def get_public_metrics(snap_name, json):
         json=json
     )
 
-    return metrics_response.json()
+    return process_response(metrics_response)
--- a/modules/public/api.py
+++ b/modules/public/api.py
@@ -1,9 +1,6 @@
 import modules.cache as cache
 import os
-from modules.exceptions import (
-    InvalidResponseContent,
-    ApiErrorResponse
-)
+from modules.exceptions import ApiResponseError
 
 SNAPCRAFT_IO_API = os.getenv(
     'SNAPCRAFT_IO_API',
@@ -61,12 +58,15 @@ def process_response(response):
     try:
         body = response.json()
     except ValueError as decode_error:
-        error_message = 'JSON decoding failed: {}'.format(decode_error)
-        raise InvalidResponseContent(error_message)
+        api_error_exception = ApiResponseError(
+            'JSON decoding failed: {}'.format(decode_error)
+        )
+        api_error_exception.status_code = 500
+        raise api_error_exception
 
     if not response.ok:
         if 'error_list' in body:
-            api_error_exception = ApiErrorResponse("Error list")
+            api_error_exception = ApiResponseError("Error list")
             api_error_exception.status_code = response.status_code
             errors = []
             for error in body['error_list']:
@@ -75,7 +75,7 @@ def process_response(response):
 
             raise api_error_exception
         else:
-            api_error_exception = ApiErrorResponse("Unknown error")
+            api_error_exception = ApiResponseError("Unknown error")
             api_error_exception.status_code = response.status_code
             raise api_error_exception
 
--- a/modules/public/api.py
+++ b/modules/public/api.py
@@ -100,7 +100,7 @@ def get_promoted_snaps():
         headers=PROMOTED_QUERY_HEADERS
     )
 
-    return promoted_response.json()
+    return process_response(promoted_response)
 
 
 def get_searched_snaps(snap_searched, size, page):
--- a/modules/public/api.py
+++ b/modules/public/api.py
@@ -91,7 +91,7 @@ def get_featured_snaps():
         headers=SEARCH_QUERY_HEADERS
     )
 
-    return featured_response.json()
+    return process_response(featured_response)
 
 
 def get_promoted_snaps():
--- a/modules/public/api.py
+++ b/modules/public/api.py
@@ -1,6 +1,10 @@
 import modules.cache as cache
 import os
-from modules.exceptions import ApiResponseError
+from modules.exceptions import (
+    ApiResponseDecodeError,
+    ApiResponseError,
+    ApiResponseErrorList
+)
 
 SNAPCRAFT_IO_API = os.getenv(
     'SNAPCRAFT_IO_API',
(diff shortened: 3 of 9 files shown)


modules/public/api.py
```diff
@@ -53,20 +57,25 @@
 }
 
 
-class InvalidResponseContent(Exception):
-    pass
-
-
-class ApiErrorResponse(Exception):
-    pass
+def process_response(response):
```
nottrobin: Is every `get` going to go through `process_response`? If so, you might as well roll all these checks into `get` - although you could rename it if you feel the name doesn't quite suite any more.

tbille: @nottrobin 
- I prefer not to put this code in the get function yet. Because it could break the api calls for the publisher pages. This should be done when I'll work on the refactor of the publishers.

- the name seems good to me for now, you thought about something else?

nottrobin: Makes sense

modules/public/views.py
```diff
@@ -284,6 +320,35 @@ def external(attrs, new=False):
...
+            metrics_query_json
+        )
+
+        users_by_country = normalize_metrics(
+            metrics_response[0]['series']
+        )
+        country_data = build_country_info(users_by_country)
+    except api.InvalidResponseContent as invalid_response_content:
+        message = str(invalid_response_content)
+        errors = [{'code': 'json-error', 'message': message}]
```
nottrobin: Why do we need this structure to the error (`code` and `message`)? Why isn't it simply a string?

In any case, the `api` layer deliberately obfuscates the fact that it's using JSON, which is why it converts a `JSONDecodeError` into a `InvalidResponseContent` error (because we shouldn't care outside the API whether it's JSON or not). Why then are we inferring again that it is a JSON problem? If we care that it's a JSON error we should bubble that up from within the API layer. But my feeling is that we shouldn't care, in which case we shouldn't mention it here.

modules/cache.py
```diff
@@ -20,12 +21,21 @@ def get(
...
+            method=method,
+            url=url,
+            headers=headers,
+            json=json,
+            files=files,
+            data=data,
+            timeout=2
+        )
+    except requests.exceptions.Timeout as exception_response:
+        api_error_exception = ApiErrorResponse("Timeout")
```
nottrobin: Since a timeout is something that the API layer should explicitly expect, I think we should be more declarative here and create an `ApiTimeoutError` class.

Then we can simply do:

``` python
return ApiTimeoutError('The request to {} timed out'.format(url))
```

And we don't need to worry about hacking `code` and `status` onto another error type.

modules/public/api.py
```diff
@@ -53,20 +57,25 @@
...
+    try:
+        body = response.json()
+    except ValueError as decode_error:
+        error_message = 'JSON decoding failed: {}'.format(decode_error)
+        raise InvalidResponseContent(error_message)
 
+    if not response.ok:
+        if 'error_list' in body:
+            api_error_exception = ApiErrorResponse("Error list")
```
nottrobin: Similarly, since only an "Error list" is going to contain an `errors` property, we should probably be explicit about this by making it its own type: `ApiResponseErrorList`.

modules/public/views.py
```diff
@@ -162,24 +162,63 @@ def convert_limit_offset_to_size_page(link):
 
 
 def homepage():
+    featured_snaps = []
+    errors = []
+    try:
+        featured_snaps = normalize_searched_snaps(api.get_featured_snaps())
+    except api.InvalidResponseContent as invalid_response_content:
+        errors = ["Error from server"]
```
nottrobin: It feels like we're losing information here if all we pass on is "Error from server". We know it's about the content of the response. Couldn't we be slightly more descriptive in the public message?

modules/public/views.py
```diff
@@ -162,24 +162,63 @@ def convert_limit_offset_to_size_page(link):
...
+    except api.ApiErrorResponse as api_error_exception:
+        if api_error_exception.errors:
+            errors = api_error_exception.errors
+        else:
+            errors = ["Unknown error"]
```
nottrobin: I believe we're masking the error message from the API layer in case it leaks secrets of some kind? It's probably worth mentioning here at least that it's an error with the API's response. We should also at least [log](http://flask.pocoo.org/docs/dev/logging/) the original error message.

modules/public/views.py
```diff
@@ -162,24 +162,63 @@ def convert_limit_offset_to_size_page(link):
...
+        errors=errors
     )
 
 
 def store():
+    featured_snaps = []
+    errors = []
+    try:
+        featured_snaps = normalize_searched_snaps(api.get_featured_snaps())
+    except api.InvalidResponseContent as invalid_response_content:
```
nottrobin: This block of code is going to get repetitive - there must be some way to encapsulate that

modules/public/views.py
```diff
@@ -193,22 +232,38 @@ def search_snap():
...
+        links = get_pages_details(
             (
                 searched_results['_links']
                 if '_links' in searched_results
                 else []
             )
         )
+    except api.InvalidResponseContent as invalid_response_content:
+        message = str(invalid_response_content)
+        flask.abort(500, message)
```
nottrobin: Are we sure we want to throw an error page here? I thought we were going to try to make APi errors look graceful?

modules/public/views.py
```diff
@@ -239,26 +336,13 @@ def snap_details(snap_name):
...
-    metrics_response = api.get_public_metrics(
-        snap_name,
-        metrics_query_json
-    )
-
-    users_by_country = normalize_metrics(
-        metrics_response[0]['series']
-    )
-    country_data = build_country_info(users_by_country)
+            errors = [
```
nottrobin: `errors.append()`?

modules/public/views.py
```diff
@@ -315,7 +432,9 @@ def external(attrs, new=False):
         'is_linux': (
             'Linux' in flask.request.headers.get('User-Agent', '') and
             'Android' not in flask.request.headers.get('User-Agent', '')
-        )
+        ),
+
+        'errors': errors
     }
 
     return flask.render_template(
```
nottrobin: If we do have an API communication error, could we please return a non-200 status? Since all the information for that page comes from the API, I don't think it counts has a successful response if the information isn't there. We can keep the page running through the same template - i.e. looking relatively pretty - but I think its status should probably be one of:

```
502 Bad Gateway
503 Service Unavailable
504 Gateway Timeout
```

modules/exceptions.py
```diff
@@ -0,0 +1,6 @@
+class InvalidResponseContent(Exception):
```
nottrobin: I just realised it would be useful to use a single base class for all API exceptions. E.g. `ApiError`. That way we can check for any deliberate `ApiError`, as opposed to random other errors - as might often be useful.

modules/public/api.py
```diff
@@ -53,20 +57,25 @@
...
+            api_error_exception.status = response.status_code
+            api_error_exception.errors = body['error_list']
+            raise api_error_exception
+        else:
+            api_error_exception = ApiErrorResponse("Unknown error")
+            api_error_exception.status = response.status_code
```
nottrobin: It's probably worth preserving the `body` on this exception

modules/exceptions.py
```diff
@@ -0,0 +1,6 @@
...
+    pass
+
+
+class ApiErrorResponse(Exception):
```
nottrobin: Could we call this `ApiResponseError`?

modules/public/views.py
```diff
@@ -284,6 +367,40 @@ def external(attrs, new=False):
...
+    except api.InvalidResponseContent as invalid_response_content:
+        message = str(invalid_response_content)
+        errors = [{'code': 'json-error', 'message': message}]
+    except api.ApiErrorResponse as api_error_exception:
+        if api_error_exception.errors:
+            errors = api_error_exception.errors
+        else:
+            errors = [
+                {
+                    'code': 'unknown-error',
```
nottrobin: Again, what is the use of these `code`s? Might this not be over-complicating things? We should only pass through to the template what the template needs to display, and I suspect all it needs is a message.

modules/public/views.py
tbille: @nottrobin we use this structure because:
- the server sends us this structure
- There could be mutlitple errors, so the frontend will iterate on those errors
- We were think about adding a mapping of all the errors to add our own messages since we have the list of possible [errors](https://dashboard.snapcraft.io/docs/api/snap.html#errors)

modules/public/views.py
tbille: I can add better errors descriptions, can we add the log once the refactor is done. Maybe as a new issue in the epic?

modules/public/views.py
tbille: done every template reterns a status code

modules/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class ApiResponseError(Exception):
```
nottrobin: You collapsed this file down to one single type? I do think it's a good idea to keep a few different types, I was just suggesting that you have a unified base-type so we could check for any type of error coming from this layer. e.g.:

``` python
class ApiError(Exception):
    pass

class ApiTimeoutError(ApiError):
    pass

class ApiResponseDecodeError(ApiError):
    pass

class ApiResponseErrorList(ApiError):
    def __init__(self, message, errors):
        self.errors = errors
        return super().__init__(message)
```

Then users of `api.py` can be general or specific when handling errors:

``` python
try:
    api.do_something()
except ApiTimeoutError:
    print('API was super-slow')
except APIError:
    print('API had some other error')
```

modules/public/views.py
```diff
@@ -162,28 +162,70 @@ def convert_limit_offset_to_size_page(link):
 
 
 def homepage():
+    featured_snaps = []
+    errors = []
+    status_code = 200
+    try:
+        featured_snaps = normalize_searched_snaps(api.get_featured_snaps())
+    except api.ApiResponseError as api_error_exception:
+        if hasattr(api_error_exception, 'errors'):
```
nottrobin: Building on my [previous comment](https://github.com/canonical-websites/snapcraft.io/pull/376/files#r174540058), we shouldn't need to infer the type of an error that we created ourselves. If we're interested in the difference between an error list and another type of error (which we are) we should just make explicit error types to test against.

modules/public/views.py
```diff
@@ -162,28 +162,70 @@ def convert_limit_offset_to_size_page(link):
...
+            errors = api_error_exception.errors
```
nottrobin: You simply want to pass all errors from the API straight onto your `errors` object? I'm not sure that's a good idea. There are a couple of ways in which you could get multiple errors, I believe:

- You may be making multiple API calls per view function
- You may be getting multiple errors back from a single API call

I don't think you should merge the two sets of errors. I think it makes more sense to preserve that hierarchy. Here's two ways I can think of to do that:

``` python
error_messages = []

try:
    api.get_one()
except ApiError as api_error:
    error_messages.append("API request 'one' gave us an error. Naughty API.")

try:
    api.get_two()
except ApiResponseErrorList as list_error:
    error_codes = ", ".join(list_error.errors.keys())
    error_messages.append("API request 'two' gave us a bunch of errors: " + error_codes)
```

or

``` python
errors = []

try:
    api.get_one()
except ApiError as api_error:
    errors.append({'message': "API request 'one' gave us an error. Naughty API."})

try:
    api.get_two()
except ApiResponseErrorList as list_error:
    errors.append({'message': "API request 'two' gave us a bunch of errors", 'errors': list_error.errors)
```

modules/public/views.py
```diff
@@ -193,29 +235,44 @@ def search_snap():
...
                 searched_results['_links']
                 if '_links' in searched_results
                 else []
             )
         )
+    except api.ApiResponseError as api_error_exception:
+        if hasattr(api_error_exception, 'errors'):
+            errors = api_error_exception.errors
+        else:
+            errors.append(str(api_error_exception))
```
nottrobin: Sorry if this was because of my suggestion, but we don't need `errors.append()` in this case - we're just making one call which will just return one error.

modules/public/views.py
```diff
@@ -284,6 +325,35 @@ def external(attrs, new=False):
...
+            metrics_response[0]['series']
+        )
+        country_data = build_country_info(users_by_country)
+    except api.ApiResponseError as api_error_exception:
+        if hasattr(api_error_exception, 'errors'):
+            errors = api_error_exception.errors
+        else:
+            errors.append(str(api_error_exception))
+
+        status_code = api_error_exception.status_code
```
nottrobin: We shouldn't just return the `status_code` from the API. This is how the *API* tells *us* what went wrong on their end, but that's not necessarily the same thing that we want to tell our client.

Basically, we need to decide if we consider ourselves a "gateway" to the API, in the HTTP sense. If we are then most API errors should be communicated as a "502 Bad Gateway" error, except for the specific case of an API timeout, in which case we should send a "504 Gateway Timeout" error.

If we don't consider ourselves a "gateway" in the technical sense, then all errors should simply be communicated as "500 Internal Server Error".

[RFC2616 sec1](https://www.w3.org/Protocols/rfc2616/rfc2616-sec1.html#sec1) describes a "gateway" as:

> A server which acts as an intermediary for some other server. Unlike a proxy, a gateway receives requests as if it were the origin server for the requested resource; the requesting client may not be aware that it is communicating with a gateway.

To me that sounds a lot like what we're doing here, so I would suggest we should use `502` in most cases, and `504` in the case of timeouts.

modules/public/views.py
```diff
@@ -227,38 +284,23 @@ def snap_details(snap_name):
...
         details = api.get_snap_details(snap_name)
-    except api.InvalidResponseContent as invalid_response_content:
-        message = str(invalid_response_content)
-        flask.abort(500, message)
-    except api.ApiErrorResponse as api_error_exception:
-        if api_error_exception.errors:
-            flask.abort(api_error_exception.status, api_error_exception.errors)
+    except api.ApiResponseError as api_error_exception:
+        if hasattr(api_error_exception, 'errors'):
+            flask.abort(
```
nottrobin: So, in this case, if you get an API error, you're sending the user to an error page, rather than sending them through the main "snap details" template. This is different from what you seem to be doing in the case of most other errors - where you're using the original view template, but with some `errors` in the context.

I understand that a `404` for snap details is a special case, but here you're catching all error types.

Is this deliberate? If there's a specific reason then let's discuss it. It may be that it's actually simplest to *always* send users to an error page on any API error, and then just make sure that that page is pretty.

modules/exceptions.py
nottrobin: You collapsed this file down to declaring one single type? I do think it's a good idea to keep a few different types, I was just suggesting that you have a unified base-type so we could check for any type of error coming from this layer. e.g.:

``` python
class ApiError(Exception):
    """
    Base exception for any errors in the API layer
    """

    pass


class ApiTimeoutError(ApiError):
    """
    Communication with the API timed out
    """

    pass


class ApiResponseDecodeError(ApiError):
    """
    We failed to properly decode the response from the API
    """

    pass


class ApiResponseError(ApiError):
    """
    The API responded with an error
    """

    def __init__(self, message, status_code):
        self.status_code = status_code
        return super().__init__(message)


class ApiResponseErrorList(ApiResponseError):
    """
    The API responded with a list of errors,
    which are included in self.errors
    """

    def __init__(self, message, status_code, errors):
        self.errors = errors
        return super().__init__(message, status_code)
```

Then users of `api.py` can be general or specific when handling errors:

``` python
try:
    api.do_something()
except ApiTimeoutError:
    print('API was super-slow')
except APIError:
    print('API had some other error')
```

modules/public/views.py
```diff
@@ -193,29 +243,39 @@ def search_snap():
...
+        "links": links,
+        "errors": errors
 
     }
 
     return flask.render_template(
         'search.html',
         **context
-    )
+    ), status_code
```
tbille: no status code

modules/public/views.py
```diff
@@ -227,38 +287,44 @@ def snap_details(snap_name):
...
-        if api_error_exception.errors:
-            flask.abort(api_error_exception.status, api_error_exception.errors)
+    except ApiTimeoutError as api_timeout_error:
+        flask.abort(
+            504,
+            str(api_timeout_error)
+        )
+    except ApiResponseDecodeError as api_response_decode_error:
+        flask.abort(
+            500,
```
tbille: 502

modules/public/views.py
```diff
@@ -227,38 +287,44 @@ def snap_details(snap_name):
...
-        metrics_query_json
-    )
-
-    users_by_country = normalize_metrics(
-        metrics_response[0]['series']
-    )
-    country_data = build_country_info(users_by_country)
+            error_messages = ', '.join(api_response_error_list.errors)
+            flask.abort(
+                500,
```
tbille: 502

modules/public/views.py
```diff
@@ -227,38 +287,44 @@ def snap_details(snap_name):
...
+                error_messages
+            )
+    except ApiResponseError as api_response_error:
+        flask.abort(
+            500,
```
tbille: 502

modules/public/views.py
```diff
@@ -227,38 +287,44 @@ def snap_details(snap_name):
...
+            str(api_response_error)
+        )
+    except ApiError as api_error:
+        flask.abort(
+            500,
```
tbille: 502

modules/public/views.py
```diff
@@ -227,38 +287,44 @@ def snap_details(snap_name):
...
-    metrics_response = api.get_public_metrics(
-        snap_name,
-        metrics_query_json
-    )
-
-    users_by_country = normalize_metrics(
-        metrics_response[0]['series']
-    )
-    country_data = build_country_info(users_by_country)
+            error_messages = ', '.join(api_response_error_list.errors)
```
tbille: concat codes

templates/404.html
```diff
@@ -10,11 +10,9 @@
     </div>
     <div class="col-6">
       <h1>404: Page not found</h1>
-      {% for error in error_list %}
-        <p class="p-notification__response">
-          {{ error.message }}
-        </p>
-      {% endfor %}
+      <p class="p-notification__response">
+        {{ errors }}
```
tbille: error

modules/public/views.py
```diff
@@ -161,29 +168,76 @@ def convert_limit_offset_to_size_page(link):
...
 
 
 def store():
+    featured_snaps = []
+    errors = []
+    status_code = 200
+    try:
+        featured_snaps = normalize_searched_snaps(api.get_featured_snaps())
+    except ApiError as api_error:
+        status_code, errors = _handle_errors(api_error)
```
nottrobin: Should be "error" - now that it's a dictionary

PR #378, opened 2018-03-12
Remove logic from views

tbille:
# Refactoto S02E03 - Logic is the answer

## Summary

- Remove all logic from view functions so that the view calls the API, the logic plays with the data and the view returns the template

## QA

- `./run`
- http://0.0.0.0:8004/toto
- The descritpion should be displayed properly, the map too
- http://0.0.0.0:8004/search?q=test&limit=10&offset=40
- Play with next, prev, first, last buttons
- Make sure nothing is broken

### DIFF
--- a/modules/public/logic.py
+++ b/modules/public/logic.py
@@ -1,0 +1,154 @@
+import flask
+import pycountry
+from urllib.parse import parse_qs, urlparse
+
+
+def calculate_colors(countries, max_users):
+    for country_code in countries:
+        countries[country_code]['color_rgb'] = [
+            calculate_color(
+                countries[country_code]['percentage_of_users'],
+                max_users,
+                8,
+                229
+            ),
+            calculate_color(
+                countries[country_code]['percentage_of_users'],
+                max_users,
+                64,
+                245
+            ),
+            calculate_color(
+                countries[country_code]['percentage_of_users'],
+                max_users,
+                129,
+                223
+            )
+        ]
+
+
+def calculate_color(thisCountry, maxCountry, maxColor, minColor):
+    countryFactor = float(thisCountry)/maxCountry
+    colorRange = maxColor - minColor
+    return int(colorRange*countryFactor+minColor)
+
+
+def normalize_metrics(geodata):
+    users_by_country = {}
+    max_users = 0.0
+    for country_counts in geodata:
+        country_code = country_counts['name']
+        users_by_country[country_code] = {}
+        counts = []
+        for daily_count in country_counts['values']:
+            if daily_count is not None:
+                counts.append(daily_count)
+
+        if len(counts) > 0:
+            users_by_country[country_code]['number_of_users'] = (
+                sum(counts)
+            )
+            users_by_country[country_code]['percentage_of_users'] = (
+                sum(counts) / len(counts)
+            )
+        else:
+            users_by_country[country_code]['number_of_users'] = 0
+            users_by_country[country_code]['percentage_of_users'] = 0
+
+        if max_users < users_by_country[country_code]['percentage_of_users']:
+            max_users = users_by_country[country_code]['percentage_of_users']
+
+    calculate_colors(users_by_country, max_users)
+
+    return users_by_country
+
+
+def build_country_info(users_by_country, display_number_users=False):
+    country_data = {}
+    for country in pycountry.countries:
+        country_info = users_by_country.get(country.alpha_2)
+        number_of_users = 0
+        percentage_of_users = 0
+        color_rgb = [247, 247, 247]
+        if country_info is not None:
+            if display_number_users:
+                number_of_users = country_info['number_of_users'] or 0
+            percentage_of_users = country_info['percentage_of_users'] or 0
+            color_rgb = country_info['color_rgb'] or [247, 247, 247]
+
+        # Use common_name if available to be less political offending (#310)
+        try:
+            country_name = country.common_name
+        except AttributeError:
+            country_name = country.name
+
+        country_data[country.numeric] = {
+            'name': country_name,
+            'code': country.alpha_2,
+            'percentage_of_users': percentage_of_users,
+            'color_rgb': color_rgb
+        }
+
+        if display_number_users:
+            country_data[country.numeric]['number_of_users'] = number_of_users
+
+    return country_data
+
+
+def normalize_searched_snaps(search_results):
+    return (
+        search_results['_embedded']['clickindex:package']
+        if '_embedded' in search_results
+        else []
+    )
+
+
+def get_pages_details(links):
+    links_result = {}
+
+    if('first' in links):
+        links_result['first'] = convert_limit_offset_to_size_page(
+            links['first']['href']
+        )
+
+    if('last' in links):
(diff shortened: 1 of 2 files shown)


modules/public/views.py
```diff
@@ -1,214 +1,122 @@
...
 
 def snaps():
+    promoted_snaps = []
+    errors = []
+    try:
+        promoted_snaps = logic.normalize_searched_snaps(
+            api.get_promoted_snaps()
+        )
+    except api.InvalidResponseContent as invalid_response_content:
+        errors = ["Error from server"]
```
nottrobin: It feels like we're losing information here if all we pass on is "Error from server". We know it's about the content of the response. Couldn't we be slightly more descriptive in the public message?

modules/public/views.py
```diff
@@ -1,214 +1,122 @@
...
+    except api.ApiErrorResponse as api_error_exception:
+        if api_error_exception.errors:
+            errors = api_error_exception.errors
+        else:
+            errors = ["Unknown error"]
```
nottrobin: It's probably worth mentioning here at least that it's an error with the API's response.

templates/index.html
```diff
@@ -310,34 +310,36 @@ <h2>The Snap Store</h2>
           <p class="p-heading--five">The Snap Store is enabled by default on tens of millions of Linux systems. It allows developers to release free or paid apps for multiple architectures, on multiple release channels from daily builds to stable releases.</p>
```
nottrobin: Could you split these template changes out into their own PR? They don't seem related to the code restructure, and I don't right now understand what they're doing, so it would be easier if they were in their own PR where I could focus on what they're doing.

modules/public/logic.py
```diff
@@ -142,6 +186,13 @@ def get_pages_details(url, links):
 
 
 def convert_limit_offset_to_size_page(url, link):
+    """Convert navigation link from offst/limit to size/page
```
nottrobin: `s/offst/offset/`

nottrobin: Could you make this one a bit more descriptive? Give examples of input and output URLs maybe?

modules/public/logic.py
```diff
@@ -162,6 +213,12 @@ def convert_limit_offset_to_size_page(url, link):
 
 
 def format_paragraphs(unformatted_description):
+    """Format paragraphs to displayable paragraph
```
nottrobin: split_description_into_paragraphs()
"""
Split a long description into a set of paragraphs. We assume each paragraph is separated by 2 or more line-breaks in the description.

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+
+        paragraph = bleach.clean(paragraph, tags=[])
+        paragraph = bleach.linkify(paragraph, callbacks=callbacks)
+
+        formatted_paragraphs.append(paragraph.replace('\n', '<br />'))
+
+    return formatted_paragraphs
+
+
+def convert_args_search(limit, offset):
```
nottrobin: This seems so trivial that it really doesn't need a function

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+    if('self' in links):
+        links_result['self'] = convert_limit_offset_to_size_page(
+            url,
+            links['self']['href']
+        )
+
+    return links_result
+
+
+def convert_limit_offset_to_size_page(url, link):
```
nottrobin: I think it would make sense to call this `convert_navigation_url`, since it takes and returns a whole URL.

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+    """Convert navigation link from offst/limit to size/page
```
nottrobin: Could you give examples of input and output URLs here?

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+
+    return host_url.format(
+        base_url=url,
+        q=q,
+        limit=size,
+        offset=size*(page-1)
+    )
+
+
+def format_paragraphs(unformatted_description):
```
nottrobin: `split_description_into_paragraphs`?

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+    """Format paragraphs to displayable paragraph
```
nottrobin: "Split a block of text into paragraphs, assuming 2 or more line breaks signifies a paragraph break."

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
+import bleach
+from math import floor
+import pycountry
+import re
+from urllib.parse import parse_qs, urlparse
+
+
+def calculate_colors(countries, max_users):
+    """Calculate the displayed colors for a list of country depending on the
+    maximum of users
```
nottrobin: "maximum number of users"

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+
+    :param countries: List of countries
+    :param max_users: Maximum of users
```
nottrobin: ditto ^

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+
+    :returns: The list of countries with a caculated color for each
```
nottrobin: s/caculated/calculated/

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+    """
+    for country_code in countries:
+        countries[country_code]['color_rgb'] = [
```
nottrobin: It's a bad pattern to directly manipulate a variable that was passed by reference - functions should [be atomic and avoid side-effects](https://softwareengineering.stackexchange.com/questions/15269/why-are-side-effects-considered-evil-in-functional-programming) to avoid confusion.

So what you should do in this case is create a new dictionary and return that instead.

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+
+    :returns: The calculated color for the country
+    """
+    countryFactor = float(thisCountry)/maxCountry
+    colorRange = maxColor - minColor
+    return int(colorRange*countryFactor+minColor)
+
+
+def transform_metrics(geodata):
+    """Transform metrics to get more displayable metrics
```
nottrobin: From reading this docstring, I still don't understand what this function does. Could you elaborate?

modules/public/logic.py
```diff
@@ -0,0 +1,258 @@
...
+        if max_users < users_by_country[country_code]['percentage_of_users']:
+            max_users = users_by_country[country_code]['percentage_of_users']
+
+    calculate_colors(users_by_country, max_users)
+
+    return users_by_country
+
+
+def build_country_info(users_by_country, display_number_users=False):
+    """Build info list for every country
```
nottrobin: Same here, could you write some more detail?

modules/public/logic.py
```diff
@@ -0,0 +1,280 @@
+import bleach
+import pycountry
+import re
+from urllib.parse import parse_qs, urlparse
+
+
+def calculate_colors(countries, max_users):
+    """Calculate the displayed colors for a list of country depending on the
```
nottrobin: s/country/countries/

PR #385, opened 2018-03-13
Show blank graphs and message when lacking data.

Lukewh:
# Done

- Implemented some styling for no data
- Added notification at the top of the page
- Made sure axis still show on graphs without data

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/SNAP_NAME/measure where SNAP_NAME is the name of a snap without any data
- Everything in 'Done' should be there.

# Screenshot

screenshot-2018-3-14 publisher metrics for lukewh-test-snap 3


templates/publisher/measure.html
```diff
@@ -1,123 +1,124 @@
...
+    </div>
+    {% endblock %}
 
-{% block scripts %}
-<script src="/static/js/modules/moment.min.js"></script>
-<script src="/static/js/modules/d3.min.js"></script>
-<script src="/static/js/modules/d3-geo.min.js"></script>
-<script src="/static/js/modules/topojson-client.min.js"></script>
-<script src="/static/js/modules/billboard.min.js"></script>
+    {% block scripts %}
```
tbille: Weird indentation

templates/publisher/measure.html
```diff
@@ -1,123 +1,124 @@
...
-                <option value="os"{% if active_device_metric == 'os' %} selected="selected"{% endif %}>By OS</option>
+          <div class="grid">
+            <div class="col-3 u-no-margin--left">
+              <select>
+                <option>by version</option>
               </select>
             </div>
           </div>
-          <div id="active_devices" class="snapcraft-metrics__graph snapcraft-metrics__active-devices"></div>
+          <p>GRAPH</p>
```
tbille: GRAPH, GRAPH, GRAPH

matthewpaulthomas:
If previously the page was empty, then this is a definite improvement and should land.

The notification text, though, is quite mysterious. Why is there no data available? Is the site malfunctioning? Would I see this if the snap still had zero users after a month? For example, [the prototype screen](https://projects.invisionapp.com/d/main#/console/13208845/276109344/preview) says “You’ll be able to see user retention, installs and active devices when people start using your snap.”. (Obviously we couldn’t use exactly that text, as long as most of those graphs aren’t implemented yet.)

In future I’d also prefer that message to be a centered overlay, rather than a notification. I think notifications should be reserved for transitory and unusual things, not something that is true for hours for every newly-published snap.

lyubomir-popov:
@Lukewh Nothing to add to what @matthewpaulthomas. Is a strip with a heading and a paragraph outside the scope of this?

PR #416, opened 2018-03-19
Market page - only save changed fields

bartaz:
Refactors market page front-end logic to allow saving only data changed by user.

Adds `changes` field to the form which contains JSON with only values changed by the user.

### NOTE:
Needs backend update to use diff instead of form input values, but can be merged for front-end changes only, as is backwards compatible (only adds new hidden data field to the form)

### QA

- ./run
- go to market page
- change some values, save
- market page should be saved as before
- additional `changes` field is sent to backend

### DIFF
--- a/static/js/publisher/market/market.js
+++ b/static/js/publisher/market/market.js
@@ -1,12 +1,24 @@
 import { initSnapScreenshotsEdit } from './screenshots';
 
-function initSnapIconEdit(iconElId, iconInputId) {
+function initSnapIconEdit(iconElId, iconInputId, state) {
   const snapIconInput = document.getElementById(iconInputId);
   const snapIconEl = document.getElementById(iconElId);
 
   snapIconInput.addEventListener("change", function(){
-    const fileList = this.files;
-    snapIconEl.src = URL.createObjectURL(fileList[0]);
+    const iconFile = this.files[0];
+    snapIconEl.src = URL.createObjectURL(iconFile);
+
+    // remove existing icon from state object
+    const images = state.images.filter(image => image.type !== 'icon');
+    // replace it with a new one
+    images.unshift({
+      url: URL.createObjectURL(iconFile),
+      file: iconFile,
+      name: iconFile.name,
+      status: 'new'
+    });
+
+    updateState(state, { images });
   });
 
   snapIconEl.addEventListener("click", function() {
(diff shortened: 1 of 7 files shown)


static/js/publisher/market/market.js
```diff
@@ -35,9 +47,110 @@ function initFormNotification(formElId, notificationElId) {
...
+    config.screenshotsWrapper,
+    state
+  );
+
+  // when anything is changed update the state
+  marketForm.addEventListener('change', function() {
+    updateState(state, new FormData(marketForm));
+  });
+
+  document.querySelector('.js-market-submit').addEventListener('click', function(event) {
```
Lukewh: Is this not a submit event on the form for a reason? What happens when people press return?

Lukewh: Currently it reverts the form!?

bartaz: Yes, both 'Revert' and 'Apply' are submitting, that's why 'Enter' picks up first and 'Reverts' it.

I wanted to have it on `submit`, but because we have 2 submit buttons I wasn't able to tell the difference in `submit` event which button was clicked (without hacks), so `click` seemed to be more reasonable.

Would be good to make 'Enter' do a submit. I guess it would require changing button order maybe, so that 'Apply' is first in markup.

Lukewh: Does making the revert a normal `button` and apply a `type="submit"` button fix that? It would be more semantic too.

templates/publisher/market.html
```diff
@@ -34,8 +34,8 @@ <h1 class="u-no-margin--top">{{ title }}</h1>
 
           <div class="row">
             <div class="col-12 u-align--right">
-                <input type="submit" class="p-button--neutral" name="submit_revert" value="Revert"/>
-                <input type="submit" class="p-button--positive" name="submit_apply" value="Apply"/>
+                <input type="submit" class="p-button--neutral js-market-revert" name="submit_revert" value="Revert"/>
```
Lukewh: If you change this type to `reset` it should reset the form and `return` should submit the form as expected.

bartaz: It seems that Vanilla doesn't like reset buttons (`display: none`).
Also, it will only reset text fields and not icon/screenshots that are custom.

This is not a new issue so if possible I would merge it as it is and think of a better solution in next PR.
Possibly I guess we can make it just a button with click handler that would reload the page (without posting).

Lukewh: :+1:

PR #419, opened 2018-03-20
CLI install command

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/389

- Removed buggy and unsupported 'Install' button
- Added CLI install instruction

# QA

- Pull this branch
- `./run`
- Visit http://0.0.0.0:8004/vlc or visit http://snapcraft.io-pr-419.run.demo.haus/vlc
- Click the copy to clipboard button
- Paste the result somewhere, it should match:
  `sudo snap install vlc`

# Screenshots

## Desktop
screenshot-2018-3-21 vlc linux software in the snap store 1

## Mobile
screenshot-2018-3-21 vlc linux software in the snap store


templates/snap-details.html
```diff
@@ -29,14 +29,28 @@
...
+                  <button
+                    class="p-code-snippet__action"
+                    id="copy"
+                    data-clipboard-target="#snap-install"
+                    onclick="if(typeof ga !== 'undefined'){ga('gtm1.send', {hitType: 'event', eventCategory: 'Snap details', eventAction: 'Copy CLI install button', eventLabel: 'CLI install {{ package_name }}'})}">
+                    Copy to clipboard
+                  </button>
+                </div>
+                <p class="p-form-help-text">
+                To install snaps on your system you'll need to <a href="https://docs.snapcraft.io/core/install">install snapd</a>.
```
Greggless: Update copy to:
>To run snaps on your system you'll need to install snapd

Lukewh: Done :)

Greggless:
Yes @bartaz Linux or no Linux

package.json
```diff
@@ -6,7 +6,7 @@
     "build": "yarn run build-js && yarn run build-css",
     "build-css": "node-sass --include-path node_modules static/sass --output static/css && postcss --use autoprefixer --replace 'static/css/**/*.css' && postcss --use cssnano --dir static/minified 'static/css/**/*.css'",
     "build-js": "yarn run copy-3rd-party-js && yarn run build-js-transpile",
-    "copy-3rd-party-js": "cp node_modules/d3/build/d3.min.js static/js/modules && cp node_modules/d3-geo/build/d3-geo.min.js static/js/modules && cp node_modules/topojson-client/dist/topojson-client.min.js static/js/modules && cp node_modules/billboard.js/dist/billboard.min.js static/js/modules && cp node_modules/moment/min/moment.min.js static/js/modules",
+    "copy-3rd-party-js": "cp node_modules/d3/build/d3.min.js static/js/modules && cp node_modules/d3-geo/build/d3-geo.min.js static/js/modules && cp node_modules/topojson-client/dist/topojson-client.min.js static/js/modules && cp node_modules/billboard.js/dist/billboard.min.js static/js/modules && cp node_modules/moment/min/moment.min.js static/js/modules && cp node_modules/clipboard/dist/clipboard.min.js static/js/modules",
```
bartaz: Oh, we should finally start bundling these modules with rollout...
Not for scope of this PR, but would be nice to do it before we need to add anything more.

templates/snap-details.html
```diff
@@ -29,14 +29,28 @@
...
             <p class="p-media-object__content--large">{{ publisher }}</p>
-
-            <p class="p-media-object__content--large">
-              {% if is_linux %}
-                <a class="p-button--positive" href="snap://{{ package_name }}" onclick="if(typeof ga !== 'undefined'){ga('gtm1.send', {hitType: 'event', eventCategory: 'Snap details', eventAction: 'Clicked Install button', eventLabel: 'Install {{ package_name }}'})}">Install</a>
-              {% else %}
-                You can install this software on a snap-powered system like <a class="p-link--external" href="https://www.ubuntu.com/download">Ubuntu</a>.
-              {% endif %}
-            </p>
+            <div class="row p-media-object__content--large">
```
bartaz: Not sure if it's common Vanilla pattern to add grid classes (like `row` here) to elements of other patterns.
But I guess it's the easiest way to achieve what we want without writing additional styling...

templates/snap-details.html
```diff
@@ -29,14 +29,28 @@
...
+                    class="p-code-snippet__input"
+                    id="snap-install"
+                    value="sudo snap install {{ package_name }}"
+                    readonly="readonly"
+                  />
+                  <button
+                    class="p-code-snippet__action"
+                    id="copy"
+                    data-clipboard-target="#snap-install"
+                    onclick="if(typeof ga !== 'undefined'){ga('gtm1.send', {hitType: 'event', eventCategory: 'Snap details', eventAction: 'Copy CLI install button', eventLabel: 'CLI install {{ package_name }}'})}">
```
caldav: Would you mind moving the event to the `p-code-snippet` div? Not everyone is using a button to select text.

evandandrea:
>Removed buggy and unsupported 'Install' button

For the curious, the context is [here](https://github.com/CanonicalLtd/snapcraft-design/issues/304#issuecomment-374145091).

If we're going to drop the green install button, I'd like to propose we also:
```window.location.replace('snap://firefox');```

This way, Ubuntu users will immediately see GNOME Software with much of the same content.

**Comment by Lukewh** [comment]
@evandandrea that would redirect all users, and those that don't have `snap://` support will be redirected to something like the screenshot below... Pressing the back button will cause the reload to happen again - thus making it impossible for anyone to see the page.

screenshot from 2018-03-21 10-19-20

File: templates/snap-details.html (hunk 4 above)
Comment (Lukewh) [reply templates/snap-details.html]: I originally thought about trying to be clever and check whether the text had been highlighted, but that seemed overkill.

Perhaps 2 events, on on the button (as is) and another on the input `onfocus`?

Comment (caldav) [reply templates/snap-details.html]: I think having 2 events will end up being quite spammy, especially since clicking the copy button is (afaict) moving the focus around. I just did some more testing and it looks like using an `oncopy` event on the `input` will be the less spammy version we can have and will only be triggered on and actual copy to clipboard action. If you agree, please also change the `eventAction` value to "Copy CLI install instructions".

File: package.json (hunk 2 above)
Comment (Lukewh) [reply package.json]: Agreed, this could start getting silly. Especially moving to squid, the http2 benefit of having separate files will go.

File: templates/snap-details.html (hunk 3 above)
Comment (Lukewh) [reply templates/snap-details.html]: Yeah, the media object pattern has some strange quirks. I want to investigate a better way of doing this kind of thing outside of this project to try and propose an improvement.

File: static/sass/_patterns_media-object.scss [hunk 5]
```diff
@@ -0,0 +1,11 @@
+@mixin p-media-object {
+  @include vf-p-media-object;
+
+  .p-media-object {
+    @extend .p-media-object;
```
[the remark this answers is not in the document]
Lukewh: Fixed... thanks

templates/snap-details.html
Lukewh: Added my proposal, let me know what you think

Lukewh:
@bartaz updated mobile view

templates/snap-details.html
Lukewh: Sorry, github didn't update... I've now implemented the oncopy as you suggested

static/sass/_snapcraft_snap-details.scss
```diff
@@ -0,0 +1,11 @@
+@mixin snapcraft-snap-details {
+  @media screen and (max-width: $breakpoint-small) {
+    .snapcraft-snap-details {
+      &__snippet {
+        left: -4.5rem;
```
bartaz: If you do it with `margin-left: -4.5rem` it adapts width automatically without whole calc magic.
Still, a hack that I don't like too much.

Would be better to move install part out of media object and move to separate grid row with a spacing col on desktop.
But this will probably not line up well with the icon above, so...

Hack is faster to implement, so probably good enough for now, and we probably should look into redoing this part of page (when there will be channel maps, badges, etc - it will get more complicated).

templates/snap-details.html
```diff
@@ -29,14 +29,28 @@
...
-            </p>
+            <div class="row p-media-object__content--large snapcraft-snap-details__snippet">
+              <div class="col-6">
+                <div class="p-code-snippet">
+                  <input
+                    class="p-code-snippet__input"
+                    id="snap-install"
+                    value="sudo snap install {{ package_name }}"
+                    readonly="readonly"
+                    oncopy="if(typeof ga !== 'undefined'){ga('gtm1.send', {hitType: 'event', eventCategory: 'Snap details', eventAction: 'Copy CLI install instructions', eventLabel: 'Copy {{ package_name }}'})}">
```
bartaz: Does `oncopy` this work (crossbrowser)?
Does copying through `clipboard` script trigger it?

lyubomir-popov:
I'm like @bartaz 's suggestion and am ok with this going out as he has it in the screenshot above. 

As a side note, I played around a bit with the styling to try and balance things better, but this is outside of the scope of the current pr and more for a separate discussion around the channel maps and making the code snippet the focal point of the header.

image

bartaz:
@lyubomir-popov We can't assume all snaps have nice and short names like Firefox does, do I believe we need to reserve much more space for title there.

templates/snap-details.html
Lukewh: I've changed this to use the built-in event for the copy button and added some wonderful event js for the input instead. I don't think there's any way around having two events @caldav...
MDN says oncopy is 'non-standard' https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/oncopy so we should avoid using it.

static/sass/_snapcraft_snap-details.scss
Lukewh: Done!

templates/snap-details.html
caldav: Ok, let's see how it goes. The likeliness of triggering both is lower than before, thanks to this creative function :)

templates/snap-details.html
```diff
@@ -101,6 +115,47 @@ <h4>Where people are using {{ snap_title }}</h4>
       whitelistUrls: ['snapcraft.io']
     }).install();
     Raven.context(function () {
+      if (typeof ga !== 'undefined') {
```
bartaz: GA is loaded asynchronously so in many cases this will run before QA is loaded and will silently fail.

We should only check for GA at the moment we attempt to use it and initialise everything else normally to have event handlers in place regardless of GA being defined.

templates/snap-details.html
```diff
@@ -101,6 +115,47 @@ <h4>Where people are using {{ snap_title }}</h4>
...
+
+        function keyDown(e) {
+          var key = e.key.toLowerCase();
+          if (key === 'control' || key === 'meta') {
+            ctrlDown = true;
+          }
+        }
+
+        function keyUp(e) {
+          var key = e.key.toLowerCase();
```
bartaz: This doesn't work on Mac.

When 'Cmd' button (meta) is held clicking other buttons (like C) doesn't trigger keyup. This makes it harder to detect.
Keydown is triggered normally when Cmd is held, so possibly, if we move `if ((key === 'c' || key === 'x') && ctrlDown)` to keydown handler it should work...

lyubomir-popov:
@bartaz fair point about the name length. I've taken that into account in the mockup here https://github.com/CanonicalLtd/snapcraft-design/issues/261#issuecomment-364902244

evandandrea:
>those that don't have snap:// support will be redirected to something like the screenshot below... 

And I take it that's what led you to trying it in an iframe, which didn't produce the desired results across all browsers. Shame.

cassidyjames:
Might I share some unsolicited feedback? It would be handy to include an install link that only shows if the reported OS is Linux, then it displayed a modal with a little more information and a button to `snap://`. This way users on those platforms still have a way to install snaps, and apps like [Snaptastic](https://github.com/bartzaalberg/snaptastic) aren't rendered completely useless.

>## Install on Ubuntu or another [supported OS](#)
>If you already have snap and a store, you can always search for `project name` from the store, or use the button here.
>
>**[Install Foo](snap://foo)**

Lukewh:
@cassidyjames thanks for the feedback. You'll be pleased to know that's essentially our current thinking. We don't want to make GUI's useless but felt it was way worse in the short term to mislead people.

Our wording is also scarily similar to what you've suggested.

PR #422, opened 2018-03-20
Use talisker

nottrobin:
QA
--

``` bash
./run clean
./run
```

Check site works. Check other ./run commands.


Dockerfile
```diff
@@ -6,19 +6,19 @@ RUN apt-get update && apt-get install --yes python3-pip libsodium-dev
 # Python dependencies
 ENV LANG C.UTF-8
 RUN pip3 install --upgrade pip
-RUN pip3 install gunicorn
-
-# Set git commit ID
-ARG COMMIT_ID
-ENV COMMIT_ID=${COMMIT_ID}
-RUN test -n "${COMMIT_ID}"
+RUN pip3 install talisker pipenv
```
WillMoggridge: I think it would be cleaner/more efficient to remove the Talisker dependency in this file and install it from the requirements file.

nottrobin: Yes. I think I did actually add it to requirements. So I don't think it's actually needed here, nor is pipenv any more.

nottrobin: Fixed I believe

PR #423, opened 2018-03-20
Extend timeout to 3 seconds

nottrobin:
We've been getting a number of api timeout errors in production, intermittently.

This should hopefully help to reduce the chance of that happening, without allowing the API to bring down the site by waiting for too long.

QA
--

`./run`, check the site works

### DIFF
--- a/modules/cache.py
+++ b/modules/cache.py
@@ -32,11 +32,11 @@ def get(
             json=json,
             files=files,
             data=data,
-            timeout=2
+            timeout=3
         )
     except requests.exceptions.Timeout:
         api_error_exception = ApiTimeoutError(
-            'The request to {} took longer than 2 seconds'.format(url),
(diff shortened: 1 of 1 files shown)


modules/cache.py
```diff
@@ -32,11 +32,11 @@ def get(
             json=json,
             files=files,
             data=data,
-            timeout=2
+            timeout=3
```
Lukewh: smells like a variable

modules/cache.py
```diff
@@ -32,11 +32,11 @@ def get(
...
         )
     except requests.exceptions.Timeout:
         api_error_exception = ApiTimeoutError(
-            'The request to {} took longer than 2 seconds'.format(url),
+            'The request to {} took longer than 3 seconds'.format(url),
```
Lukewh: that can be used here too

PR #427, opened 2018-03-21
Use default talisker prometheus endpoint

nottrobin:
We were exporting prometheus metrics on port 9990. But now that we use
talisker, it has its own prometheus endpoint at /_status/metrics. We
should now switch to using this endpoint, which uses the
TALISKER_NETWORKS environment variable to only show the endpoints to
specific IPs.

A review of this PR effectively constitutes a review of https://github.com/canonical-webteam/generator-canonical-webteam/pull/66, so you could maybe approve that one too if this one is okay.

QA
--

``` bash
./run --env TALISKER_NETWORKS=172.16.0.0/12
```

Go to http://127.0.0.1:8004/, browse around a bit, then go to http://127.0.0.1:8004/_status/metrics and see the metrics.

Try out all other `./run` commands.


Lukewh:
I can confirm it didn't work when I tried to access the endpoint on the demo, from home.

PR #428, opened 2018-03-21
Snap details screenshots

### LINKED ISSUE #344: Screenshots missing from directory snap page
Screenshots are shown in clients such as Gnome Software, but they are not shown on directory pages such as [snapcraft.io/skype](https://snapcraft.io/skype).

[Originally reported by “NeoTheThird” in [LP#1752296](https://launchpad.net/bugs/1752296).

bartaz:
Fixes #344

Adds screenshots view to snap details page.

### QA

- ./run
- go to snap detail page of a snap with screenshots
- screenshots should be visible
- click on a screenshot - lightbox should be opened
- if there is more screenshots you should be able to scroll through them with buttons
- go to snap without screenshots - screenshots should not be rendered
- try to disable JavaScript
- go to snap with screenshots
- screenshots should still be displayed correctly (without buttons) and should be scrollable with mouse scroll

screen shot 2018-03-21 at 18 03 45
screen shot 2018-03-21 at 18 32 48


Lukewh:
Spacing below looks a bit large and on mobile the next/ previous arrows aren't vertically centre.
Could we rely on swiping on mobile?
screenshot-2018-3-21 atom linux software in the snap store

static/sass/_snapcraft_details_screenshots.scss
```diff
@@ -0,0 +1,103 @@
...
+
+      display: flex;
+      flex-wrap: nowrap;
+      overflow-x: auto;
+      overflow-y: hidden;
+    }
+
+    .p-screenshot {
+      flex: 0 0 auto;
+      height: 240px;
```
Lukewh: `max-height` maybe?

bartaz:
@Lukewh Mobile seems fine in Chrome: 

screen shot 2018-03-21 at 18 47 59

but yes, it's off in Firefox:

screen shot 2018-03-21 at 18 47 49

I'll check what to do with that.

caldav:
When hovering a screenshot, an "enlarge" icon appears and my first instinct was to click on it, but clicking on the icon itself doesn't do anything, which is a bit confusing.

bartaz:
@caldav Good catch, thanks.

static/sass/_snapcraft_details_screenshots.scss
bartaz: Tried that, but it causes some issues.
First of all layout 'jumps' before images load, because they don't have any height until they load.
And there are some flexbox issues in Firefox. It seems like it messes up the size of this element when it doesn't have a fixed height given in CSS.

Lukewh: Hmm, that's annoying - can we init the screenshots UI after the final image has loaded? That way it should all be based on the final sizes (unless you resize the browser like a madm'n)

static/sass/_patterns_maas_modal.scss
```diff
@@ -63,7 +63,7 @@ $color-modal-controls: rgba($color-x-dark, .2);
 
   // circle preloader
   .vbox-preloader {
-    background-image: url('preload-circle.png');
+    // background-image: url('preload-circle.png'); // it's 404... on MAAS as well
```
Lukewh: Just delete this line.

modules/public/views.py
```diff
@@ -200,6 +200,11 @@ def snap_details(snap_name):
     except ApiError as api_error:
         status_code, error_info = _handle_errors(api_error)
 
+    # filter out banner and banner-icon images from screenshots
+    screenshot_urls = [
```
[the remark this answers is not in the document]
nottrobin: Yep looks perfect. :+1:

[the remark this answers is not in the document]
nottrobin: Actually thinking about it just a touch more, it might be nice to call it something other than `screenshot_urls`, just to make it clear that it's not *exactly* the same as the `screenshot_urls` set that comes from the API.

static/sass/_snapcraft_details_screenshots.scss
bartaz: We can, but what would be the benefit of doing that?

If we don't set the size of screenshots in CSS it will render as nothing until images start popping up which will make the whole page jump. Waiting for last image to load would not change that.

These are screenshots, apart from some weird cases they will be above 240px, so what's the point of using `max-height`?

matthewpaulthomas:
I see some tiny quirks, but nothing worth blocking landing. Nice work.

nottrobin:
@matthewpaulthomas maybe file issues for those quirks?

lyubomir-popov:
looks good @bartaz, just one thing - the screenshot arrows disappear < 768, shouldn't p-screenshots__wrapper have overflow: scroll instead of auto? SO you can scroll with the mouse?

bartaz:
@lyubomir-popov Yes, it should work exactly as you describe. When buttons disappear it should be scrollable.
Unless there is some bug in some case, but it's the same breakpoint in media query that switches between scrolling and showing buttons.

matthewpaulthomas:
> @matthewpaulthomas maybe file issues for those quirks?

Reported #455 and #456.

PR #435, opened 2018-03-22
Remove flask dependency from validation

tbille:
# Refactoto S02E06 - The validation

~~This PR should be merged before: https://github.com/canonical-websites/snapcraft.io/pull/431~~

## Summary

- Remove last flask dependency from publisher API module
- Remove useless validation checks for the frontend. Only check if the macaroon needs to be refreshed in the headers. If yes raise an exception to handle it.

## QA

- `./run`
- http://0.0.0.0:8004/account/
- Should work as usual

Hard to QA because the macaroon is valid for 1 year...

### DIFF
--- a/modules/authentication.py
+++ b/modules/authentication.py
@@ -88,36 +88,6 @@ def request_macaroon():
     return response.json()['macaroon']
 
 
-def verify_macaroon(root, discharge, url):
-    """
-    Submit a request to verify a macaroon used for authorization.
-    Returns the response.
-    """
-    authorization = get_authorization_header(root, discharge)
-    url = ''.join([
-        DASHBOARD_API,
-        'acl/verify/',
-    ])
-    response = requests.request(
-        url=url,
-        method='POST',
-        json={
-            'auth_data': {
-                'authorization': authorization,
-                'http_uri': url,
-                'http_method': 'GET'
-            }
-        },
-        headers={
-            'Accept': 'application/json, application/hal+json',
-            'Content-Type': 'application/json',
-            'Cache-Control': 'no-cache',
-        }
-    )
-
-    return response.json()
-
-
 def get_refreshed_discharge(discharge):
     """
     Get a refresh macaroon if the macaroon is not valid anymore.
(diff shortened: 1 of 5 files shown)


modules/authentication.py
```diff
@@ -169,11 +170,11 @@ def verify_response(
             session['macaroon_discharge']
```
nottrobin: The docstring above needs updating to reflect the new behaviour of this function

modules/authentication.py
```diff
@@ -169,11 +170,11 @@ def verify_response(
...
         )
 
-        return {
-            'status_code': 307,
-            'redirect': url_calling,
-            'reason': 'Macaroon discharge refreshed'
-        }
+        raise ApiVerifyResponseError(
```
nottrobin: I don't think checking if the macaroon needs refreshing should happen in the same function that we check for other general errors. And I'm also not sure that a Exception is the best way to communicate this state, or things like URLs. I haven't understood this code well enough to provide concrete suggestions, but I think probably we should have a "check authorization" step somewhere which is where things like macaroon freshness should be checked.

modules/authentication.py
```diff
@@ -185,24 +186,24 @@ def verify_response(
...
         # Not authorized content
         if response.status_code == 401 and not verified['allowed']:
-            return {
-                'status_code': 401,
-                'redirect': url_account,
-                'reason': 'Not authorized'
-            }
+            raise ApiVerifyResponseError(
+                'Not authorized',
+                401,
```
nottrobin: Could you write this as `status_code=response.status_code`? Firstly, I think it's helpful to use named arguments if you can because it helps to clarify how the function might be using that argument. Secondly, I think we should make it clear that this API library is never simply choosing status_codes to pass back itself - whenever an ApiError has a status_code property it's because that status_code came directly from the API itself.

modules/authentication.py
```diff
@@ -185,24 +186,24 @@ def verify_response(
...
+            )
 
         # The package doesn't exist
         if verified['account'] is not None and response.status_code == 404:
-            return {
-                'status_code': 404,
-                'redirect': url_account,
-                'reason': 'Not found'
-            }
+            raise ApiVerifyResponseError(
```
nottrobin: Actually all we're basically doing in this function, if you remove the macaroon refresh stuff, is pass request errors back to the client. Maybe it's not even worth wrapping these - we should simply do `response.raise_for_status` and let the client catch them? Or do you think there's an argument for wrapping it in our own error type?

If we do wrap it in our own error type, couldn't it be generic? E.g.:

if response.status_code >= 400:
    raise ApiResponseError(
        "The API returned a bad response",
        response=response
    )

modules/publisher/api.py
```diff
@@ -1,11 +1,11 @@
-import flask
 import modules.authentication as authentication
 import modules.cache as cache
 import os
 from modules.exceptions import (
     ApiResponseDecodeError,
     ApiResponseError,
     ApiResponseErrorList,
+    MacaroonNeedsRefresh
```
WillMoggridge: Minor nitpick on name, I think "requires" is better than "needs".
I'm also wondering if it sounds better as `MacaroonRefreshRequired`. Then it feels namespaced and could do `MacaroonRefreshFailed` and similar things if you needed.

I really like using Exceptions to broadcast the state! :+1:

modules/publisher/views.py
```diff
@@ -12,44 +13,50 @@
...
-            else:
-                error_list.append(error)
-
-    user_snaps = []
-    if '16' in account['snaps']:
-        user_snaps = account['snaps']['16']
-
-    flask_user = flask.session['openid']
+    try:
+        account = api.get_account(flask.session)
```
WillMoggridge: Does the whole function need to be in a try block or can it be limited to a smaller section, maybe this line?

tbille: For now, yes because this part is not totally refactored yet

modules/publisher/api.py
tbille: `MacaroonRefreshFailed` will be handled in another PR when I arrive on the refactor of this particular file

modules/publisher/api.py
```diff
@@ -127,18 +108,8 @@ def get_account(session):
...
-        ACCOUNT_URL,
-        '/account',
-        '/login',
-    )
-
-    if verified_response is not None:
-        return {
-            'redirect': verified_response
-        }
+    if authentication.verify_headers(response.headers):
```
WillMoggridge: This function name is pretty confusing but that may be a subject for a different PR

PR #442, opened 2018-03-23
Defend against undefined vars

Lukewh:
# Done

Wrapped everything with `if` statements.

A small number of browsers (IE) seem to support `.key` but provides strings from an older spec.... However looking at the shim (https://github.com/shvaikalesh/shim-keyboard-event-key/blob/master/index.js) we're using none of the strings effected.

### DIFF
--- a/templates/snap-details.html
+++ b/templates/snap-details.html
@@ -138,44 +138,53 @@
       var ctrlDown = false;
 
       function keyDown(e) {
-        var key = e.key.toLowerCase();
-        if (key === 'control' || key === 'meta') {
-          ctrlDown = true;
-        }
-        if ((key === 'c' || key === 'x') && ctrlDown) {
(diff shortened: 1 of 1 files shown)


templates/snap-details.html
```diff
@@ -138,44 +138,55 @@ <h4>Where people are using {{ snap_title }}</h4>
...
-        if (typeof ga !== 'undefined') {
-          ga('gtm1.send', {
-            hitType: 'event',
-            eventCategory: 'Snap details',
-            eventAction: 'Copy CLI install instructions',
-            eventLabel: 'Copy {{ package_name }} button'
-          });
-        }
-      });
+      snapcraft.public.map('#js-snap-map', {{ countries|tojson }});
```
bartaz: This is a duplicated like.
We set up screenshots and map already in the beginning of this script.

PR #444, opened 2018-03-26
Fixes screenshots JS errors from sentry

bartaz:
There was a 'bunch' of errors in Sentry cased by recent public screenshots release.

Most common are 2 types:

- `screenshotsEl` is null - very common case because it happens when snap doesn't have any screenshots, in such case we should not try to run screenshots related code...

- `Array.from` is not a function - it seems that we still have quite a lot of people with older or not mainstream browsers that don't yet support this function, so we need to get back to older solutions

This PR addresses both of issues listed above.

### QA

- ./run
- go to a public snap page with screenshots (everything should work as expected, no errors in console)
- go to a public span page without screenshots (everything should work, no errors in console)


Lukewh:
Screenshots still work with the changes thanks @bartaz QA +1

PR #451, opened 2018-03-26
Update search results

Lukewh:
# Done

Fixes https://github.com/canonicalltd/snapcraft-design/issues/301

- Removed left padding from publisher name
- Removed description

# QA

- Pull this branch
- `./run`
- Visit http://0.0.0.0:8004/search?q=test and http://0.0.0.0:8004/search?q=lukewh
- Ensure the description is not there and the publisher name is left aligned with the snap name

### DIFF
--- a/static/sass/_patterns_media-object.scss
+++ b/static/sass/_patterns_media-object.scss
@@ -7,5 +7,9 @@
     &__details {
       flex-grow: 1;
     }
+
+    &__meta-list-item {
+      padding-left: 0;
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


Greggless:
# YAH BOY 👨‍🎤 🐉 :shipit:

static/sass/_patterns_media-object.scss
```diff
@@ -7,5 +7,9 @@
     &__details {
       flex-grow: 1;
     }
+
+    &__meta-list-item {
```
bartaz: Maybe something we should report to Vanilla. Having only meta data without icons looks misaligned.

PR #453, opened 2018-03-26
Tests account detail page

tbille:
# Summary

- Now the tests are run in the `package.json`
- Keep basic tests for the index page
- Add entire integration test suite for the /account page

# QA

- `./run test`
The tests should pass!


tbille:
@WillMoggridge @nottrobin 

The idea here is to do integration tests. We will check that all the layers are well integrated together. To check if everything works, we want to make sure that the right data is sent to the function `flask.render_template`.

Will that be enought for now?

PR #458, opened 2018-03-27
Fix width of empty screenshots box on market page

bartaz:
### QA
- ./run
- go to snap market page that doesn't have screenshots
- 'Add images' grey field should have width of the page not whole screen


Greggless:
# Heroic victory ⚔️ 
image

Issue #399, opened 2018-03-15
Market page validation UI (marking invalid fields)


bartaz:

matthewpaulthomas:
[Draft design](https://docs.google.com/document/d/1PQzECKRrVZ-y3W7DcdkfeZipBVr5Xl2mpMj5Vqhux2E/edit#heading=h.88kyurq00o04) for the text fields:
> Title, Developer Web site, and Developer contact are [editable only on this page](https://docs.google.com/document/d/1PQzECKRrVZ-y3W7DcdkfeZipBVr5Xl2mpMj5Vqhux2E/edit#heading=h.7g4zasx003xn), while Summary and Description are [also editable in snapcraft.yaml](https://docs.google.com/document/d/1PQzECKRrVZ-y3W7DcdkfeZipBVr5Xl2mpMj5Vqhux2E/edit#heading=h.7g4zasx003xn). For all five:
> - If you enter any disallowed character, it should be silently discarded.
> - Whenever the field contains more than the maximum number of characters, the excess characters should not be discarded (because the eventual value will likely include some of them). Instead, the field should be [highlighted as an error](https://docs.vanillaframework.io/en/base/forms#validation-classes), and the excess number of characters should be overlaid in the error color inside the end of the field. (The meaning of this number is obvious from how it changes after any edit.)
> - Whenever the field contains fewer than the minimum _or_ more than the maximum number of characters, the Apply button should be disabled.

Does this seem reasonable?

PR #460, opened 2018-03-27
Adds basic Vanilla style validation to market input fields

bartaz:
First pass of #399 

This PR adds basic marking of invalid fields with Vanilla styling.

Possible improvements for next PRs:
- showing error messages next to fields instead of top notification

### QA
- ./run
- go to market page
- update any of text fields to invalid value (make title or summary too long, make contact or website invalid URL)
- save the form (click `Apply`)
- Error notification should show up on top of the form, invalid field should be marked red
- Change the value in invalid field
- Field should be unmarked when you leave it

screen shot 2018-03-27 at 15 37 02

### DIFF
--- a/static/js/publisher/market/market.js
+++ b/static/js/publisher/market/market.js
@@ -50,7 +50,21 @@ function initFormNotification(formElId, notificationElId) {
 
 
 
-function initForm(config, initialState) {
+function initForm(config, initialState, errors) {
+  // if there are errors mark fields as invalid
+  if (errors && errors.length) {
+    errors.forEach((error) => {
+      if (error.code === 'invalid-field') {
+        const name = error.extra.name;
+        if (name) {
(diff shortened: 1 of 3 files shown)


Greggless:
Superb. We can make these notifications more human friendly by saying something along the lines of:
> The summary must be less than 128 characters ❗️ 

@matthewpaulthomas is working on https://github.com/CanonicalLtd/snapcraft-design/issues/180 which may change how we present these notifications and how they work so we should merge, and then return to these

bartaz:
@Greggless These error messages come from API currently, so we don't have direct control on what they say.

Currently the work is focused mostly on making what we have more user friendly without big changes.
When we move Market page into editing by one field model after 18.04 we may have more control on how to present errors.

Also, next thing to do is simply limiting fields max length properly, so users can't enter too long data or invalid URLs. This should make it much less likely for users to see these errors from API.

PR #461, opened 2018-03-28
Polyfill for Array.from and Element.closest

bartaz:
Because usage of `Array.from` is causing us some sentry errors and possibly `Element.closest` can also be problematic in older browsers we can add a polyfill for those functions.

This will allow us to use new DOM API without loosing compatibility with older browsers.

NOTE: Possibly we could use `babel-polyfill` which should include these, but it also includes many more that may make our files unnecessary big at the moment. Also I had some troubles with using it in our current rollup setup.

### QA
- ./run
- go to snap details page (with screenshots)
- everything should work normally
- try it in older browser (possibly IE) still there should be no errors in console

### DIFF
--- a/static/js/polyfills.js
+++ b/static/js/polyfills.js
@@ -1,0 +1,101 @@
+// Array.from
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from
+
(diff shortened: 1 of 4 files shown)


Lukewh:
#QA good

On Chrome 38 on Windows7 snapcraft.io has errors, the demo does not!

Issue #436, opened 2018-03-22
Add toggles to market page


Lukewh:
- [ ] Global `public_metrics_enabled`
- [ ] One for each public metrics shown
- [ ] If global toggle is off, disable list of other metrics

Field names:
`public_metrics_enabled`
`public_metrics_blacklist`

PR #462, opened 2018-03-28
Public metrics toggles

Lukewh:
# Done
Fixes https://github.com/canonical-websites/snapcraft.io/issues/436 and https://github.com/canonical-websites/snapcraft.io/issues/443

- Added Public Metrics global checkbox and Territories specific checkbox to the Market page
- Made the backend pass the data through

# QA

- Pull this branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/<snapname>/market
- Scroll down to the bottom and see the checkboxes
- When toggling the public metrics checkbox the items below should be disabled
- If you disable all metrics but have a specific metric checked, the specific metric should still be checked, but disabled
- Save changes and check the state when the page reloads

# Screenshots

## Nothing selected
screenshot-2018-3-28 market details for lukotron

## Global selected
screenshot-2018-3-28 market details for lukotron 1

## Global and individual selected
screenshot-2018-3-28 market details for lukotron 2

## Global not selected and individual selected
screenshot-2018-3-28 market details for lukotron 3

## Potential future when distro graph is implemented
screenshot-2018-3-28 market details for lukotron 4

### DIFF
--- a/app.py
+++ b/app.py
@@ -69,6 +69,21 @@ def inject_template_variables():
     }
 
 
+# Template helpers
+@app.context_processor
+def utility_processor():
+    def contains(arr, str):
+        return str in arr
+
+    def join(arr, str):
+        return str.join(arr)
+
+    return {
+        'contains': contains,
+        'join': join
+    }
+
+
 def login_required(func):
     """
     Decorator that checks if a user is logged in, and redirects
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -260,7 +260,9 @@ def get_market_snap(snap_name):
         "publisher_name": snap_details['publisher_name'],
         "screenshot_urls": snap_details['screenshot_urls'],
         "contact": snap_details['contact'],
-        "website": snap_details['website']
+        "website": snap_details['website'],
+        "public_metrics_enabled": snap_details['public_metrics_enabled'],
+        "public_metrics_blacklist": snap_details['public_metrics_blacklist'],
     }
 
     return flask.render_template(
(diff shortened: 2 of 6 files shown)


Greggless:
# :octopus:

Super de duper. I'll give some UX recommendations here, and we can pick it up later.

+1 for now, though
<hr>

## UX recommendations
- Use a switch
- Place it to the right of a label
- saying: 'Display public metric charts' 
  - Maybe with a tooltip or text saying something along the lines of: 'Does not disclose any usage numbers for your application
- Hide individual graph options when primary switch is of
- Use checkboxes for each of the graphs
- Provide a little context for what each of the graphs are
  - Terrotories: Displays where you snap is being used by country
  - Distributions: Displays the spread of users across distributions

Lukewh:
Thanks @Greggless.

I originally used a switch to the left of the label, but I figure you have to click 'Apply' for the changes to take effect so it might make people think the change has taken effect when it hadn't.
I also originally implemented hiding the individual items but because you can have them set to 'on' when the global one is 'off' it's nice to see what you're going to enable by switching metrics on, though this isn't as much of an issue as I thought, as the change to the global state doesn't take effect immediately.
Yeah I think some wording around what each thing is and what it does would be really useful.
I'll make the change to the 'Public metrics' label that you suggested and ping when it's done.

modules/publisher/views.py
```diff
@@ -423,7 +441,9 @@ def post_market_snap(snap_name):
                 "screenshot_urls": snap_details['screenshot_urls'],
                 "contact": snap_details['contact'],
                 "website": snap_details['website'],
-                "error_list": error_list
+                "error_list": error_list,
+                "public_metrics_enabled": details_metrics_enabled,
+                "public_metrics_blacklist": details_blacklist
```
tbille: please keep `error_list` at the end

Greggless:
Ah, that's a good point on the `apply`.

This depends on how @matthewpaulthomas  implements the [market page revisions](https://github.com/canonicalltd/snapcraft-design/issues/180), specifically the general interaction model for each editable thing in the listing

app.py
```diff
@@ -69,6 +69,21 @@ def inject_template_variables():
     }
 
 
+# Template helpers
+@app.context_processor
+def utility_processor():
```
tbille: I am not a big fan of passing functions like this in the template. There is no other way of processing the data before sending to the template?

modules/publisher/views.py
Lukewh: Done

Lukewh:
@Greggless yeah I think when we have individually edited inputs, we can do some nicer things

app.py
Lukewh: I figure these are 'mission critical' for anything other than input values/ display so was better suited to be in the view layer - could definitely add some helper functions somewhere and do it in the template.

Lukewh: Actually, I'd have to do it for every metric added so would have to add a `is_{metric}_metric_enabled` for every metric we added to the context... Seems overkill for checking whether an array contains an item.

Lukewh:
Merging @Greggless have created https://github.com/CanonicalLtd/snapcraft-design/issues/309 for the UX things

PR #467, opened 2018-03-29
Show validation errors next to input fields

### LINKED ISSUE #399: Market page validation UI (marking invalid fields)

bartaz:
Fixes #399 

- Shows validation errors next to input fields (or in notification in case of icon or screenshots)
- When there is validation error the invalid value is shown in the input, so user understands what caused the error
- Focuses first invalid input (so that the page scrolls down when error is in contact or website)

### QA
- ./run or http://snapcraft.io-pr-467.run.demo.haus/
- go to market page of any snap
- enter invalid title (longer then 64 letters) and Apply
- Title should be marked invalid with message under the input
- Invalid title should still be the value of the field
- Page title header should show old title (as from API), not invalid one
- Edit the title, when focus moves out of the field it should be not marked as error anymore
- Add invalid icon (bigger then 256px) or invalid screenshot (SVG), Apply
- Error should show in notification next to icon or screenshots
- Enter invalid website (invalid URL), Apply
- Website should be marked invalid, field should be focused and page scrolled so first error is visible
- Click `Revert`
- All changes should be reverted to values from the store

screen shot 2018-03-29 at 11 11 57

### DIFF
--- a/static/js/publisher/market/market.js
+++ b/static/js/publisher/market/market.js
@@ -54,9 +54,12 @@ function initFormNotification(formElId, notificationElId) {
 function initForm(config, initialState, errors) {
   // if there are errors focus first error
   if (errors && errors.length) {
-    const errorInput = document.querySelector('.is-error input');
+    // find input with error or error notification and scroll it into view
+    const errorInput = document.querySelector('.is-error input')
+      || document.querySelector('.p-notification--negative');
+
     if (errorInput) {
-      errorInput.focus();
+      errorInput.scrollIntoView();
     }
   }
 
--- a/static/js/publisher/market/market.js
+++ b/static/js/publisher/market/market.js
@@ -59,7 +59,11 @@ function initForm(config, initialState, errors) {
       || document.querySelector('.p-notification--negative');
 
     if (errorInput) {
-      errorInput.scrollIntoView();
+      errorInput.focus();
+
+      if (errorInput.scrollIntoView) {
+        errorInput.scrollIntoView();
+      }
     }
   }
 
--- a/static/js/publisher/market/market.js
+++ b/static/js/publisher/market/market.js
@@ -52,17 +52,12 @@ function initFormNotification(formElId, notificationElId) {
 
 
 function initForm(config, initialState, errors) {
-  // if there are errors mark fields as invalid
+  // if there are errors focus first error
   if (errors && errors.length) {
-    errors.forEach((error) => {
-      if (error.code === 'invalid-field') {
-        const name = error.extra.name;
-        if (name) {
-          const input = document.querySelector(`[name=${name}]`);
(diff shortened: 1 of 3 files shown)


Greggless:
Nice! It feels much more robust now. 

#### A couple of points:
> Edit the title, when focus moves out of the field it should be not marked as error anymore

This should only be true when the title is <64 char's. Ideally we'd validate on blur as the user moves through the fields. But I think this will be an enhancement 🏅 for the next round 😄.

The screenshot notification is a little odd. A user would have scrolled up to click `Apply` If they don't scroll down they would miss the error and move on with their life, never knowing the mistakes they have made 😭 

This will be solved when we introduce the 'staging' thingy to validate the image as it is uploaded. For now I think it would be best to scroll to the error as the behaviour for the other fields. (This won't work if there are multiple errors). 

I'm sure MPT has a plan for all of these validations, which I'll catch up with him later about, I'll invite you along too 

Thanks

# 🦑

bartaz:
@Greggless Thanks for comments

> This should only be true when the title is <64 char's. Ideally we'd validate on blur as the user moves through the fields. But I think this will be an enhancement 🏅 for the next round 😄.

Yes, ideally it could work like that but at this point it would mean implementing additional validation logic on client side that does the same as backend.
Hopefully it will be not needed, because we can easily add length limit on the fields and simple checks for valid URLs, so it will be much harder for user to enter invalid data into the form.
Also, some simple image checks should also be possible.
This should make sure that in most cases users will not be able to enter and post invalid data and will be much less likely to see these validation errors from API.

> The screenshot notification is a little odd. A user would have scrolled up to click Apply If they don't scroll down they would miss the error and move on with their life, never knowing the mistakes they have made 😭

Good point. I haven't noticed that on my big screen ;) I updated the script so now it should scroll first error to view (input or screenshots notification).

static/js/publisher/market/market.js
```diff
@@ -52,17 +52,15 @@ function initFormNotification(formElId, notificationElId) {
...
-          input.closest('.p-form-validation').classList.add('is-error');
-        }
-      }
-    });
+    // find input with error or error notification and scroll it into view
+    const errorInput = document.querySelector('.is-error input')
+      || document.querySelector('.p-notification--negative');
+
+    if (errorInput) {
+      errorInput.scrollIntoView();
```
Lukewh: This will probably cause errors in older browsers, but worth going - maybe a check around it to check scrollIntoView is defined? Not the worst thing if the view doesn't scroll to the input.

Lukewh:
The `scrollIntoView` isn't working if I change the contact info to be invalid, it just jumps here:
screenshot-2018-3-29 market details for lukotron

Lukewh:
Sorted thanks @bartaz

Issue #437, opened 2018-03-22
Handle public `status: NO DATA` metrics on the front end


Lukewh:
Currently if the map doesn't load you still see the title, this should be handled.

cprov:
@Lukewh The was a slightly change in the spec, we are simply reusing `status: NO DATA` instead of a new `FORBIDDEN` there is no reason to put publisher on the spotlight because they decided to not show a metric.

Lukewh:
Thanks @cprov updated title!

PR #468, opened 2018-03-29
Don't display map when 'NO DATA'

Lukewh:
# Done
Fixes https://github.com/canonical-websites/snapcraft.io/issues/437

- Wrapped map initialiser with `try... catch` to stop displaying territories title if there's a JS error
- Made the backend return an empty object if there's `NO DATA` from the API

# QA

- Pull this branch
- `./run`
- Disable public metrics on a snap
- Visit the snap details
- The map will not be displayed
BONUS:
- If you have a snap without any users but public metrics turned on
- Visit the snap details and see no map or title

# Screenshots

## Public metrics on
screenshot-2018-3-29 lukotron linux software in the snap store

## Public metrics off
screenshot-2018-3-29 lukotron linux software in the snap store 1

### DIFF
--- a/templates/snap-details.html
+++ b/templates/snap-details.html
@@ -95,7 +95,7 @@
       </div>
     </div>
     {% if countries %}
-      <div class="row" style="margin-top: 2rem">
+      <div class="row js-snap-map-holder" style="margin-top: 2rem;">
         <div class="col-12">
           <h4>Where people are using {{ snap_title }}</h4>
           <div id="js-snap-map" class="snapcraft-territories"></div>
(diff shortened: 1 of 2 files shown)


templates/snap-details.html
```diff
@@ -133,7 +133,15 @@ <h4>Where people are using {{ snap_title }}</h4>
...
       snapcraft.public.screenshots('#js-snap-screenshots');
-      snapcraft.public.map('#js-snap-map', {{ countries|tojson }});
+      try {
+        var mapHolder = document.getElementById('js-snap-map');
+        if (mapHolder) {
+          snapcraft.public.map(mapHolder, {{ countries | tojson }});
+        }
+      } catch(e) {
+        Raven.captureException(e);
+        document.querySelector('.js-snap-map-holder').style.display = 'none';
```
bartaz: I think it may be possible for this line to cause error when there is no`js-snap-map-holder`.

If `countries` is empty, map container would not be rendered in HTML.
So rendering map will fail, but also there would be nothing to set display none on.

Ideally we could possibly not render the piece of JS that tries to render the map if we know that it's disabled.

Lukewh: Fixed! Thanks @bartaz

PR #471, opened 2018-03-29
Return 404 on invalid snap name

tbille:
# Summary 

- When requesting invalid snap name return 404
- Add /favicon.ico endpoint

# QA

- `./run`
- http://0.0.0.0:8004/robots.txt
- Should return 404 with message: Invalid snap name
- http://0.0.0.0:8004/toto
- Should work as usual
- http://0.0.0.0:8004/favicon.ico
- Should redirect to https://assets.ubuntu.com/v1/fdc99abe-ico_16px.png

### DIFF
--- a/app.py
+++ b/app.py
@@ -22,6 +22,7 @@ from werkzeug.contrib.fixers import ProxyFix
 import prometheus_flask_exporter
 
 import modules.authentication as authentication
+import modules.helpers as helpers
 import modules.public.views as public_views
 import modules.publisher.views as publisher_views
 from modules.macaroon import (
@@ -34,6 +35,7 @@ talisker.flask.register(app)
 app.wsgi_app = ProxyFix(app.wsgi_app)
 app.secret_key = os.environ['SECRET_KEY']
 app.url_map.strict_slashes = False
+app.url_map.converters['regex'] = helpers.RegexConverter
 sentry = Sentry(app)
 
 metrics = prometheus_flask_exporter.PrometheusMetrics(
@@ -251,7 +253,7 @@ def search_snap():
     return public_views.search_snap()
 
 
-@app.route('/<snap_name>')
+@app.route('/<regex("[a-z0-9-]*[a-z][a-z0-9-]*"):snap_name>')
 def snap_details(snap_name):
     return public_views.snap_details(snap_name)
 
--- a/app.py
+++ b/app.py
@@ -177,6 +177,11 @@ def create_redirect():
     return flask.redirect('https://docs.snapcraft.io/build-snaps')
 
 
+@app.route('/favicon.ico')
+def favicon():
+    return flask.redirect('https://assets.ubuntu.com/v1/fdc99abe-ico_16px.png')
+
+
 # Login handler
 # ===
 @app.route('/login', methods=['GET', 'POST'])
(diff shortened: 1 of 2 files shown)


static/robots.txt
```diff
@@ -0,0 +1,4 @@
+User-agent: *
+Allow: /
+Allow: /account*
+Allow: /^[a-z0-9-]*[a-z][a-z0-9-]*$
```
bartaz: I'm not sure if regexp are allowed by standard, although it's possible that google/bing would recognise that.

On the other hand, I think `Allow: /` is enough to allow anything, so we don't need to specify snap details URL specifically.

static/robots.txt
```diff
@@ -0,0 +1,4 @@
+User-agent: *
+Allow: /
+Allow: /account*
```
bartaz: I'm not sure if we want or need robots traffic on account pages.
In this case I guess we should `Disallow: /account`

static/robots.txt
tbille: I don't really know I took it from Celso's comment I'll check this

templates/_layout.html
```diff
@@ -14,8 +14,8 @@
       <meta property="twitter:site" content="@snapcraftio" />
     {% endblock %}
 
-    <link rel="icon" type="image/png" href="https://assets.ubuntu.com/v1/fdc99abe-ico_16px.png" sizes="16x16" />
```
nottrobin: Don't do this. It's much better to link people directly to the correct file than to send them to a different URL that then redirects.

static/robots.txt
```diff
@@ -0,0 +1,4 @@
+User-agent: *
```
nottrobin: If all we're doing is allowing things, we don't need a `robots.txt`.

app.py
```diff
@@ -142,6 +142,25 @@ def apply_caching(response):
     return response
 
 
+# Static files
+# ===
+@app.route('/robots.txt')
```
nottrobin: We don't need a robots.txt

app.py
```diff
@@ -142,6 +142,25 @@ def apply_caching(response):
...
+def robots():
+    return flask.send_from_directory(
+        app.static_folder, 'robots.txt')
+
+
+@app.route('/favicon.ico')
+def favicon():
+    size = flask.request.args.get('size', default=16, type=int)
```
nottrobin: Given that `/favicon.ico` should *only* be called by browsers following the old standard (as they mostly still do), there's no use for `/favicon.ico?size=XX`. If we were deliberately asking for a specific favicon we should simply use the real URL - e.g. https://assets.ubuntu.com/v1/fdc99abe-ico_16px.png.

This means that in fact `/favicon.ico` should just redirect to one place.

modules/public/views.py
```diff
@@ -151,6 +152,10 @@ def snap_details(snap_name):
     with appropriate sanitation.
     """
 
+    pattern = re.compile("^[a-z0-9-]*[a-z][a-z0-9-]*$")
+    if not pattern.match(snap_name):
```
nottrobin: It would be neater to use the Regex URL converter, as [used in insights.ubuntu.com](https://github.com/canonical-websites/insights.ubuntu.com/blob/master/app.py#L22), so that it will only hit this route for valid snap names - then other URLs will 404 automatically.

nottrobin:
@tbille I do think it's worth leaving in your `/favicon.ico` redirect to redirect to the assets server. That was a nice touch.

Issue #438, opened 2018-03-22
Badly cached CSS (and possibly JS)


cprov:
Edited by @bartaz:

Currently our static resources (such as CSS and JS bundles) are simply cached by the browser.
When we release new version to production most often browsers will have old version of CSS and JS cached and will not request a new one.

This leads to a broken experience. Such as recent issue with CLI instructions:

<img width="576" alt="screen shot 2018-03-22 at 9 03 02 am" src="https://user-images.githubusercontent.com/3610997/37769504-f8dfbc30-2daf-11e8-8be8-0b5f994bfde7.png">

Full browser reload is needed to fix this from user side, but this should not be accepted as a workaround.

We need to make sure browsers request new versions of files when new version is deployed.

cprov:
Using Chrome on Mac OSX

bartaz:
Old CSS cached. Full reload should fix that.

There must be something wrong with cache headers that browser didn't detect the change in CSS.

cprov:
@bartaz right, full reload solved it, thanks. Any ideas for improving the CSS caching ?

bartaz:
Files affected are CSS bundle:

`<link rel="stylesheet" href="/static/minified/styles.css" />`

Our JS bundles:
`<script src="/static/js/dist/public.js"></script>`
`<script src="/static/js/dist/publisher.js"></script>`

And quite possibly any 3rd party JS lib that we load (not included in the bundle). They don't update often, but if we update any of these at some point they may be affected as well:
Public snap details:
```
<script src="/static/js/modules/clipboard.min.js"></script>
<script src="/static/js/modules/d3.min.js"></script>
<script src="/static/js/modules/d3-geo.min.js"></script>
<script src="/static/js/modules/topojson-client.min.js"></script>
```
Measure page:
```
<script src="/static/js/modules/moment.min.js"></script>
<script src="/static/js/modules/d3.min.js"></script>
<script src="/static/js/modules/d3-geo.min.js"></script>
<script src="/static/js/modules/topojson-client.min.js"></script>
<script src="/static/js/modules/billboard.min.js"></script>
```

Lukewh:
Could we also include the version in an environment var so we can pass it through to Sentry to tag by build?

bartaz:
We have a `x-vcs-revision` header sent with build number which maybe could be used for sentry as well. But I don't see it anywhere in the app, so possibly it's handled by kubernetes.

tbille:
[...]

PR #473, opened 2018-03-29
Add commit_id to css file in template

tbille:
# Summary

Add commit id to styles.css in template

# QA 

- `./run`
- http://0.0.0.0:8004
- Make sure in the html received the commit id is here


nottrobin:
As mentioned [on the issue](https://github.com/canonical-websites/snapcraft.io/issues/438#issuecomment-377294110), we shouldn't be adding the `COMMIT_ID` to the URLs.

PR #474, opened 2018-04-02
Scoping public API calls to specific fields.

cprov:
Requesting explicit fields will make it easier to migrate to new APIs (similarly to what is already done in snapd).

Transitioning snap-details to `media` (instead of `icon_url` and `screenshot_urls`) which makes it ready for "banner" and "banner_icon" support. Search/promoted/featured-snaps remain using `icon_url` mostly for convenience, but eventually it should also be ported to `media`.

### DIFF
--- a/modules/public/api.py
+++ b/modules/public/api.py
@@ -15,6 +15,11 @@ SNAP_DETAILS_URL = ''.join([
     SNAPCRAFT_IO_API,
     'snaps/details/{snap_name}',
     '?channel=stable',
+    '&fields=snap_id,package_name,title,summary,description,license,contact,'
+    'website,publisher,prices,media,'
+    # Released (stable) revision fields will eventually be replaced by
+    # `channel_maps_list` contextual information.
+    'revision,version,binary_filesize,last_updated'
 ])
 DETAILS_QUERY_HEADERS = {
     'X-Ubuntu-Series': '16',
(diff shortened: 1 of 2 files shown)


modules/public/views.py
```diff
@@ -202,14 +202,16 @@ def snap_details(snap_name):
 
     # filter out banner and banner-icon images from screenshots
     screenshots = [
-        url for url in details['screenshot_urls'] if "banner" not in url
+        m['url'] for m in details['media']
+        if m['type'] == "screenshot" and "banner" not in m['url']
```
nottrobin: @cprov, @bartaz is `"banner" not in m['url']` still relevant? Is `m['type'] == "screenshot"` not enough to filter out banners?

cprov: @nottrobin the filtering on file name still needed until the existing (~5) snaps migrate their banner* and banner-icon* out of the `screenshot` type. But they will only do that once they have UI available.

modules/public/api.py
```diff
@@ -34,6 +39,7 @@
     'snaps/search',
     '?q={snap_name}&page={page}&size={size}',
     '&confinement=strict,classic',
+    '&fields=package_name,title,summary,icon_url'
```
tbille: the publisher is missing

PR #478, opened 2018-04-03
Refresh macaroon on every api call if needed

tbille:
# Summary

Refresh macaroon if needed on every api call

# QA 

- `./run`
- http://0.0.0.0:8004/account
- browse the `/account`s endpoints


bartaz:
QA +1, market page with screenshots works as expected

Issue #441, opened 2018-03-23
Market page should only save data modified by user [back-end]


bartaz:
Continue work on #396 (PR: #416)

PR #483, opened 2018-04-04
Market page - save only changes

bartaz:
Makes sure we only save in API data that was changed by user.

Fixes: #441 

### QA
- ./run
- go to market page of any snap
- change some text value (title, summary, description)
- Apply - API should only save given value (print in backend to debug)
- change some image (icon, screenshot)
- Apply - API should only save images

### DIFF
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -322,7 +322,7 @@ def post_market_snap(snap_name):
     changes = loads(flask.request.form['changes'])
 
     if changes:
-        snap_id = api.get_snap_id(snap_name, flask.session)
+        snap_id = flask.request.form['snap_id']
         error_list = []
         info = []
         images_files = []
@@ -407,7 +407,7 @@ def post_market_snap(snap_name):
                 body_json['public_metrics_blacklist'] = metrics_blacklist
 
             metadata = api.snap_metadata(
-                flask.request.form['snap_id'],
+                snap_id,
                 flask.session,
                 body_json
             )
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -319,7 +319,11 @@ def build_image_info(image, image_type):
 
 
 def post_market_snap(snap_name):
-    changes = loads(flask.request.form['changes'])
+    changes = None
+    changed_fields = flask.request.form['changes']
+
+    if changed_fields:
+        changes = loads(changed_fields)
 
     if changes:
         snap_id = flask.request.form['snap_id']
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -319,12 +319,10 @@ def build_image_info(image, image_type):
 
 
 def post_market_snap(snap_name):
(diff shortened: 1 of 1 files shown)


modules/publisher/views.py
```diff
@@ -319,56 +319,60 @@ def build_image_info(image, image_type):
...
 
 def post_market_snap(snap_name):
-    snap_id = api.get_snap_id(snap_name, flask.session)
+    changes = loads(flask.request.form['changes'])
 
-    if 'submit_revert' in flask.request.form:
-        flask.flash("All changes reverted.", 'information')
-    else:
+    if changes:
+        snap_id = api.get_snap_id(snap_name, flask.session)
```
bartaz: @tbille I moved this line from above (as it is only needed for screenshots API call), but is it needed at all?
I think we have snap_id from the form? At least we use it later when posting body_json.

bartaz: OK. Done, seems to work as it should.

modules/publisher/views.py
```diff
@@ -319,56 +319,60 @@ def build_image_info(image, image_type):
 
 
 def post_market_snap(snap_name):
-    snap_id = api.get_snap_id(snap_name, flask.session)
+    changes = loads(flask.request.form['changes'])
```
tbille: I did a little hack :)
- In console enable Apply button, and remove `event.preventDefault();`
- Submit
- This causes a but because `flask.request.form['changes']` doesn't exist

modules/publisher/views.py
```diff
@@ -471,6 +491,8 @@ def post_market_snap(snap_name):
             )
 
         flask.flash("Changes applied successfully.", 'positive')
+    else:
+        flask.flash("No changes to save.", 'information')
```
tbille: With the little hack I did, once fixed. Test this if it is really useful, because of the redirection the line after

modules/publisher/views.py
bartaz: Thanks, fixed

modules/publisher/views.py
bartaz: Yes, this message may be unnecessary. But at least it gives some message to user when empty changes are posted for some reason.

tbille: Ok the message stays, I wasn't sure, but it's ok!

modules/publisher/views.py
```diff
@@ -386,28 +394,29 @@ def post_market_snap(snap_name):
...
+                if len(metrics_blacklist) > 0:
+                    metrics_blacklist = metrics_blacklist.split(',')
+                else:
+                    metrics_blacklist = []
 
-        body_json['public_metrics_blacklist'] = metrics_blacklist
+                body_json['public_metrics_blacklist'] = metrics_blacklist
 
-        metadata = api.snap_metadata(
-            flask.request.form['snap_id'],
```
tbille: You are right, here I get the id from the form ...

PR #485, opened 2018-04-05
Fix empty website address when value is None

### LINKED ISSUE #350: 'market' page validation issue with 'developer website' field
To reproduce:

1) Upload a new snap to the store without setting a developer website in the yaml.
2) open a browser and go to: https://snapcraft.io/account/snaps/<snapname>/market
3) Edit the title field for the snap.
4) click 'apply'

At this point, I expect the title to be updated.

What actually happens is that I get an error message something along the lines of "URL must begin with http(s)'. The validation is complaining about the developer website field, which I haven't edited.

Further, if I repeat steps 3-4, the title is updated.

I suspect the issue here is that the validation on the new publisher dashboard is more strict than the validation on the backend services.

bartaz:
Fixes #350 

For newly added snaps (with empty website field) API returns value `None` for website which was displayed in the market form as "None" string and failed validation (because it's not valid URL).

To fix this we make sure in template that `None` website is displayed as empty string.

### QA

- ./run
- go to a snap with market data that was not edited before (has empty website)
- website field should be empty (not show `None`)
- you should be able to change any values and save them without validation errors about invalid URL

### DIFF
--- a/templates/publisher/market.html
+++ b/templates/publisher/market.html
@@ -234,7 +234,7 @@
             <div class="col-8">
               <div class="p-form-validation u-no-margin {% if field_errors and field_errors['website'] %}is-error{% endif %}">
                 <div class="p-form-validation__field">
-                  <input class="p-form-validation__input" type="url" name="website" value="{{ website }}" maxlength="256"/>
+                  <input class="p-form-validation__input" type="url" name="website" value="{{ website if website is not none else '' }}" maxlength="256"/>
                 </div>
                 {% if field_errors and field_errors['website'] %}
                   <p class="p-form-validation__message">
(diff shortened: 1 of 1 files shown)


templates/publisher/market.html
```diff
@@ -234,7 +234,7 @@ <h1 class="u-float--left p-heading--three">{% if display_title %}{{ display_titl
             <div class="col-8">
               <div class="p-form-validation u-no-margin {% if field_errors and field_errors['website'] %}is-error{% endif %}">
                 <div class="p-form-validation__field">
-                  <input class="p-form-validation__input" type="url" name="website" value="{{ website }}" maxlength="256"/>
+                  <input class="p-form-validation__input" type="url" name="website" value="{{ website if website is not none else '' }}" maxlength="256"/>
```
bartaz: @nottrobin 

I'm not sure if it's the best place to fix that.
I considered doing it on python side in views, but it means doing it in even more places (separate views for get/post, etc). So maybe we should fix it in out API layer? Or maybe it's an API bug?

Lukewh: I think this is good, but we should discuss with Toto about a better global solution

nottrobin:
Superseded by https://github.com/canonical-websites/snapcraft.io/pull/489

PR #491, opened 2018-04-06
Publisher header template

Lukewh:
# Done

- Created `publisher/_header.html` template
- Modified `publisher/market.html` and `publisher/measure.html` to use the new template
- Made sure `snap_title` is consistent throughout (some places was `title`)

# QA

- Pull this branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/SNAP_NAME/market and http://0.0.0.0:8004/account/snaps/SNAP_NAME/measure and ensure the title and tab bar still display correctly

### DIFF
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -284,7 +284,7 @@ def get_market_snap(snap_name):
     context = {
         "snap_id": snap_details['snap_id'],
         "snap_name": snap_details['snap_name'],
-        "title": snap_details['title'],
+        "snap_title": snap_details['title'],
         "summary": snap_details['summary'],
         "description": snap_details['description'],
         "license": snap_details['license'],
@@ -532,7 +532,7 @@ def post_market_snap(snap_name):
                 "public_metrics_blacklist": details_blacklist,
                 "display_title": snap_details['title'],
                 # values posted by user
-                "title": (
+                "snap_title": (
                     changes['title'] if 'title' in changes
                     else snap_details['title']
                 ),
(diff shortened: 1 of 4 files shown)


modules/publisher/views.py
```diff
@@ -254,7 +254,7 @@ def get_market_snap(snap_name):
     context = {
         "snap_id": snap_details['snap_id'],
         "snap_name": snap_details['snap_name'],
-        "title": snap_details['title'],
+        "snap_title": snap_details['title'],
```
bartaz: API calls it `title`, so why renaming to `snap_title` on our side?
Why summary, description, etc don't have `snap_` prefix is that's what we go for?

templates/publisher/_header.html
```diff
@@ -0,0 +1,44 @@
...
+        <div class="col-6">
+          <ul class="p-tabs__list u-float--right" role="tablist">
+            <li class="p-tabs__item" role="presentation">
+              <a
+                href="/account/snaps/{{ snap_name }}/market"
+                class="p-tabs__link"
+                tabindex="0"
+                role="tab"
+                {% if selected_tab == 'market' %}
+                aria-selected="true"
```
bartaz: Indent would make it bit easier to read.

modules/publisher/views.py
Lukewh: Yeah I thought about this, however I can imagine some cases where `title` would want to be the title of the page, in that case I'd want `snap_title` and `page_title` to clearly indicate what each title is. I want to avoid `title` sometimes being the page title and other times the title of the snap. 

Ideally I'd like everything to do with a snap to be namespaced under `snap` like `snap.title`, `snap.description` etc. but I don't think it's necessary.

templates/publisher/_header.html
Lukewh: Fixed

modules/publisher/views.py
tbille: > I don't think it's necessary.

This is a scenario where it seems necessary. I will try to see if I can modify this in one of my refactor PR.

PR #493, opened 2018-04-06
Cache-control for public and private endpoints

### LINKED ISSUE #424: Add cache-control headers, including stale-if-error
We should add caching headers *for public pages only* to reduce load on the server. This will be initially honoured by visitors' browsers, and in the future by any caching layers we place in front of the website application (e.g. squid), to reduce load on the applications themselves.

nottrobin:
Add standard caching headers to public routes:

- Cache pages for 30 seconds
  (allows Squid to take significant load of the app)
- Serve stale pages while revalidating the cache for 5 minutes
  (allows Squid to response instantly while doing cache refreshes)
- Show stale pages if the app is erroring for 1 day
  (gives us a 1 day buffer to fix errors before they are publicly visible)

These caching headers should now be included on /, /store, /search,
/{snap-name}

All private routes should now have `Cache-Control: private`.

I also restructured the code a bit to separate out decorators and template functions from `app.py`, to keep things more focused.

Fixes #424

QA
--

``` bash
$ curl -sI http://127.0.0.1:8004/ | grep 'Cache-Control'
Cache-Control: max-age=30, stale-while-revalidate=300, stale-if-error=86400
$ curl -sI http://127.0.0.1:8004/store | grep 'Cache-Control'
Cache-Control: max-age=30, stale-while-revalidate=300, stale-if-error=86400
$ curl -sI http://127.0.0.1:8004/search | grep 'Cache-Control'
Cache-Control: max-age=30, stale-while-revalidate=300, stale-if-error=86400
$ curl -sI http://127.0.0.1:8004/skype | grep 'Cache-Control'
Cache-Control: max-age=30, stale-while-revalidate=300, stale-if-error=86400
$ curl -sI http://127.0.0.1:8004/lxd | grep 'Cache-Control'
Cache-Control: max-age=30, stale-while-revalidate=300, stale-if-error=86400
$ curl -sI http://127.0.0.1:8004/account  # Should have no cache-control
HTTP/1.1 302 FOUND
Server: gunicorn/19.7.1
Date: Fri, 06 Apr 2018 17:28:27 GMT
Connection: close
Content-Type: text/html; charset=utf-8
Content-Length: 245
Location: http://127.0.0.1:8004/login?next=/account
X-Hostname: 8cc34cf58ef7
X-VCS-Revision: 0811fe1175be2c42754b91b0dadd2e0deb9a3bcb
X-Request-Id: fad12c89-ae6e-4791-af16-664e716e2cba
```

Check cache-control settings make sense ^.

Then browse to http://127.0.0.1, check the headers in the Network inspector.

Then browse to http://127.0.0.1/account, log in, then check the headers in the network inspector include `Cache-Control: private`.

### DIFF
--- a/decorators.py
+++ b/decorators.py
@@ -1,0 +1,50 @@
+# Core packages
+import functools
+
+# Third party packages
+import flask
+import modules.authentication as authentication
+
+
+def login_required(func):
+    """
+    Decorator that checks if a user is logged in, and redirects
+    to login page if not.
+    """
+    @functools.wraps(func)
+    def is_user_logged_in(*args, **kwargs):
+        if not authentication.is_authenticated(flask.session):
+            return flask.redirect('login?next=' + flask.request.path)
+
+        return func(*args, **kwargs)
+    return is_user_logged_in
+
+
+def public_cache_headers(decorated):
+    """
+    Add standard caching headers to a public route:
+
+    - Cache pages for 30 seconds
+      (allows Squid to take significant load of the app)
+    - Serve stale pages while revalidating the cache for 5 minutes
+      (allows Squid to response instantly while doing cache refreshes)
+    - Show stale pages if the app is erroring for 1 day
+      (gives us a 1 day buffer to fix errors before they are publicly visible)
+    """
+
+    cache_control_headers = [
+        'max-age=30',
+        'stale-while-revalidate=300',
+        'stale-if-error=86400',
+    ]
+
+    @functools.wraps(decorated)
+    def decorated_function(*args, **kwargs):
+        response = flask.make_response(
+            decorated(*args, **kwargs)
+        )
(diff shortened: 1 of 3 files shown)


decorators.py
```diff
@@ -0,0 +1,77 @@
...
+    Return a decorator function to add a set of Cache-Control to the response
+    """
+
+    @functools.wraps(route_function)
+    def decorated_function(*args, **kwargs):
+        response = flask.make_response(
+            route_function(*args, **kwargs)
+        )
+
+        if response.status_code == 200:
```
tbille: `if response.ok:`

nottrobin: Actually it's worse than that - ["respose.ok: Returns True if status_code is less than 400"](http://docs.python-requests.org/en/master/api/#requests.Response.ok). I *definitely* don't want 3** responses to be cached.

Issue #394, opened 2018-03-14
Add cross link notification


Lukewh:
Depends on https://github.com/CanonicalLtd/snapcraft-design/issues/298

PR #495, opened 2018-04-09
Cross link notification

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/394

- Updated vanilla to 1.6.7
- Added `_beta-notification.html` template for the notification

Depends on https://github.com/canonical-websites/snapcraft.io/pull/491

# QA

- Pull this branch
- `./run`
- See the notification, click the links in the notification, close the notification
- Visit other account pages and see the notification

# Screenshot
screenshot-2018-4-12 market details for lukotron

### DIFF
--- a/templates/developer_programme_agreement.html
+++ b/templates/developer_programme_agreement.html
@@ -5,49 +5,48 @@
 {% endblock %}
 
 {% block content %}
-<section class="p-layout__main">
+{% include "publisher/_beta-notification.html" %}
+<section class="p-strip">
   <div class="row">
-    <h1>Developer Programme Agreement</h1>
+    <h1 class="p-heading--three">Developer Programme Agreement</h1>
   </div>
 
   <div class="row">
     <div class="p-card">
-      <h2>Canonical Terms of Service</h2>
+      <h2 class="p-heading--four">Canonical Terms of Service</h2>
       <p class="p-card__content">
         <a class="p-link--external" href="https://www.ubuntu.com/legal/terms-and-policies/developer-terms-and-conditions" target="_blank">Developer Terms and Conditions</a>
       </p>
     </div>
   </div>
 
-  <div class="p-strip is-shallow u-no-padding--top" >
-    <div class="row u-no-margin--top">
-      <form method="POST" action="/account/agreement">
-        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
-        <div class="row">
-          <div class="col-8">
-            <div class="p-form-validation">
-              <input name="i_agree" id="id_i_agree" class="p-form-validation__input" type="checkbox">
-              <label for="id_i_agree"">I agree to these terms</label>
-            </div>
-          </div>
-          <div class="col-4 u-align--right">
-            <a class="p-button--neutral" href="/">Cancel</a>
-            <button class="p-button--positive" disabled="disabled" type="submit">Continue</button>
+  <div class="row u-no-margin--top">
+    <form method="POST" action="/account/agreement">
+      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+      <div class="row">
+        <div class="col-8">
+          <div class="p-form-validation">
+            <input name="i_agree" id="id_i_agree" class="p-form-validation__input" type="checkbox">
+            <label for="id_i_agree"">I agree to these terms</label>
           </div>
         </div>
-        <script>
-         var checkbox = document.querySelector('#id_i_agree');
-         var button = document.querySelector('button[type="submit"]');
-         checkbox.addEventListener('change', function(event) {
-           if (this.checked) {
-             button.removeAttribute('disabled');
-           } else {
-             button.setAttribute('disabled', 'disabled');
-           }
-         });
-        </script>
-      </form>
-    </div>
+        <div class="col-4 u-align--right">
+          <a class="p-button--neutral" href="/">Cancel</a>
+          <button class="p-button--positive" disabled="disabled" type="submit">Continue</button>
+        </div>
+      </div>
+      <script>
+       var checkbox = document.querySelector('#id_i_agree');
+       var button = document.querySelector('button[type="submit"]');
+       checkbox.addEventListener('change', function(event) {
+         if (this.checked) {
+           button.removeAttribute('disabled');
+         } else {
+           button.setAttribute('disabled', 'disabled');
+         }
+       });
+      </script>
+    </form>
   </div>
 </section>
 {% endblock %}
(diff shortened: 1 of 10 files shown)


Greggless:
**A tweak:** If a user dismisses the notification, it should not reappear when they move to another page on snapcraft.io

bartaz:
@Lukewh I'm not sure if close button makes sense in a way it works now.

Currently user can close the notification if they feel they don't need it, but if they navigate to another page they see it again. So it seems there is no really point in closing the notification.

On the other hand we probably shouldn't be hiding the notification permanently, so that users do have a way to go back to dashboard if they need it later.

I would say we shouldn't allow closing this notification at all or closing it should be tied to user session (for example stored in browser sessionStorage), so that when user comes back to the site later, they see notification again.

templates/publisher/_beta-notification.html
```diff
@@ -0,0 +1,12 @@
+<section class="p-strip is-shallow u-no-padding--bottom" id="beta-notification">
+  <div class="row">
+    <div class="p-notification--note">
+      <div class="p-notification__response">
+        <button class="p-icon--close u-float--right" aria-label="Close notification" onclick="document.getElementById('beta-notification').remove()">Close</button>
```
bartaz: Not a critical thing, but `remove` may not be supported in some browsers (IE and mobile Safari) which may lead to some exceptions (sentry).

Greggless:
> I would say we shouldn't allow closing this notification at all or closing it should be tied to user session (for example stored in browser sessionStorage), so that when user comes back to the site later, they see notification again.

I like the idea of storing it in the browser session, but I think it'd be frustrating to not be able to close the notification once on the site.

templates/publisher/_beta-notification.html
Lukewh: Maybe just `style.display = 'none'` instead, do you think?

Lukewh:
Re. closing the notification, I'm not sure what the best solution is. I don't think closing it forever is great - someone closes it within the first 10 seconds, then visit a page that doesn't let them do what they want, now they're at a dead-end

@bartaz @Greggless

Greggless:
If `my snaps` linked to dashboard whilst it has the majority of the functionality, that would mean closing the notification wouldn't be a problem...

Lukewh:
`My snaps` will link to https://snapcraft.io/account/snaps from the get-go @Greggless unless there's a reason we shouldn't?

Greggless:
OK in that case, we should keep it ever present. 

But can we reduce the padding from the header?

Greggless:
OK in that case, we should keep it ever present. 

But can we reduce the padding from the header?

templates/publisher/_beta-notification.html
Lukewh: This is now not needed

Lukewh:
@Greggless I've removed some of the padding and updated the demo for your viewing pleasure.

lyubomir-popov:
@bartaz @Lukewh Might be late for alternative suggestions, if so lmk and I'll plus one it. But I agree with @Greggless this is too much space for something you can't close. For comparison, consider this:
screen shot 2018-04-11 at 09 59 22

So how about something like this applied to both new and old pages. On new ones, it gives you the options above. On old pages, it suggests switching to the new ui. And you can go back and forth as much as you please.

image

bartaz:
@Greggless @Lukewh 

One thing that I find a bit confusing is that with current text: "Snapcraft.io BETA will be the new publisher UI for your snaps." it's not exactly clear to me if THIS page that I'm looking at is this new snapcraft BETA or is it trying to tell me that there is some OTHER snapcraft beta...

screen shot 2018-04-12 at 10 26 33

Greggless:
@bartaz 

> Snapcraft.io is in `BETA`. Can't find something? **Report an issue** or **go back to the old version**

lyubomir-popov:
The screenshot looks good @Lukewh

templates/publisher/_beta-notification.html
```diff
@@ -0,0 +1,12 @@
+<section class="p-strip--light u-no-margin--top u-no-padding">
+  <div class="row p-beta-notification">
+    <div class="col-12">
+      <p>
+        <span class="p-muted-heading"><b>Beta</b></span>
+        Snapcraft.io is in Beta. Can't find something?
+        <a href="https://dashboard.snapcraft.io/snaps">Go back to the old version</a> or
```
bartaz: Not sure from design point of view, but maybe it would be worth making these links external.

templates/username.html
```diff
@@ -5,10 +5,13 @@
 {% endblock %}
 
 {% block content %}
+
+{% include "publisher/_beta-notification.html" %}
+
 <section class="p-layout__main u-no-margin--top">
-  <div class="p-strip is-shallow">
+  <div class="p-strip is-shallow u-no-margin--top">
```
bartaz: Possibly not related directly to this PR, but layout on this page is strange.

There are no styles for `p-layout__main` at all, `u-no-margin--top` there is needed only to cancel Vanilla infamous `* + *`, but it seems that this whole `<section>` is not needed at all and p-strip with a row inside would be just fine (as it is in all other pages).

Also `u-no-margin--top` seems to not be needed (?)

templates/developer_programme_agreement.html
```diff
@@ -5,14 +5,15 @@
 {% endblock %}
 
 {% block content %}
-<section class="p-layout__main">
+{% include "publisher/_beta-notification.html" %}
+<section class="p-layout__main u-no-margin--top">
```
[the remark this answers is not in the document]
lyubomir-popov: @bartaz @Lukewh On the topic of strips, I think in most cases a strip should just separate body content from header and footer. Given the current amount of white space, I think these should be consistently the deep strips on all pages. 
This way, white space is always bigger between the main wrappers on the page (header, body, footer) and smaller between subsections of the body. I.e.:
```header ______ row __ row __ row ______ footer```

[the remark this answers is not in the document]
Comment (Lukewh) [reply templates/developer_programme_agreement.html]: Fixed

File: templates/username.html (hunk 3 above)
Comment (Lukewh) [reply templates/username.html]: Fixed

File: templates/publisher/_beta-notification.html (hunk 2 above)
Comment (Lukewh) [reply templates/publisher/_beta-notification.html]: I agree, fixed

File: templates/publisher/_beta-notification.html [hunk 5]
```diff
@@ -0,0 +1,12 @@
+<section class="p-strip--light u-no-margin--top u-no-padding">
+  <div class="row p-beta-notification">
+    <div class="col-12">
+      <p>
+        <span class="p-muted-heading"><b>Beta</b></span>
+        Snapcraft.io is in Beta. Can't find something?
+        <a href="https://dashboard.snapcraft.io/snaps" class="p-link--external">Go back to the old version</a> or
+        <a href="https://github.com/canonical-websites/snapcraft.io" class="p-link--external">report an issue</a>.
```
bartaz: This link could be directly to `/issues` (as one in the footer).

Lukewh: Good plan, thanks @bartaz

PR #498, opened 2018-04-10
Add sign in and account dropdown to navigation

### LINKED ISSUE #393: Add Login and account dropdown to navigation
Logged Out:
- 'Login' -> /account

Logged In:
- '\<username\> \\/'
  - 'Account' -> /account/details
  - 'Logout' -> /logout

bartaz:
Fixes #393

Adds sign in link and account dropdown to navigation.

### QA

- ./run or http://snapcraft.io-pr-498.run.demo.haus/
- /logout (just in case you are logged in from before)
- go to any page, 'Sign in' should be in header
- click 'Sign in'
- go through login screens
- you should be redirected back to snapcraft
- your user name should be in right side of header
- click you user name, dropdown will appear with links to account page and sign out
- go to 'Account details' ( account page should be opened)
- click 'Sign out' in dropdown, you should be signed out

Also check if it works on smaller screens.
screen shot 2018-04-18 at 10 14 27
screen shot 2018-04-12 at 15 23 51
screen shot 2018-04-12 at 15 23 58
screen shot 2018-04-11 at 14 11 29

With long user name:
screen shot 2018-04-17 at 12 52 42


app.py
```diff
@@ -106,12 +106,18 @@ def static_url(filename):
 
         return url + '?v=' + file_hash.hexdigest()[:7]
 
+    if 'openid' in flask.session:
```
bartaz: @tbille Not sure if that's the best way to expose signed in user to all templates.

@Lukewh Also, as far as I remember we shouldn't be showing user name that comes from SSO, but the one from store API, right? So we would need to get it here?

templates/_layout.html
```diff
@@ -44,6 +44,7 @@
 
     {% include "_footer.html" %}
 
+    <script src="{{ static_url('js/dist/base.js') }}"></script>
```
bartaz: @Lukewh I added the `base.js` script to layout.
But, as sentry is included in the template, I guess we would need to include this script after that, right?
So that will mean we need to move setting up raven to layout. To make sure raven is loaded even if page itself doesn't have any custom scripts?

app.py
Lukewh: @bartaz correct, we should be using the name provided by the store API

bartaz: @tbille Any suggestions how to do that? I guess we don't want to call user name from API every time any page with user is loaded? I guess instead (or together) with storing 'openid' in session we should load and keep store user name in session once user logs in?

templates/_layout.html
Lukewh: Yeah we could do something like:
`_layout.html`:
```
<script src="base"></script>
{% block scripts %}{% endblock %}
<script src="raven"></script>
<script>
 Raven.setupStuff(config, function() {
   {% block script_execs %}{% endblock %}
 });
</script>
```

`market.html` etc.:
```
{% block scripts %}
<script src="d3"></script>
<script src="publisher.js"></script>
{% endblock %}
{% block script_execs %}
  snapcraft.publisher.market(...)
{% endblock %}
```

Not great, but means we can wrap what we need to with the raven stuff?

app.py
tbille: we could store some information in the session? We will have to make that when we logout we delete everything.

tbille: mmm but we will have a problem. If a users logsin and goes to directly to the market page (without passing via the acocunt page) we will need a wrapper that would get the username from the API

tbille: @bartaz @Lukewh I did a small PR to fix this problem https://github.com/canonical-websites/snapcraft.io/pull/500

app.py
```diff
@@ -88,12 +88,18 @@ def utility_processor():
     can be used in all templates
     """
 
+    if 'openid' in flask.session:
+        user_name = flask.session['openid']['fullname']
+    else:
+        user_name = None
+
```
bartaz: @tbille Is this piece OK to get user name to all pages?

bartaz:
Marking as `Don't merge` as we need it for QA session next week, but we don't want to merge it until go live.

Greggless:
# `Super_cat` 🐱 🚀 
image
**Some tweaks**
- [ ] Left align Account details and sign out with `display_name` 
- [ ] Set width of dropdown to match `| display_name |`
- [ ] Use a the same colour for the dropdown as the header? (@lyubomir-popov can/should we do this?)

lyubomir-popov:
This is what Karl kas done for Maas, which I think is better than the speech bubble menu here: 

image

lyubomir-popov:
I'd suggesrt doing it like this here and proposing it as a enhancement to vanilla.

bartaz:
@Lukewh @lyubomir-popov Was this already built for maas, so we could reuse, or is it something new we need to build for snapcraft.io?

lyubomir-popov:
@bartaz @Lukewh https://github.com/maas/maas/blob/master/src/maasserver/static/scss/_patterns_navigation.scss
it is currently in master

app.py
tbille: i'd rather do something like this: 
```python
        user_name = flask.session['openid']['fullname']
    else:
        user_name = None
```

I might add in the future a module to handle our session variables, but for now let's use this

lyubomir-popov:
@bartaz just a couple of things - 

- [ ]  ideally, mobile menu should match the dark header colour
- [ ]  dividers after forum and account name are not necessary - pls ignore if vanilla makes these difficult to remove, as the dividers will be dropped very soon anyway

bartaz:
@lyubomir-popov 

@bartaz just a couple of things - 

> ideally, mobile menu should match the dark header colour

Vanilla is giving mobile menu light colour (regardless of colour of the header). It's really hard to override Vanilla styling in navigation, so I'd rather want to avoid it here.

> dividers after forum and account name are not necessary - pls ignore if vanilla makes these difficult to remove, as the dividers will be dropped very soon anyway

Same applies to the dividers. These styles come from the Vanilla and are quite hard to override (SCSS produces long nested selectors there). 
Also, removing just last border - while it looks good for 'Forum' is seems bit strange on account menu on the right (as it still as border on one side):
screen shot 2018-04-13 at 11 55 04

matthewpaulthomas:
I haven’t finished testing this yet, but I have found #516.

Greggless:
Much better, 
- [ ] If a name is truncated, we should use an ell... 
- [ ] A little bug 🐛  when I over on the menu items the `username` whites-out
image

bartaz:
@Greggless For the ellipsis in the long name it should be there:

screen shot 2018-04-17 at 13 30 01

Unless your long name is not long enough ;)

And good catch with the hover bug.
I don't know why it happened, I suspect recent Vanilla update done in other branch :/

Greggless:
### My name in login.ubuntu
> FLAMMENWERFERwerferflammenlongnamelongnametryingtobreakthething

### What I see at 1440px
image

### What I see at 1111px
image

Greggless:
## 🌐 Chrome Version 65.0.3325.181
## 💻 On Mac Version 10.13.4

bartaz:
@Greggless exactly same OS and browser for me... 😕

templates/_header.html
```diff
@@ -21,10 +21,29 @@
           </a>
         </li>
         <li class="p-navigation__link" role="menuitem"><a href="https://build.snapcraft.io">Build</a></li>
-        <li class="p-navigation__link" role="menuitem"><a class="p-link--external" href="https://dashboard.snapcraft.io/snaps">My snaps</a></li>
+        <li class="p-navigation__link" role="menuitem"><a href="/account">My snaps</a></li>
```
tbille: `/account` redirects to `/account/snaps` but can you use `/account/snaps` to avoid problem in the future

tbille: Forget it, I mixed this PR with the split of the pages :)

templates/_header.html
```diff
@@ -22,11 +22,30 @@
         </li>
         <li class="p-navigation__link" role="menuitem"><a href="https://build.snapcraft.io">Build</a></li>
         <li class="p-navigation__link" role="menuitem">
-          <a class="p-link--external{% if page_slug and page_slug == 'my-snaps' %} is-selected{% endif %}" href="https://dashboard.snapcraft.io/snaps">My snaps</a>
+          <a class="{% if page_slug and page_slug == 'my-snaps' %} is-selected{% endif %}" href="/account/snaps">My snaps</a>
```
Lukewh: if we comment out the login/ logout we should also add this in a comment and replace it with the 'external-link' to dashboard

templates/_header.html
```diff
@@ -22,11 +22,30 @@
...
         </li>
         <li class="p-navigation__link" role="menuitem"><a href="https://docs.snapcraft.io">Docs</a></li>
         <li class="p-navigation__link" role="menuitem"><a href="https://forum.snapcraft.io/categories">Forum</a></li>
       </ul>
+      <ul class="p-navigation__links--right" role="menu">
```
Lukewh: Comment it out and gtg

PR #500, opened 2018-04-10
Use username, fullname from API instead of l.u.com

tbille:
# Summary

Instead of using the data received from login.ubuntu.com we use the data from dashboard
This will fix https://github.com/canonical-websites/snapcraft.io/pull/498#discussion_r180426317

# QA

- `./run`
- http://0.0.0.0:8004/logout
- http://0.0.0.0:8004/account
- The username and display name should be the same
- http://0.0.0.0:8004/logout
- http://0.0.0.0:8004/account/snaps/toto/market
- login
- http://0.0.0.0:8004/account
- The username and displayname should be OK

### DIFF
--- a/app.py
+++ b/app.py
@@ -27,6 +27,7 @@ import modules.authentication as authentication
 import modules.helpers as helpers
 import modules.public.views as public_views
 import modules.publisher.views as publisher_views
+import modules.publisher.api as publisher_api
 import template_functions
 import decorators
 from modules.macaroon import (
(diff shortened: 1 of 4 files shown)


app.py
```diff
@@ -189,15 +190,26 @@ def login():
...
-        'fullname': resp.fullname,
-        'image': resp.image,
-        'email': resp.email
-    }
-
     flask.session['macaroon_discharge'] = resp.extensions['macaroon'].discharge
+
+    try:
+        account = publisher_api.get_account(flask.session)
+        flask.session['openid'] = {
```
bartaz: I'd say that in this case calling it `openid` is a bit misleading, because there is more information from account API then from openid itself.

So maybe it should be `flask.session['user']` or something similar?

PR #502, opened 2018-04-10
Stop using snaps `uploaded` flag, it's deprecated.

cprov:
Checking the `latest_revisions` list directly.

### DIFF
--- a/templates/account.html
+++ b/templates/account.html
@@ -47,7 +47,7 @@ Account page — Linux software in the Snap Store
             {% for snap in snaps|sort %}
               <tr>
                 <td class="u-vertically-center p-snap-list">
-                  {% if snaps[snap].uploaded %}
+                  {% if snaps[snap].latest_revisions %}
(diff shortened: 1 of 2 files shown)


tbille:
@cprov thanks for the update. One little detail:

In `tests_publisher.py` could you update the mock in the function `def test_account_logged_in(self):` [this line](https://github.com/canonical-websites/snapcraft.io/blob/master/tests_publisher.py#L124)

cprov:
@tbille good point, but fixing the test is pro-forma, since it is only checking that the `snaps` object is passed blindly to the template.

In order to test this properly we would have to render the template and set `app.jinja_env.undefined = jinja2.StrictUndefined`. If we do that right now, the template explodes because of earlier problems:

```
  ...
  File "/Users/cprov/Canonical/snapcraft.io/templates/_header.html", line 19, in top-level template code
    {% if page_slug and page_slug == 'store' %} class="is-selected"{% endif %}
jinja2.exceptions.UndefinedError: 'page_slug' is undefined
```

PR #508, opened 2018-04-11
Test get market page

### LINKED ISSUE #480: test /account/snaps/snap/market (GET)
From https://github.com/CanonicalLtd/snappy-design-squad/issues/419:

> We should test the functions that the `@route`s are applied to, with a mocked API response (using `responses` most likely). We should test for every conceivable condition from the API:
> 
> - A connection timeout from the API
> - A DNS lookup failure
> - A connection failure
> - A 404 response when we weren't expecting one
> - A 404 response if we were expecting one (if relevant)
> - A redirect from the API
> - A list of errors from the API
> - A internal server error from the API
> 
> We can test the output of this function in one of two ways:
> 
> 1. Call the function directly and check that it tries to `render_template` with the expected context (although this feels a little like testing internals); or
> 2. Call the URL endpoint (with `self.app.get`) and check the rendered markup contains the expected elements

tbille:
# Summary

- Add tests for the market endpoint Fixes #480 
- Refactor test to get a generic test class for authenticated endpoints

# QA

- `./run test`

### DIFF
--- a/tests/endpoint_testing.py
+++ b/tests/endpoint_testing.py
@@ -1,0 +1,232 @@
+import requests
+
+import pymacaroons
+import responses
+
+from app import app
+from flask_testing import TestCase
+from modules.authentication import get_authorization_header
+
+
+# Make sure tests fail on stray responses.
+responses.mock.assert_all_requests_are_fired = True
+
+
+class BaseTestCases:
+    '''
+    This class has a set of test classes that should be inherited by endpoint
+    that have authentication.
+
+    It is also used to avoid unittest to run this tests file.
+    '''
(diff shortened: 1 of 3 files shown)


tests/endpoint_testing.py
```diff
@@ -0,0 +1,222 @@
...
+from app import app
+from flask_testing import TestCase
+from modules.authentication import get_authorization_header
+
+
+# Make sure tests fail on stray responses.
+responses.mock.assert_all_requests_are_fired = True
+
+
+class BaseTestCases:
```
tbille: add docstring

PR #509, opened 2018-04-12
Market toggles UX tweaks

Lukewh:
# Done

Partially fixes https://github.com/canonicalltd/snapcraft-design/issues/309
 - Toggle doesn't make sense yet, until we have single field saving
 - Distribution chart is not implemented yet, so the switch doesn't exist

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/SNAP_NAME/market
- Ensure the market toggles work as expected in reference to https://github.com/canonicalltd/snapcraft-design/issues/309

### DIFF
--- a/templates/publisher/market.html
+++ b/templates/publisher/market.html
@@ -254,8 +254,9 @@
                   checked="checked"
                 {% endif %}
                 />
-              <label class="u-no-margin--top" for="public-metrics-enabled">Display public metric charts</label>
-              <p class="p-form-help-text">Does not disclose any usage numbers for your application</p>
+              <label class="u-no-margin--top p-tooltip p-tooltip--top-left" for="public-metrics-enabled">Display public popularity charts
+                <span class="p-tooltip__message">These display relative popularity, not exact numbers.</span>
+              </label>
               <input
                 type="hidden"
                 name="public_metrics_blacklist"
@@ -267,7 +268,7 @@
           </div>
 
           <div class="row">
-            <div class="col-4">
+            <div class="col-6">
               <input
                 type="checkbox"
                 name="public_metrics_territories"
@@ -280,7 +281,7 @@
                 {% endif %}
                 />
               <label class="u-no-margin--top" for="public-metrics-territories">Territories</label>
-              <p class="p-form-help-text">Shows usage percentage by country on a map</p>
+              <p class="p-form-help-text">Displays where your snap is being used by country</p>
             </div>
           </div>
         </form>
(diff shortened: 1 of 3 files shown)


static/js/publisher/market/publicMetrics.js
```diff
@@ -4,6 +4,7 @@ const NAMES = {
 
 function publicMetrics(form) {
   const publicMetricsEnabled = form['public_metrics_enabled'].checked;
+  const publicMetricsEle = document.querySelector('.js-public-metrics');
```
bartaz: `publicMetricsEle` - why `e` in the end? typo?

templates/publisher/market.html
```diff
@@ -280,7 +281,7 @@
                 {% endif %}
                 />
               <label class="u-no-margin--top" for="public-metrics-territories">Territories</label>
-              <p class="p-form-help-text">Shows usage percentage by country on a map</p>
+              <p class="p-form-help-text">Displays where you snap is being used by country</p>
```
bartaz: `s/you/your/`

static/js/publisher/market/publicMetrics.js
Lukewh: Just habit - Ele(ment) always done it :D

templates/publisher/market.html
Lukewh: good catch

Greggless:
When I check the '`[ ]` Display public...' the screen jumps around a little. Can we fix the position of the screen so the selected checkbox remains in the same position?

bartaz:
@Greggless I guess it happens only when you are scrolled all the way to the bottom of the page.

We can't do much about it - we are hiding part of the page that is near the bottom, there is nothing else to fill this space, so browser needs to scroll some part of the page from the top.

Lukewh:
@Greggless what @bartaz said... We could just `opacity: 0` the element, but then we'd have loads of white space at the bottom before the footer.

Greggless:
Two options:
### 0 opacity
image

### greyed out
image

bartaz:
@Greggless @Lukewh It was grayed out (disabled) before, not sure why it was changed to hiding it.

But maybe we could add a grayed out style - it seems better for be because it reveals a bit what options are available when charts are enabled.

Lukewh:
@Greggless @bartaz updated screenshot (demo is updated too):
screenshot-2018-4-13 market details for lukotron

bartaz:
@Lukewh Looks good, thanks.

But for me (Chrome Mac OS) help text unnecessarily wraps (doesn't fit `col-4`):

screen shot 2018-04-13 at 10 12 35

Lukewh:
Updated @bartaz I've made it a `col-6` as for the foreseeable future we'll only have 2 toggles anyway.

PR #512, opened 2018-04-13
Remove useless exceptions catch

### LINKED ISSUE #507: Bookmarked URLs don't work after 24h
### Expected behaviour

When accessing protected URLs from browser bookmark or history, it should redirect users to refresh their credentials (the macaroon) instead of freeze with a 502 

screen shot 2018-04-11 at 5 32 26 am

The store recently started enforcing the 24h expiration time for the discharge-macaroon and this error indicated the app is not handling `WWW-Authenticate: Macaroon needs_refresh=1` properly.

If I go back to /account, the app will appropriately refresh the session (the discharge macaroon in it, actually) and then I can access the bookmarked URL without problems.

### Specs

- _URL:_ https://snapcraft.io/account/snaps/spotify/measure (or market)

tbille:
# Summary

Fixes #507 

- Remove useless exeption catch. This exception is never raised
- Because of exception inheritance `ApiError` was blocking the excpetion `MacaroonNeedsRefresh` to be catched.

# QA

- `./run`
- On a brower that you don't use for dev login on http://0.0.0.0:8004/account
- Go on a snap you own
- Bookmark the page
- Reopen the page in more than 24hours
- It should redirect you to the page you bookmarked

### DIFF
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -8,7 +8,6 @@ from dateutil import relativedelta
 from json import dumps, loads
 from operator import itemgetter
 from modules.exceptions import (
-    ApiError,
     ApiTimeoutError,
     ApiConnectionError,
     ApiResponseDecodeError,
@@ -140,8 +139,6 @@ def publisher_snap_measure(snap_name):
             flask.abort(502, error_messages)
     except ApiResponseError as api_response_error:
         flask.abort(502, str(api_response_error))
-    except ApiError as api_error:
-        flask.abort(502, str(api_error))
     except MacaroonRefreshRequired:
         return refresh_redirect(
             flask.request.path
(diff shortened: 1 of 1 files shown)


modules/publisher/views.py
```diff
@@ -140,8 +139,6 @@ def publisher_snap_measure(snap_name):
             flask.abort(502, error_messages)
     except ApiResponseError as api_response_error:
         flask.abort(502, str(api_response_error))
-    except ApiError as api_error:
```
bartaz: Should it be removed, or just moved after `MacaroonRefreshRequired`?
So if the error is `ApiError` but is not `MacaroonRefreshRequired` it will still be caught?

bartaz:
I think it works. At least I visited bookmarked URL after (hopefully) more then 24h and I didn't see any error.

PR #521, opened 2018-04-17
Add tests post market page

tbille:
# Summary

- Add tests for POST endpoint
- Seperated in 2 files:
  - One for fields updates
  - One for screenshots/icon uploads

# QA

- `./run test`

### DIFF
--- a/tests/tests_post_binary_metadata.py
+++ b/tests/tests_post_binary_metadata.py
@@ -1,0 +1,440 @@
+import unittest
+import responses
+import json
+import io
+from tests.endpoint_testing import BaseTestCases
+
+
+class PostBinaryMetadataMarketPage(BaseTestCases.BaseAppTesting):
+    def setUp(self):
+        self.snap_id = 'complexId'
+        snap_name = "test-snap"
+        endpoint_url = '/account/snaps/{}/market'.format(snap_name)
+
+        super().setUp(snap_name, None, endpoint_url)
+        self.authorization = self._log_in(self.client)
+
+    def _get_redirect(self):
+        return (
+            'http://localhost'
+            '/account/snaps/{}/market'
+        ).format(self.snap_name)
+
+    @responses.activate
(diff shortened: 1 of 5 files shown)


tests/tests_post_market.py
```diff
@@ -0,0 +1,568 @@
...
+        called = responses.calls[1]
+        self.assertEqual(
+            'https://login.ubuntu.com/api/v2/tokens/refresh',
+            called.request.url)
+
+        assert response.status_code == 302
+        assert response.location == self._get_redirect()
+
+    @responses.activate
+    def test_update_valid_fied(self):
```
bartaz: Typo in function name `fied`

tests/tests_post_market.py
```diff
@@ -0,0 +1,568 @@
...
+    def test_post_no_data(self):
+        response = self.client.post(
+            self.endpoint_url,
+        )
+
+        assert response.status_code == 302
+        assert response.location == self._get_location()
+
+    @responses.activate
+    def test_update_invalid_fied(self):
```
bartaz: Another typo: `fied`

tests/tests_post_binary_metadata.py
```diff
@@ -0,0 +1,440 @@
+import unittest
+import responses
+import json
+import io
+from tests.endpoint_testing import BaseTestCases
+
+
+class PostBinaryMetadataMarketPage(BaseTestCases.BaseAppTesting):
```
bartaz: There are no tests for removing screenshots?
We could also have a test for removing icon (even if we don't have UI for it yet).

I'm not sure if I seen a test for when images fail to upload (icon is wrong size or screenshots in wrong format).

Also, I think we still have a bug that if you add twice same image (which is possible in our UI) upload will fail. Maybe we could test this specific edge case (and fix it). But not sure if it needs to be part of this PR if it's too big.

nottrobin:
@bartaz can you add issues for the missing cases?

PR #525, opened 2018-04-18
Give all session responses Cache-Control: private

nottrobin:
Make the logic to apply Cache-Control headers apply to *all* responses,
and use the existence of a `flask.session` cookie to decide whether it's
"public" or "private".

QA
--

``` bash
./run
```

Now check the headers in the command-line:

``` bash
$ curl -I http://0.0.0.0:8004 | grep 'Cache'
Cache-Control: max-age=61, stale-while-revalidate=300, stale-if-error=86400, public
```

Now open the browser, go to the homepage, check the "network" tab in developer tools to inspect the response headers. Check again the cache-control is as above.

Now to go http://0.0.0.0:8004/account and login. Check the response headers now have "Cache-Control: private".

Go back to the homepage, check the response headers, should still have "Cache-Control: private".

### DIFF
--- a/app.py
+++ b/app.py
@@ -29,7 +29,7 @@ import modules.public.views as public_views
 import modules.publisher.views as publisher_views
 import modules.publisher.api as publisher_api
 import template_functions
-import decorators
+from decorators import login_required
 from modules.macaroon import (
     MacaroonRequest,
     MacaroonResponse,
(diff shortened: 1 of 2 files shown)


nottrobin:
This solves the problem of users' logged-in state in the header potentially being cached.

However, it leaves an issue:

If a user visits and caches the homepage, then logs in, then returns to the homepage with 1 minute, then they will see the cached (logged-out) page status, until they refresh or revisit after a minute.

This PR should still be merged.

tbille:
can you replace the `/logout` endpoint with this code:
The session wasn't properly emptied

```python
@app.route('/logout')
def logout():
    flask.session.clear()
    return flask.redirect('/')
```

nottrobin:
Actually @tbille let's briefly talk about this. I'm not quite sure if emptying the session is quite the right approach. Also the existing `/logout` route does seem to successfully empty the session for me.

Issue #531, opened 2018-04-18
Remove 'License' field from publisher pages


Lukewh:
Interim fix for: https://github.com/canonicalltd/snapcraft-design/issues/386

PR #532, opened 2018-04-18
Remove license field from market page

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/531

- Removed non-editable 'License' on the market page.

# QA

- Pull the branch
- `./run`
- Visit `/account/snaps/<snap_name>/market`
- Make sure license decleration isn't there

### DIFF
--- a/templates/publisher/market.html
+++ b/templates/publisher/market.html
@@ -189,21 +189,6 @@
 
           <div class="row">
             <div class="col-2">
-              <label>License: </label>
-            </div>
-            <div class="col-8">
-              {{ license }}
-            </div>
-          </div>
-
-          <div class="row">
-            <div class="col-12">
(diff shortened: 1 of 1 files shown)


templates/publisher/market.html
```diff
@@ -187,15 +187,6 @@
...
-          <div class="row">
-            <div class="col-2">
-              <label>License: </label>
-            </div>
-            <div class="col-8">
-              {{ license }}
-            </div>
-          </div>
-
           <div class="row">
```
bartaz: You should also remove one of the rows with `<hr/>` from above or below license.

Lukewh: Thanks @bartaz I thought I had removed it! Fixed.

PR #536, opened 2018-04-19
Open tutorial in new window

Lukewh:
# Done

Fixes https://github.com/CanonicalLtd/snapcraft-design/issues/358
- Added `target="_blank"` to tutorials link on 'no snaps' page

# QA

- Pull the branch
- `./run`
- Create new account/ modify `publisher/views.py` for a user/ object with no snaps
- Visit http://0.0.0.0:8004/account/snaps
- Click the 'Try the tutorial' link, it should open a tutorial in a new window/ tab.

### DIFF
--- a/templates/publisher/account-snaps.html
+++ b/templates/publisher/account-snaps.html
@@ -25,7 +25,16 @@ My snaps — Linux software in the Snap Store
       <div class="col-6">
         <h3 class="p-heading--five">'Hello world' with snapcraft</h3>
         <p>Try this 15 minute tutorial to get to know the basics of the snap format and snapcraft tool.</p>
-        <p><a href="https://tutorials.ubuntu.com/tutorial/create-your-first-snap#0" class="p-button--neutral p-link--external">Try the tutorial</a></p>
+        <p>
+          <a
+            href="https://tutorials.ubuntu.com/tutorial/create-your-first-snap#0"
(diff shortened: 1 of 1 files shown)


templates/publisher/account-snaps.html
```diff
@@ -25,7 +25,7 @@ <h2 class="p-heading--four">Get started&hellip;</h2>
       <div class="col-6">
         <h3 class="p-heading--five">'Hello world' with snapcraft</h3>
         <p>Try this 15 minute tutorial to get to know the basics of the snap format and snapcraft tool.</p>
-        <p><a href="https://tutorials.ubuntu.com/tutorial/create-your-first-snap#0" class="p-button--neutral p-link--external">Try the tutorial</a></p>
+        <p><a href="https://tutorials.ubuntu.com/tutorial/create-your-first-snap#0" target="_blank" class="p-button--neutral p-link--external">Try the tutorial</a></p>
```
bartaz: It's a good security practice to add  `rel="noreferrer noopener"` as well:
https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/

Although in this case we fully control this link and it's our site, so it's not a deal breaker.

Lukewh: definitely worth adding though, thanks @bartaz

PR #558, opened 2018-04-20
Make sure to ignore clicks on disabled buttons

### LINKED ISSUE #518: Can choose more than five screenshots
1\. Go to the “Market” (or equivalent) page for a snap.
2\. Click the “+” button and choose a screenshot, until there are five.
3\. Click “+” again and choose another screenshot.

What happens: You can choose one.

What should happen: the “+” button should be disabled.

bartaz:
Fixes #518 

Because we use event delegation to capture clicks on screenshots and buttons we need to make sure disabled buttons are ignored.

## QA

- ./run or http://snapcraft.io-pr-558.run.demo.haus/
- go to market page
- add 5 screenshots (add button should be disabled)
- click on disabled button
- nothing should happen, you should not be able to add more screenshots

### DIFF
--- a/static/js/publisher/market/screenshots.js
+++ b/static/js/publisher/market/screenshots.js
@@ -127,9 +127,11 @@ function initSnapScreenshotsEdit(screenshotsToolbarElId, screenshotsWrapperElId,
   // delegated click handlers
   document.addEventListener("click", function(event){
     // Delete screenshot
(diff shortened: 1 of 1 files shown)


anthonydillon:
Out of curiosity, why doesn't the event trigger from the anchor tag or button?

PR #561, opened 2018-04-20
Market > Listings, Measure > Metrics

Lukewh:
# Done

- Renam the `market` endpoint `listing`
- Rename `market.html` to `listing.html`
- Rename the `measure` endpoint `metrics`
- Rename `measure.html` to `metrics.html`
- Update python function names
- Update tests
- Update references in templates

# Not done

- Update JS variables and function names

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/account and click around:
- There should be no reference to 'market' or 'measure'
- Visit http://0.0.0.0:8004/account/snaps/<snap_name>/market and http://0.0.0.0:8004/account/snaps/<snap_name>/measure, you should be redirected to the appropriate page

### DIFF
--- a/app.py
+++ b/app.py
@@ -327,16 +327,16 @@ def post_account_name():
     return publisher_views.post_account_name()
 
 
-@app.route('/account/snaps/<snap_name>/measure')
+@app.route('/account/snaps/<snap_name>/metrics')
 @login_required
-def publisher_snap_measure(snap_name):
-    return publisher_views.publisher_snap_measure(snap_name)
+def publisher_snap_metrics(snap_name):
+    return publisher_views.publisher_snap_metrics(snap_name)
 
 
-@app.route('/account/snaps/<snap_name>/market', methods=['GET'])
+@app.route('/account/snaps/<snap_name>/listing', methods=['GET'])
 @login_required
-def get_market_snap(snap_name):
-    return publisher_views.get_market_snap(snap_name)
+def get_listing_snap(snap_name):
+    return publisher_views.get_listing_snap(snap_name)
 
 
 @app.route('/account/snaps/<snap_name>/release')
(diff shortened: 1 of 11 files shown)


tests/test_get_metrics.py
```diff
@@ -8,7 +8,7 @@
 class MeasurePageNotAuth(BaseTestCases.EndpointGetLoggedOut):
```
tbille: Update class name

Lukewh: Done!

app.py
```diff
@@ -345,7 +345,25 @@ def snap_release(snap_name):
...
-@app.route('/account/snaps/<snap_name>/market', methods=['POST'])
+@app.route('/account/snaps/<snap_name>/listing', methods=['POST'])
+@login_required
+def post_listing_snap(snap_name):
+    return publisher_views.post_listing_snap(snap_name)
+
+
+@app.route('/account/snaps/<snap_name>/market')
+@login_required
+def get_market_snap(snap_name):
```
tbille: let's make it more readable: 
```python
def get_market_snap(snap_name):
    return flask.redirect(
        "/account/snaps/{snap_name}/listing".format(
            snap_name=snap_name))
```

tbille:
@Lukewh a little bit annoying but there is still a lot of `market` vars in the JS. (file market.js, in the html ...) Do you wanna update this now? Because I have the feeling that it will never be updated if we don't do it now...

Lukewh:
@tbille I decided not to do it now (as mentioned in this PR description) because there are a lot of variables and functions that are 'marketX' which feels fragile to change when rushing for a release, especially when we're hopefully refactoring the listings page js in the coming month or two.

app.py
```diff
@@ -345,7 +345,23 @@ def snap_release(snap_name):
...
+    return flask.redirect(
+        "/account/snaps/{snap_name}/listing".format(
+            snap_name=snap_name))
+
+
+@app.route('/account/snaps/<snap_name>/measure')
 @login_required
-def post_market_snap(snap_name):
-    return publisher_views.post_market_snap(snap_name)
+def get_measure_snap(snap_name):
```
bartaz: Shouldn't it be `get_metrics_snap` `get_listing_snap` - or maybe even `get_snap_metrics` `get _snap_listing`?

Lukewh: that's just the redirect to metrics, figured would leave them as the legacy name and keep the name formats the same throughout

PR #563, opened 2018-04-23
Get channel map and send to template

tbille:
# Summary

Send channel map to template on snap details page

# QA 

- `./run`
- 0.0.0.0:8004/toto
- should work

This is the structure sent to the template: https://pastebin.canonical.com/p/d4qs6fMRZJ/

### DIFF
--- a/modules/public/logic.py
+++ b/modules/public/logic.py
@@ -281,3 +281,45 @@ def split_description_into_paragraphs(unformatted_description):
         formatted_paragraphs.append(paragraph.replace('\n', '<br />'))
 
     return formatted_paragraphs
+
+
+def convert_channel_maps(channel_maps_list):
+    """
+    Converts channel maps list to format easier to manipulate
+
+    Example:
+    - Input:
+    [
+      {
+        'architecture': 'arch'
+        'map': [{'info': 'release', ...}, ...],
+        'track': 'track 1'
+      },
+      ...
+    ]
+    - Output:
+    {
+      'arch': {
+        'track 1': [{'info': 'release', ...}, ...],
+        ...
+      },
+      ...
+    }
+
+    :param channel_maps_list: The channel maps list returned by the API
+
+    :returns: The channel maps reshaped
+    """
+    channel_maps = {}
+    for channel_map in channel_maps_list:
+        arch = channel_map.get('architecture')
+        track = channel_map.get('track')
+        if arch not in channel_maps:
+            channel_maps[arch] = {}
+        channel_maps[arch][track] = []
+
+        for channel in channel_map['map']:
+            if channel.get('info'):
+                channel_maps[arch][track].append(channel)
+
+    return channel_maps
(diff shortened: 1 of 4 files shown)


bartaz:
We want to display a release date for each channel, is it possible with current data from API or do we need to request an API change for that?

Other thing (could be done separtely) would be to change the structure of data a bit.
In UI we are displaying channel map by architecture and by track, so it would be nice to have the data structure reflect that.
Currently it's a flat list with `architecture` and `track` fields.

Ideally structure of the data would be:

```
channel_map = {
   'amd64': {
       'latest': [
           { channel: 'stable', release: 1, version: '1.0', ... }
           { channel: 'candidate', release: 2, version: '1.1', ... }
           ...
       ]
   }
   'armhf': {
       'latest': [ ... ]
       'othertrack': [ ... ]
   }
}
```

tbille:
@bartaz I can update the structure of the json sent. But it doesn't seem that we can get the release date. This is something that should be asked to the store team.

bartaz:
@tbille Yes, I think it would be helpful to have data structured in a way that will make it easier to access it in template/front-end.

And for the date - we can do a first implementation without it I think, but we should request this to be added to API.

tbille:
@bartaz the channel maps sent to the template is as you asked for. Check the documentation added in the function added in `logic.py`

modules/public/logic.py
```diff
@@ -281,3 +281,44 @@ def external(attrs, new=False):
         formatted_paragraphs.append(paragraph.replace('\n', '<br />'))
 
     return formatted_paragraphs
+
+
+def convert_channel_maps(channel_maps_list):
```
bartaz: No tests for this?

modules/public/logic.py
```diff
@@ -281,3 +281,44 @@ def external(attrs, new=False):
...
+
+    :param channel_maps_list: The channel maps list returned by the API
+
+    :returns: The channel maps reshaped
+    """
+    channel_maps = {}
+    for channel_map in channel_maps_list:
+        arch = channel_map['architecture']
+        track = channel_map['track']
+        channel_maps[arch] = {}
```
bartaz: Does this work properly?
If there are 2 tracks for same architecture the second one will remove the first one if I read the code correctly.

bartaz: I QA'ed it and it doesn't work for example for data like this:

```
    channel_map = [
        {
            'architecture': 'arch',
            'map': [
                {'info': 'release', 'channel': 'stable'},
                {'info': 'release', 'channel': 'beta'}
            ],
            'track': 'latest'
        },
        {
            'architecture': 'arch',
            'map': [
              {'info': 'release', 'channel': 'stable'},
              {'info': 'release', 'channel': 'beta'}
            ],
            'track': 'track2'
        },
        ]
```

modules/public/logic.py
tbille: tests added

modules/public/logic.py
tbille: done

PR #564, opened 2018-04-23
Handle errors account details and snap page

tbille:
# Summary

- Handle errors on account endpoint
- handle errors on `/account/snaps` and `/account/details` endpoint
- Update tests by using BaseTesting structure
Fixes canonical-websites/snapcraft.io#539 

# QA

- `./run`
- 0.0.0.0:8004/account/snaps
- 0.0.0.0:8004/account/details
- Try with different accounts (just registered etc.)

### DIFF
--- a/modules/publisher/api.py
+++ b/modules/publisher/api.py
@@ -64,12 +64,11 @@ def process_response(response):
 
     if not response.ok:
         if 'error_list' in body:
-            api_error_exception = ApiResponseErrorList(
+            raise ApiResponseErrorList(
                 'The api returned a list of errors',
                 response.status_code,
                 body['error_list']
             )
-            raise api_error_exception
         else:
             raise ApiResponseError(
                 'Unknown error from api',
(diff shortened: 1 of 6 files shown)


modules/publisher/api.py
```diff
@@ -64,12 +64,11 @@ def process_response(response):
 
     if not response.ok:
         if 'error_list' in body:
-            api_error_exception = ApiResponseErrorList(
+            raise ApiResponseErrorList(
```
Lukewh: Nice

modules/publisher/api.py
```diff
@@ -91,16 +90,7 @@ def get_authorization_header(session):
 
 
 def get_account(session):
-    authorization = authentication.get_authorization_header(
-        session['macaroon_root'],
-        session['macaroon_discharge']
-    )
-
-    headers = {
-        'X-Ubuntu-Series': '16',
```
Lukewh: I'm guessing we don't need the X-Ubuntu headers anymore?

tbille: nope, here we are requiring the account information this is not needed

PR #575, opened 2018-04-24
Listing page notice/ link to public store

Lukewh:
# Done

Partially fixes https://github.com/CanonicalLtd/snapcraft-design/issues/405
Fixes https://github.com/CanonicalLtd/snapcraft-design/issues/328 and https://github.com/CanonicalLtd/snapcraft-design/issues/383

- Changed padding around the buttons
- Changed 'Apply' to 'Save'
- Added notification text for the following cases:
  - Is not released to stable
  - Is released to stable but is not public
  - Is released to stable and is public

**Note** There is currently no way of knowing which store the snap is released to on this page. Once the field is added we can do some magic with that information. I'm not sure it's an issue for now though as there is only a small number of brand stores with exclusive snaps, afaik.

# QA

You'll need:
- A snap that has not been released to stable, but has the .snap file uploaded
- A snap that has been released to stable and is private
- A snap that has been released to stable and is public

Now:
- Pull this branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/snap_name_from_above_list/listing
- Ensure the correct message is displayed and links go to the appropriate places

# Screenshots
## Public and released to stable
screenshot-2018-4-23 listing details for lukotron test test

## Private and released to stable
screenshot-2018-4-25 listing details for lukotron test test

## Not released to stable
screenshot-2018-4-23 listing details for test-lukewh

### DIFF
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -376,6 +376,14 @@ def get_listing_snap(snap_name):
             flask.request.path
         )
 
+    is_on_stable = False
+    for series in snap_details['channel_maps_list']:
+        for series_map in series['map']:
+            is_on_stable = (
+                is_on_stable or
+                'channel' in series_map and
+                series_map['channel'] == 'stable')
+
     # Filter icon & screenshot urls from the media set.
     icon_urls = [
         m['url'] for m in snap_details['media']
@@ -396,9 +404,11 @@ def get_listing_snap(snap_name):
         "publisher_name": snap_details['publisher']['display-name'],
         "screenshot_urls": screenshot_urls,
         "contact": snap_details['contact'],
+        "private": snap_details['private'],
         "website": snap_details['website'] or '',
         "public_metrics_enabled": snap_details['public_metrics_enabled'],
         "public_metrics_blacklist": snap_details['public_metrics_blacklist'],
+        "is_on_stable": is_on_stable,
     }
 
     return flask.render_template(
@@ -629,6 +639,14 @@ def post_listing_snap(snap_name):
                 else snap_details['website']
             )
 
+            is_on_stable = False
+            for series in snap_details['channel_maps_list']:
+                for series_map in series['map']:
+                    is_on_stable = (
+                        is_on_stable or
+                        'channel' in series_map and
+                        series_map['channel'] == 'stable')
+
             # Filter icon & screenshot urls from the media set.
             icon_urls = [
                 m['url'] for m in snap_details['media']
@@ -666,7 +684,9 @@ def post_listing_snap(snap_name):
                     changes['contact'] if 'contact' in changes
                     else snap_details['contact']
                 ),
+                "private": snap_details['private'],
                 "website": website or '',
+                "is_on_stable": is_on_stable,
                 # errors
                 "error_list": error_list,
                 "field_errors": field_errors,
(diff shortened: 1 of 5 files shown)


modules/publisher/views.py
```diff
@@ -353,6 +353,12 @@ def get_listing_snap(snap_name):
             flask.request.path
         )
 
+    is_on_stable = False
```
tbille: please this code in a function in `public/logic.py` instead

tests/tests_listing.py
```diff
@@ -62,6 +62,7 @@ def test_account_logged_in(self):
             'publisher': {
                 'display-name': 'The publisher'
             },
+            'channel_maps_list': [{'map': [{}]}],
```
tbille:
```
'channel_maps_list': [{'map': [{'info':'info'}]}],
```

Greggless:
## Great work 🏗 

> This listing is not public as it is set to [private](https://dashboard.snapcraft.io/snaps/surl)

I think when a snap is in private, we should link them to d.s.io/snaps/`snap_name` where the publisher can change it's privacy setting. When we [create the admin page](https://github.com/CanonicalLtd/snapcraft-design/issues/423), we can then port this functionality over for a nice shiny experience.

@matthewpaulthomas might want to weigh in on why we aren't linking to dashboard in this instance.

Lukewh:
@Greggless we now have the ability to set whether a snap is private or not, I was thinking we could add it to the listing page, that way we wouldn't need to link anywhere?

tests/tests_listing.py
Lukewh: done, but still fails

Greggless:
@Lukewh If we can do that on snapcraft.io, then great! Plz ignore. 😁

matthewpaulthomas:
> @matthewpaulthomas might want to weigh in on why we aren't linking to dashboard in this instance.

I had written: “This listing is not public because the snap is set to private. Settings”. I hadn’t specified where “Settings” should link to. I’d prefer it to go to the Dashboard UI for now, so that existing users only need to learn two places for the function over time (Dashboard now, the new admin page later) rather than three (Dashboard up to now, the Store Listing page from now, the new admin page later).

Lukewh:
@Greggless there's no way to tell whether or not it's in the global store or not currently - see the **note**  in the description for more details

Greggless:
Ah, sorry I missed that. Ta.

@tbille do we have an API we can call on for which store a snap is in or does this need to be raised with the store team?

Lukewh:
@greggless I've asked celso already

Lukewh:
@Greggless @matthewpaulthomas the demo should be updated now if you want to have another look. I've also updated the screenshot in the description, including the 'settings' link do dashboard.snapcraft.io

Greggless:
@Lukewh thanks

Does this work when the snap title is in the `[Apply] [Revert]` panel? I would suggest the title takes it's place if we don't have space (especially on smol screens)

modules/publisher/views.py
```diff
@@ -656,6 +672,7 @@ def post_listing_snap(snap_name):
                     else snap_details['contact']
                 ),
                 "website": website or '',
+                "is_on_stable": is_on_stable,
```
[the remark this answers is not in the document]
Lukewh: Fixed

Lukewh:
Fixed those bits @bartaz was baaaad code. Should be sorted now. Yeah there's disparity between the /info and /account-info endpoints with the privacy! Filing a bug on launchpad

Issue #534, opened 2018-04-19
Implement distro chart on public pages


Lukewh:
By the power vesting in you

construct this wonderful element and bring forth the snap users

https://github.com/CanonicalLtd/snapcraft-design/issues/308#issuecomment-382703143

Issue #524, opened 2018-04-17
Two checkboxes for one map doesn’t make sense


matthewpaulthomas:
1\. Go to the “Market” page for any snap you admin.
2\. Scroll to the bottom of the page.

What you see:
> ☑ Display public popularity charts
> 
> ☑ Territories
> Displays where your snap is being used by country

What’s wrong with this: It’s two checkboxes controlling the visibility of a single map.

What you should see: either
- one checkbox (since there’s only one chart), or
- no checkboxes at all (since, to my knowledge, no publisher has ever asked to turn off the map, except because of #279, which we should fix anyway).

“Oh, but we might add more charts later!” If that ever happens, add checkboxes for them then. Users shouldn’t have to read our minds, or predict the future, to understand the current UI.

Lukewh:
However, unchecking the initial checkbox will essentially allow publishers to 'opt-out' of any future metrics being added. A single checkbox requires those who are nervous to constantly check the page.

matthewpaulthomas:
Yes, I know that’s what it’s for … but only because I’m on the snapcraft.io development team. To expand on my last paragraph, that the initial checkbox “will essentially allow publishers to 'opt-out' of any future metrics” is understandable to people who:
- have been a member of the snapcraft.io development team for some period longer than two weeks (as demonstrated by [the duplicate](https://github.com/CanonicalLtd/snapcraft-design/issues/376)); or
- have discussed this particular issue with one of those team members; or
- have read this bug report; or
- can read our minds or predict the future.

Right now, those four categories might cover all publisher page users. But very soon, those four categories will cover a tiny proportion of publisher page users. To everyone else, it will look like two checkboxes controlling one chart, which is just nonsense.

Now, a first attempt at making it not nonsense might add a caption to the initial checkbox. Something like this:
> ☑ Display public popularity charts
> Includes charts that may be added in the future.
>
> ☑ Territories
> Where the snap is being used across the world.

But that makes explicit the underlying problem: How on earth am I supposed to know whether it’s a good idea to opt out of charts that may be added in the future, when I have no idea what they will be? I can’t.

> A single checkbox requires those who are nervous to constantly check the page.

That logic would require dozens of checkboxes, to let nervous publishers opt out of other questionable things that we might do in the future: adding a graph of snap size over time, randomizing the background colour of the snap page, introducing a “Most uninstalls this week” leaderboard … I mean, we _probably_ won’t do any of those things. But who can say for sure?

[...]

Issue #313, opened 2018-02-16
Map colors different Measure page and detail page


tbille:
## Problem
For the same snap the colors are different between the measure page and the detail page.

## Test
- `./run`
- http://0.0.0.0:8004/SNAP_YOU_OWN
- http://0.0.0.0:8004/account/snaps/SNAP_YOU_OWN/measure
- Compare colors

## Screenshots
**Publisher metrics:**

![screenshot-2018-2-16 publisher metrics for toto](https://user-images.githubusercontent.com/2707508/36303420-ff4dbb60-1303-11e8-9fe1-3d2cb67b1c45.png)

**Detail page:**

![screenshot-2018-2-16 toto linux software in the snap store](https://user-images.githubusercontent.com/2707508/36303422-ff673f2c-1303-11e8-94e8-31e9e3b52663.png)

Issue #306, opened 2018-02-15
 Weekly active devices != Total Number of users in map 


tbille:
In publisher metrics page there is little difference between the number of Weekly active devices and the ttal number of users in the map:

- Weekly active devices in the sum of every users in all channels
- Map displays only the number of users in the stable channel.

(on a small snap it looks strange)

matthewpaulthomas:
One of [the intended segmentation options for “Weekly active devices”](https://docs.google.com/document/d/1PQzECKRrVZ-y3W7DcdkfeZipBVr5Xl2mpMj5Vqhux2E/edit#heading=h.rtb5soqn673v) is “by risk level”. This is because it’s important to know, for example, what percentage of your users will be affected by a bug that makes it as far as “beta” before it’s fixed. It may influence your QA processes, or the way you advertise the various channels to users, or both.

For that reason, I think “Weekly active devices” should continue including all channels. Therefore, for consistency, the map should include all channels too.

Lukewh:
@tbille is this an API error or the call that's been made to the API?

tbille:
This issue seems fixed.

PR #580, opened 2018-04-24
Distro chart

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/534
Fixes https://github.com/canonical-websites/snapcraft.io/issues/533

# Done

- Added distro graph - without JS :open_mouth: shocker
- Added checkbox to listings page (fixes: https://github.com/canonical-websites/snapcraft.io/issues/524)

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/snap_name
- Make sure the distro chart shows at the bottom of the page
- Visit Some other snaps
- See if the chart shows at the bottom of the page or not (won't show if no users)

- Go to http://0.0.0.0:8004/account/snaps/snap_name/listing
- Check/ uncheck the checkbox at the bottom of the page - visit the public page and see what happens with the graphs

# Screenshots
screenshot-2018-4-24 listing details for toto
screenshot-2018-4-24 toto linux software in the snap store
screenshot-2018-4-24 toto linux software in the snap store 1

### DIFF
--- a/modules/public/logic.py
+++ b/modules/public/logic.py
@@ -53,6 +53,19 @@ def calculate_color(thisCountry, maxCountry, maxColor, minColor):
     return int(colorRange*countryFactor+minColor)
 
 
+def find_metric(full_response, name):
+    """Find a named metric in a metric response
+
+    :param full_response: The JSON response from the metrics API
+    :name: Name of the metric to find
+
+    :returns: A dictionnary with the metric information
+    """
+    for metric in full_response:
+        if metric['metric_name'] == name:
+            return metric
+
+
 def calculate_metrics_countries(geodata):
     """Calculate metrics per countries:
     - Number of users
@@ -158,6 +171,38 @@ def build_country_info(users_by_country, display_number_users=False):
             country_data[country.numeric]['number_of_users'] = number_of_users
 
     return country_data
+
+
+def build_os_info(series):
+    """Build information for OS distro graph
+
+        Input:
+        ```
+        [
+            {
+                'values': [0.4],
+                'name': 'fedora/25'
+            }
+            ...
+        ]
+
+        :param series: A series list from the metrics endpoint
+
+        :returns: A list with the sorted distros
+        """
+    oses = []
+
+    for distro in series:
+        if distro['values'][0]:
+            name = distro['name'].replace('/-', '')
+            oses.append({
+                'name': name.replace('/', ' - '),
+                'value': distro['values'][-1]
+            })
+
+    oses.sort(key=lambda x: x['value'], reverse=True)
+
+    return oses
 
 
 def get_searched_snaps(search_results):
--- a/modules/public/views.py
+++ b/modules/public/views.py
@@ -179,14 +179,22 @@ def snap_details(snap_name):
     country_data = []
     try:
         today = datetime.datetime.utcnow().date()
-        week_ago = today - relativedelta.relativedelta(weeks=1)
+        yesterday = today - relativedelta.relativedelta(days=1)
+        country_metric_name = 'installed_base_by_country_percent'
+        os_metric_name = 'weekly_installed_base_by_operating_system_normalized'
 
         metrics_query_json = [
             {
-                "metric_name": "installed_base_by_country_percent",
+                "metric_name": country_metric_name,
                 "snap_id": details['snap_id'],
-                "start": week_ago.strftime('%Y-%m-%d'),
-                "end": today.strftime('%Y-%m-%d')
+                "start": yesterday.strftime('%Y-%m-%d'),
+                "end": yesterday.strftime('%Y-%m-%d')
+            },
+            {
+                "metric_name": os_metric_name,
+                "snap_id": details['snap_id'],
+                "start": yesterday.strftime('%Y-%m-%d'),
+                "end": yesterday.strftime('%Y-%m-%d')
             }
         ]
 
@@ -195,12 +203,17 @@ def snap_details(snap_name):
             metrics_query_json
         )
 
-        users_by_country = logic.calculate_metrics_countries(
-            metrics_response[0]['series']
-        )
-        country_data = logic.build_country_info(users_by_country)
     except ApiError as api_error:
         status_code, error_info = _handle_errors(api_error)
+
+    oses = logic.find_metric(metrics_response, os_metric_name)
+    normalized_os = logic.build_os_info(oses['series'])
+
+    territories = logic.find_metric(metrics_response, country_metric_name)
+    users_by_country = logic.calculate_metrics_countries(
+        territories['series']
+    )
+    country_data = logic.build_country_info(users_by_country)
 
     # filter out banner and banner-icon images from screenshots
     screenshots = [
@@ -236,6 +249,7 @@ def snap_details(snap_name):
 
         # Data from metrics API
         'countries': country_data,
+        'normalized_os': normalized_os,
 
         # Context info
         'is_linux': (
--- a/templates/snap-details.html
+++ b/templates/snap-details.html
@@ -97,14 +97,34 @@
         </table>
       </div>
     </div>
-    {% if countries %}
-      <div class="row js-snap-map-holder" style="margin-top: 2rem;">
-        <div class="col-12">
+    <div class="row" style="margin-top: 2rem;">
+      {% if countries %}
+        <div class="{% if normalized_os %}col-8{% else %}col-12{% endif %} js-snap-map-holder">
           <h4>Where people are using {{ snap_title }}</h4>
           <div id="js-snap-map" class="snapcraft-territories"></div>
         </div>
-      </div>
-    {% endif %}
+      {% endif %}
+      {% if normalized_os %}
+        <div class="col-4">
+          <h4>Users by distribution (log)</h4>
+          <div class="snapcraft-distro-chart">
+            <div class="snapcraft-distro-chart__names">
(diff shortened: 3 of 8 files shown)


Greggless:
We need (log) next to users, that'll make it more _true_
### Spotify for reference
image

bartaz:
Maybe we should consider hiding tail by default, as it's getting weirdly long:

screen shot 2018-04-24 at 17 24 54

Lukewh:
Hmm @bartaz why does yours not line up? Or did you zoom out?

templates/snap-details.html
```diff
@@ -97,14 +97,35 @@ <h1 class="p-heading--two p-snap-heading__name">{{ snap_title }}</h1>
...
+            <div class="snapcraft-distro-chart__names">
+              {% for distro in normalized_os %}
+                <div class="snapcraft-distro-chart__name">{{ distro.name }}</div>
+              {% endfor %}
+            </div>
+            <div class="snapcraft-distro-chart__bars">
+              {% for distro in normalized_os %}
+                <div
+                  class="snapcraft-distro-chart__bar snapcraft-distro-chart--color-{{ (loop.index % 12) + 1 }}"
+                  title="{{ distro.value * 100 }}"
```
bartaz: Tooltip shows a number, but we don't explain anywhere what this number means.
I can just guess that it's kind of a percent, as the top one says '100', but it's confusing.

screen shot 2018-04-24 at 17 30 07

We should either remove tooltip or make it clearer what the number there means. Also if we want a tooltip it should be a vanilla tooltip like on the map.

templates/snap-details.html
```diff
@@ -97,14 +97,35 @@ <h1 class="p-heading--two p-snap-heading__name">{{ snap_title }}</h1>
         </table>
       </div>
     </div>
-    {% if countries %}
-      <div class="row js-snap-map-holder" style="margin-top: 2rem;">
-        <div class="col-12">
+    <div class="row" style="margin-top: 2rem;">
```
bartaz: It's a bit weird to put margin in inline styles here.

I guess the easiest way to get additional spacing would be to wrap charts into `p-strip is-shallow`, or have some custom class that adds necessary spacing.

bartaz:
Spacing is bit weird with short os names right aligned - especially on smaller screens:

screen shot 2018-04-24 at 17 46 22
screen shot 2018-04-24 at 17 45 45

bartaz:
@Lukewh ah yes, it was zoomed out to fit on the screen ;) so no worries about labels not in line

bartaz:
@Greggless `core` snap is fortunately quite extreme example - but the page there looks really broken because of this long chart.

Even Spotify (the longest one from features snaps I seen) is already too long (being twice as high as map).

Greggless:
If you run through the featured snaps, even popular ones like Skype don't break this (too much)

We have 3 options:
1. Cut the long tail (has been spoke about, and vetoed by the powers that be, but we can try again)
2. Put it in a scrollable window (other powers that be are against this)
3. Place the bars on the X axis (vertical) will look bad in many cases...

bartaz:
@Greggless 

I agree that option 2 (scrollable window) wouldn't be good UX.
For option 1 - sure, we don't want to hide any data, but maybe it would be OK to make it expandable. By default show top 20 and if there is more have a button 'Show all' below the chart that would expand it?

Anyway, we can always release it as it is and try to fix it once it becomes visible issue.

Greggless:
@bartaz 

>  By default show top 20 and if there is more have a button 'Show all' below the chart that would expand it?

I like that idea. Let's merge this, and I can discuss with stakeholders about this option. It'll need some visual design input as I'm not show on if we have a 'show more' pattern.

anthonydillon:
Maybe scrollable with search/filter input like assigning users in GH?

Lukewh:
If I remember correctly, stakeholder said 'scrolling area', but @lyubomir-popov said 'no scrolling area, just let it carry on beyond the bottom of the graph' @Greggless @bartaz @anthonydillon .

@bartaz I've updated the spacing to be consistent, if you could take another look, please?

modules/public/views.py
```diff
@@ -179,14 +179,22 @@ def snap_details(snap_name):
...
+        os_metric_name = 'weekly_installed_base_by_operating_system_normalized'
 
         metrics_query_json = [
             {
-                "metric_name": "installed_base_by_country_percent",
+                "metric_name": country_metric_name,
                 "snap_id": details['snap_id'],
-                "start": week_ago.strftime('%Y-%m-%d'),
-                "end": today.strftime('%Y-%m-%d')
+                "start": yesterday.strftime('%Y-%m-%d'),
```
bartaz: Why was query start/end changed from week/today to yesterday/yesterday?
Wouldn't it affect the data displayed on map and chart? Is it desired?

modules/public/logic.py
```diff
@@ -160,6 +166,21 @@ def build_country_info(users_by_country, display_number_users=False):
     return country_data
 
 
+def build_os_info(series):
```
bartaz: Would it make sense to add some unit tests for this, to make sure we are transforming the data in an expected way?

modules/public/views.py
Lukewh: While talking to Celso he said: today doesn't have any data
I also did a test, and the shading didn't change when I switched to yesterday/yesterday and it makes the request/response from the api a little bit smaller and quicker!

modules/public/logic.py
```diff
@@ -53,6 +53,12 @@ def calculate_color(thisCountry, maxCountry, maxColor, minColor):
     return int(colorRange*countryFactor+minColor)
 
 
+def find_metric(full_response, name):
```
tbille: please add docstring

modules/public/logic.py
tbille: and add docstring please

Lukewh:
@Greggless 
screenshot-2018-4-26 spotify linux software in the snap store

modules/public/logic.py
Lukewh: Added

modules/public/logic.py
```diff
@@ -53,6 +53,19 @@ def calculate_color(thisCountry, maxCountry, maxColor, minColor):
...
+    """Find a named metric in a metric response
+
+    :full_response: The JSON response from the metrics API
```
tbille: `:param full_response`

tests/tests_public_logic.py
```diff
@@ -79,3 +79,29 @@ def test_multiple_track_different_arch_channel_map(self):
         }
 
         self.assertEqual(result, expected_result)
+
+    def test_build_os_info(self):
```
tbille: Can you add just a comment before this function: 
```python
    # build_os_info
    # ===
```

Greggless:
From a human more cleverer than I:
> The label should describe the numbers, so if the axis is scaled logarithmatically, but the numbers are not logarithms, then the label should be x, not log x.
- _[Source](https://www.researchgate.net/post/Should_the_word_log_be_presented_in_the_axis-labels_of_a_graph_or_in_the_figure_legend)_

Lukewh:
screenshot-2018-4-26 spotify linux software in the snap store 1
@Greggless @lyubomir-popov

tests/tests_public_logic.py
Lukewh: Done

modules/public/logic.py
Lukewh: Done

bartaz:
@Greggless @Lukewh I'm not sure if `(log)` in chart title is helpful. I don't think I would know what it means if I didn't know that the chart is using logarithmic scale.

For an outside user 'log' may be not understood at all. And even if they figure out that it means logarithmic scale is it helpful in understanding the chart?

If we need to explain that the chart is logaritmic maybe we should have some help icon (?) with a tooltip giving more descriptive explanation what the chart represents?

bartaz:
Oh, I was too late :) It got merged under my fingers - so maybe something to consider as separate improvement.

Greggless:
These are all valid points I raised in the lead up to the implementation.

If we don't display it as log, then the chart is flat-out lying and people will say "No way is this _nearly_ as popular on Fedora as Ubuntu"

If we display it (algebraically correctly, AIUI) then it doesn't make sense "What the hell are 'log users', and why are they using Linux?"

I like the idea of a tooltip, where we have enough space to explain this chart:
> This chart represents users spread across distributions and scaled logarithmically.

My English and math are bad, so maybe @matthewpaulthomas can use his grammatical wizardry so this makes sense to humans.

matthewpaulthomas:
> maybe @matthewpaulthomas can use his grammatical wizardry so this makes sense to humans

Sorry, but I agree with @bartaz here: I don’t think any brief title can explain to people what a log scale is. Putting a log grid on the graph might help, though it would be very noisy.

If this isn’t changed, at some point some snap publisher will write a muckraking post about how snapcraft.io’s public charts can’t be trusted. They’ll be able to show that the map is wrong thanks to #306 and #313, and they’ll be able to show that this OS chart is wrong because it’s contradicted by the OS view of the active-device chart for the same snap.

Greggless:
> ...maybe @matthewpaulthomas can use his grammatical wizardry so this makes sense to humans.

As in, I am for @bartaz's tooltip:
> "Users are plotted on a logarithmic scale to show the long tail of distributions this snap is being used on, it is not designed to show relative popularity across distributions"

PR #586, opened 2018-04-25
Handle errors get_snap_status

tbille:
# Summary

- Handle errors get snap status
#544 

# QA

- `./run`
- http://0.0.0.0:8004/account/snaps/toto/release

- `run test`


tbille:
No need for it anymore
We are removing the endpoint #594

Issue #587, opened 2018-04-25
/account/snaps page gives odd "congratulations" message


nottrobin:
The account page at https://snapcraft.io/account/snaps says:

> Congratulations!
> You've just released documentation-builder to the "stable" channel!

I actually released documentation-builder to the stable channel over a year ago.

This could be interpreted as patronising for someone who's been successfully using snaps for a long time, but is seeing this (beta) interface for the first time. I think the messaging should be tweaked.

bartaz:
I think it's only shown for users with one snap.

But yes, having one snap is the only condition - so as long as you have one snap only you will always see this.

nottrobin:
> We won't update the copy as the new logic put in should fix this issue

What's the new logic? Is it now based on the updated time?

> (you should publish some more, Robin grinning )

Because, this is not a good argument :stuck_out_tongue_closed_eyes:

matthewpaulthomas:
When I wrote the current text:
- Lyubo and I were doing rapid prototyping of the first-use case, in a hurry
- I failed to consider the case where you come back later
- I did not yet know that outside of BSI, when you first push a snap, it is not released in any channel
- I intended the text to appear only when you had one snap in `edge`, definitely not in `stable`.

So I would be quite discombobulated if “We won't update the copy”.

I’ve suggested a short-term fix in #597:
> - Drop the words “Congratulations!” and “just”.
> - If the snap is public and released in any channel, it’s not classed as “new”.

Longer-term, I think the suggestions we show here should vary depending on whether the snap:
- is released in stable yet
- is public yet
- has screenshots etc yet
- has automated builds.

Lukewh:
I've done a PR to try and address the immediate problem: https://github.com/canonical-websites/snapcraft.io/pull/597

@nottrobin the dumb logic for now https://github.com/canonical-websites/snapcraft.io/pull/597/files#diff-087e2aed723e3c7e3e9d5b69fbbab7eaR114

Greggless:
> If the snap is public and released in any channel, it’s not classed as “new”.

Many of the language guides end with a publish snap so this would skip past the first time experience...

therealjuan:
Merged so closed

PR #597, opened 2018-04-26
Improved first snap experience

Lukewh:
# Done

- Added some more logic: If the oldest upload was over 30 days ago - it's not longer classed as 'new'
- Added close button that's persisted with a cookie
- Improved 'what to do next' from a list to paragraphs

# QA

You'll need:
- An account with 1 newly created snap
- An account with 1 snap that's over 30days old (or hack the code as described below)

To QA:
- Pull the branch
- `./run`
- Visit https://0.0.0.0:8004/account/snaps

To test different cases, edit the following code:
`/modules/publisher/logic.py:26` add `user_snaps = {'snap_you_have': user_snaps['snap_you_have']}`

### DIFF
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -82,6 +82,7 @@ def get_account_snaps():
     user_snaps, registered_snaps = logic.get_snaps_account_info(account)
 
     flask_user = flask.session['openid']
+
     context = {
         'page_slug': 'my-snaps',
         'snaps': user_snaps,
--- a/static/sass/_snapcraft_snap-list.scss
+++ b/static/sass/_snapcraft_snap-list.scss
@@ -11,4 +11,23 @@
       margin-right: $sp-medium;
     }
   }
+
+  .p-snapcraft-first-snap {
+    position: relative;
+
+    &__notification {
+      margin-top: 1rem;
+    }
+
+    &__rocket {
+      margin: {
+        top: -1rem;
+        bottom: -1rem;
+      }
+
+      img {
+        height: 130px;
+      }
+    }
+  }
 }
--- a/templates/publisher/account-snaps.html
+++ b/templates/publisher/account-snaps.html
@@ -65,39 +65,35 @@ My snaps — Linux software in the Snap Store
   {% else %}
     {% if snaps|length == 1 %}
       {% for snap in snaps %}
-      <section class="p-strip--light is-shallow">
-        <div class="row">
-          <div class="col-3">
-            <h3 class="p-heading--four">Congratulations!</h3>
-          </div>
-          <div class="col-6">
-            {% if snaps[snap].latest_revisions[0].channels[0] %}
-              <p>
-                You've just released {{ snap }} to the "{{ snaps[snap].latest_revisions[0].channels[0] }}" channel!
-              </p>
-            {% else %}
-              <p>
-                You've just uploaded {{ snap }}!
-              </p>
-            {% endif %}
-            <ul>
-              <li>
-                Want to improve the listing in stores?
-                <a href="/account/snaps/{{ snap }}/market">Edit Store listing</a>
-              </li>
-              {% if not snaps[snap].latest_revisions[0].channels[0] %}
-                <li>
-                  Is your snap ready to release?
-                  <a href="https://docs.snapcraft.io/build-snaps/release" target="_blank">Release it</a>
-                </li>
-              {% endif %}
-            </ul>
-          </div>
-          <div class="col-3" style="margin-top: -25px">
-            <img alt="Rocket among stars" src="https://assets.ubuntu.com/v1/320a508d-rocket-icon-with-stars.png" />
-          </div>
-        </div>
-      </section>
+        {% if snaps[snap].is_new %}
+          <section class="p-strip--light is-shallow p-snapcraft-first-snap__notification">
+            <div class="row">
+              <div class="col-9">
+                {% if snaps[snap].latest_revisions[0].channels[0] %}
+                  <p>
+                    You've released {{ snap }} to the "{{ snaps[snap].latest_revisions[0].channels[0] }}" channel!
+                  </p>
+                {% else %}
+                  <p>
+                    You've uploaded {{ snap }}!
+                  </p>
+                {% endif %}
+                <p>
+                  Want to improve the listing in stores?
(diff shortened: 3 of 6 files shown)


matthewpaulthomas:
> Added some more logic: If the oldest upload was over 30 days ago - it's not longer classed as 'new'
> Added close button that's persisted with a cookie

This would certainly be less annoying. However, keeping it around for 30 days would mean that almost everyone would feel the need to close the message when there is no practical benefit of closing it (which is why Lyubo and I didn’t include a close button in the first place). It’s a make-work project.

I suggest instead changing the current behavior in two ways:
- Drop the words “Congratulations!” and “just”.
- If the snap is public and released in any channel, it’s not classed as “new”.

That way it doesn’t matter how fast or slow the publisher is, you aren’t reliant on a cookie that might get wiped, you don’t need a close button, and people can’t close it by mistake with no way to get it back.

templates/publisher/account-snaps.html
```diff
@@ -170,3 +172,20 @@ <h4>My registered snap names</h4>
...
 {% endblock %}
+
+{% block scripts %}
+<script>
+  Raven.context(function () {
+    var message = document.querySelector('.js-close-message');
+    if (message) {
+      message.addEventListener('click', function(e) {
+        e.preventDefault();
+        var parent = message.parentNode;
```
bartaz: We have a polyfill for `el.closest()` so may be better to use that instead of assuming it will be always 2 parents.

modules/publisher/views.py
```diff
@@ -111,12 +111,30 @@ def get_account_snaps():
...
+    if len(user_snaps) == 1:
+        for snap in user_snaps:
+            snap_info = user_snaps[snap]
+            revisions = snap_info['latest_revisions']
+            oldest_revision = revisions[-1]
+            revision_since = datetime.datetime.strptime(
+                oldest_revision['since'],
+                '%Y-%m-%dT%H:%M:%SZ')
+
+            print(abs((revision_since - now).days))
```
tbille: :shirt:

Lukewh:
@matthewpaulthomas as @Greggless mentioned on https://github.com/canonical-websites/snapcraft.io/issues/587#issuecomment-384620059

> > If the snap is public and released in any channel, it’s not classed as “new”.

> Many of the language guides end with a publish snap so this would skip past the first time experience...

Does this change your opinion?

modules/publisher/views.py
Lukewh: Don't know what you're talking about.

templates/publisher/account-snaps.html
Lukewh: Done, thanks @bartaz

modules/publisher/views.py
```diff
@@ -111,12 +111,28 @@ def get_account_snaps():
             else:
                 user_snaps[snap] = snaps[snap]
 
+    now = datetime.datetime.utcnow()
+
+    if len(user_snaps) == 1:
+        for snap in user_snaps:
```
nottrobin: This `for` is pointless. Just do `user_snaps[0]`.

nottrobin: w/evs :man_shrugging:

matthewpaulthomas:
> > Many of the language guides end with a publish snap so this would skip past the first time experience...
>
> Does this change your opinion?

Huh. Yes it does. In that case, how about this:
- Drop the words “Congratulations!” and “just”.
- If the snap is public and released in beta/candidate/stable, it’s not classed as “new”.

matthewpaulthomas:
@Lukewh Thank you. I think with the new criteria the 30-day limit is no longer necessary, but that needn’t hold this up.

static/sass/_snapcraft_snap-list.scss
```diff
@@ -11,4 +11,30 @@
       margin-right: $sp-medium;
     }
   }
+
+  .p-snapcraft-first-snap {
+    position: relative;
+
+    &__close {
```
bartaz: Is this still needed?

bartaz:
Spacing is weird on this page:

screen shot 2018-04-27 at 15 35 28

It feels like there should be some spacing between first snap strip and My snaps heading above.

Also (maybe related to that), it's strange to have `p-strip` inside of `p-strip`:
screen shot 2018-04-27 at 15 38 14

templates/publisher/account-snaps.html
```diff
@@ -63,41 +63,37 @@ <h3 class="p-heading--five">Join the community</h3>
...
-            </ul>
-          </div>
-          <div class="col-3" style="margin-top: -25px">
-            <img alt="Rocket among stars" src="https://assets.ubuntu.com/v1/320a508d-rocket-icon-with-stars.png" />
-          </div>
-        </div>
-      </section>
+        {% if snaps[snap].is_new %}
+          <section class="p-strip--light is-shallow">
+            <div class="row u-no-margin--top">
```
bartaz: `u-margin-top` doesn't seem to be needed (as this is first row in the strip)

bartaz:
I guess 'close' related QA steps are no longer relevant?

Lukewh:
@bartaz updated.

New screenshot:
screenshot-2018-5-8 my snaps linux software in the snap store

bartaz:
screen shot 2018-05-08 at 15 47 49

tbille:
rebase on master and move a part of the code added to `publisher/logic.py`

templates/publisher/account-snaps.html
```diff
@@ -65,39 +65,35 @@ <h3 class="p-heading--five">Join the community</h3>
...
+                  <a href="/account/snaps/{{ snap }}/listing">Edit Store listing</a>
+                  {% if not snaps[snap].latest_revisions[0].channels[0] %}
+                    <br />
+                    Is your snap ready to release?
+                    <a href="https://docs.snapcraft.io/build-snaps/release" target="_blank">Release it</a>
+                  {% endif %}
+                </p>
+              </div>
+              <div class="col-3 p-snapcraft-first-snap__rocket">
+                <img alt="Rocket among stars" src="https://assets.ubuntu.com/v1/abd70be7-Rocket+V2.png" />
```
bartaz: Not sure if `alt` is needed on this decorative image, but it does sound quite poetic ;)

Lukewh: :grinning:

PR #605, opened 2018-04-30
Refactor errors catching account details

### LINKED ISSUE #549: Move logic out of get_account_details

tbille:
# Summary

- Refactor errors catching from API
Fixes #549 

# QA 

- `./run`
- http://127.0.0.1:8004/account/details
- Should work as usual

### DIFF
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -8,6 +8,7 @@ from dateutil import relativedelta
 from json import dumps, loads
 from operator import itemgetter
 from modules.exceptions import (
+    ApiError,
     ApiTimeoutError,
     ApiConnectionError,
     ApiResponseDecodeError,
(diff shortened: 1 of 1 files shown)


modules/publisher/views.py
```diff
@@ -26,37 +27,42 @@ def refresh_redirect(path):
...
+        )
+    else:
+        return flask.abort(502, str(api_error))
+
+
+def _handle_error_list(errors):
+    codes = []
+    for error in errors:
+        if error['code'] == 'user-not-ready':
+            if 'has not signed agreement' in error['message']:
```
anthonydillon: This seems a little flaky. Could you create a list of internal error codes to ref here instead of checking the error messages? This would fail and be difficult to catch in QA if someone changes the message. Also, in the future, we may need to localise the messages which will make this logic difficult.

tbille: @anthonydillon unfortunatly we can't really do in another way. This errors code is not unique... We discussed about it with @WillMoggridge I added a comment and an issue so that we move this piece of code in the API module since this is a response handled in the view module.

bartaz: Yep, we had similar issue in BSI - error needs to be identified by message: https://github.com/canonical-websites/build.snapcraft.io/blob/master/src/common/actions/auth-store.js#L248

PR #606, opened 2018-04-30
Add help tooltips to channel map

### LINKED ISSUE #529: Channel map fixes/updates (tooltips)

bartaz:
Fixes #529 

Adds Vanilla help tooltips to explain what 'Architectures' 'Tracks' and 'Channels' are to channel map.

### QA
- ./run or http://snapcraft.io-pr-606.run.demo.haus/
- go to a public page of a snap (ideally one with several tracks) like skype or webstorm
- open channel map (by clicking on 'Show all versions')
- help icons should be visible next to 'Track' and 'Channel' labels
- hovering these icons should reveal tooltip with explanation

screen shot 2018-04-30 at 11 18 58
screen shot 2018-04-30 at 11 18 51

### DIFF
--- a/static/sass/_snapcraft_details_heading.scss
+++ b/static/sass/_snapcraft_details_heading.scss
@@ -148,4 +148,20 @@
       opacity: 0;
     }
   }
+
+  // because of Vanilla `* + *` top margin
+  .p-tooltip__message {
+    margin: 0;
+  }
+
+  // positioning fixes for track tooltip
+  .p-tooltip.is-icon {
+    .p-tooltip__message {
+      left: -3rem; // position in bottom-left under the label and icon
+
+      &::before {
+        left: 3rem; // reposition arrow correctly under the icon
+      }
+    }
+  }
 }
--- a/templates/snap-details/_channel_map.html
+++ b/templates/snap-details/_channel_map.html
@@ -3,12 +3,21 @@
     <div class="col-3">
       <form>
         <div class="js-channel-map-arch-field">
-          <label for="js-channel-map-track-select">Architecture</label>
+          <label for="js-channel-map-track-select">
+            Architecture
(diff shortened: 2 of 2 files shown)


static/sass/_snapcraft_details_heading.scss
```diff
@@ -148,4 +148,20 @@
...
     }
   }
+
+  // because of Vanilla `* + *` top margin
+  .p-tooltip__message {
+    margin: 0;
+  }
+
+  // positioning fixes for track tooltip
+  .p-tooltip--track {
```
anthonydillon: This seems like an issue when applying a tooltip to an icon. It's unlikely any text will be smaller then tooltip icon offset. It would be nice if this could solve that issue by changing the position if on a tooltip or creating an icon related override for all the tooltip directions. 

Therefore, it would be good to have this renamed to `.is-icon` or something and create an issue on Vanilla.

Greggless:
@bartaz I think we can remove the Architecture `[i]`.

bartaz:
@Greggless OK, there was no 'Architecture' tooltip in designs, but it was defined in the spec. But I can remove it.

static/sass/_snapcraft_details_heading.scss
bartaz: @anthonydillon Yes, it's because of using tooltip on the icon.
I updated class name to make it more explicit what it's applied to an icon. I can't make it a fully applicable Vanilla override, because adding custom `is-icon` styles also overrides every tooltip varian, when it's really needed for default (bottom left without any modifier class).

bartaz: Also added Vanilla bug: https://github.com/vanilla-framework/vanilla-framework/issues/1791

PR #615, opened 2018-04-30
Use function to catch exceptions API calls

tbille:
# Summary

- Use function `_handle_errors` to catch exceptions

# QA

- `./run`
- Post a new username
- Post a new agreement
- Should work as usual

### DIFF
--- a/modules/publisher/views.py
+++ b/modules/publisher/views.py
@@ -116,22 +116,12 @@ def post_agreement():
     if agreed == 'on':
         try:
             api.post_agreement(flask.session, True)
-        except ApiTimeoutError as api_timeout_error:
-            flask.abort(504, str(api_timeout_error))
-        except ApiConnectionError as api_connection_error:
-            flask.abort(502, str(api_connection_error))
-        except ApiResponseDecodeError as api_response_decode_error:
-            flask.abort(502, str(api_response_decode_error))
         except ApiResponseErrorList as api_response_error_list:
             codes = [error['code'] for error in api_response_error_list.errors]
             error_messages = ', '.join(codes)
             flask.abort(502, error_messages)
-        except ApiResponseError as api_response_error:
-            flask.abort(502, str(api_response_error))
-        except MacaroonRefreshRequired:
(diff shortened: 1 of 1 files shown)


modules/publisher/views.py
```diff
@@ -116,22 +116,12 @@ def post_agreement():
...
         try:
             api.post_agreement(flask.session, True)
-        except ApiTimeoutError as api_timeout_error:
-            flask.abort(504, str(api_timeout_error))
-        except ApiConnectionError as api_connection_error:
-            flask.abort(502, str(api_connection_error))
-        except ApiResponseDecodeError as api_response_decode_error:
-            flask.abort(502, str(api_response_decode_error))
         except ApiResponseErrorList as api_response_error_list:
             codes = [error['code'] for error in api_response_error_list.errors]
```
bartaz: Why is not `_handle_errors_list` used here as well?

modules/publisher/views.py
```diff
@@ -149,20 +139,10 @@ def post_account_name():
...
         try:
             api.post_username(flask.session, username)
-        except ApiTimeoutError as api_timeout_error:
-            flask.abort(504, str(api_timeout_error))
-        except ApiConnectionError as api_connection_error:
-            flask.abort(502, str(api_connection_error))
-        except ApiResponseDecodeError as api_response_decode_error:
-            flask.abort(502, str(api_response_decode_error))
         except ApiResponseErrorList as api_response_error_list:
             errors = errors + api_response_error_list.errors
```
bartaz: Why is not `_handle_errors_list` used here as well?

modules/publisher/views.py
tbille: `_handle_errors_list` handles known errors and returns an `abort` while for those functions I don't too.

modules/publisher/views.py
tbille: same answer

bartaz:
Wasn't able to test any API errors, but standard flow seems to still work.

PR #617, opened 2018-05-02
'My snaps' > 'Dashboard'; 'Dashboard' and 'Forum' external links

### LINKED ISSUE #522: Forum is a separate site, but snapcraft.io header doesn’t show it as one
1\. Go to [snapcraft.io](https://snapcraft.io/).
2\. Look at the “My snaps” link.
3\. Look at the “Forum” link.

What you see:
2\. “My snaps” has the external-link icon, because it is a separate Web site: different header, color scheme, sign-in system, etc.
3\. “Forum” does not have the external-link icon, despite being a separate Web site: different header, color scheme, sign-in system, etc.

What you should see:
3\. “Forum” should have the external-link icon, the same as “My snaps” does.

We did not include this when the new header was implemented, because we thought the Forum was going to adopt the new header, and thereby become part of the same site. But that was abandoned as “not possible” in [snapcraft-design#276](https://github.com/CanonicalLtd/snapcraft-design/issues/276). So, we should continue showing it as an external site instead.

### LINKED ISSUE #595: “My snaps” links to a mostly-useless UI
1\. Go to [snapcraft.io](https://snapcraft.io/).
2\. Click “My snaps”, and sign in if necessary.

What happens: You end up at [snapcraft.io/account/snaps](https://snapcraft.io/account/snaps), which is roughly 10% done, and therefore mostly not useful to publishers. There’s no way to see which channels any of your snaps is released in, no way to see the revision history, no way to see any information about a release (architecture, core version, confinement, etc), no way to make a new release, no way to add or remove a collaborator, no way to add a video link, no way to set a distribution whitelist/blacklist, no way to make the snap private or public, etc.

What should happen: You end up at [dashboard.snapcraft.io/snaps/](https://dashboard.snapcraft.io/snaps/), where you can do all of those things.

It doesn’t matter that we plan to reimplement all of those things in the future: the people following this link each day over the next few months can’t use something that doesn’t exist yet. And it doesn’t matter that the new pages link back to Dashboard: that’s a good idea regardless of which UI is more functional, but users still shouldn’t be taken to a mostly-useless UI by default.

matthewpaulthomas:
- Fixes #595: Changes “My snaps” back to “Dashboard”, linking to dashboard.snapcraft.io/snaps until snapcraft.io/account becomes more functional.
- Fixes #522: Styles both “Dashboard” and “Forum” as external links, since both are presented as different sites (different header navigation, sign-in sessions, etc.

### DIFF
--- a/templates/_header.html
+++ b/templates/_header.html
@@ -22,10 +22,12 @@
         </li>
         <li class="p-navigation__link" role="menuitem"><a href="https://build.snapcraft.io">Build</a></li>
         <li class="p-navigation__link" role="menuitem">
-          <a class="{% if page_slug and page_slug == 'my-snaps' %} is-selected{% endif %}" href="/account/snaps">My snaps</a>
+          <a class="{% if page_slug and page_slug == 'my-snaps' %} is-selected{% endif %} p-link--external"
+          href="https://dashboard.snapcraft.io/snaps">Dashboard</a>
         </li>
         <li class="p-navigation__link" role="menuitem"><a href="https://docs.snapcraft.io">Docs</a></li>
-        <li class="p-navigation__link" role="menuitem"><a href="https://forum.snapcraft.io/categories">Forum</a></li>
+        <li class="p-navigation__link" role="menuitem"><a href="https://forum.snapcraft.io/categories"
+        class="p-link--external">Forum</a></li>
       </ul>
 
       <ul class="p-navigation__links--right" role="menu">
--- a/templates/_header.html
+++ b/templates/_header.html
@@ -22,7 +22,7 @@
         </li>
         <li class="p-navigation__link" role="menuitem"><a href="https://build.snapcraft.io">Build</a></li>
         <li class="p-navigation__link" role="menuitem">
-          <a class="{% if page_slug and page_slug == 'my-snaps' %} is-selected{% endif %} p-link--external"
+          <a class="p-link--external"
           href="https://dashboard.snapcraft.io/snaps">Dashboard</a>
         </li>
         <li class="p-navigation__link" role="menuitem"><a href="https://docs.snapcraft.io">Docs</a></li>


bartaz:
Two issues:
- when user signs in and is redirected to new 'My snaps' page external link to Dashboard is selected in the main nav
- when user is signed in and goes out from 'My snaps' page (to account details page or to any public snapcraft.io page) they have no way to get to 'My snaps'

matthewpaulthomas:
Thanks @bartaz.
- I think the highlight problem is fixed now.
- I don’t think “no way to get to 'My snaps'” is a problem, given that Dashboard will link to it from every page, so it will be one click from there ([snappy-design-squad#405](https://github.com/CanonicalLtd/snappy-design-squad/issues/405)). Basically, the effect of this change is:

Navigation | Before | After
-|-|-
Get to our complete publisher UI|One or two clicks, depending where you are|One click
Get to our new, incomplete UI|One click|One or two clicks, depending where you are

matthewpaulthomas:
@therealjuan I’ve mailed you and Carla with background for this change.

matthewpaulthomas:
@Lukewh Sure. It’s pretty much a summary of #595, but anyway:
> We’d like your approval on a snapcraft.io header navigation issue.
>
> As you know, since 2016, developers have used Dashboard to publish and
manage snaps. <https://dashboard.snapcraft.io/snaps> Since late last
year, we’ve been developing a replacement for Dashboard. We’ve just
released the first, small portion of it, covering store listings and
metrics. <https://snapcraft.io/account/snaps>
>
> But that’s only a small fraction of what Dashboard does. Many vital
functions, we haven’t implemented (or even designed) replacements for
yet: for example, registering a snap name, making a release, managing
collaborators, tracks and channels, revision history, distribution
whitelist/blacklist, setting a snap as private/public, and viewing
release info. Adding these features will take several more months.
>
> From now on, Dashboard will have a banner inviting you to try the new
pages. <https://github.com/CanonicalLtd/snappy-design-squad/issues/405>
Conversely, snapcraft.io/account has a banner guiding you back to
Dashboard. No-one disagrees on those. And maybe we will decide that a
few Dashboard features can be retired and not replaced. That’s not the
issue either.
>
> The issue is which UI we should link to, right now, in the snapcraft.io
header. Though we’re all proud of the new pages, Greg and I think the
best user experience will be if the header continues linking to
Dashboard, *until* snapcraft.io/account becomes more useful. That
doesn’t require reimplementing every last feature of Dashboard — just
that the new pages be more useful, for more people, than the old ones.

Lukewh:
Thanks @matthewpaulthomas

shivjoshi1996:
not doing it this way

PR #621, opened 2018-05-02
Metrics classes

tbille:
# Summary

Metrics are super confusing in the code. I want to create a module that helps us managing all the metrics that we play with. 

- THIS PR is still in dev 
- This is a test: transform matrics handling into its own module, since the API is used in public and publisher view.
- Metrics have the same structure : create Metric class
- Create Metric class depending on the kind of metrics required: Os, ActiveDevices, Coutries

# QA

- Make sure everything works as usual


modules/metrics/metrics.py
```diff
@@ -0,0 +1,266 @@
...
+
+class Metric:
+    def __init__(
+            self, name, series, buckets, status):
+        self.name = name
+        self.series = series
+        self.buckets = buckets
+        self.status = status
+
+    def to_dictionnary(self):
```
WillMoggridge: You could refer to this through builtin functions, or use them directly:
```
def __dict__(self):
    return self.to_dictionary()
```

So that `Dict(metric_object)` works

modules/publisher/views.py
```diff
@@ -219,27 +219,31 @@ def publisher_snap_metrics(snap_name):
...
-    country_data = public_logic.build_country_info(
-        users_by_country,
-        True)
+    territories_total = 0
+    if country_devices.has_data():
+        territories_total = country_devices.get_number_territories()
 
-    territories_total = logic.get_number_territories(country_data)
+    # TODO this is stupid
+    nodata = not (country_devices.has_data() or active_devices.has_data())
```
WillMoggridge: If you wanted, `has_data` function could be used in `__bool__` which Python calls in truthy checks. Then an alternative to this is:
```
not any([country_devices, active_devices])
```

I'm not sure that solves your comment here but it is something fun.

modules/metrics/metrics.py
tbille: done

modules/publisher/views.py
tbille: done

modules/metrics/metrics.py
```diff
@@ -0,0 +1,281 @@
+from operator import itemgetter
+import pycountry
+
+
+class Metric(object):
+    """This is a basic class for metrics
+
+    :var name: The name of the metrics
```
bartaz: Nit picking, but name of the `metric`.

modules/metrics/metrics.py
```diff
@@ -0,0 +1,281 @@
...
+    :var series: The series dictionnary from the metric
+    :var buckets: The buckets dictionnary from the metric
+    :var status: The status of the metrics"""
```
bartaz: `metric` as well.

modules/metrics/metrics.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def __iter__(self):
+        yield('name', self.name)
+        yield('series', self.series)
+        yield('buckets', self.buckets)
+
+    def __bool__(self):
+        """Verifies if one of the metrics has no data
+
+        :returns True if one of the metrics has no data, False if
```
bartaz: "one of the metrics" or this metric represented by the class?

modules/metrics/metrics.py
```diff
@@ -0,0 +1,281 @@
...
+                latest_active_devices += values[len(values)-1]
+
+        return latest_active_devices
+
+
+class CountryDevices(Metric):
+    """Metrics for the devices in countries.
+
+    :var name: The name of the metrics
+    :var series: The series dictionnary from the metric
```
bartaz: Typo in `dictionnary` - also repeated in some other places.

modules/metrics/metrics.py
```diff
@@ -0,0 +1,281 @@
...
+    :var buckets: The buckets dictionnary from the metric
+    :var status: The status of the metrics
+    :var private: Boolean to add data to the informations returned
```
bartaz: I don't understand what this means.

modules/metrics/metrics.py
```diff
@@ -0,0 +1,281 @@
...
+    to the user
+    :var users_by_country: Dictionnary with aditionnal metrics per country
```
bartaz: `additional`

modules/metrics/helper.py
```diff
@@ -0,0 +1,71 @@
...
+        snap_id, installed_base, metric_period=30,
+        metric_bucket='d'):
+    """Build the json that will be requested to the API
+
+    :param snap_id The snap id
+    :param installed_base_metric The base metric requested
+    :param metric_period The metric period requested, by default 30
+    :param metric_bucket The metric bucket, by default 'd'
+
+    :returns A dictionnary with the filters for the metrics API, by default
```
bartaz: `dictionnary`

modules/metrics/helper.py
```diff
@@ -0,0 +1,71 @@
...
+                end=end)]}
+
+
+def find_metric(full_response, name):
+    """Find a named metric in a metric response
+
+    :param full_response: The JSON response from the metrics API
+    :name: Name of the metric to find
+
+    :returns: A dictionnary with the metric information
```
bartaz: `dictionnary`

modules/metrics/helper.py
tbille: nice catch, thanks

modules/metrics/helper.py
tbille: nice catch, thanks

PR #622, opened 2018-05-02
Remove Talisker

tbille:


WillMoggridge:
This was related to a possible theory about slowdown in the environment but this currently looks like it isn't related to Talisker.

PR #623, opened 2018-05-02
Focusable screenshots in listing form

### LINKED ISSUE #414: Screenshots - tabbing through images
It'd be helpful if the user was able to tab through the screenshots individually, rather than skipping the whole section
### Before `tab`
screen shot 2018-03-19 at 10 12 01
### After `tab`
screen shot 2018-03-19 at 10 13 03

#
triaged from #365

bartaz:
Fixes #414

Makes screenshots in listing form focusable (also with keyboard).
Focused screenshot has blue border (just like any other focused input). To make screenshots selection visible checkboxes where added on top of screenshots, so that selected screenshots have checkbox checked.
Multiple screenshots can be selected (multiple checkboxes checked) and therefore multiple screenshots can be deleted at once (by selecting them and clicking delete button).

### QA

- ./run or  http://snapcraft.io-pr-623.run.demo.haus/
- go to listing page of any snap (ideally with screenshots already)
- use keyboard tab to tab through the form (and screenshots)
- once screenshot is focused hit space to select/unselect it
- use mouse to click on screenshots to select/unselect it
- select multiple screenshots and delete them
- make sure any other screenshots functionality (adding screenshots, full screen, removing screenshots before uploading, etc) is still working as before

#### Not selected, not focused
screen shot 2018-05-02 at 16 43 33
#### Focused
screen shot 2018-05-02 at 16 44 13
#### Multiple selected
screen shot 2018-05-02 at 16 44 24

### DIFF
--- a/static/js/publisher/market/screenshots.js
+++ b/static/js/publisher/market/screenshots.js
@@ -8,13 +8,16 @@ const templates = {
 
   screenshot: (screenshot) => `
     <div class="col-2">
-      <div class="p-screenshot__holder ${screenshot.status === 'delete' ? 'is-deleted' : ''}">
+      <label
+        ${screenshot.status === 'delete' ? '' : 'tabindex="0"'}
+        class="p-screenshot ${screenshot.status === 'delete' ? 'is-deleted' : ''} ${screenshot.selected ? 'is-selected' : ''}">
+        <input class="p-screenshot__checkbox" tabindex="-1" type="checkbox" ${screenshot.selected ? 'checked="checked"' : ''} >
         <img
-          class="p-screenshot ${screenshot.selected ? 'selected' : ''}"
+          class="p-screenshot__image"
           src="${screenshot.url}"
           alt=""
         />
-      </div>
+      </label>
     </div>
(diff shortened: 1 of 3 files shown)


Greggless:
Beautiful! A couple things,

### Design:
- Can we hide the unselected checkbox? It may be a bit odd to see a white square on a image

### Current behaviour:
1. Double clicking unselected screenshot, selects then deselects
1. Double clicking selected screenshot, opens in fullscreen
1. If I have two screenshots (A & B), and B is selected, when I double click A, B is displayed fullscreen

### Expected behaviour
1. Double clicking any screenshot should fullscreen it
1. See **1.**
1. If multiple screenshots are selected, it should still behave as **1.**

bartaz:
@Greggless Good catch with double clicking - should be fixed now.

Also I hide checkboxes by default - but instead of showing them only when selected I show them when hovered/focused. I think it nicely shows that interaction is possible. Let me know if it OK to keep it this way.

Greggless:
That's great, thanks @bartaz

PR #633, opened 2018-05-07
Make sure to warn users when leaving page during upload/post

### LINKED ISSUE #603: Lose saved changes when navigating away from Listing page
### Expected behaviour
- I upload some hefty screenshots
- I click save
- save takes time because of my slow interent
- move away from page
- I get a warning about: **Upload in progress** `[leave anyway]` or `[wait]`

bartaz:
Fixes #603 

Warn users that they are leaving a page when there is upload/post in progress (after clicking on Save on Listing page).

### Note:

There is a [bug in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=152649) that causes upload/post to be canceled regardless of the choice user makes in the warning dialog.
So in Chrome we **don't** show warning dialog when user tries to leave during upload, because it would be canceled anyway.

### QA

Use Firefox:

- ./run or  http://snapcraft.io-pr-633.run.demo.haus/
- go to Listing page of any snap
- be on a slow network, or throttle the network speed in developer tools, or choose a large screenshot to upload
- click Save
- while browser is uploading click on any other link on the page (to Measures, main nav)
- warning dialog should be shown that you are leaving page with unsaved changes
- you can choose to leave the page (and cancel ongoing upload/post) or stay on the page (and continue upload)


bartaz:
@Greggless @therealjuan It seems that there is no reliable way to tell if user decided to stay on the page or leave it (in the dialog), so in Chrome that would leave them with canceled upload anyway.

Should we then disable this for Chrome? Meaning: when user has a 'Save' in progress and they navigate away or close the tab, they will not be warned about loosing changes.
It will still work in Firefox (or others).
And it will still work (in all browsers) if there are unsaved changes (before save was clicked).

PR #634, opened 2018-05-08
Add register name endpoint with basic template

tbille:
# Summary

Add register name endpoint with basic template
Fixes https://github.com/CanonicalLtd/snappy-design-squad/issues/481

# QA

- `./run`
- http://127.0.0.1:8004/account/register-name
- Try to register a snap
- Go on my snaps and check if name is well registered

### DIFF
--- a/templates/publisher/register-name.html
+++ b/templates/publisher/register-name.html
@@ -1,0 +1,70 @@
+{% extends "_layout.html" %}
+
+{% block title %}
+Register new Snap name
+{% endblock %}
+
+{% block content %}
+<div class="p-strip">
+
+  <div class="row">
+    <h1>Register a snap name</h1>
+  </div>
+  {% if errors %}
+      <div class="row">
+        <div class="col-12">
+          {% for error in errors %}
+            <div class="p-notification--negative">
+              <p class="p-notification__response">
+                  {{ error.message }}
+              </p>
+            </div>
+          {% endfor %}
+        </div>
+      </div>
+  {% endif %}
+
+  <form method="POST">
+    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+
+    <div class="row">
+      <div class="col-8">
+        <div class="p-form-validation">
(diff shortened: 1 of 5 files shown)


tbille:
@Lukewh 

> It would be good to signify that the snap will be registered on the global store.

This a really temporary basic template. I am working on the new template right after: https://raw.githubusercontent.com/CanonicalLtd/snapcraft-design/master/dashboard%20legacy%20work/01%20-%20Register%20snap.png?token=AClQNHoxVsQrYY96EfzqmuEIY8q6aojJks5a-qRwwA%3D%3D

> Can we do 'dry_run' the first time to make sure the name is available too? 

This could be cool, but we need to check how we want to manage this.
It could be nice that when the user goes out of the text input a call would be made to the store to check if the name is available.

templates/publisher/register-name.html
```diff
@@ -0,0 +1,50 @@
...
+        </div>
+      </div>
+    </div>
+
+    <div class="row">
+      <div class="col-8">
+        <div class="p-form-validation">
+          <label for="snap-title">Registrant Comment:</label>
+          <div class="p-form-validation__field">
+            <textarea class="p-form-validation__input u-no-margin" name="registrant_comment" rows="10" required></textarea>
```
bartaz: Comment is not required, shouldn't be marked as so.

bartaz:
It would be good to at least show some error messages when registering fails. For example when you try to register already registered or reserved name (like: `firefox`). Currently there is no info at all.

Greggless:
If a name is taken, then they should see something like this:

<hr>

image

<hr>

The 'file a dispute' form is similar to the 'register name' form, but includes a registrant comment and the button is labelled differently

tbille:
@Greggless @Lukewh 
The objective of this PR was mainly to build the backend.
The rest of the template design will be done in this issue: https://github.com/CanonicalLtd/snappy-design-squad/issues/482

PR #636, opened 2018-05-08
Make sure nodejs shows up on the store

Lukewh:
# Done

Semi-temporary hack to ensure the latest nodejs shows on snapcraft.io.

## Why?

Node don't use the 'Latest' track for their stable track, instead they use `:version:/stable` (currently '10/stable'). This is temporary until default tracks are implemented in the store.

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/node or http://snapcraft.io-pr-636.run.demo.haus/node
- Ensure the page loads

### DIFF
--- a/modules/public/views.py
+++ b/modules/public/views.py
@@ -153,8 +153,11 @@ def snap_details(snap_name):
     """
 
     error_info = {}
+    default_channel = logic.get_default_channel(snap_name)
+
     try:
-        details = api.get_snap_details(snap_name)
+        details = api.get_snap_details(
+                snap_name, default_channel)
     except ApiTimeoutError as api_timeout_error:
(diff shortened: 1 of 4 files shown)


modules/public/views.py
```diff
@@ -153,8 +153,12 @@ def snap_details(snap_name):
     """
 
     error_info = {}
+    default_channel = logic.get_default_channel(snap_name)
+
     try:
-        details = api.get_snap_details(snap_name)
+        details = api.get_snap_details(
```
tbille:
```python
details = api.get_snap_details(
    snap_name, default_channel)
```

Lukewh: Done xox

Issue #612, opened 2018-04-30
Account dropdown changes size for short names


Lukewh:
Closed:
![screenshot-2018-4-30 my snaps linux software in the snap store](https://user-images.githubusercontent.com/479384/39433545-0d09620e-4c8e-11e8-82ba-abc64eda6543.png)

Open:
![screenshot-2018-4-30 my snaps linux software in the snap store 1](https://user-images.githubusercontent.com/479384/39433559-1199d9c0-4c8e-11e8-84b1-a3e3f4400e1f.png)

lyubomir-popov:
I don't think the menu should match the width of the dropdown. It should be width of name (up to a certain max-width, then truncated with ellipsis) + padding on both sides. Then the dropdown should be right aligned so that the right edge of menu and dropdown coincide, with the left end free to be as wide as needed. 

This will make the menu not align with the keyline of the dropdown menu items, and that's ok.

Would be nice if on mobile the dropdown snaps to full width.

therealjuan:
Fixed on latest vanilla update

PR #640, opened 2018-05-09
Moves “My snaps” and adds “Build with GitHub” to account menu

matthewpaulthomas:
Implements @therealjuan’s [proposal from #595](https://github.com/canonical-websites/snapcraft.io/issues/595#issuecomment-386573579) so it can be demoed:
> 1. Remove 'My Snaps' from the top navigation
> 2. Include 'My published Snaps' in the account navigation as well as "Build with GitHub"

I assume this means remove both “My snaps” and “Build” from the header, and add both to the account menu instead, rather than moving one and duplicating the other.

Also changes “My snaps” to “My published snaps” to match, in both `account.html` and `account-snaps.html`.

I’m marking this as “Don’t merge” for now because I think three design issues would need addressing first:
- You need to sign in to an Ubuntu One account merely to navigate from snapcraft.io to build.snapcraft.io, because otherwise the header doesn’t contain the menu that contains the “Build with GitHub” item.
- On any build.snapcraft.io page (assuming this was implemented there too), no section in the header would be highlighted, because you’d be in a section that’s inside the account menu.
- On any publisher page, no section in the header would be highlighted, because you’d be in a section that’s inside the account menu.


therealjuan:
Some comments looking at the demo:

### Keep Build in the top navigation
Sorry if it wasn't completely explained but the idea was to include not to remove from the top navigation. 

Obviously we would need to apply the changes to the build codebase but having the build homepage on snapcraft.io codebase will help us to maintain consistency across headers and footers of the site. Probably to move this, we need a proper issue to be raised to also consider the different use cases.

### Highlight the account dropdown
This is actually happening at the current implementation of snapcraft.io if you visit your account section and visit account details.  We can highlight the account dropdown so users are aware they are within their account section.

### Account dropdown styling
It seems different when opening and closing the account dropdown, margin on the right jumps quite a bit

therealjuan:
> @therealjuan are you suggesting we add the dropdown to build.snapcraft.io too?

Probably will be better to create a new issue to "integrate build.snapcraft.io" homepage to our Snapcraft.io codebase and have a think about the later journey on build.

> If we're adding more to the dropdown maybe we want to add some delimiter before the 'Sign out' button to separate it?

Probably a horizontal line would make sense to distinguish **Sign out** from the rest

matthewpaulthomas:
> ### Keep Build in the top navigation
> Sorry if it wasn't completely explained but the idea was to include not to remove from the top navigation.

Done. As implied above, though, I think it’s quite strange to link to have two links to “Build” in the header.

> ### Highlight the account dropdown
> This is actually happening at the current implementation of snapcraft.io if you visit your account section and visit account details.

That does not happen on snapcraft.io for me, and I don’t see any code in the template that would do it.

I’ve added code that will highlight the menu whenever the URL contains “`account`”. However, this needs tweaking so that it does not trigger in the edge case where you’re looking at the Store page for a snap that happens to have “`account`” in its name. The fix for that will look like the fix for #658.

therealjuan:
> I think it’s quite strange to link to have two links to “Build” in the header.

The idea was to be able to access the repos from your account dropdown, not redirecting the user to the build homepage as that is not useful at all. So ideally you would be able from the account dropdown to Sign in with your Github and (after signing in with Github) to View your repos

Lukewh:
Assigning @bartaz and @therealjuan for reviews - @matthewpaulthomas and I both worked on it, so shouldn't really review it :thinking:

templates/account.html
```diff
@@ -35,7 +35,7 @@ <h1 class="u-no-margin">{{ username }}</h1>
   <section class="p-strip is-shallow">
     <div class="row">
       <div class="col-12">
-        <h3>My snaps</h3>
```
bartaz: @Lukewh Is this template used anywhere? I suppose it was replaced by `account-details` and `account-snaps` (?)

bartaz:
Because menu items have now longer text the issue with dropdown width (#612) is even more visible now:

dropdown

Probably no need to fix it in this PR, but we may want to increase priority of this one.

Lukewh:
@bartaz I've updated the PR to remove the account.html template and to improve the jumping dropdown!

PR #653, opened 2018-05-11
Adds install tab to channel map overlay

### LINKED ISSUE #619: Add install tab to channel map overlay

bartaz:
Fixes #619 

Adds new tab to channel map overlay with install instructions for stable version and 'View in store' install button, that on supported system will open native store application.

### QA

- ./run or http://snapcraft.io-pr-653.run.demo.haus/
- go to details page of any snap
- open channel map via 'Show all versions' button
- on top there is a new tab with snap name and current version
- click it to open install tab
- you should see CLI install instructions for stable version (same as on snap details page)
- you should see 'View in store' button
- if you are on supported system clicking on a button should open store app for given snap

screen shot 2018-05-14 at 15 01 51

screen shot 2018-05-14 at 15 02 09

### DIFF
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -152,7 +152,7 @@ export default function initChannelMap(el, packageName, channelMapData) {
   const channelMapEl = document.querySelector(el);
   const channelOverlayEl = document.querySelector('.p-channel-map-overlay');
 
-  const defaultTab = channelMapEl.querySelector('.p-tabs__link[aria-controls=channel-map-tab-versions]');
+  const defaultTab = channelMapEl.querySelector('.p-tabs__link[aria-controls=channel-map-tab-install]');
   initTabs(channelMapEl);
 
   let closeTimeout;
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -123,7 +123,31 @@ function initTabs(el) {
   });
 }
 
+function initOpenSnapButtons() {
+  document.addEventListener('click', (event) => {
+    const openButton = event.target.closest('.js-open-snap-button');
+
+    if (openButton) {
+      const name = openButton.dataset.snap;
+      let iframe = document.querySelector('.js-snap-open-frame');
+
+      if (iframe) {
+        iframe.parentNode.removeChild(iframe);
+      }
+
+      iframe = document.createElement('iframe');
+      iframe.className = 'js-snap-open-frame';
+      iframe.style.position = 'absolute';
+      iframe.style.top = '-9999px';
+      iframe.style.left = '-9999px';
+      iframe.src = `snap://${name}`;
+      document.body.appendChild(iframe);
+    }
+  });
+}
+
 export default function initChannelMap(el, packageName, channelMapData) {
+  initOpenSnapButtons();
 
   const channelMapEl = document.querySelector(el);
   const channelOverlayEl = document.querySelector('.p-channel-map-overlay');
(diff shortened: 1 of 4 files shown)


Lukewh:
Just thinking to make it a nicer experience for those that the button doesn't work for, we could:

- Add some kind of spinner/ change the text
- After a second or 2, reset the text and...
- Add a message below the button along the lines of 'Didn't work? Please use the CLI snippet to the right >'

If it did work, people won't care about the message.
If it didn't work, people have a next action

@therealjuan @Greggless wdyt?

Lukewh:
@bartaz I think so. It's why I want @therealjuan and @Greggless to chime in

evandandrea:
It was pointed out in the Product sprint that we overuse the copy button pattern, particularly in cases where the shell commands you're copying are not complex.

"Or a snap enabled store" does not make sense. Isn't this the Snap Store?

Also, it's not enough to just have snapd installed. You need snapd and whatever [plugin](https://packages.ubuntu.com/artful/gnome-software-plugin-snap) is needed for your desktop storefront to communicate with snapd.

Lukewh:
I agree with @evandandrea regarding the "Or a snap enabled store", just above we have "Store" highlighted in the website navigation. Surely something more generic like 'GUI' or 'desktop store' is better?

I'm not sure about your second point @evandandrea - that help text is for the CLI install instructions, not the GUI part

Lukewh:
I also think the button should be something like "More..." and opens the dropdown in the first tab by default.

evandandrea:
Yes, ignore the latter point I made. That's my mistake.

therealjuan:
What about tweaking the wording to say:

> Got Ubuntu 16.04 or later? <- Headline
> **View in Desktop store** <- Call to action
> Link not working? Make sure a Desktop store is installed <- Caption

We can display the caption message after the user pressed on **View in Desktop store** as we are not able to detect if the action worked or not.

I couldn't find any link that might be useful to point users towards to in the caption message, but we can always update afterwards if we find / create one if need be  ✨

Lukewh:
@bartaz I think we should do a release this morning, then merge this afterwards, if that's ok?

bartaz:
@Greggless @therealjuan Quick mock how Juan's suggestion would look:

screen shot 2018-05-14 at 09 36 17

From layout point of view it looks a bit misaligned compared to CLI instructions - because they have additional caption above. So maybe we can add something above 'View in Desktop store' button as well? or remove from above CLI?

Also 'Link didn't work'... It's a button that user clicks, so maybe 'Button didn't work...' instead?

therealjuan:
Thanks for the mock @bartaz 

What about...
* Rephrase _Link not working?..._ to "Make sure you have a Desktop store installed" with no need to display only after clicking
* Remove _(or prefer the command line)_  below the snapd section

Might be a good idea to talk about Desktop store somewhere where we can link to here but for now, we can just leave it as is with no need to show the caption only after clicking the button.

evandandrea:
@therealjuan would you be amenable to "Link not working? Make sure [snap support]() is enabled in your Desktop store."

therealjuan:
Sure thing. That makes sense but _Link not working?_ might not be needed and we can go with:

> Make sure snap support is enabled in your Desktop store.

bartaz:
@evandandrea Anywhere where 'snap support' can be linked to? Your current link is to this PR.

evandandrea:
We (@popey @flexiondotorg) can update the individual [install docs pages](https://docs.snapcraft.io/core/install-elementary-os) to mention how to get the plugin installed.

bartaz:
@therealjuan @Greggless Updated to the latest comments: http://snapcraft.io-pr-653.run.demo.haus/skype

Could you give it a review and UX+1 if it's fine to merge?

therealjuan:
This is fine to merge for now but we need to iterate the installation flow design. Right now, as it stands we have duplicate information (command line instructions) that is displayed (again) to the user after pressing Install...

Let's merge this and continue to improve the Installation flow

PR #667, opened 2018-05-17
Make Install more prominent and remove duplication #666

### LINKED ISSUE #666: Make Install more prominent and remove duplication
This is an iteration from #620 and #653 and based on [this design](https://github.com/CanonicalLtd/snapcraft-design/blob/master/public-facing/install-button/Snap%20details%20with%20install%20button.png)

## Current status

image

On the top section of a Snap details page, we are directing users to either install using Command line or View All versions. We recently added the ability to install with one button using a Desktop store with snap support (i.e. GNOME Software).

image

As it currently stands the command line stays on the top of the page and if a user clicks on Install, we are showing the same information again as well the ability to install from the store (and visit all versions).

## How to iterate the current solution?

- Move `command-line instructions` from the header to below the snap description
- Make `Install` a primary call to action
- Reduce copy from `Show All Versions` to `All Versions`
- `Install` and `All versions` open their relevant tabs on the Channel map

image

therealjuan:
# Done

Fixes #666  

- Move `command-line instructions` from the header to below the snap description
- Make `Install` a primary call to action
- Reduce copy from `Show All Versions` to `All Versions`
- `Install` and `All versions` open their relevant tabs on the Channel map

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/vlc
- Ensure copy and layout is updated
- Ensure `Install` and `All versions` open the relevant tabs on the channel map

# Screenshot

image

### DIFF
--- a/static/sass/_snapcraft_details_heading.scss
+++ b/static/sass/_snapcraft_details_heading.scss
@@ -46,7 +46,7 @@
   .p-snap-install {
 
     &__versions {
-      margin-top: 73px; // align with install instructions
+      margin-top: 68px; // align with install instructions
       text-align: right;
     }
 
--- a/static/sass/_snapcraft_details_heading.scss
+++ b/static/sass/_snapcraft_details_heading.scss
@@ -43,21 +43,9 @@
     }
   }
 
-  .p-snap-install {
-
-    &__versions {
-      margin-top: 68px; // align with install instructions
-      text-align: right;
-    }
-
-    .p-code-snippet {
-      font-size: 1rem;
-      margin-top: $sp-x-small;
-    }
-
-    @media screen and (min-width: $breakpoint-medium) {
-      margin-top: 41px; // align with channel map
-    }
+  .p-snap-install-buttons {
+    margin-top: 68px; // align with snap name (60px icon + 8px margin)
+    text-align: right;
   }
 
   .p-view-store-button,
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -1,3 +1,5 @@
+/* global ga */
+
 const LATEST = 'latest';
 
 function setTrack(arch, track, packageName, channelMap) {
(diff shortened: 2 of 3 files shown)


Greggless:
the `copy-to-clipboard` button highlights the code snippet in the `install-panel` and `all versions` but not in the snap details page. Anyone know why this would be happening? @bartaz @Lukewh ?

therealjuan:
As pointed out by @matthewpaulthomas removing `command-line instructions` from the page seems like a good move to avoid confusion on which version you are installing and be true to the name of the issue: remove duplication. Also, it would solve the copy action problem pointed by @Greggless.

In regards to `All versions` it feels like something to take into account in the future work to be done on the channel map so will reference this comment on the later issue.

static/sass/_snapcraft_details_heading.scss
```diff
@@ -46,7 +46,7 @@
   .p-snap-install {
 
     &__versions {
-      margin-top: 73px; // align with install instructions
+      margin-top: 68px; // align with install instructions
```
bartaz: @therealjuan Is this 68px used to align buttons with something? Snap name baseline I guess? - then we should update the comment in code, as it still says about install instructions.

Lukewh:
Can we add some GA events to these CTA's too before we merge this - would be good to see what people are using to help drive decisions in the future.

bartaz:
@Lukewh GA events added (to buttons that open overlay dialog and to 'View in store' button): https://github.com/canonical-websites/snapcraft.io/pull/667/commits/c0bfaacb73478a17d40328ebe1607d2e33298c63

bartaz:
@therealjuan Didn't know that we can set it up 'remotely' via tag manager. Although, I generally don't trust such third party injected code. I feel like keeping it in our codebase makes sure it works when we want it to work and is up to date when we do changes.

therealjuan:
That makes sense @bartaz. We can always use Tag Manager to trial events ahead of adding to the codebase or testing events.

Probably event tracking should be done on every feature implementation to identify how effective and keep track of this functionality.

lyubomir-popov:
I'm adding design plus 1, but I do have one small concern - magic numbers are prone to causing misalignments. Would be better to reorganize the markup so the heights of whatever you're aligning with naturally push things down.

static/sass/_snapcraft_details_heading.scss
lyubomir-popov: magic number alert :)

bartaz:
@lyubomir-popov This section is evolving a lot recently and possibly another changes are coming here, so I guess it's easier to keep the 'magic' margin for now.

Hopefully we can update it at least when we move to newest Vanilla.

PR #670, opened 2018-05-17
Add style to reserve name templates

tbille:
# Summary

Fixes https://github.com/CanonicalLtd/snap-squad/issues/482
- Implement design for reserve name endpoint
- Add template to claime a dispute

- Stores not handled yet

Designs provided:

**Register name**
image
**Name dispute**
image

# QA

- `./run`
- http://127.0.0.1:8004/account/register-name
- Register a new snap
- Should redirect to `/account`

- http://127.0.0.1:8004/account/register-name?snap_name=SNAPNAME&is_private=True
- Fields should be filled
- Register a new snap
- Should redirect to `/account`

- http://127.0.0.1:8004/account/register-name
- Register a snap that already exists and that you don't own (I made one for you :heart: `test-register-snap-qa`)
- Should redirect your to: http://127.0.0.1:8004/account/register-name?snap_name=test-register-snap-qa&is_private=False&conflict=True
- Claim a dispute
- THERE IS A PROBLEM WITH THE API SO FOR NOW YOU WILL BE REDIRECTED TO THE SAME PAGE: https://bugs.launchpad.net/snapstore/+bug/1771767
- Should redirect you to `/account`

Love and prosperity to you

### DIFF
--- a/templates/publisher/register-name.html
+++ b/templates/publisher/register-name.html
@@ -8,8 +8,21 @@ Register new Snap name
 <div class="p-strip">
 
   <div class="row">
-    <h1>Register a snap name</h1>
+    {% if conflict %}
+      <h1 class="p-heading--three">Request reserved name</h1>
+    {% else %}
+      <h1 class="p-heading--three">Register a snap name</h1>
+    {% endif %}
   </div>
+
+  <div class="row">
+    {% if conflict %}
+      <h2 class="p-heading--five">You can request a transfer of ownership of this name or <a href="/account/register-name">try a different name</a></h2>
+    {% else %}
+      <h2 class="p-heading--five">Before you can push your snap to the store, its name must be registered</h2>
+    {% endif %}
+  </div>
+
   {% if errors %}
       <div class="row">
         <div class="col-12">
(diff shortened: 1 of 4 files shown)


Greggless:
+8,237,204

templates/publisher/register-name.html
```diff
@@ -10,6 +10,11 @@
   <div class="row">
     <h1>Register a snap name</h1>
```
bartaz: `<h1>` seems too large comparing to other headings used around snapcraft.io

'My snaps' main heading uses: `<h1 class="p-heading--three">My snaps</h1>`, so I guess we should use same sizing here.

bartaz:
@tbille I don't see the 'Store' radio button or dropdown - should I?

tbille:
@bartaz No it's normal for now we don't have an enpoint the get the stores so we don't add them to template until we can start playing with them

bartaz:
@tbille OK, got confused by the screenshot

templates/publisher/register-name.html
```diff
@@ -8,8 +8,13 @@
 <div class="p-strip">
 
   <div class="row">
-    <h1>Register a snap name</h1>
+    <h1 class="p-heading--three">Register a snap name</h1>
   </div>
+
+  <div class="row">
+    <h2 class="p-heading--four">Before you can push your snap to the store, its name must be registered</h2>
```
bartaz: `p-heading--four` seems almost the same as `--three`. Reducing it to `--five` looks better I think.

templates/publisher/register-name.html
```diff
@@ -41,27 +46,26 @@ <h1>Register a snap name</h1>
...
-      </div>
+      <hr/>
     </div>
 
     <div class="row">
-      <div class="col-4">
+      <div class="u-align--right">
+
+        <a class="p-button--neutral" href="/account/snaps">Cancel</a>
         <input type="submit" class="p-button--positive" value="Register"/>
```
bartaz: Second button has unnecessary margin top (because Vanilla). Adding `u-no-margin--top` to it fixes it.

bartaz:
I think there is a UX issue here when trying to register the name owned by someone else.

For example, go to http://snapcraft.io-pr-670.run.demo.haus/account/register-name?snap_name=firefox and try to register firefox name.

The page refreshes to 'Request reserved name', but there is no error or any other indication that something went wrong. The page looks almost exactly the same (with one additional field) which makes it look like it's the same page that just got refreshed and nothing happened.

I think we need to make it more clear (with error or info notification) that requested name is reserved and if user thinks they should own it they can submit 'Request reserved name'. It should be more visible that 'Requesting reserved name' is different and longer process then just registering new name.

Here is how it currently looks on dashboard when you try to register 'firefox':
screen shot 2018-05-18 at 14 30 06

/cc @Greggless @therealjuan

matthewpaulthomas:
> The page looks almost exactly the same (with one additional field) which makes it look like it's the same page that just got refreshed and nothing happened.

Relatedly, you can change the “Snap name” value to a name that isn’t taken. If you do that, the rest of the form still assumes that you’re requesting a reserved name, when you no longer are.

A minimal way to fix this would be for the “Request reserved name” form not to contain a “Snap name” field at all. (And since that was the only place the name was disclosed, the intro text would need changing to remind you what name you’re trying to request.)

Greggless:
Further to our discussion today @ retro, I think we can close this and then create an enhancement issues with @mpt's comments in. For that reason, I'll remove the `UX-1` label and we can merge with the `code` and `QA` are done.

Lukewh:
As decided in standup, we should merge this and track improvements in another issue... Perhaps [this one](https://github.com/canonical-websites/snapcraft.io/issues/678)?

Issue #599, opened 2018-04-26
Single snap table looks like it has extra margin at the bottom of the row


anthonydillon:
When you have a single snap the `<hr/ >` below the table looks like the bottom of the table:
![screenshot-2018-4-26 my snaps linux software in the snap store 1](https://user-images.githubusercontent.com/1413534/39303219-37eaa580-494d-11e8-8cf2-267a140ebc07.png)

Greggless:
@anthonydillon the version is a string, typically it's some form of semantic versioning with `+git298347274`, so it'll stay left aligned.

Lukewh:
Removing from the current iteration. To be addressed when Vanilla.1.7.* is applied to snapcraft.io (see https://github.com/canonical-websites/snapcraft.io/pull/681)

PR #681, opened 2018-05-22
Remove hr below snap list table if only 1 snap

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/599

If there's more than 1 snap in the table, show the hr above the Registered names table, otherwise don't show the hr.

This issue only affects those with registered names but only 1 published snap.

# Screenshot

screenshot-2018-5-22 my published snaps linux software in the snap store

### DIFF
--- a/templates/publisher/account-snaps.html
+++ b/templates/publisher/account-snaps.html
@@ -151,7 +151,9 @@ My published snaps — Linux software in the Snap Store
   {% if registered_snaps %}
     <div class="row">
       <div class="col-12">
-        <hr />
+        {% if snaps|length > 1 %}
+          <hr />
+        {% endif %}
         <h4>My registered snap names</h4>
       </div>
     </div>


bartaz:
Do we need this HR at all?

Even with more snaps on the list it kind of looks like misaligned part of the table, because it has exactly the same style as border in the table:

screen shot 2018-05-23 at 11 03 34

Lukewh:
@bartaz I forward your question on to @lyubomir-popov

lyubomir-popov:
In my opinion the issue here is lack of precise control over white space, rather than the hr itself. As an example, I've done a few quick changes to the [PR on vanilla 1.7.0](http://snapcraft.io-pr-672.run.demo.haus/account/snaps) in dev tools to better group the hr with the heading it goes with. I've also used the same heading, as the registered names section is a sibling, not a child of the Published ones above it. I've also reduced the spacing in the table and added generous white space between groups of related elements:

snapcraft io-pr-672 run demo haus_account_snaps 1 


snapcraft io-pr-672 run demo haus_account_snaps

I also think it is worth reviewing whether a one line table needs a black line under the thead, or whether tables in general should have a black border strongerthan any other border we use. WIll follow up on the last two points with @kwm14 and @spencerbygraves

Lukewh:
@lyubomir-popov @bartaz do you both think this can be merged as is (as an improvement on what we currently have) or does it need further discussion?

lyubomir-popov:
@Lukewh I think this pr does rthe wrong thing. In my opinion we should address this issue once 1.7.0 is in, when we will be able to do what the mockups above suggest. 

I'm ok to merge if we feel it currently looks unacceptable, but I don't want it to set a precedent for removing hrs indiscriminately.

Lukewh:
Closing this PR, awaiting Vanilla 1.7.*.

Thanks @lyubomir-popov

Lukewh:
FYI I've also created the following epic to keep track of vanilla migration issues: https://github.com/CanonicalLtd/snap-squad/issues/563

Issue #673, opened 2018-05-17
Unnecessary horizontal line when no screenshots


bartaz:
Go to snap details that doesn't have screenshots: https://staging.snapcraft.io/core

New updated layout (with horizontal lines) doesn't work well if there is no screenshots. There should be no line on top.

<img width="1066" alt="screen shot 2018-05-17 at 13 41 25" src="https://user-images.githubusercontent.com/83575/40175533-b20183ac-59d8-11e8-8bae-e2d5b2a517d3.png">

Lukewh:
Fixed by https://github.com/canonical-websites/snapcraft.io/pull/683

PR #683, opened 2018-05-22
Small tidy up for unnecessary hrs on snap pages

Lukewh:
# Done

- Added hr's to conditionals

# Screenshots

## Before
screenshot-2018-5-22 fabulous peesh - linux software in the snap store

## After
screenshot-2018-5-22 fabulous peesh - linux software in the snap store 1

### DIFF
--- a/templates/snap-details.html
+++ b/templates/snap-details.html
@@ -79,13 +79,13 @@
           </div>
         </div>
       </div>
-    {% endif %}
 
-    <div class="row">
-      <div class="col-12">
-        <hr />
+      <div class="row">
+        <div class="col-12">
+          <hr />
+        </div>
       </div>
-    </div>
+    {% endif %}
 
     <div class="row u-no-margin--top">
       <div class="col-8">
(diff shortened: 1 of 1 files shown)


lyubomir-popov:
Looks good, and as a general rule - hrs separate same-level groups of elements, so there should never be one one above the first or under the last.

In vanilla 1.7.0 we have a %pseudo-hr placeholder, which allows to define these as:
el + el {@extend %pseudo-hr--top;} 

This way, they can never show up in places they are not needed.

bartaz:
Fixes #673

bartaz:
@lyubomir-popov Are things like `%pseudo-hr--top` documented anywhere - or are there plans for it? It seems there are quite significant internal changes to Vanilla in terms of spacing variables and other things. Knowing about them would be very helpful for migration to 1.7.1 that we should do quite soon.

lyubomir-popov:
@bartaz I can see @Caleb-Ellis is doing a lot of work to explain the changes in the docs, and I'm writing an overview as a blog post. Might be quickest to just have a catchup or a workshop to discuss these if there is interest.

PR #687, opened 2018-05-23
Small fixes for s.io

### LINKED ISSUE #638: 502 when uploading malformed icon
### Expected behaviour

I tried to upload a screenshot instead of an icon at https://snapcraft.io/account/snaps/dosbox-x/listing and it let me upload it, but when i pressed save, it gave me a 502.

Go to listing page for a snap, click icon, find an image which isn't 256x256 (a screenshot might do) and then press save. Get 502 error.

### Specs

- _URL:_ https://snapcraft.io/account/snaps/dosbox-x/listing
- _Operating system:_ Ubuntu 16.04
- _Browser:_ Firefox

tbille:
# Summary

Fixes #638 
- Fix empty fields case if error on listinig page
- Remove tests that are calling the real API. Today API was slow this means that nothing can be merged until the API is not back in normal again...

# QA

- `./run`
- http://127.0.0.1:5000/account/snaps/tbmb-bsi-test-2/listing
- Upload an icon >256x256px
- Check if all empty fields are empty and not `None`


brunob:
I don't know if this pull was pushed on production but i still got a 502 error when uploading a 48px png as an icon on https://snapcraft.io/account/snaps/xxxx/listing

tbille:
@brunob 
Thank you for the report.
Icons should be 256x256px
You should receive an error message when you try to upload the image with the wrong size, not a 502. We will investigate on it.

PR #689, opened 2018-05-25
Limit intent GA to first click

Lukewh:
# Done

A quick fix after looking at some stats with Greg. Each time the install button is clicked we send an event, however, if the button didn't work and a user perceives this as "oh, it didn't work I need to press it again maybe?" we'll get 2 events when really there was only a single intent.

# QA

You might want to install the browser extension 'Google analytics debugger' for your browser

- Pull the branch
- `./run`
- Visit a snap details page
- Open the overlay and click the 'View in Desktop store' button
- See the GA event being fired
- Click the 'View in Desktop store' button again
- See that another GA event isn't being sent
- Repeat until your mouse breaks or you get RSI.

### DIFF
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -126,6 +126,8 @@ function initTabs(el) {
 }
 
 function initOpenSnapButtons() {
+  let attempt = 1;
+
   document.addEventListener('click', (event) => {
     const openButton = event.target.closest('.js-open-snap-button');
 
@@ -146,13 +148,26 @@ function initOpenSnapButtons() {
       document.body.appendChild(iframe);
 
       if (typeof ga !== 'undefined') {
+        // The first attempt should be counted towards the 'intent'
+        let label = 'Snap install intent';
+        let value = `${name}`;
+
+        // Subsequent attempts should still be tracked, but not as 'intent'
+        if (attempt > 1) {
+          label = 'Snap install click';
+          value += ` - click ${attempt}`;
+        }
+
         ga('gtm1.send', {
           hitType: 'event',
           eventCategory: 'Snap details',
           eventAction: 'Click view in desktop store button',
-          eventLabel: `Click view in desktop store for ${name} snap`
+          eventLabel: label,
+          eventValue: value
         });
       }
+
+      attempt += 1;
     }
   });
 }


bartaz:
@Lukewh @Greggless Don't we want to know that user clicked couple of times? Maybe we could send a different event/label for any subsequent attempts, so that we can identify 'intent' as one event, but we will still be able to see how many people are clicking multiple times (possibly due to failure).

Of course it only makes sense to do that if we find this information useful.

static/js/public/snap-details/channelMap.js
```diff
@@ -145,14 +147,18 @@ function initOpenSnapButtons() {
       iframe.src = `snap://${name}`;
       document.body.appendChild(iframe);
 
-      if (typeof ga !== 'undefined') {
+      // Only send the QA event on first click, any subsequent clicks
+      // on this page can be assumed as the same intent.
+      if (typeof ga !== 'undefined' && attempted === false) {
```
bartaz: Any reason for strong comparison with `false`? Wouldn't `!attempted` be enough?

Lukewh: not specifically I just like :muscle: things, do you think I should change it?

bartaz: I would say that `!attempted` reads better ("not attempted"). And `attempted === false` makes me think that there is some hidden reason why this needs to be `false` and nothing else. But maybe it's just me ;)

Greggless:
That's a very good point @bartaz 
A user clicking the button multiple times suggests that the button did not work, therefore we can loosely track failed `[Install]` button attempts

Lukewh:
@caldav is there value in changing the way we label these events as I believe your dashboards just takes all events as 'intent'.

Perhaps:
```
{
  hitType: 'event',
  eventCategory: 'Snap details',
  eventAction: 'Click view in desktop store button',
  eventLabel: 'Snap install intent'/ 'Snap install click',
  eventValue: 'SNAP_NAME' / 'SNAP_NAME - click COUNT'
}
```

Lukewh:
I've implemented the change I suggested above, so this is ready for a rereview.

static/js/public/snap-details/channelMap.js
```diff
@@ -146,13 +148,26 @@ function initOpenSnapButtons() {
...
 
       if (typeof ga !== 'undefined') {
+        // The first attempt should be counted towards the 'intent'
+        let label = 'Snap install intent';
+        let value = `${name}`;
+
+        // Subsequent attempts should still be tracked, but not as 'intent'
+        if (attempt > 1) {
+          label = 'Snap install click';
+          value += ` - click ${attempt}`;
```
bartaz: We don't want snap `name` in next attempts label?

Lukewh: `value +=`

bartaz: Ah, it's `+=`, so never mind...

Issue #684, opened 2018-05-23
Improve listing page contact fields


Lukewh:
- Allow users to enter email addresses without `mailto:` which can be added by the client
- It allows `mailto:@domain.com` which shouldn't be allowed
- Allow users to enter domains without `http://` or `https://`

PR #708, opened 2018-06-07
Better experience for contact inputs

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/684

As a compromise I've implemented the following:
- When a user enters a website such as 'snapcraft.io', automatically prepend 'http://' in the form field (we should perhaps default to https?)
- When a user enters something that looks like an email address 'email@email.com' (containing 1 @ and at least 1 .) prepend 'mailto:'

Also added `Array.includes` and `String.includes` polyfills

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/snap_name/listing mess around with the contact details. Even if the JS lets you enter crazy data that doesn't seem correct the API should correct it and provide an error.

### DIFF
--- a/static/js/publisher/market/market.js
+++ b/static/js/publisher/market/market.js
@@ -3,8 +3,10 @@ import { updateState, diffState } from './state';
 import { publicMetrics } from './publicMetrics';
 
 // https://gist.github.com/dperini/729294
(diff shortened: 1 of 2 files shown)


therealjuan:
This works great. 

I wonder if we should show `mailto` to users or we just prepend when sending data to the API? 

`https` might be different as some URLs might be `http` instead

Lukewh:
@therealjuan I thought that too initially, then thought it's better to be transparent and show that it's a mailto: as some urls could technically be http://luke:password@website.com

bartaz:
When value is automatically fixed (website or email) it doesn't save properly.

Example:
1. Put website without http (field is marked as valid, Save is enabled)
2. Click 'Save' without focusing anywhere else (https gets added, form is submitted)
3. Error is returned (from API) that URL doesn't have http
4. Field is invalid, edit it to revalidate (but keep it without http)
5. Save works, says that changes are saved, but there are not (URL doesn't change)

website

bartaz:
One small bug:
1. enter website URL without http
2. don't leave the field, hit Enter to submit
3. Submit is prevented, browser validation finds URL invalid (because it doesn't have http)

It's a small issue and I don't know how to fix it (apart from taking off url type to disable browser validation). So I wouldn't block merging this PR on that, but keep it as a low priority known issue.

static/js/publisher/market/market.js
```diff
@@ -320,6 +322,34 @@ function initForm(config, initialState, errors) {
...
+  function prefixInput(input) {
+    if (['website', 'contact'].includes(input.name)) {
+      if (
+        validation[input.name].isValid &&
+        input.value.length > 0 &&
+        !input.value.includes('http') &&
+        !input.value.includes('mailto')
+      ){
+        if (input.name === 'website') {
+          // Default to http as most sites redirect to https if available
```
bartaz: Commment says we default to `http` while we default to `https`

static/js/publisher/market/market.js
```diff
@@ -320,6 +322,34 @@ function initForm(config, initialState, errors) {
...
+          input.value = `https://${input.value}`;
+        } else if (input.name === 'contact' && input.value.includes('@') && !input.value.includes('mailto')) {
```
bartaz: Maybe instead of just checking that it includes '@' we could use `MAILTO_REGEXP` to make sure it's a mail?

Issue #645, opened 2018-05-10
When tracks are numbers, order in descending values


Greggless:
1\. Go to [snapcraft.io/node](https://snapcraft.io/node)
2\. Click on 'Tracks' dropdown

### Expected behaviour
See Node's tracks listed like this:
- Latest
- 10
- 9 
- 8
- 6
### Current behaviour
![image](https://user-images.githubusercontent.com/23459065/39859318-6b58a662-5431-11e8-979d-50f3b5168632.png)

matthewpaulthomas:
Most file managers sort filenames the way that we want here. For example, “blah(9)” < “blah(10)” < “blah(11)” etc.

Basically, any group of consecutive integers is treated as a number.

PR #709, opened 2018-06-07
Order tracks on channel map

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/645

- Implemented `localeCompare` rather than a standard comparison

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/node, http://0.0.0.0:8004/webstorm and http://0.0.0.0:8004/etcd (for a mix of string and number tracks)
- Open the channel map
- Look at the order in the dropdown

I can't add a screenshot because my laptop won't let me open a select input and then take a screenshot... sozzles.

### DIFF
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -69,19 +69,39 @@ function setArchitecture(arch, packageName, channelMapData) {
 
   const tracks = Object.keys(channelMapData[arch]);
 
-  // sort tracks alphabetically with 'latest' always first
-  tracks.sort((a, b) => {
-    if (a === LATEST) {
-      return -1;
+  // split tracks into strings and numbers
+  let numberTracks = [];
+  let stringTracks = [];
+  tracks.forEach(track => {
+    // numbers are defined by any string starting any of the following patterns:
+    //   just a number – 1,2,3,4,
+    //   numbers on the left in a pattern – 2018.3 , 1.1, 1.1.23 ...
+    //   or numbers on the left with strings at the end – 1.1-hotfix
+    if (isNaN(parseInt(track.substr(0, 1)))) {
+      stringTracks.push(track);
+    } else {
+      numberTracks.push(track);
     }
-    if (b === LATEST) {
-      return 1;
-    }
-    return a <= b ? -1 : 1;
   });
 
-  // by default take first track (which should be 'latest')
-  const track = tracks[0];
+  // Ignore case
+  stringTracks.sort(function (a, b) {
+    return a.toLowerCase().localeCompare(b.toLowerCase());
+  });
+
+  const latestIndex = stringTracks.indexOf(LATEST);
+  stringTracks.splice(latestIndex, 1);
+  stringTracks = [LATEST].concat(stringTracks);
+
+  // Sort numbers (that are actually strings)
+  numberTracks.sort((a, b) => {
+    return b.localeCompare(a, undefined, { numeric: true, sensitivity: 'base' });
+  });
+
+  // Join the arrays together again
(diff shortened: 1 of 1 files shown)


therealjuan:
👍

bartaz:
This now sorts standard names in reverse alphabetical order, while we wanted them to be in alphabetical order.

bartaz:
QA: when version numbers have 2 parts (1.2.3) they are not treated as numbers and are sorted separately from numbers:

screen shot 2018-06-11 at 12 31 02

bartaz:
@Lukewh @therealjuan 
Thats a UX question - we are overcomplicating sorting of tracks here to fix the fact that numeric values were out of order (because of the bug). Original issue suggested to order numbers in descending order, but names look weird in descending order so we wanted to separately order names in ascending order - but this causes issues in detecting what is number what is not and properly build the list (see bug in comment above when snap mixes numbers and non-numbers in tracks).

It would simplify both the code and sorting rules if we would just accept to sort everything in ascending order:
screen shot 2018-06-11 at 12 37 06

It would be also simple to sort everything descending (but this seems unnatural order for names):
screen shot 2018-06-11 at 12 42 07

therealjuan:
Discussion from #snappy-squad on IRC:

## General 
- Show `latest` first as it is the default selection
- Divide the sorting of tracks into `numbers` and `characters` groups with different sorting criteria
- Normally publishers decide to either have `numbers` or `characters` for their track labeling
- When we have `numbers` and `characters` within tracks on a given snap, we sort in the following order: (1) `characters` (ordered in ascending order) then (2) `numbers` (in descending order)

## Numbers

- Group of `numbers` are sorted in descending order
- `numbers` are defined by any string starting any of the following patterns: `just a number – 1,2,3,4`, `numbers on the left in a pattern – 2018.3 , 1.1, 1.1.23 ...` or `numbers on the left with strings at the end – 1.1-hotfix`

## Characters
- Group `characters` are sorted in ascending order to ease readability within the tracks dropdown

bartaz:
@Lukewh Current implementation fails to treat 3 part versions (1.2.3) as numbers. Instead of using `isNaN` as a check we should probably just check if the first character is a number (or some simple regexp to detect if string starts with semver).

static/js/public/snap-details/channelMap.js
```diff
@@ -69,31 +69,52 @@ function setArchitecture(arch, packageName, channelMapData) {
...
+  // Sort numbers (that are actually strings)
+  numberTracks.sort((a, b) => {
+    return b.localeCompare(a, undefined, { numeric: true, sensitivity: 'base' });
+  });
+
+  // Join the arrays together again
+  const sortedTracks = stringTracks.concat(numberTracks);
+
+  // by default take first track (which should be 'latest') unless it's node
+  const track = packageName === 'node' ? sortedTracks[sortedTracks.indexOf('10')] : sortedTracks[0];
```
bartaz: I know that node is a special case, but adding special treatment in different places (first backend, now channel map) will make it much more difficult to track them down later (especially that this is added in totally unrelated PR).

bartaz: The issue that `node` has 'latest' selected by default instead of '10' is separate from sorting, and maybe at least it would be good to solve it separately (for it to be found easier later). Or ideally to solve it in a different way that doesn't require hardcoding it into JS. Would be good if backend could give us a current track name, so we would default to this instead of 'latest'.

Lukewh: Forgot I added that. Did it as a little test to see if it would work as expected. Removed.

PR #714, opened 2018-06-11
Only use lightbox if on a larger screen

Lukewh:
# Done
Fixes https://github.com/canonicalltd/snap-squad/issues/581

If the screenshots previous button is not visible when attempting to open the lightbox, open in a new tab instead.
This works because on mobile we rely on the horizontal scrolling, so the previous/ next buttons aren't loaded.

# QA
- Pull the branch
- `./run`
- View a snap detail page on desktop, click an image to open in the lightbox, it should work
- View a snap detail page on mobile, click an image and it should open in another tab

### DIFF
--- a/static/js/public/snap-details/screenshots.js
+++ b/static/js/public/snap-details/screenshots.js
@@ -1,4 +1,5 @@
 import lightbox from './../../publisher/market/lightbox';
+import { isMobile } from '../../libs/mobile';
 
 export default function initScreenshots(screenshotsId) {
   const screenshotsEl = document.querySelector(screenshotsId);
@@ -13,7 +14,12 @@ export default function initScreenshots(screenshotsId) {
     const url = event.target.src;
 
     if (url) {
-      lightbox.openLightbox(url, images);
+      if (isMobile()) {
+        window.open(url, '_blank');
+        window.focus();
+      } else {
+        lightbox.openLightbox(url, images);
+      }
     }
   });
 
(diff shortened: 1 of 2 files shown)


static/js/public/snap-details/screenshots.js
```diff
@@ -13,7 +13,12 @@ export default function initScreenshots(screenshotsId) {
     const url = event.target.src;
 
     if (url) {
-      lightbox.openLightbox(url, images);
+      if (prev.offsetWidth === 0 && prev.offsetHeight === 0) {
```
bartaz: Feels like a strange way to detect 'small screen'. Why is `prev.offsetWidth` used for that?

Lukewh: It's a way to check visibility of an element, I should probably add a comment.

static/js/public/snap-details/screenshots.js
```diff
@@ -13,7 +13,14 @@ export default function initScreenshots(screenshotsId) {
     const url = event.target.src;
 
     if (url) {
-      lightbox.openLightbox(url, images);
+      // prev.offsetWidth and prev.offsetHeight are used to detect the visibility
+      // of the prev element.
+      if (prev.offsetWidth === 0 && prev.offsetHeight === 0) {
```
bartaz: I've written the code that hides prev/next buttons on mobile, but it took me quite a while to get what is going on here.

Lukewh: I'd rather find a better way to detect it via an element's visibility, rather than setting numbers - perhaps whether the mobile menu is visible?

I can definitely make it as a function so it can be used in other places too in the future.

PR #715, opened 2018-06-11
Add publisher blueprint

tbille:
# Summary

- Add publisher bluepirint
- Fix redirections with function `url_for` to avoid hardcoded urls
- Fix tests to apply with the new redirections (`url_for` adds automatically a `/` at the end of the url)

# QA

- `./run`
- Navigate in publiser pages and make sure everything works as usual
http://127.0.0.1:8004/account
http://127.0.0.1:8004/account/details
http://127.0.0.1:8004/account/snaps
http://127.0.0.1:8004/account/username
http://127.0.0.1:8004/account/agreement
http://127.0.0.1:8004/account/snaps/toto/listing
http://127.0.0.1:8004/account/snaps/toto/market
http://127.0.0.1:8004/account/snaps/toto/measure

List of the routes (by running `flask routes`
```
account_page.get_account             GET        /account/
account_page.get_account_details     GET        /account/details
account_page.get_account_name        GET        /account/username
account_page.get_account_snaps       GET        /account/snaps
account_page.get_agreement           GET        /account/agreement
account_page.get_listing_snap        GET        /account/snaps/<snap_name>/listing
account_page.get_market_snap         GET        /account/snaps/<snap_name>/market
account_page.get_measure_snap        GET        /account/snaps/<snap_name>/measure
account_page.get_register_name       GET        /account/register-name
account_page.post_account_name       POST       /account/username
account_page.post_agreement          POST       /account/agreement
account_page.post_listing_snap       POST       /account/snaps/<snap_name>/listing
account_page.post_register_name      POST       /account/register-name
account_page.publisher_snap_metrics  GET        /account/snaps/<snap_name>/metrics
```

### DIFF
--- a/webapp/publisher/views.py
+++ b/webapp/publisher/views.py
@@ -16,8 +16,8 @@ from canonicalwebteam.snapstoreapi.exceptions import (
 )
 
 
-account_page = flask.Blueprint(
-    'account_page', __name__,
+account = flask.Blueprint(
+    'account', __name__,
     template_folder='/templates', static_folder='/static')
 
 
@@ -35,10 +35,10 @@ def _handle_errors(api_error: ApiError):
         return flask.abort(504, str(api_error))
     elif type(api_error) is MissingUsername:
         return flask.redirect(
-            flask.url_for('account_page.get_account_name'))
+            flask.url_for('account.get_account_name'))
     elif type(api_error) is AgreementNotSigned:
         return flask.redirect(
-            flask.url_for('account_page.get_agreement'))
+            flask.url_for('account.get_agreement'))
     elif type(api_error) is MacaroonRefreshRequired:
         return refresh_redirect(
             flask.request.path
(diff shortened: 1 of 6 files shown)


webapp/publisher/views.py
```diff
@@ -15,6 +16,11 @@
 )
 
 
+account_page = flask.Blueprint(
```
Lukewh: Any reason not to call it account, rather then account_page?

tbille: @Lukewh the problem here is that I already have some vars with this name, I think it might be less confusing by having a specific name for the blueprint (we have the same var name for the store and the blog blueprints

Lukewh: Does it not then make more sense for the other vars to be more specific? Account is quite a general term for single vars whereas `account_page` is, I'd say, too specific for a bucket of sub-resources.

Another reason is to match the name of the blueprint with the endpoint, having a different name could cause confusion

PR #717, opened 2018-06-11
Prevent 'store' being highlighted in nav when on publisher pages

Lukewh:
# Done
- Fixed small bug in navigation

# QA
- Pull the branch
- `./run`
- Visit public and publisher pages
- Check the navigation is correctly highlighted

### DIFF
--- a/templates/_header.html
+++ b/templates/_header.html
@@ -16,8 +16,11 @@
       <ul class="p-navigation__links" role="menu">
         <li class="p-navigation__link" role="menuitem">
           <a
-            {% if (path and (path.startswith('/store') or path.startswith('/search'))) or snap_title %} class="is-selected"{% endif %}
-            href="/store">Store
+            {% if page_slug == 'store' %}
+              class="is-selected"
+            {% endif %}
+            href="/store">
+            Store
           </a>
         </li>
         <li class="p-navigation__link" role="menuitem">
@@ -30,7 +33,7 @@
         {% if user_name %}
           <li role="menuitem" class="p-navigation__link p-dropdown">
               <a
-                {% if path and path.startswith('/account') %}
+                {% if page_slug == 'account' %}
                   class="p-dropdown__toggle is-selected"
                 {% else %}
                   class="p-dropdown__toggle"
--- a/webapp/handlers.py
+++ b/webapp/handlers.py
@@ -1,6 +1,6 @@
 import flask
 import socket
-import webapp.template_functions as template_functions
+import webapp.template_utils as template_utils
 from urllib.parse import (
     unquote,
     urlparse,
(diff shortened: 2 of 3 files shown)


templates/_header.html
```diff
@@ -16,7 +16,14 @@
...
           <a
-            {% if (path and (path.startswith('/store') or path.startswith('/search'))) or snap_title %} class="is-selected"{% endif %}
+            {% if
+              path and (
+                (
+                  path.startswith('/store') or path.startswith('/search')
+                ) or (
+                  snap_title and not path.startswith('/account')
+                )
+              ) %} class="is-selected"{% endif %}
```
tbille: Works for me but it seems that there could be a much easier way of doing this

Lukewh: Yeah, this has got a bit complicated now tbh, but at least it is automatic.
I didn't really like 'page_slug' as it depended on adding the right string to the views context for each page. It seemed more error prone. Using the built-in flask stuff is more reliable then humans IMO

bartaz: Now it's a complicated logic that depends on route path name (that can potentially change) and existence of snap_title in snap details (but in theory it can be used on any other page as well - as the example shows).

bartaz: Maybe we could at least move this logic out of template to context processor, so it's automatically added to all templates? So it would be one place that would have the logic for creating something like `page_slug` (instead of doing it manually)?

Lukewh: Yeah that makes sense - generate the page_slug based on some rules that are in a specific file... It either does something special (as in this case) or just passes through the flask path.

webapp/template_functions.py
```diff
@@ -49,3 +48,19 @@ def format_number(number: int):
     with a comma between every thousands
     """
     return '{:,}'.format(number)
+
+
+def generate_slug(path):
```
bartaz: It's a bit confusing that this function is here, because it seems all the other functions from here are meant to be used in templates, while this one is only used in the preprocessor.

templates/_header.html
```diff
@@ -15,9 +15,7 @@
       </span>
       <ul class="p-navigation__links" role="menu">
         <li class="p-navigation__link" role="menuitem">
-          <a
-            {% if (path and (path.startswith('/store') or path.startswith('/search'))) or snap_title %} class="is-selected"{% endif %}
-            href="/store">Store
+          <a{% if page_slug == 'store' %} class="is-selected"{% endif %} href="/store">Store
```
bartaz: It would read a bit better if the conditional statement was multiline (as with the dropdown below).

webapp/template_functions.py
Lukewh: @bartaz because it's a var in the templates I wanted to keep all that bizniz together. @tbille what do you think?

templates/_header.html
Lukewh: Nice - done.

webapp/template_functions.py
Lukewh: @bartaz After discussion with @tbille we decided to rename the file and add comments to differentiate the different types of functions related to templates.

PR #718, opened 2018-06-11
Small fixes for blueprints

tbille:
# Summary

- Rename blog blueprint to be more readable (`blog_page` -> `blog`)
- Rename store blueprint to be more readable (`store_page` -> `store`)
- Remove useless promoted view because it wasn't linked anywhere
- use `url_for` for redirects in the store view
- Make `url_for` more readable, instead of writing `blueprint.function` -> `.function` (only if this redirect is in the blueprint)
- Remove unsued endpoint (`/status` is now `/_status` provided by talisker)

# QA

- `./run`
- Visit
- http://127.0.0.1:8004/
- http://127.0.0.1:8004/store
- http://127.0.0.1:8004/discover
- http://127.0.0.1:8004/search
- http://127.0.0.1:8004/toto
- http://127.0.0.1:8004/blog
- Those urls should work as usual


bartaz:
Everything works.

But I'm not sure if the `/blog` is in shape we want to put live:
screen shot 2018-06-11 at 17 32 10

I know it's in progress, but if it's merged it potentially can go live. And even if it's not linked anywhere it could at least be empty.

PR #724, opened 2018-06-12
Use factories to launch snapcraft

tbille:
# Summary

To help launching the brandstores I added 2 factories to build the app:

- create_snapcraft
- create_brandstore

Those 2 factories uses the same base `create_app` but they instanciate different blueprints.
To help running the tests those factories have a testing parameter so that not all the extensions added to snapcraft are run (prometheus, sentry, csrf)

The functions used in all the project (context_processor, 404) are moved into it's own file.

# QA

- `./run`
- Should have access to the usual snapcraft website
- Edit `.env` and change `FACTORY=create_snapcraft` to `FACTORY=create_brandstore`
- `./run`
- The home page should be the search page (might need to `ctrl + shift + r` (because of cached pages)

### DIFF
--- a/.env
+++ b/.env
@@ -3,4 +3,4 @@ FLASK_DEBUG=true
 SECRET_KEY=local_development_fake_key
 ENVIRONMENT=devel
 DEVEL=True
-WEBPAPP=snapcraft.io
+WEBAPP=snapcraft.io
--- a/.env
+++ b/.env
@@ -3,4 +3,4 @@ FLASK_DEBUG=true
 SECRET_KEY=local_development_fake_key
 ENVIRONMENT=devel
 DEVEL=True
-FACTORY=create_snapcraft
+WEBPAPP=snapcraft.io
--- a/.env
+++ b/.env
@@ -3,3 +3,4 @@ FLASK_DEBUG=true
 SECRET_KEY=local_development_fake_key
 ENVIRONMENT=devel
 DEVEL=True
+FACTORY=create_snapcraft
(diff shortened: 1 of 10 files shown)


.env
```diff
@@ -2,4 +2,5 @@ PORT=8004
 FLASK_DEBUG=true
 SECRET_KEY=local_development_fake_key
 ENVIRONMENT=devel
-DEVEL=True
\ No newline at end of file
+DEVEL=True
+FACTORY=create_snapcraft
```
WillMoggridge: Missing a new line!

We didn't really talk much about the new env format but I think this is too implementation specific. I think we should look at it as what makes a good readable configuration item. This currently has to accept an exact function name but what we really want is "What site should we load?"

In that sense, we could use an environment variable which is more like:
```
WEBAPP=snapcraft
```
Then the code would look for that variable and load the correct items.

Alternatively maybe it could be `BRANDSTORE=storename` and if that isn't set, it loads snapcraft by default.

.env
```diff
@@ -2,4 +2,5 @@ PORT=8004
 FLASK_DEBUG=true
 SECRET_KEY=local_development_fake_key
 ENVIRONMENT=devel
-DEVEL=True
\ No newline at end of file
+DEVEL=True
+WEBPAPP=snapcraft.io
```
WillMoggridge: s/WEBPAPP/WEBAPP/

Also... New line!

tests/blog/tests_index.py
```diff
@@ -14,10 +14,10 @@ class BlogPage(TestCase):
     render_templates = False
 
     def create_app(self):
-        my_app = app
-        my_app.testing = True
+        app = create_app(testing=True)
+        app.testing = True
```
WillMoggridge: Would setting `app.testing` inside the `create_app(testing=True)` make more sense?

PR #742, opened 2018-06-15
[DEMO] Brand store

bartaz:
Demo brand store custom header

screen shot 2018-06-15 at 15 44 04

### DIFF
--- a/.env
+++ b/.env
@@ -3,4 +3,4 @@ FLASK_DEBUG=true
 SECRET_KEY=local_development_fake_key
 ENVIRONMENT=devel
 DEVEL=True
-WEBAPP=snapcraft
+WEBAPP=brand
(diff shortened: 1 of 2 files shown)


.env
```diff
@@ -3,4 +3,4 @@ FLASK_DEBUG=true
 SECRET_KEY=local_development_fake_key
 ENVIRONMENT=devel
 DEVEL=True
-WEBAPP=snapcraft
+WEBAPP=brand
```
WillMoggridge: I think you may have been told about this already but we don't want to commit this.

You can run without modifying this with `./run --env WEBAPP=brand`, or if you aren't using the run script `WEBAPP=brand yarn run serve`

bartaz: @WillMoggridge Yes, sure - we just want to have a PR for iteration demo

WillMoggridge: Aha, OK. Great. :)

Issue #693, opened 2018-05-29
Register a new snap UX improvements


Lukewh:
From https://github.com/canonicalltd/snap-squad/issues/536

# Items to discuss

- [ ] There is no way to access from Snaps list to **register a new name**

- [ ] When a user tries to register a Snap that they own, we display a message acknowledging ownership from the publisher. Perhaps we can suggest the user the ability to visit the details of that specific Snap

![image](https://user-images.githubusercontent.com/8167677/40644129-29704514-631a-11e8-9f76-7024c01b7e6e.png)

- [ ] Clearer message if a user tries to register a Snap they don't own, so they can decide to request reserved name. 

Can we display a dialog with... ?
**Title:** <%SNAP NAME%> is already taken
**Message:** Another publisher already registered <%SNAP NAME%>. You can file a dispute to request transfer ownership of the name or register a new name
**Actions:** Request reserved name | Register a new name

- [ ] Provide better explanation on what **public** and **private** means under the label 
- [ ] Be able to go back to **Snaps list**
- [ ] Confirmation message after registering a new name about what you can do

Lukewh:
Relevant: https://bugs.launchpad.net/snapstore/+bug/1771767

PR #743, opened 2018-06-15
Better registration

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/693

- Removed supporting message box and link to dashboard.snapcraft.io form instead
- Updated messaging at the top of the page
- Added flash message to top of snap list page when name successfully registered

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/account/
- Click the 'Register name' at the top right of the snap list
- Try registering a name that you already have access to, make sure a message is displayed
- Try and register a name that is registered to someone else 'toto' for example, make sure a message is displayed with a link to the dashboard form
- Register a new name, that is not taken (prefixed with `test-`), make sure a message is displayed when you return to the snap list

### DIFF
--- a/templates/publisher/account-snaps.html
+++ b/templates/publisher/account-snaps.html
@@ -7,11 +7,26 @@ My published snaps — Linux software in the Snap Store
 {% block content %}
 {% include "publisher/_beta-notification.html" %}
 <section class="p-strip is-shallow">
+  {% with messages = get_flashed_messages() %}
+    {% if messages %}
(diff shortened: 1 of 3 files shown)


therealjuan:
A few comments:
- Might be good to link to "How to upload a Snap?" after registration completion on the snackbar
- In case a publisher tries to register a snap they already own, can we add a link to the store listing page for that given snap?

PR #773, opened 2018-06-19
Add store conf to have specific ubuntustore

tbille:
# Summary

Fixes https://github.com/CanonicalLtd/snap-squad/issues/588
Add possiblity to store to add a header in the API calls with `X-Ubuntu-Store`.

To do this I had to refactor the code so that blueprints have a constructor: `store_blueprint()` so that we can store parameters (for example `store_query` which is the store to query)

Same thing has been done for the API module which has been transformed into a class so that the header is built once.

# QA

- `./run`
- Should have snapcraft as usual
- In the folder `webapp/configs`, add the file `lime.py`:
```python
WEBAPP_CONFIG = {
    'STORE_QUERY': 'LimeNET',
    'LAYOUT': '_layout-brandstore.html',
    'STORE_NAME': 'LimeNET'
}
```
- ~In `.env` file change `WEBAPP`: `WEBAPP=lime`~
- ~`./run`~
- `./run --env WEBAPP=lime`
- http://0.0.0.0:8004
- Search for lime, you should have a list of snaps
- Search for toto, no snap found

### DIFF
--- a/webapp/config.py
+++ b/webapp/config.py
@@ -24,3 +24,4 @@ SENTRY_CONFIG = {
 }
 
 WEBAPP = os.getenv('WEBAPP', 'snapcraft')
+WEBAPP_EXTRA_HEADERS = {}
--- a/.env
+++ b/.env
@@ -3,5 +3,4 @@ FLASK_DEBUG=true
 SECRET_KEY=local_development_fake_key
 ENVIRONMENT=devel
 DEVEL=True
-WEBAPP=snapcraft
 BLOG_ENABLED=true
--- a/webapp/api/public.py
+++ b/webapp/api/public.py
@@ -55,9 +55,9 @@ PROMOTED_QUERY_URL = ''.join([
 class StoreApi:
     headers = {'X-Ubuntu-Series': '16'}
 
-    def __init__(self, store_query=None):
-        if store_query:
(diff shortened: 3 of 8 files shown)


Lukewh:
Looks good but maybe get @WillMoggridge to take a look too

webapp/config.py
```diff
@@ -24,3 +24,4 @@
 }
 
 WEBAPP = os.getenv('WEBAPP', 'snapcraft')
+WEBAPP_EXTRA_HEADER = {}
```
WillMoggridge: Can this be made plural please? As the content is not a single item. `WEBAPP_EXTRA_HEADERS`

WillMoggridge:
@tbille You don't need to modify the .env file, that is the beauty of environment variables!

You can run: `./run --env WEBAPP=lime` or `WEBAPP=lime yarn run serve`

WillMoggridge:
I don't think you pass in the required headers anymore, the searches show the normal results.


I think I may have misunderstood the requirement and use of the header. I thought we were talking about misc debug headers like `COMMIT_ID`.
If it is a header that is part of the API specification, it makes sense to set it with an init argument such as:
```python3
api = StoreApi(store='lime')
```

This PR can continue forward. That could be discussed and updated in a follow up PR.

Issue #669, opened 2018-05-17
Put popularity charts checkboxes in one column


bartaz:
From @lyubomir-popov https://github.com/canonical-websites/snapcraft.io/pull/664#issuecomment-389795162

On listing page popularity charts checkboxes should be in one column

lyubomir-popov:
I think it would be better to tighten the vertical spacing between them, rather than using multiple columns. There's a lot of research against mutliptl columns ([one random article](https://baymard.com/blog/avoid-multi-column-forms)). So maybe tight like this?

<img width="310" alt="screen shot 2018-05-18 at 11 47 10" src="https://user-images.githubusercontent.com/2741678/40230981-52b09762-5a91-11e8-9ff4-e7de5495689c.png">

lyubomir-popov:
Agreed to address this once the site is on vanilla 1.7.*

Lukewh:
@bartaz made a valid comment regarding single column here: https://github.com/canonical-websites/snapcraft.io/pull/776#issuecomment-398734616

Any thoughts?

lyubomir-popov:
@Lukewh Replied there.

PR #776, opened 2018-06-20
Single column checkboxes

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/669

- Made chcekboxes a single column

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/SNAP_NAME/listing
- Ignore all the other style issues, scroll down to the bottom of the page
- The metrics checkboxes should be nicely aligned

### DIFF
--- a/templates/publisher/listing.html
+++ b/templates/publisher/listing.html
@@ -316,21 +316,9 @@
                   checked="checked"
                 {% endif %}
                 />
-              <label class="u-no-margin--top p-tooltip p-tooltip--top-left" for="public-metrics-enabled">Display public popularity charts
+              <label class="p-tooltip p-tooltip--top-left" for="public-metrics-enabled">Display public popularity charts
                 <span class="p-tooltip__message">These display relative popularity, not exact numbers.</span>
               </label>
-              <input
-                type="hidden"
-                name="public_metrics_blacklist"
-                {% if public_metrics_blacklist|length > 0 %}
-                  value="{{ join(public_metrics_blacklist, ',')}}"
-                {% endif %}
-                />
-            </div>
-          </div>
-
-          <div class="row">
-            <div class="col-6">
               <input
                 type="checkbox"
                 name="public_metrics_territories"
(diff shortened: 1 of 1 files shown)


templates/publisher/listing.html
```diff
@@ -316,21 +316,9 @@
                   checked="checked"
                 {% endif %}
                 />
-              <label class="u-no-margin--top p-tooltip p-tooltip--top-left" for="public-metrics-enabled">Display public popularity charts
-                <span class="p-tooltip__message">These display relative popularity, not exact numbers.</span>
+              <label class="p-tooltip p-tooltip--top-left" for="public-metrics-enabled">Display public popularity charts
+                <span class="p-tooltip__message u-sv-1">These display relative popularity, not exact numbers.</span>
```
bartaz: This seems to be a bug with Vanilla tooltip (that there is more padding in the bottom). Or maybe it's intended. I wouldn't fix that with `u-sv-1`, as it's a hack.

bartaz:
Maybe it would be worth indenting lower checkboxes a bit to show that they depend on the top one?

Now it looks like you have 3 options to choose from:

screen shot 2018-06-20 at 14 40 28

templates/publisher/listing.html
Lukewh: Agreed, I've removed this and the spacing utility and created https://github.com/vanilla-framework/vanilla-framework/issues/1882

lyubomir-popov:
@bartaz How about keeping the rest of the checkboxes in a disabled state until you tick the first one?

bartaz:
@lyubomir-popov It works this way already:
screen shot 2018-06-22 at 10 44 39

So it's only confusing when first one is checked so they are all enabled then.

But I guess we can merge it as it is for now and update as small task later when we have decision how to do it.

lyubomir-popov:
@bartaz ah, I see. I think you're right, they should be indented in that case; Could you indent them so that the left edge of the nested checkboxes lines up with the first letter of the parent checkbox label?

Lukewh:
Added to https://github.com/canonical-websites/snapcraft.io/pull/793

PR #786, opened 2018-06-20
Update to latest run with unset array fix

WillMoggridge:
This is the latest run script from https://github.com/canonical-webteam/generator-canonical-webteam/pull/74
This PR can be reviewed alongside the upstream one.

## QA

Make sure a Mac user reviews this!

`./run` - Site should function as normal


therealjuan:
Just tried and it worked perfectly on my machine 👍

tbille:
@WillMoggridge I also reverted the update you did in the `Dockerfile` to keep the run script in the same environment as the docker image. Could you update it in this PR?

PR #792, opened 2018-06-21
Fixes selected navigation styles (on mobile)

bartaz:
We had some custom navigation styling that broke (for mobile) after Vanilla 1.7 update. This PR removes unnecessary styling to fix the issue.

### QA

- go to store page
- make sure Store is selected in navigation
- make screen small (so navigation turns into dropdown menu)
- make sure Store is still correctly selected

screen shot 2018-06-21 at 13 52 40

### DIFF
--- a/static/sass/_snapcraft_p-navigation.scss
+++ b/static/sass/_snapcraft_p-navigation.scss
@@ -4,34 +4,11 @@
 
   .p-navigation {
     z-index: 100; // appear over channel map
+
     // custom styles for dark header with hover
-    @media (min-width: $breakpoint-navigation-threshold + 1) {
-      .p-navigation__link {
-        & > a:hover,
-        .p-contextual-menu__toggle:hover,
-        .is-selected {
-          background: $navigation-hover-color;
-        }
-      }
-
-      .p-navigation__links,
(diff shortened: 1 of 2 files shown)


Lukewh:
Demo seems a bit broken for me:
screenshot-2018-6-21 snapcraft - snaps are universal linux packages

bartaz:
@Lukewh good catch, it seems to happen in Firefox. Will check if that's us or Vanilla bug.

bartaz:
@Lukewh @Caleb-Ellis This seems to be a Vanilla issue, not related to this PR (we have this bug anyway). https://github.com/vanilla-framework/vanilla-framework/issues/1892

Unfortunately quite a serious one, as we have a broken navigation (unless we stop using sticky footer).

Caleb-Ellis:
@bartaz Ahhh I was wondering what was causing https://github.com/canonical-websites/vanillaframework.io/issues/136. After playing around with it a but a quick fix is:
``` scss
.p-navigation__banner {
  flex-basis: auto;
}
```

@lyubomir-popov Do you remember why mobile nav has flex-basis 100%? Would this break anything?

bartaz:
@Lukewh Updated with workaround from Caleb. Nav should work now on Firefox.

PR #793, opened 2018-06-21
Snap listing

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/761

- Updated markup, moved things around
- Used `p-strips` for spacing between sections
- Removed `u-margin*` and `u-padding*` classes
- Added some `u-margin*` and `u-padding*` classes

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/SNAP_NAME/listing
- Use the form and ensure everything works and looks as expected.


static/sass/_snapcraft_p-tabs.scss
```diff
@@ -8,6 +8,21 @@
     &__list {
       padding-right: 0;
       width: auto;
+      margin-bottom: -1px !important;
```
bartaz: Seems that `!important` is needed only because we add `u-no-margin--bottom`. No need to have `u-no-margin` if we set margin ourselves.

static/sass/_snapcraft_p-tabs.scss
```diff
@@ -8,6 +8,21 @@
...
+    }
+
+    & .is-bordered {
```
bartaz: This seems too hacky to me. We are using Vanilla built in class name `is-bordered`, but we disable what it originally does to add our own behaviour. Maybe we just should use custom style.

static/sass/_snapcraft_p-tabs.scss
bartaz: I'm also worried that we are changing built in vanilla classes with custom styles here (`p-tabs`, `p-tabs__list`) that will be affected in other places as well - like on channel map (where we probably don't want `margin-bottom: -1px`.

templates/publisher/listing.html
```diff
@@ -5,218 +5,220 @@
...
+              {% if field_errors and field_errors['summary'] %}
+                <p class="p-form-validation__message">
+                  <strong>Error:</strong> {{ field_errors['summary'] }}
+                </p>
+              {% endif %}
             </div>
           </div>
-          <div class="row">
-            <div class="col-2">
-              <label class="p-label--grid-baseline">Description: </label>
```
bartaz: If `p-label--grid-baseline` is not needed anymore it should be removed from CSS as well.

static/sass/_snapcraft_p-tabs.scss
Lukewh: Yeah, I agree. There's a few bits coming up that use tabs. I've changed it to be `.p-tabs-bordered` and added an extra class for `tab_group`

templates/publisher/listing.html
Lukewh: Ah, thought it was part of vanilla, didn't realise we had an override. Removed!

static/sass/_snapcraft_p-tabs.scss
```diff
@@ -1,11 +1,21 @@
 @mixin snapcraft-p-tabs {
   @include vf-p-tabs;
 
-  .p-tabs {
+  .p-tabs-bordered {
     @extend .p-tabs; // sass-lint:disable-line placeholder-in-extend
     box-shadow: none;
```
bartaz: I think that Vanilla tabs don't have shadow anymore, so we don't need this anymore.

Lukewh: You mean in new designs? Will ask Lyubo when he's around

Lukewh: Removed! Didn't understand what you meant, was looking at the `-bordered` addition highlighted in green, not the line that you added the comment to. Silly me

bartaz:
@Lukewh What happened to this branch? There seem to be some unrelated commits here

Lukewh:
@bartaz as mentioned on IRC "I accidentaly rebased https://github.com/canonical-websites/snapcraft.io/pull/793 on master, before the vanilla branch... but nothing seems to have broken"

Lukewh:
Moved to https://github.com/canonical-websites/snapcraft.io/pull/797

PR #811, opened 2018-06-25
Update API module with caching on public pages

WillMoggridge:
Update the cache module to a more generic requests module (open to other naming ideas).

I modified this code to extend the requests session object with some custom code to manage our custom API exceptions. It includes a factory to set the session up and I left the `get` function available, which uses the new session in the background.

I updated the Public API modules to use the cached session. In turn I simplified them taking advantage of the Requests session API. Setting default headers and removing the duplicate code.

## QA

To disable internet to test, you can turn your internet off or you can modify the `webapp/api/public.py` file. If you change the URL on line 11 to `'https://google.com:81',` all the API traffic will time out.

 - `./run test` - All should be OK (As it is in the CI test)
 - `./run` and try the site in general, it should all work well.
 - Go to http://localhost:8004/search?q=toto and see some results.
 - Disable the internet and still see the results
 - Restart with `./run --env WEBAPP=lime`
 - Go to the same URL and there should be no results.
 - Turn the internet back on and still no results but searching for "lime" returns results.


webapp/api/blog.py
```diff
@@ -1,9 +1,12 @@
-from webapp.api import cache
+from webapp.api import requests as api_requests
```
nottrobin: Couldn't you do `from webapp import api` and then refer to `api.requests`? This would achieve the unique naming you're looking for without fundamentally renaming the function.

webapp/api/public.py
```diff
@@ -1,5 +1,5 @@
 import os
-from webapp.api import cache
+from webapp.api import requests as api_requests
```
nottrobin: Same point as before, why not:

``` python
from webapp import api
# ...
self.session = api.requests.CachedSession()
```

webapp/api/dashboard.py
```diff
@@ -107,7 +107,7 @@ def get_authorization_header(session):
 def get_account(session):
     headers = get_authorization_header(session)
 
-    response = cache.get(
+    response = requests.get(
```
nottrobin: Why use a session in `blog.py` but not here?

webapp/api/requests.py
```diff
@@ -0,0 +1,138 @@
...
+        latency_histogram.labels(
+            domain=domain, code=request.status_code
+        ).observe(
+            request.elapsed.total_seconds()
+        )
+
+        return request
+
+
+class Session(BaseSession, _requests.Session):
```
nottrobin: Since this class isn't doing anything, why not just rename `class BaseSession()` to `class Session(_requests.Session)`?

webapp/api/requests.py
```diff
@@ -0,0 +1,138 @@
...
+    'feed_latency_seconds',
+    'Feed requests retrieved',
+    ['domain', 'code'],
+    buckets=[0.25, 0.5, 0.75, 1, 2, 3],
+)
+
+
+class TimeoutHTTPAdapter(_requests.adapters.HTTPAdapter):
+    def __init__(self, timeout=None, *args, **kwargs):
+        self.timeout = timeout
```
nottrobin: This seems insanely complex, but I agree there doesn't seem to be another way. *sigh*

webapp/api/requests.py
```diff
@@ -0,0 +1,138 @@
...
+class Session(BaseSession, _requests.Session):
+    pass
+
+
+class CachedSession(BaseSession, requests_cache.CachedSession):
+    def __init__(self, *args, **kwargs):
+        # Set cache defaults
+        options = {
+            'backend': 'sqlite',
+            'expire_after': 5,
```
nottrobin: The QA steps didn't work for me because the cache expires after 5 seconds - so once I'd turned off my internet and reloaded the pages, the cache had already expired.

I increased it to `60`, and then I saw cached pages.

I think this logic is correct, I'm just pointing out that the QA steps didn't work as written.

tbille:
could you add a test in `tests/endpoint_testing.py` like you did in `tests_publisher.py`  just to make sure that all the endpoint are not cached when called?

webapp/api/requests.py
WillMoggridge: The problem is that we want to use our base class on top of one of two different classes.

So the BaseSession is used on top of Requests or RequestsCache sessions.

WillMoggridge:
OK, I updated the naming importing set up as it is nicer to read I think. Though to import the api like that, I have to create an `__init__.py` which pulls the module in.

I also removed the backwards compatible function. It was causing me problems before but this attempt worked pretty well.

tests/tests_requests.py
```diff
@@ -0,0 +1,31 @@
+import responses
+import unittest
+from requests.exceptions import ConnectionError, Timeout
+
+from webapp.api import requests
```
nottrobin: `from webapp import api` ?

tests/tests_requests_cache.py
```diff
@@ -0,0 +1,65 @@
+import requests_cache
+import responses
+import unittest
+
+from webapp.api import requests
```
nottrobin: `from webapp import api`?

webapp/api/__init__.py
```diff
@@ -0,0 +1 @@
+from . import requests # noqa
```
nottrobin: Do you really need this? Wouldn't simply the existence of the `__init__.py` be enough?

webapp/api/requests.py
nottrobin: Oh fair enough.

nottrobin:
I'm now having trouble browsing to http://0.0.0.0:8004/search?q=toto with the internet off:

image

WillMoggridge:
I rebased and I added a commit to handle the unbound variable. It looks like an existing problem which hasn't come up much

PR #822, opened 2018-06-27
Add releases page with table of revisions

bartaz:
Fixes https://github.com/CanonicalLtd/snap-squad/issues/599

- Adds releases page to publisher snap pages, with (react-rendered) table of revisions.
- Removes dependency on vanilla-framework-react (as there were couple of issues when using it (https://github.com/vanilla-framework/vanilla-framework-react/issues/132, https://github.com/vanilla-framework/vanilla-framework-react/issues/133)

### QA

- ./run or http://snapcraft.io-pr-822.run.demo.haus
- Sign in
- Go to any snap
- "Releases" should be in tabs navigation
- go to Releases page, list of revisions should render
screen shot 2018-06-27 at 16 26 19


static/js/publisher/release/revisionsList.js
```diff
@@ -0,0 +1,46 @@
...
+          <td>{ uploadDate.format("YYYY-MM-DD HH:mm") }</td>
+          <td className="u-align--right">{ uploadDate.fromNow() }</td>
+        </tr>
+      );
+    });
+  }
+
+  render() {
+    return (
+      <table>
```
Lukewh: We should use the vanilla-react sortable table here, I think

therealjuan:
What about...

- Include the precise date and time of submission as a tooltip within the `x hours ago` column
- Include filesize in the table
- Increase version width (a tiny bit) to allow bigger version names as in node, spotify and others
- Rename `release` to `revisions` as right now we only display revisions including released and non-released ones
- Rename `release history` to `revision history`

static/js/publisher/release/revisionsList.js
bartaz: I tried to use Table from vanilla-framework-react and found 2 issues that make it really hard to use, so went with simple approach instead. The only sorting we need for now is by revision number I guess, so should be fine like this for a while.

bartaz:
@therealjuan Fixed all apart from adding filesize (which this API doesn't return).

PR #824, opened 2018-06-28
Updates empty search results to show featured snaps

### LINKED ISSUE #768: No results text whilst searching long names looks strange
### What is happening

When adding a term with quotes ("") to search on the Snapcraft.io store we add the name at the end and using long names can look a little bit strange

image

### How can we solve this?

Change no results text to a more generic text like 'No snaps found for your search' and perhaps provide showcase featured snaps to avoid having an empty page...

browse copy

bartaz:
Fixes #768 

### QA

- ./run
- go to Store (or search) and search for something that doesn't exist
- search term should not be repeated in heading
- if there are no results featured snaps should be shown

screen shot 2018-06-28 at 13 19 19


webapp/public/views.py
```diff
@@ -108,13 +110,18 @@ def search_snap():
                     else []
                 )
             )
+
+            featured_snaps = logic.get_searched_snaps(
```
Lukewh: We should only get featured snaps if `snap_results` is empty.

Lukewh:
There's also different spacing and heading size between the /store page (first screenshot) and empty search results (second screenshot)

screenshot-2018-6-28 install linux apps in seconds snap store
screenshot-2018-6-28 snap search results for

caldav:
Fails when the query starts with `:`

therealjuan:
👍

Lukewh:
@caldav I've created this to track that issue - I imagine it'll need some @tbille input (but he's away until Monday)!

Issue #733, opened 2018-06-14
View snap from different categories on snapcraft.io


therealjuan-zz:
We want to allow visitors to be able to browse all snaps available from different categories of the Snapcraft.io store.

## What can we do?
- Change the title of **Store** to **Browse**
- Include the ability to specify which categories to search snaps
- Display all categories in a carousel style on the **Browse** homepage
- The carousel will include up to 24 snaps per category and the ability to view all the remaining snaps
- Visiting a snap category page will display all snaps under that category to show the breadth of the store
- Search results will be displayed underneath and we should have a loading state whilst searching...

![store](https://user-images.githubusercontent.com/8167677/41404891-4e372cf2-6fc0-11e8-919d-c465034af152.png)

**Note:** We discussed the list of categories during the GNOME Software sprint and it might take some time to get all planned but we can start implementing how categories will surface on snapcraft.io

PR #828, opened 2018-06-28
Show categories on the store page

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/733

- Added extra call on the backend to get the list of categories from the API
- Modified the search API to include the category
- Updated templates to show the list of categories
- Added JS to fetch snaps for the listed categories
- Added JS `Carousel` class (generic so we can replace the screenshots one with it) for scrolling through category snaps
- Added extra sass component

# QA

- Pull the branch
- `./run`
- Make sure the categories load
- Click a category heading
- Make sure you get the list of snaps in that category (search results page, for now)
- Paginate through
- Use the search input to search for a snap name that exists
- Results should show, as expected
- Paginate through
- Search for a snap that doesn't exist
- Error message should show followed by the categories
- Clear the text search and select a category from the dropdown
- Click search
- Category should show
- Search for the name of a snap in that category
- The snap should show
- Search for a snap that doesn't exist
- Error message should display with the carousel for the selected category

# Screenshots
## Store page
screenshot-2018-6-28 install linux apps in seconds snap store 1
## Category page
screenshot-2018-6-28 games linux software in the snap store
## Search results (All snaps)
screenshot-2018-6-28 snap search results for 1
## Search results (Games)
screenshot-2018-6-28 snap search results for 2
## No results (All snaps)
screenshot-2018-6-28 snap search results for 3
## No results (Games)
screenshot-2018-6-28 snap search results for 4


bartaz:
Couple of visual issues:

### Carousel buttons feel misaligned (they seem to be lower then they should be):
screen shot 2018-06-29 at 09 54 48

I guess it's because carousel has spacing below snaps, but not above it - so when button is centered is seems off:
screen shot 2018-06-29 at 09 57 19

### Page jumps a lot when loading, showing weirdly aligned spinners

categories

### On mobile search inputs and button have no spacing

screen shot 2018-06-29 at 09 54 10

bartaz:
Browsing through categories feels weird.

- I go to the Store and click on one of the categories titles (say "Games")
- I'm now on 'Games' page that is really an empty search in Games category
- I want to change the category, I see that there is a dropdown with categories, I click it and change to Featured
- Nothing happens
- I have to click 'Search' (without entering any search) to get to Featured category page

There should be a better way to navigate between categories. It could be a breadcrumb or this search dropdown should reload given category on input change (not when empty search is clicked).

bartaz:
When showing category snaps when search returns no results it should be more clear that these are not search results, because now it seems like we say 'we couldn't find anything', but we show something anyway:

screen shot 2018-06-29 at 10 10 16

Maybe the heading above snaps instead of saying just 'Productivity' should say something like 'Browse other snaps in Productivity category'

static/js/public/carousel.js
```diff
@@ -0,0 +1,130 @@
...
+
+    this.scroller = this.holder.querySelector('.p-carousel__scroller');
+    this.items = this.holder.querySelectorAll('.p-carousel__item');
+    this.nextButton = this.holder.querySelector('.p-carousel__next');
+    this.prevButton = this.holder.querySelector('.p-carousel__prev');
+
+    this.currentIndex = 0;
+    this.itemWidth = this.items[0].scrollWidth;
+    this.scrollerWidth = (this.itemWidth) * this.items.length;
+    this.visibleItems = Math.ceil(this.holder.scrollWidth / this.itemWidth);
```
bartaz: Calculating this only when Carousel is created may lead to some cases when it will start acting weird after screen is resized (when number of visible items changes).

Not sure if we need to listen on the resize event. Quite possibly it would be enough to recalculate `visibleItems` just before they are used.

static/js/public/store-categories.js
```diff
@@ -0,0 +1,47 @@
...
+function getCategory(holder) {
+  const category = holder.dataset.category;
+
+  // Write the html and init the carousel
+  const writeCategory = function(response) {
+    holder.innerHTML = response;
+    new Carousel(holder.querySelector('.p-carousel'));
+  };
+
+  const url = `/store/categories/${category}`;
```
bartaz: This looks like an URL to a webpage, not internal API. If we start having APIs, we should consider having `/api` URL namespace for it or something.

static/sass/_patterns_carousel.scss
```diff
@@ -0,0 +1,95 @@
+@mixin snapcraft-p-carousel {
```
bartaz: This looks a lot like `_snapcraft_details_screenshots.scss`. Once this lands it would be good to redo the screenshots using the carousel.

static/sass/_patterns_matrix-object.scss
```diff
@@ -0,0 +1,57 @@
+// Based on the matrix component https://github.com/vanilla-framework/vanilla-framework/blob/develop/scss/_patterns_matrix.scss
+@mixin snapcraft-p-matrix-object {
+  $matrix-img-width: map-get($icon-sizes, thumb--small);
+  $matrix-img-gutter: $sph-inter;
+
+  .p-matrix-object {
```
bartaz: It's a bit confusing that it has `matrix` in the name while not being used in matrix pattern at all.

I think we should consider it as a variant of media object pattern and call it accordingly, something like `media-object--stacked` or something? Or just `media-object--snap` if it only fits into our specific use case.

templates/search.html
```diff
@@ -3,98 +3,117 @@
...
-                    {% endif %}
-                  </span>
-                  <span class="p-media-object__details">
-                    <h4>{{ snap.title }}</h4>
-                  </span>
-                </a>
-              {% endfor %}
-            {% endif %}
+          <div data-category="{{ category.slug }}" class="row js-store-category">
+            {# _category-partial.html loaded via JS #}
```
bartaz: It's just one category to show here. We should be able to render it in HTML here easily and not involve JS (apart from running the carousel).

lyubomir-popov:
@Lukewh @bartaz 

For the slider position mentioned above, how about using the spacing applied between elements to correct for the margin-bottom of the content? like ```top: calc(50% - #{$spv-inter--scaleable});```

• In my opinion, in this case it would be better to make the slider arrows visible at all times - otherwise you may not understand there's more
• To get an idea of how much more, maybe add pagination dots underneath?
• The slider slides one snap at a time, it should slide the number of items that are visible
• Once they slide in batches, easing will probably be needed

In my opinion, the round buttons borrowed from the screenshots slider don't work well - they overlap the icons too much:
screen shot 2018-06-29 at 10 31 51

Since this is not a vanilla component yet, I'll make a design and talk to @spencerbygraves next week to get signoff.

• The matrix component in the vanilla example contains paragraph copy, and the spacing is optimised for that use. In this case:

(1) With a tiny label underneath, the spacing between the blue link and the publisher grey name is too much. I would increase the negative margin on .p-matrix--desc to -1.5rem; (Ideally we should provide for this in vanilla, maybe have a .p-matrix--label class which can substitute the .p-matrix--desc to indicate you want tighter spacing. I'll make an issue for this in vanilla.)

(2) For the image (again, since this use of the matrix is quite different than the one in the [example](https://vanilla-framework.github.io/vanilla-framework/examples/patterns/matrix/)), I would add margin-bottom: $spv-inter--scaleable.

(3) And finally since the slider is a section by itself, I would apply $spv-inter--scaleable or wrap it in a shallow strip:

screen shot 2018-06-29 at 10 57 52

therealjuan:
Few comments:

## Suggestions
* To include a breadcrumb to `categories` when you are on a category page so you can quickly go back above the category title
* On the carousel of snaps in the all categories page we can increase the image size of each Snap to be similar to the ones on search results
* Probably we need new carousel arrows as the screenshot ones are rounded and borrow too much icon image
* As @bartaz is suggesting we might want to include a headline when you search for a specific name within a category that suggests: `Other productivity snaps`

## Questions:
* Do we need pagination at all or maybe we can have a higher limit of 25, 50 or even a 100? The pagination visually looks off and floating in space but not sure if that is a vanilla comment

Greggless:
We want the page to feel more 'alive' with snaps, this build shows a lot of white space (on the category pages). Could we adjust and use a larger icon size (as on the search results page) and reduce spacing between the 'tiles' in anyway?

A common pattern used to show that there are more 'things' in a list is to have one more item 'peeking' out from the edge of the grid.

Another point is the larger type size for the snap title, this increase the chance that a title will need to wrap or truncate. Can we use an alternative treatment to avoid this?

I think Microsoft do a fair job of solving all three issues I outlined above. take a little look at this screenhshot

image

static/sass/_patterns_carousel.scss
Lukewh: that's what I was thinking - generalise the code and styles as it seems carousels are the future of snapcraft....

static/js/public/carousel.js
Lukewh: Done

static/sass/_search-form.scss
```diff
@@ -4,6 +4,10 @@
 
     .p-form__group {
       flex-grow: 1;
+
+      &:first-child {
```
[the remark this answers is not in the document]
Lukewh: Done

static/js/public/store-categories.js
Lukewh: I agree. Originally I did it as an 'API' like endpoint, but because I'm passing HTML through I didn't want to label it as 'API'. We will definitely want to make it a better API in the future

static/js/public/store-categories.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // Use fetch ¯\_(ツ)_/¯
+  return fetch(url).then(response => {
```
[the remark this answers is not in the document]
Lukewh: Added the github `whatwg-fetch` pollyfill ([from here](https://github.com/whatwg/fetch)).

static/sass/_patterns_matrix-object.scss
Lukewh: dun

templates/search.html
Lukewh: For this first version I just wanted to reuse what I had already built, without needing to add extra calls to the search endpoint. If we decide this doesn't work very well long term, we can definitely implement it a better way.

Lukewh:
@therealjuan @Greggless regarding the design, I'll wait for @lyubomir-popov's designs. In the meantime is there anything stopping this being merged from a usability perspective?

lyubomir-popov:
@therealjuan @Lukewh I think the dots are due to collapsed height of the inline anchor with class "p-media-object--snap". It needs display: block to wrap its contents.

therealjuan:
All good but maybe we can leave pagination for now until we know the number of total snaps. Sorry for the back and forth.

Everything else is good to go! 👍

webapp/public/logic.py
```diff
@@ -76,17 +76,28 @@ def convert_navigation_url(url, link):
...
+    else:
+        category = ''
+
     size = int(url_queries['size'][0])
     page = int(url_queries['page'][0])
 
     return host_url.format(
         base_url=url,
         q=q,
+        category=category,
```
tbille: I think you should add the category after the url is built, otherwise it makes an ugly url if you are not searching for particular categories: http://127.0.0.1:5000/search?category=&q=toto

static/js/public/store-categories.js
```diff
@@ -0,0 +1,43 @@
...
+    if (!response.ok) {
+      throw `Response not ok for ${category} category \`${url}\``;
```
tbille: I don't know what is best in case an api call doesn't work, diplay a message or not?

**Example all api fail**
image

Lukewh: I specifically made it like this so that it would just be an 'ok' list if the API fails - interested in a UX perspective though

therealjuan: Probably makes sense to have this list still and work out a little bit more on the edge cases as we go along, as this is a first iteration pass of the store categories feature

lyubomir-popov:
Sorry for butting in in the last possible moment, and feel free to ignore all of the below - I think it looks ok already. 

I liked @Greggless's example from microsoft, especially reducing the snap name to ellipsize less of it.
modifications in this image:
• featured category names h3 > h4
• spacing under icon - .5rem instead of 1.5rem
• p-media-object--snap__title extends h5 instead of h4

image

webapp/public/views.py
```diff
@@ -57,69 +57,102 @@ def discover():
...
 
-            if not snaps_results:
-                featured_snaps = logic.get_searched_snaps(
-                    api.get_featured_snaps()
-                )
-        except ApiError as api_error:
-            status_code, error_info = _handle_errors(api_error)
+        if not snaps_results:
+            featured_snaps = logic.get_searched_snaps(
+                api.get_featured_snaps()
```
tbille: this was in a `try ... except` can you surround it again?

webapp/public/views.py
```diff
@@ -263,4 +301,33 @@ def snap_details(snap_name):
...
+        except ApiError as api_error:
+            status_code, error_info = _handle_errors(api_error)
+
+        snaps_results = logic.get_searched_snaps(category_results)
+
+        context = {
+            "category": category,
+            "snaps": snaps_results,
+            "error_info": error_info
+
```
tbille: remove line

webapp/public/logic.py
```diff
@@ -181,3 +196,26 @@ def get_default_channel(snap_name):
...
 
     return 'stable'
+
+
+def get_categories(categories_json):
+    """
+    Retrieve and flatten the nested array from the legacy API response.
+
+    :param categories_json: The returned json
+    :return: A list of categories
```
tbille: `returns`

webapp/public/logic.py
```diff
@@ -181,3 +196,26 @@ def get_default_channel(snap_name):
         return '10/stable'
 
     return 'stable'
+
+
+def get_categories(categories_json):
+    """
+    Retrieve and flatten the nested array from the legacy API response.
```
tbille: text should be on same line as `"""`

PR #857, opened 2018-07-05
Make details endpoint case insensitive

### LINKED ISSUE #614: snap name in URL should be case insensitive
E.g. https://snapcraft.io/Firefox should either load the same page as https://snapcraft.io/firefox, or redirect there. Currently it leads to a 404.

tbille:
# Summary

Fixes #614
- Make Details endpoint case insensitive
- When accessing details endpoint with a Uppercase in the snap name, you should be redirect to the details endpoint with lower case snap name

# QA

- `./run`
- http://127.0.0.1:8004/firEfox
- http://127.0.0.1:8004/Firefox
- http://127.0.0.1:8004/FirefoX
- Should redirect to: - http://127.0.0.1:8004/firefox

### DIFF
--- a/webapp/store/views.py
+++ b/webapp/store/views.py
@@ -306,6 +306,11 @@ def store_blueprint(store=None):
             **context
         ), status_code
 
+    @store.route('/<regex("[A-Za-z0-9-]*[A-Za-z][A-Za-z0-9-]*"):snap_name>')
+    def snap_details_case_sensitive(snap_name):
+        return flask.redirect(
+            flask.url_for('.snap_details', snap_name=snap_name.lower()))
+
     @store.route('/store/categories/<category>')
     def store_category(category):
         status_code = 200


nottrobin:
Is this okay because it's impossible to create snap names with uppercase characters? Do we know that for sure?

nottrobin:
@tbille I understand what's happening. What I'm saying is, your work will redirect snapcraft.io/FiReFoX to snapcraft.io/firefox, which for works great.

What I'm asking is, what if there was a snap called "Robin", with a capital? Then presumably snapcraft.io/Robin would show it to you and snapcraft.io/robin would show a 404. But now, snapcraft.io/Robin would redirect to snapcraft.io/robin and everyone would see a 404.

I'm just asking - do we know for sure that all snap names are forced to be lowercase?

Lukewh:
@nottrobin we are: https://bazaar.launchpad.net/~canonical-ca-hackers/software-center-agent/trunk/view/head:/src/devportal/utilities.py#L83

This is the code used to validate the names on the API end xox

nottrobin:
Great, thanks @Lukewh, that's what I wanted to know :)

PR #865, opened 2018-07-06
Add get all snaps api endpoint

tbille:
# Summary

- add funciton to get all snaps

# QA

- `./run`
- try function on an endpoint


webapp/api/store.py
```diff
@@ -92,6 +98,13 @@ def process_response(self, response):
 
         return body
 
+    def get_all_snaps(self):
```
Lukewh: we should probably add a `size` param to this, some stores might have 100s or 1000s.

Issue #731, opened 2018-06-14
Simplify choice on channel map


therealjuan:
## What is happening?

Right now, snapcraft.io is asking too much on snap consumers to decide or view which versions are available for them to install on a given snap. Before making the decision, they should aware of what track, architecture or risk levels are which is a lot to ask before installing a snap.

## What can we do?
To help with the decision making, we can ...

- Default the **main install version** to the lowest risk level of the default track (latest) 
- Show an option to install other versions next to the main install action being conscious of the install being more prominent
- Display **other relevant releases to install by default** in a list to ease choosing. These relevant releases include (a) all risk levels available on the default track (latest) (b) highlighting its lowest risk available and (c) lowest risk level available for the remaining available tracks
- Allow users to **view all other releases available** if they wish to
- (optional) Surface other relevant data on releases to help users make a decision like published date snap size or revision number
- Remove size and version from the details as they are attached to the snap version you would install

## How to install latest/stable using main CTA

![collapsed](https://user-images.githubusercontent.com/8167677/41963388-5812bdb6-79ef-11e8-9797-ec104ba73bb6.png)

![install stable](https://user-images.githubusercontent.com/8167677/41963389-582705d2-79ef-11e8-8250-1250f8dd65f6.png)

## See other versions and install alternative channel

![other versions with hover](https://user-images.githubusercontent.com/8167677/41963387-57f73dd4-79ef-11e8-9757-e9b3608ef887.png)

![install 10 stable](https://user-images.githubusercontent.com/8167677/41963419-732f98d0-79ef-11e8-9c84-c4981dcc1229.png)

PR #881, opened 2018-07-10
Surface non-stable snaps in search

Lukewh:
# Done

- Enables the `wide` `scope` on the search endpoint
- Hides the install button and associated overlay if there is not a stable version

# QA

- Pull the branch
- `./run`
- Search for `hello-roadmr`
- You should get many more results than https://snapcraft.io/search?category=&q=hello-roadmr
- Click on `hello-roadmr-comm4`
- You should not have the install button in the header and clicking 'all versions' should only have 1 tab

### DIFF
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -126,8 +126,15 @@ function setArchitecture(arch, packageName, channelMapData) {
 function selectTab(tabEl, tabsWrapperEl) {
   const selected = tabEl.getAttribute('aria-selected');
   if (!selected) {
-    tabsWrapperEl.querySelector('.p-channel-map__tab.is-open').classList.remove('is-open');
-    tabsWrapperEl.querySelector('.p-tabs__link[aria-selected]').removeAttribute('aria-selected');
+    const selectedTabEle = tabsWrapperEl.querySelector('.p-channel-map__tab.is-open');
+    if (selectedTabEle) {
+      selectedTabEle.classList.remove('is-open');
+    }
+
+    const selectedTabLinkEle = tabsWrapperEl.querySelector('.p-tabs__link[aria-selected]');
+    if (selectedTabLinkEle) {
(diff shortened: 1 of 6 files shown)


Greggless:
Is the UI/UX for this void as the behaviour will be different taking into consideration the new channel map updates? 

There's an interesting issue happening now; we 'special-cased' Node to display **`10/stable`** by default. Now that the `wide==` parameter has been brought in, Node's page defaults to the nightly build. Do we still need to 'special-case' Node?

We also need to change the publisher pages a little, but that's probably for another PR
image

matthewpaulthomas:
👍

> Is the UI/UX for this void as the behaviour will be different taking into consideration the new channel map updates?

As far as I can see, this change alters only the “Install” button. The channel map appears when you click the “All versions” button, which is unaffected.

> Now that the `wide==` parameter has been brought in, Node's page defaults to the nightly build.

I don’t see this behavior, because I see #891 instead. (Initially I thought that bug was node-specific.)

Greggless:
> As far as I can see, this change alters only the “Install” button. The channel map appears when you click the “All versions” button, which is unaffected.

I was thinking more along the lines of when [these changes](https://github.com/canonical-websites/snapcraft.io/issues/731) come in. I'm not sure what it's final form is, but it might be worth considering the affect non-stable snaps have

bartaz:
Code looks good. But I'm not sure how `node` should work now. We have some custom code for it that was meant to make it default to `10` track (which has stable). This seems to be messing it up a bit by showing empty channel in default install instructions.

Juans PR done separately (https://github.com/canonical-websites/snapcraft.io/pull/893) 'fixes' it, but is just removes channel from install instructions and this is wrong for node which should show install instructions from `10/stable`.

bartaz:
So (after fixing conflict and merging master) there is a small issue with node install instructions:

screen shot 2018-07-16 at 15 40 07

Channel was removed from install instructions and for `node` (that is a special case for not having stable in latest track) install instructions are wrong. But we may want to fix it separately.

PR #897, opened 2018-07-13
Updated channel map

Lukewh:
# Done

 Fixes https://github.com/canonical-websites/snapcraft.io/issues/731

- Refactored `channelMap.js` to be a class
- Refactored `_channel_map.html`
- Added extra styles to `_snapcraft_details_heading.scss`
- Included `vf-u-hide`
- Removed `Size` and `Version` but added `Last updated` in the details table on the snap details page
- Better implemented `defaultTrack` throughought (hardcoded to 'node' for now, until implemented store-side)
- Made `copySuccess` a method on the global `snapcraft` so as to initialize within JS
- Added `get_lowest_available_risk` to find the lowest available risk for the default track

# QA

- Pull the branch
- `./run`
- Click on some snaps, search for snaps and click on them
- For each snap click the 'Install' button, make sure the overlay works and everything within it does
- For each snap click the 'Other versions' button, make sure the overlay works and everything in does
- The default channel should always be at the top of the table
- The default channel should always have a bold name
- The 'Overview' tab (if present) should show all available risks within 'latest' and the lowest available risk for other tracks
- The 'All releases' tab (if present) should show all available tracks

### DIFF
--- a/static/js/libs/events.js
+++ b/static/js/libs/events.js
@@ -1,0 +1,64 @@
+const Events = {
+  constructor: function() {
+    this.events = {};
+    this.availableHandles = [];
+
+    return this;
+  },
+
+  _addListener: function(type, bindWindow) {
+    const bindTarget = bindWindow ? window : document.body;
+    bindTarget.addEventListener(type, this._handleEvent.bind(this, type));
+  },
+
+  _handleEvent: function(type, event) {
+    const eventTarget = event.target;
+
+    this.events[type].forEach(ev => {
+      // A bit hacky but this allows using the same api to bind to window for key, resize and scroll events
+      const target = ev.selector === 'window' ? window : eventTarget.closest(ev.selector);
+
+      if (target) {
+        ev.func(target, event);
+      }
+    });
+  },
+
+  /**
+   * Add an event listener
+   *
+   * @param {String} type The event type (click, change, keyup etc.)
+   * @param {String} selector CSS Selector to trigger the event
+   * @param {Function} func The function to trigger
+   */
+  addEvent: function(type, selector, func) {
+    if (!this.events[type]) {
+      this.events[type] = [];
+    }
+
+    this.events[type].push({
+      selector: selector,
+      func: func
+    });
+
+    if (!this.availableHandles.includes(type)) {
+      this._addListener(type, selector === 'window');
+      this.availableHandles.push(type);
+    }
+  },
+
+  /**
+   * Add events by type
+   *
+   * @param {Array.<{eventType: {selector: Function}}>} types
+   */
+  addEvents: function(types) {
+    Object.keys(types).forEach(type => {
+      Object.keys(types[type]).forEach(selector => {
+        this.addEvent(type, selector, types[type][selector]);
+      });
+    });
+  }
+};
+
+export default Events.constructor();
--- a/static/js/libs/events.js
+++ b/static/js/libs/events.js
@@ -1,37 +1,38 @@
-const Events = {
-  constructor: function() {
+class Events {
+
+  constructor(defaultBindTarget) {
+    this.defaultBindTarget = defaultBindTarget || document.body;
     this.events = {};
     this.availableHandles = [];
 
     return this;
-  },
+  }
 
-  _addListener: function(type, bindWindow) {
-    const bindTarget = bindWindow ? window : document.body;
+  _addListener(type, selector) {
+    const bindTarget = typeof(selector) === 'string' ? this.defaultBindTarget : selector;
     bindTarget.addEventListener(type, this._handleEvent.bind(this, type));
-  },
+  }
 
-  _handleEvent: function(type, event) {
+  _handleEvent(type, event) {
     const eventTarget = event.target;
 
     this.events[type].forEach(ev => {
-      // A bit hacky but this allows using the same api to bind to window for key, resize and scroll events
-      const target = ev.selector === 'window' ? window : eventTarget.closest(ev.selector);
+      const target = typeof(ev.selector) === 'string' ? eventTarget.closest(ev.selector) : ev.selector;
 
       if (target) {
-        ev.func(target, event);
+        ev.func(event, target);
       }
     });
-  },
+  }
 
   /**
    * Add an event listener
    *
    * @param {String} type The event type (click, change, keyup etc.)
-   * @param {String} selector CSS Selector to trigger the event
+   * @param {String|Element} selector CSS Selector or element to trigger the event
    * @param {Function} func The function to trigger
    */
-  addEvent: function(type, selector, func) {
+  addEvent(type, selector, func) {
     if (!this.events[type]) {
       this.events[type] = [];
     }
@@ -42,23 +43,23 @@ const Events = {
     });
 
     if (!this.availableHandles.includes(type)) {
-      this._addListener(type, selector === 'window');
+      this._addListener(type, selector);
       this.availableHandles.push(type);
     }
-  },
+  }
 
-  /**
-   * Add events by type
-   *
-   * @param {Array.<{eventType: {selector: Function}}>} types
-   */
-  addEvents: function(types) {
-    Object.keys(types).forEach(type => {
-      Object.keys(types[type]).forEach(selector => {
-        this.addEvent(type, selector, types[type][selector]);
+  addEvents(eventTypes) {
+    /**
+     * Add events by type
+     *
+     * @param {{eventType: {selector: Function}}} eventTypes
+     */
+    Object.keys(eventTypes).forEach(type => {
+      Object.keys(eventTypes[type]).forEach(selector => {
+        this.addEvent(type, selector, eventTypes[type][selector]);
       });
     });
   }
-};
+}
 
-export default Events.constructor();
+export default Events;
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -7,6 +7,7 @@ class ChannelMap {
   constructor(selectorString, packageName, channelMapData, defaultTrack) {
     this.RISK_ORDER = ['stable', 'candidate', 'beta', 'edge'];
     this.INSTALL_TEMPLATE = document.querySelector('[data-js="install-window"]').innerHTML;
+    this.CHANNEL_ROW_TEMPLATE = document.querySelector('[data-js="channel-map-row"]').innerHTML;
 
     this.packageName = packageName;
     this.currentTab = 'overview';
@@ -315,13 +316,14 @@ class ChannelMap {
         rowClass.push('no-border');
       }
 
-      let _row = `
-<tr class="${rowClass.join(' ')}" data-js="slide-install-instructions" data-channel="${row[0]}/${row[1]}">
-  <td>${row[0]}/${row[1]}</td>
-  <td>${row[2]}</td>
-  <td class="u-hide--medium u-hide--small">${row[3]}</td>
-  <td class="u-align--center"><a href="#" class="p-channel-map__version-table-install">Install &rsaquo;</a></td>
-</tr>`;
+      let _row = this.CHANNEL_ROW_TEMPLATE
+        .split('${rowClass}')
+        .join(rowClass.join(' '));
+
+      row.forEach((val, index) => {
+        _row = _row.split('${row[' + index + ']}').join(val);
+      });
+
       cache = row[0];
       return _row;
     });
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -1,6 +1,7 @@
-/* global ga, ClipboardJS, snapcraft */
+/* global ga */
 
 import moment from '../../../../node_modules/moment/src/moment';
+import globalEvents from '../../libs/events';
 
 class ChannelMap {
   constructor(selectorString, packageName, channelMapData, defaultTrack) {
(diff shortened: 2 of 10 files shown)


matthewpaulthomas:
As far as I can tell, this change doesn’t “simplify choice on channel map”. There’s no difference in the choices offered. The main difference is that, where the current UI has two buttons opening different tabs in the same popover, this UI has two buttons opening two separate popups, and one of the popups is a page stack with one child page per channel in the selected architecture. That’s quite a lot _more_ complex.

Now, that’s not necessarily a bad thing, if it allows bigger improvements in other aspects. So, reviewing this on what it does do…

Steps forward:
- The popovers no longer span the entire viewport, making it much more obvious that they _are_ popovers.
- Both new popovers are much smaller than the previous one, preserving more context around them.
- The popovers no longer cover their opener buttons, so you can open then close them without moving.
- The “Other versions” popover shows all tracks at once, improving scannability.
- “Install using the command line” is more understandable than “Install using snapd”.

Steps back:
- The secondary button is now even larger — more than twice the size of the primary button! — though people will want to use it much less often than the primary button.
- “Other versions” is odd, since the non-other version isn’t described at the same level of the UI.
- “Other versions” is also incorrect, in that the list includes the same version that “Install” offers.
- “Other versions” has its chevron to the left, when it’s much more common for it to go on the right (cf. [macOS](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/pull-down-buttons/), [Windows UWP](https://docs.microsoft.com/en-gb/windows/uwp/design/controls-and-patterns/menus#menus-vs-context-menus), [Windows desktop](https://docs.microsoft.com/en-us/windows/desktop/uxguide/ctrl-command-buttons#usage-patterns), and snapcraft.io right now).
- The popups are different colours (one grey, one white) for no obvious reason.
- The two labels “All available options to install this snap Architecture” are confusingly close together.
- “Published date” is odd, implying that it might not be the real date. (Perhaps “Date released”?)
- There’s no clue that the list items are clickable unless you hover over the list.
- It’s unusual and fiddly for a popover to contain a page stack at all. Expanding table rows are a possible alternative; probably there are others.

So, the smaller popovers are a big improvement in a few ways, but at the moment I think they’re outweighed by issues with the buttons and popover contents.

therealjuan:
Thanks for your points @matthewpaulthomas, looking at your subjective comments, some can be minimised updating copy and adjusting the visual design.

In regards to a popover not being a "simplified solution to the channel map" I would rather launch this design we worked out with different stakeholders from the wider team and iterate based on feedback.

templates/store/snap-details.html
```diff
@@ -255,9 +266,7 @@ <h4>Users by distribution (log)</h4>
       }
 
       if (typeof ClipboardJS !== 'undefined') {
-        var copy = new ClipboardJS('.js-clipboard-copy');
-
-        copy.on('success', function () {
+        snapcraft.copySuccess = function() {
```
bartaz: If we start using it as a global utility we should have it properly moved to own file/module. We shouldn't have channel map code dependent on a code that is inlined in the template.

webapp/store/views.py
```diff
@@ -333,6 +333,12 @@ def snap_details(snap_name):
         icons = [
             m['url'] for m in details['snap']['media'] if m['type'] == "icon"]
 
+        default_track = '10' if details['name'] == 'node' else 'latest'
```
bartaz: Some code comment would be nice around that, so it's easier to discover that that's a node-only workaround.

webapp/store/logic.py
```diff
@@ -248,3 +248,27 @@ def has_stable(channel_maps_list):
                         return True
 
     return False
+
+
+def get_lowest_available_risk(channel_map, default_track):
```
bartaz: We are adding these logic utility functions - it would be nice to have some basic unit tests for them to make sure that give expected results in different cases.

bartaz:
Couple of small issues:

1. When channel map is opened and window is resized channel map is not positioned relative to button anymore, but moves into weird position

screen shot 2018-07-17 at 13 27 00

2. On mobile channel map has shadow which suggests that it covers something, while it really doesn't (it's just part of the flow of the page)

screen shot 2018-07-17 at 13 26 24

static/js/public/snap-details/channelMap.js
```diff
@@ -1,294 +1,418 @@
...
-        versionEl.innerHTML = `No release in ${risk} channel, using ${fallbackRisk} release.`;
+    archSelect.addEventListener('change', ()=> {
+      this.prepareTable(this.channelMapData[archSelect.value]);
+    });
+
+    this.arch = this.channelMapData['amd64'] ? 'amd64' : architectures[0];
+
+    // To ensure the scope of the event handlers are bound to the class (not the element)
+    // and that 'removeEventListener' works
+    this.closeOnClick = function(event) {
```
bartaz: Is `this.closeOnClick` used anywhere? I found references for `_closeOnClick` but not to the `closeOnClick` one.

static/js/public/snap-details/channelMap.js
```diff
@@ -1,294 +1,418 @@
...
+      this._closeOnClick(event);
+    }.bind(this);
+
+    this.closeOnEscape = function(event) {
```
bartaz: Wouldn't `this._closeOnEscape.bind(this)` do exactly the same?

static/js/public/snap-details/channelMap.js
```diff
@@ -1,294 +1,418 @@
...
-      // open tab based on button click (or install tab by default)
-      const openTabName = openButton.getAttribute('aria-controls') || 'channel-map-tab-install';
+      if (isSameTrack) {
+        rowClass.push('no-border');
+      }
 
-      // clear hiding animation if it's still running
-      clearTimeout(closeTimeout);
+      let _row = `
+<tr class="${rowClass.join(' ')}" data-js="slide-install-instructions" data-channel="${row[0]}/${row[1]}">
```
bartaz: Does this need to be this weirdly indented? I get that this is not to have unnecessary whitespace in HTML, but I guess from browser/DOM point of view it doesn't matter and the code would read better when indented in standard way.

static/js/public/snap-details/channelMap.js
```diff
@@ -1,294 +1,418 @@
...
 
-    if (openButton) {
-      // open tab based on button click (or install tab by default)
-      const openTabName = openButton.getAttribute('aria-controls') || 'channel-map-tab-install';
+      if (isSameTrack) {
+        rowClass.push('no-border');
+      }
 
-      // clear hiding animation if it's still running
-      clearTimeout(closeTimeout);
```
bartaz: On the other note - it's a bit pity that while we started to use HTML template in other place (for install window) we still keep a bunch of quite complicated HTML in the JS.

static/js/public/snap-details/channelMap.js
```diff
@@ -1,294 +1,418 @@
...
+</tr>`;
+      cache = row[0];
+      return _row;
+    });
+
+    el.innerHTML = tbody.join('');
+  }
 
-      const openTab = channelMapEl.querySelector(`.p-tabs__link[aria-controls=${openTabName}]`);
+  prepareTable(archData) {
```
bartaz: This method contains quite some logic that is hard to follow without better context. I don't fully understand what it does (not knowing what is the shape of the data or what does 'trimmed' tracks mean here). Possibly it would be good to add some comments to give some context on what kind of preparations on the channel map data are done here.

static/js/public/snap-details/channelMap.js
```diff
@@ -1,294 +1,418 @@
...
+    let paramString = '';
 
-        ga('gtm1.send', {
-          hitType: 'event',
-          eventCategory: 'Snap details',
-          eventAction: 'Click view in desktop store button',
-          eventLabel: label,
-          eventValue: value
-        });
+    if (channel.indexOf('latest') === 0) {
```
bartaz: I have no idea what this is trying to do... Index equals 0? or 7? ... is channel a table or string? why are we looking for such specific index of substrings?

templates/store/snap-details.html
Lukewh: After you said about this, I looked in to it some more and it turns out clipboardjs uses event delegation, so no need to re-init after the initial page load

webapp/store/logic.py
Lukewh: Added!

webapp/store/views.py
Lukewh: Added a comment

static/js/public/snap-details/channelMap.js
Lukewh: Yeah it's purely so it's not crazy in the output for debugging, I can indent it if you feel passionately about it?

static/js/public/snap-details/channelMap.js
Lukewh: This is removed now for events module

static/js/public/snap-details/channelMap.js
Lukewh: This is removed now for events module

static/js/public/snap-details/channelMap.js
Lukewh: I've moved it :+1:

static/js/public/snap-details/channelMap.js
```diff
@@ -302,6 +302,7 @@ class ChannelMap {
     holder.innerHTML = template;
   }
 
+
```
tbille: remove line

bartaz:
There is a JS error in the console when resizing the window. I think it happens when channel map is opened and then closed:

screen shot 2018-07-20 at 16 04 26

Lukewh:
Thanks @bartaz, updated

static/js/libs/events.js
```diff
@@ -0,0 +1,64 @@
+const Events = {
```
bartaz: Not sure why do we need that... I see that it helps with assigning many handlers for different selectors at once, but I'm just not sure that creating our own event abstraction is that helpful.

static/js/libs/events.js
```diff
@@ -0,0 +1,64 @@
...
+    const bindTarget = bindWindow ? window : document.body;
+    bindTarget.addEventListener(type, this._handleEvent.bind(this, type));
+  },
+
+  _handleEvent: function(type, event) {
+    const eventTarget = event.target;
+
+    this.events[type].forEach(ev => {
+      // A bit hacky but this allows using the same api to bind to window for key, resize and scroll events
+      const target = ev.selector === 'window' ? window : eventTarget.closest(ev.selector);
```
bartaz: Yep, using 'window' as string selector seems a bit hacky. Would be better if selector could be either string (used as a selector) or a DOM object (like window, or just a DOM element) to be used directly.

static/js/libs/events.js
```diff
@@ -0,0 +1,64 @@
...
+    if (!this.availableHandles.includes(type)) {
+      this._addListener(type, selector === 'window');
+      this.availableHandles.push(type);
+    }
+  },
+
+  /**
+   * Add events by type
+   *
+   * @param {Array.<{eventType: {selector: Function}}>} types
```
bartaz: Is param really an Array or and object with nested objects?

static/js/libs/events.js
```diff
@@ -0,0 +1,64 @@
...
+   */
+  addEvents: function(types) {
+    Object.keys(types).forEach(type => {
+      Object.keys(types[type]).forEach(selector => {
+        this.addEvent(type, selector, types[type][selector]);
+      });
+    });
+  }
+};
```
bartaz: I see that it's possibly not needed yet, but it would be good to have a way to remove an event handler. Right now it seems you can only add events and there is no way to remove an event handler if needed.

static/js/libs/events.js
```diff
@@ -0,0 +1,64 @@
...
+
+  _handleEvent: function(type, event) {
+    const eventTarget = event.target;
+
+    this.events[type].forEach(ev => {
+      // A bit hacky but this allows using the same api to bind to window for key, resize and scroll events
+      const target = ev.selector === 'window' ? window : eventTarget.closest(ev.selector);
+
+      if (target) {
+        ev.func(target, event);
```
bartaz: It is a little bit confusing that event handler uses non standard parameters, so instead of just having event (or at least having event as first param) it has target first and then event.

static/js/public/snap-details/channelMap.js
```diff
@@ -1,294 +1,425 @@
...
-    // numbers are defined by any string starting any of the following patterns:
-    //   just a number – 1,2,3,4,
-    //   numbers on the left in a pattern – 2018.3 , 1.1, 1.1.23 ...
-    //   or numbers on the left with strings at the end – 1.1-hotfix
-    if (isNaN(parseInt(track.substr(0, 1)))) {
-      stringTracks.push(track);
-    } else {
-      numberTracks.push(track);
+  bindEvents() {
+    globalEvents.addEvents({
```
bartaz: Events module quite nicely cleans up the events code here, I'm just not sure if we need all of these events handled globally on `body`? Event delegation has obvious benefits (especially when HTML is created dynamically and events don't need to be recreated), but it would be good to have control over it - for example to handle them on the root element of channel map instead of the body.

static/js/public/snap-details/channelMap.js
```diff
@@ -1,294 +1,425 @@
...
-    if (openButton) {
-      // open tab based on button click (or install tab by default)
-      const openTabName = openButton.getAttribute('aria-controls') || 'channel-map-tab-install';
+  writeInstallInstructions(channel) {
+    let paramString = '';
 
-      // clear hiding animation if it's still running
-      clearTimeout(closeTimeout);
+    if (channel.indexOf('latest') === 0) {
+      if (channel.indexOf('stable') !== 7) {
```
bartaz: I think I finally figure out that you are checking here if track is `latest` and channel `stable` in `'latest/stable'` string, but figuring it out from numbers 0 and 7 was not that easy. At least a bit of comment would be nice for our future selves.

Lukewh:
@bartaz yeah sounds good. Part of the reason I did it was to start the conversation, so we can pick it up next week.

static/js/libs/events.js
Lukewh: I've switched these around, I originally put target first, as more of the code seemed to use target (`event` was optional) but it makes more sense to follow the standard of `event` first, as it's a more complete resource.

We could technically get rid of the `target`, but as we do the lookup to closest on `trigger`, we may as well pass it through to save doing it again in most cases.

static/js/libs/events.js
Lukewh: Agreed, figured I'd wait until we needed it

static/js/libs/events.js
Lukewh: Updated this to check for type. Not 100% safe, but better then this hack!

static/js/public/snap-details/channelMap.js
Lukewh: I've made the events module a class so we can define a default element to bind to - though in the case of the channelMap I'm just letting it bind to the default body still

Lukewh:
Thanks all

PR #902, opened 2018-07-18
Fix the <span> to be a <div> to increase standards compliance

sparkiegeek:
Minor fix to front page to avoid putting block level elements inside an inline element

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -304,9 +304,9 @@
                 <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/6fbb3483-snapcraft-default-snap-icon.svg" alt="">
               {% endif %}
             </span>
-            <span class="p-media-object__details">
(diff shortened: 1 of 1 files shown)


templates/index.html
```diff
@@ -304,9 +304,9 @@ <h3>Featured snaps</h3>
                 <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/6fbb3483-snapcraft-default-snap-icon.svg" alt="">
               {% endif %}
             </span>
```
Lukewh: Can you update the span above too please for consistency?

sparkiegeek: Yes, done.

sparkiegeek:
https://validator.w3.org/nu/?doc=http%3A%2F%2Fsnapcraft.io-pr-902.run.demo.haus%2F is looking much healthier

PR #904, opened 2018-07-19
Add simple release table to revisions page

### LINKED ISSUE #894: Create static releases matrix
As a first step we should create a read-only version of the release matrix

bartaz:
Fixes #894 

Adds simple static releases table to publishers snap revisions page

### QA

- ./run or demo
- go to publisher snap revisions page
- above revisions list there should be a table of released revisions
- table should only contain architectures that were released

screen shot 2018-07-23 at 11 40 34

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -12,6 +12,5 @@
     "test": {
       "presets": ["env"]
     }
-  },
-  "plugins": ["external-helpers"]
+  }
 }
--- a/.babelrc
+++ b/.babelrc
@@ -12,5 +12,6 @@
     "test": {
       "presets": ["env"]
     }
-  }
+  },
+  "plugins": ["external-helpers"]
 }
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   "dependencies": {
     "autoprefixer": "^6.3.1",
     "babel-core": "^6.26.0",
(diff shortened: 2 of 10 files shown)


therealjuan:
Looking good and probably would need some style tweaking but we can cover that later.

* Table header is uppercasing the architecture, but might be good to keep them in lower-case as we show in the revisions table
* Might be better to use version number instead of revision number as it will be more meaningful to the publisher. We have the case for node having big version strings but it is not 
* Would it be possible to show only 1 track at a time and be able to switch?
* Can we include always the name of the track + risk level as the pair is what constructs the channel even on the latest track (i.e. `latest/stable`)?

There are a few more features we might want to include as part of the first version but we can cover it next week.

bartaz:
@therealjuan I updated the branch with changes you suggested (showing 'latest/', versions, lowercase architectures and dropdown to choose the track).

therealjuan:
Looking good! Would it be possible to...
* ... make architecture columns wider and reduce the width of channel column?
* ... move track selector to the right-hand side of the headline of the release map?
* ... remove # from revisions table?

bartaz:
@therealjuan Implemented your suggestions, so you can have another look, thanks!

PR #922, opened 2018-07-24
WIP Update 0 results for search

Greggless:
Making a PR, innit

🐐 

<hr>

1. search for [something odd](http://snapcraft.io-pr-922.run.demo.haus/search?category=&q=helicpter-hormones)
1. See there are no results
1. See there is some help text
1. See there are no other snaps displayed on the page

### DIFF
--- a/templates/store/search.html
+++ b/templates/store/search.html
@@ -90,14 +90,6 @@
         <h2>Sorry, we couldn’t find anything for your search "{{ query }}"</h2>
         <h4>Try using some different, or more general keywords.</h4>
       </div>
-          <div class="row">
-            <h3>{{ category.name }}</h3>
-          </div>
-          <div data-category="{{ category.slug }}" class="row js-store-category">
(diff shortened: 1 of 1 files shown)


bartaz:
screen shot 2018-07-24 at 10 54 53

One of previously mentioned issues of search results page was that we unnecessarily repeat (especially long) search query. Here we do it again. We have a query in the input and again in quite large heading.

Personally, I don't have much against it, but we are going back to something we have been asked to fixed already.

Greggless:
Ah yes I remember that at one of the Sprints. I think it's good to repeat the search term, in the cases where we are purposefully using a very long search term then it's always going to look bad because it's an abuse of the system out of the normal terms of use. 

A conflict between teams. Will discuss with other UXer and see if it's worth fighting for ⚔️

Greggless:
Closing PR and starting fresh...

PR #923, opened 2018-07-24
limenet update

Lukewh:
# QA

- Pull the branch
- Change `.env` store to `limenet`
- `./run`
- Go to the homepage and see limenet store

### DIFF
--- a/webapp/configs/limenet.py
+++ b/webapp/configs/limenet.py
@@ -1,11 +1,23 @@
 WEBAPP_CONFIG = {
     'LAYOUT': '_layout-brandstore.html',
-    'STORE_NAME': 'Limestore',
+    'STORE_NAME': 'LimeNET',
     'STORE_BRAND': {
-        'LOGO': None,
-        'COLOUR': None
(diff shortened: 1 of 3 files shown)


therealjuan:
LGTM from your screen in the hangout

webapp/configs/limenet.py
```diff
@@ -1,11 +1,23 @@
...
+        {
+            'title': 'Lime Micro',
+            'href': 'http://www.limemicro.com/'
+        },
+        {
+            'title': 'Myriad-RF',
+            'href': 'http://www.limemicro.com/'
+        }
+    ],
+    'CUSTOM_STYLES': '.p-navigation__link a {color: #000;}\
```
bartaz: As much as it probably works in short term, CUSTOM_STYLES should be a URL or file name to CSS file.

Lukewh: Agreed, ideally we wouldn't do any 'custom styles' at all (just config for colours) but at this point i don't want to make things too complicated as Lime might say they hate it and don't want any custom styles :D!

PR #925, opened 2018-07-24
Simplify install intent GA

Lukewh:
This PR removes the complicated intent tracking to only fire an event when the green install button is clicked.

@caldav the important part:
```
{
    hitType: 'event',
    eventCategory: 'Snap details',
    eventAction: 'Open install dialog',
    eventLabel: `Open install dialog screen for ${this.packageName} snap`
}
```

### DIFF
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -12,8 +12,6 @@ class ChannelMap {
     this.packageName = packageName;
     this.currentTab = 'overview';
 
-    this.openDesktopAttempt = 1;
-
     if (!defaultTrack) {
       this.defaultTrack = 'latest';
     } else {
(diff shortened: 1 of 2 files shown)


caldav:
Working well, +1

static/js/public/snap-details/channelMap.js
```diff
@@ -177,12 +175,13 @@ class ChannelMap {
     this.channelOverlayEl.style.display = 'block';
     this.channelMapEl.classList.remove('is-closed');
 
-    if (typeof ga !== 'undefined') {
+    // As discussed with David Calle, we should just track 'install' button clicks
+    if (this.openScreenName === 'channel-map-install' && typeof ga !== 'undefined') {
```
bartaz: If it's meant to be tracking clicking on 'Install' button I think a better place for this would be a click event handler, not a function that renders channel map dialog.

Lukewh: Moved to the click handler, thanks!

PR #926, opened 2018-07-24
0 search results

Greggless:
1. [Search for something that will lead to 0 results](http://snapcraft.io-pr-926.run.demo.haus/search?category=&q=asdfg)
1. See the help text
1. Do not see the categories

### Screenshot
image

### Success
image

### DIFF
--- a/static/sass/_search-form.scss
+++ b/static/sass/_search-form.scss
@@ -24,5 +24,5 @@
   overflow: hidden;
   white-space: nowrap;
   font-weight: 400;
-  }
+ }
 }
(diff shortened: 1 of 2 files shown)


therealjuan:
All well and good but if you add `"` and large words to the term...

image

To solve this we can ... 

- **(a)** strip the quotes or 
- **(b)** leave the search as is and have a different style for the search term, as well as use [ellipsis](https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/) to avoid the layout being broken

Greggless:
It would be nice to differentiate the system text from the query text. If quotes are a no-go we can use Ubuntu regular for the query text. But that might look kinda weird and oddly sized with it being a `H2`. So I'd suggest using `H3` (and reflect changes to other pages of a similar template). 

Ellipsis are cool, though 😄

PR #927, opened 2018-07-24
Update 404.html

Greggless:
1. Try looking for a snap (or page) that doesn't exist: http://snapcraft.io-pr-927.run.demo.haus/helipcopterhormones
1. Get a helpful 404
image

### DIFF
--- a/templates/404.html
+++ b/templates/404.html
@@ -5,12 +5,21 @@
 {% block content %}
 <div class="p-strip">
   <div class="row">
-    <div class="col-6">
+    <div class="col-10">
       <h2>404: Aw, snap! We couldn't find that page</h2>
       <h4>Looking for a snap? Try searching:</h4>
-      <h5>
+      <form action="/search" class="p-form p-form--inline p-form--search">
+     <div class="p-form__group">
+        <label for="search-input" class="u-off-screen">Search</label>
+        <div class="p-form__control u-clearfix">
+          <input class="u-no-margin--bottom" id="search-input" type="search" name="q" value="">
+        </div>
+      </div>
+      <button type="submit" alt="search" class="p-button--positive">Search</button>
+    </form>
+      <p>
         <p> Or try searching the <a href="https://forum.snapcraft.io">Snapcraft Forum</a> Snapcraft Forum or <a href="https://docs.snapcraft.io">Documentation</a> </p>
-      </h5>
+      </p>
     </div>
   </div>
 </div>
--- a/templates/404.html
+++ b/templates/404.html
@@ -7,15 +7,15 @@
   <div class="row">
     <div class="col-10">
       <h2>404: Aw, snap! We couldn't find that page</h2>
-      <h4>Looking for a snap? Try searching:</h4>
-      <form action="/search" class="p-form p-form--inline p-form--search">
+        <h4>Looking for a snap? Try searching:</h4>
+          <form action="/search" class="p-form p-form--inline p-form--search">
      <div class="p-form__group">
         <label for="search-input" class="u-off-screen">Search</label>
-        <div class="p-form__control u-clearfix">
-          <input class="u-no-margin--bottom" id="search-input" type="search" name="q" value="">
-        </div>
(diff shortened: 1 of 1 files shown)


therealjuan:
Only one tiny comment... Can we add an external link icon to the forum link? Then LGTM 👍

templates/404.html
```diff
@@ -5,10 +5,20 @@
...
   <div class="row">
-    <div class="col-6">
-      <h1>404: Page not found</h1>
-      <p class="p-notification__response">
-        {{ error }}
+    <div class="col-10">
+      <h2>404: Aw, snap! We couldn't find that page</h2>
+      <h4>Looking for a snap? Try searching:</h4>
+      <form action="/search" class="p-form p-form--inline p-form--search">
+     <div class="p-form__group">
```
Lukewh: Indentation sun!?

templates/404.html
```diff
@@ -5,10 +5,20 @@
 {% block content %}
 <div class="p-strip">
   <div class="row">
-    <div class="col-6">
-      <h1>404: Page not found</h1>
-      <p class="p-notification__response">
-        {{ error }}
```
Lukewh: Do we not care if there's a specific error any more?

templates/404.html
```diff
@@ -5,10 +5,20 @@
...
+     <div class="p-form__group">
+        <label for="search-input" class="u-off-screen">Search</label>
+        <div class="p-form__control u-clearfix">
+          <input class="u-no-margin--bottom" id="search-input" type="search" name="q" value="">
+        </div>
+      </div>
+      <button type="submit" alt="search" class="p-button--positive">Search</button>
+    </form>
+      <p>
+        <p> Or try the Snapcraft <a href="https://forum.snapcraft.io">Forum</a> or <a href="https://docs.snapcraft.io">Documentation</a> </p>
```
tbille: `p`-ception

templates/404.html
```diff
@@ -5,10 +5,20 @@
...
+      <form action="/search" class="p-form p-form--inline p-form--search">
+     <div class="p-form__group">
+        <label for="search-input" class="u-off-screen">Search</label>
+        <div class="p-form__control u-clearfix">
+          <input class="u-no-margin--bottom" id="search-input" type="search" name="q" value="">
+        </div>
+      </div>
+      <button type="submit" alt="search" class="p-button--positive">Search</button>
+    </form>
+      <p>
```
Lukewh: Yo dawg, I heard you liked `<p>`'s so I wrapped your `<p>` in a `<p>` so you could paragraph while you paragraph.

You should remove the outer `<p></p>`

templates/404.html
```diff
@@ -5,10 +5,20 @@
...
+        <p> Or try the Snapcraft <a class="p-link--external" href="https://forum.snapcraft.io/categories">Forum</a> or <a href="https://docs.snapcraft.io">Documentation</a> </p>
```
tbille: `p`-ception ep. 2

templates/404.html
Lukewh: `<p>Or....</a></p>` no spaces after the `<p>` or before the `</p>`

templates/404.html
tbille: + no need for spaces after opening tab, same for closing tab

templates/404.html
Greggless: What's the standard? Should I continue to indent subsequent items further in?

Lukewh: All children should be indented. So every time you have an opening tag `<tag>` you should indent. Every time you have a closing tag `</tag>` you should outdent. You don't indent on elements that are self closing `` `<input />` `<br />`

templates/404.html
Greggless: I don't think so. But I am probably wrong. 

What were the differentiating factors @tbille?

I understood there was a 'page not found' and a 'snap not found'. The latter use case happened because only stable snaps were surfaced and certain architecture dependencies. Both of those have been fixed. 

This new page _hopefully_ allows an onward journey for users to find what they need

PR #930, opened 2018-07-24
Updated whitelisting to make it more specific

Lukewh:
This PR makes snapcraft.io js whitelisting more specific - by setting the path we should remove unrelated errors on the page, but it does mean that any inline code we have written will not be caught.


bartaz:
But inline stuff wrapped in `Raven.context` will be caught? I guess we wrap everything (at least everything worth it), so hopefully we will be fine.

Lukewh:
@bartaz yeah exactly, we just have make sure we wrap the things that are important

Issue #592, opened 2018-04-26
Header items placed inconsistently on different sections of the site


matthewpaulthomas:
The Snapcraft logo and header links have three different horizontal placements on [snapcraft.io](https://snapcraft.io/), [build.snapcraft.io](https://build.snapcraft.io/), and [docs.snapcraft.io](https://docs.snapcraft.io/). This is sloppy and disorienting.

The items are positioned the same if the window is narrow, but diverge as the window gets wider.

As [I wrote in snapcraft-design#294](https://github.com/CanonicalLtd/snapcraft-design/pull/294#issuecomment-369528129):
>There is a lot of left and right padding in the header, apparently to line up with the page content … But the page content of what? Currently:
> - snapcraft.io is `max-width: 1030px`
> - build.snapcraft.io is `max-width: 1070px`
> - forum.snapcraft.io is `max-width: 1110px;` … `padding: 0 8px`
> - docs.snapcraft.io is implicitly `width: 100%`, and explicitly `margin: 0;` … `padding: 0`
>
> Either the header shouldn’t try to line up with the page content; or the domains all need changes to their page layout to line up with the headers. I’d suggest the first, with the navigation links flush left and the sign-in/sign-out flush right. Not because it’s less work, though it is — but because, for example, it’s quite reasonable for a Snapcraft brochure page to have much bigger left and right margins than a page of snap download charts does.

See hat PR for discussion with @lyubomir-popov about whether to fix this by left-aligning the header, or by making all snapcraft.io pages the same width.

[Noticed again by @therealjuan in [snapcraft-design#410](https://github.com/CanonicalLtd/snapcraft-design/issues/410).]

Lukewh:
maas.io, vanillaframework.io, landscape.canonical.com and jujucharms.com all have 'full width' navigation bars, so we should reflect this with snapcraft.io and build.snapcraft.io

lyubomir-popov:
@Lukewh I may be unaware of some discussions, but I don't think an argument like "landscape or vanilla is full width, so snapcraft should also be" is a sound one.

To take that logic to an extreme, pls consider the following examples:
• In the cases of docs sites, including the snapcraft docs, the full width applies to the body of the site too. Does that mean we need to make the body of snapcraft.io and all other product sites full width?
• all of the above sites are on vanilla 1.6, does that mean we need to roll back any updates to vanilla 1.8/2.0?

[...]

PR #931, opened 2018-07-24
Make navigation full width

Lukewh:
Partially fixes https://github.com/canonical-websites/snapcraft.io/issues/592

build.snapcraft.io will be sorted by https://github.com/CanonicalLtd/snap-squad/issues/640

# QA

- Pull the branch
- `./run`
- Click around (within this codebase)
- The navigation should be full width

### DIFF
--- a/templates/_header.html
+++ b/templates/_header.html
@@ -1,74 +1,72 @@
 <header id="navigation" class="p-navigation">
-  <div class="row">
-    <div class="p-navigation__banner">
-      <div class="p-navigation__logo">
-        <a class="p-navigation__link" href="/">
-          <img class="p-navigation__image" src="https://assets.ubuntu.com/v1/7f93bb62-snapcraft-logo--web-white-text.svg" alt="Snapcraft" />
-        </a>
-      </div>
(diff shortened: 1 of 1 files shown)


therealjuan:
All well and good but if you add `"` and large words to the term...

image

To solve this we can ... 

- **(a)** strip the quotes or 
- **(b)** leave the search as is and have a different style for the search term, as well as use [ellipsis](https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/) to avoid the layout being broken

Greggless:
It would be nice to differentiate the system text from the query text. If quotes are a no-go we can use Ubuntu regular for the query text. But that might look kinda weird and oddly sized with it being a `H2`. So I'd suggest using `H3` (and reflect changes to other pages of a similar template). 

Ellipsis are cool, though 😄

Greggless:
GTML :shipit:

lyubomir-popov:
@Greggless I believe the design +1 should be applied by someone on the visual design side of things, rather than the UX side. Maybe the label needs renaming to avoid confusion?

Greggless:
Ah, apologies @lyubomir-popov. That was actually supposed to be a `UX+1` rather than a `Design +1`

PR #936, opened 2018-07-25
[HACKDAY] Add copydocs metadata

bartaz:
Adds copy doc metadata to snapcraft.io home page.

### QA

- go to https://github.com/bartaz/ubuntu-copy-docs and install extension for Firefox or Chrome
- with extension installed visit snapcraft home page http://snapcraft.io-pr-936.run.demo.haus
- there should be a link at the bottom that goes to copy doc

screen shot 2018-07-26 at 10 08 09

### DIFF
--- a/templates/_layout.html
+++ b/templates/_layout.html
@@ -6,6 +6,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="description" content="{% block meta_description %}Snaps are containerised software packages that are simple to create and install. They auto-update and are safe to run. And because they bundle their dependencies, they work on all major Linux systems without modification.{% endblock %}">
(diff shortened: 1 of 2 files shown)


templates/_layout.html
```diff
@@ -6,6 +6,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="description" content="{% block meta_description %}Snaps are containerised software packages that are simple to create and install. They auto-update and are safe to run. And because they bundle their dependencies, they work on all major Linux systems without modification.{% endblock %}">
+    <meta name="copydoc" content="{% block meta_copydoc %}https://drive.google.com/drive/folders/1q9ia7iDuWQlCaJ7nfKVeQ5G-FPyJbP-H{% endblock %}">
```
Lukewh: maybe put the `{% block meta_copydoc %}` around the whole line? Or does the extension rely on having an empty `copydoc` meta tag?

Lukewh: Ignore me, that's the default to the folder.

bartaz: Yep, the idea is to have a folder as default link in case some page doesn't define a specific copy doc.

PR #940, opened 2018-07-26
WIP: Prototype UI for promoting released revisions

bartaz:
UI prototype for promoting released revisions

- adds a button next to released revisions that allow to promote given revision to lower risk channel
- when revision is promoted it's shown in the UI in target channel
- if there is any revision waiting to be promoted message is shown with Apply/Cancel buttons
- Apply is currently disabled (backend not ready), so changes can't be saved

screen shot 2018-07-26 at 17 22 57


therealjuan:
There are a few adjustments we can do:

* Show a summary of changes you accumulate when promoting to have a better overview
* Rename the tab from `Revisions` to `Release` as per the URL indicates
* `Undo` seems like a necessary action to have, as there is no way to go back to a previous state unless you reload and then `undo` the last step will `Cancel` all promotions
* Adding some visual indication on which one is the `old` version with some opacity on the version perhaps
* We probably might want to move the button on the left hand side of the version number / release number. Do we actually need the release number altogether or could it live under a tooltip with the published date and other details?

Some further next steps but we can talk in more detail in our meeting Tomorrow
* Probably including guardrails feels like a necessary task to perform after this, as we might want to define `edge` or `stable`
* There is an arrow pointing up when you look on the release table at the CLI so might be good to replicate the same behaviour which is tightly bounded to closing channels
* Maybe we can even look at transversing the matrix to be horizontal instead of vertical but might be better to achieve consistency with the current implementation on CLI. Risk levels are always going to be four of them but architectures might vary... I can look into that this week

therealjuan:
We are going to ...

* Design a horizontal version to include all future actions we need – Juan
* Move this to a feature flag – Bartek
* Implement `Undo` under the javascript state – Bartek
* Move to the new v2 API endpoint – Bartek
* Promote and close channels – Juan & Bartek
* View guardrails and warning users – Juan & Bartek
* Grade level warning – Juan & Bartek
* Present on cross-team snapcraft sync
* Release history prior stakeholder showcase

.env
```diff
@@ -5,3 +5,4 @@ ENVIRONMENT=devel
 DEVEL=True
 BLOG_ENABLED=true
 WEBAPP=snapcraft
+RELEASE_UI_ENABLED=True
```
Lukewh: Maybe set it to false by default?

PR #944, opened 2018-07-27
WIP Organise the JS

Lukewh:
# Done

- Included most 3rd party libraries via imports (reducing the number of files requested)
- Included d3 version of metrics graphs
- This means 3rd party billboard and local module has been removed
- Updated `clean` in `package.json` to clear out the `dist` and `modules` folders

# TODO

- Active Devices graph:
  - Refactor and add more comments
  - Redraw graph on resize
  - Make tooltip show even when not hovering over an area


package.json
```diff
@@ -9,13 +9,13 @@
...
+    "copy-3rd-party-js": "cp node_modules/clipboard/dist/clipboard.min.js static/js/modules/ && cp node_modules/raven-js/dist/raven.min.js static/js/modules/",
     "build-js-transpile": "rollup -c",
     "serve": "./entrypoint.sh 0.0.0.0:${PORT}",
     "watch": "watch -p 'static/sass/**/*.scss' -p 'static/js/**/*.js' -c 'yarn run build'",
     "watch-scss": "watch -p 'static/sass/**/*.scss' -c 'yarn run build-css'",
     "watch-js": "watch -p 'static/js/**/*.js' -c 'yarn run build-js'",
-    "clean": "rm -rf node_modules yarn-error.log css static/css *.log *.sqlite _site/ build/ .jekyll-metadata .bundle"
+    "clean": "rm -rf node_modules yarn-error.log css static/css *.log *.sqlite _site/ build/ .jekyll-metadata .bundle static/js/modules/* static/js/dist/*"
```
bartaz: I'm not sure about adding `static/js/modules` to clean, especially now when we removed it.

bartaz: Also, while we are getting rid of `modules` probably it should be removed from .gitignore as well

Lukewh: There's still a couple of modules we need so maybe we don't need to clean it out on second thoughts, wdyt?

bartaz: Ah, true, didn't notice that we still keep some around. I'm fine with cleaning, so unnecessary files are removed, but I guess this script will remove the README.md that is in that folder now (and that's what happened in the PR I guess - that's why I thought it's not needed anymore).

bartaz: Anyway, keeping readme in this folder is probably unnecessary anyway, we could just put the whole folder into gitignore, so nothing gets added there.

static/js/public/snap-details/map.js
```diff
@@ -1,29 +1,33 @@
-/* global d3, topojson */
-
 import SnapEvents from '../../libs/events';
+import { select as d3Select, mouse as d3Mouse } from 'd3-selection';
```
bartaz: Not sure if we need `d3...` prefixing of the names. In scope of the module it will probably be clear enough what given `select`, `json`, `geoPath`, etc is. Or it will be easy enough to check where it's imported from.

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -1,201 +1,223 @@
...
 function showGraph(el) {
-  formatAxis(el);
   el.style.opacity = 1;
 }
 
-export default function activeDevices(days, activeDevices, type) {
-  const el = document.getElementById('active_devices');
+export default function activeDevices(activeDevices) {
+  // Basic svg setup
+  const holder = document.querySelector('svg').parentNode;
```
bartaz: It seems like we are assuming here that there is only one `svg` in the document. This may break if something is added, even in completely unrelated part of the page. Would be good to have more specific component related selector.

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -1,201 +1,223 @@
...
+    return d3Format('.2s')(number);
+  };
+
+  const commaValue = number => {
+    return d3Format(',')(number);
+  };
+
+  // Prepare the axis
+  const x = d3ScaleLinear().
+    rangeRound([padding.left, width - padding.left - padding.right]);
```
bartaz: That's an odd way to do chained functions. Without . in the beginning it looks like independent function call that was wrongly indented. I guess more common way is 

```
something()
  .somethingElse()
  .somethingCompletelyDifferent()
```

static/js/publisher/metrics/graphs/activeDevices.js
Lukewh: Thanks, this was a bad copy/paste that I did change, but then pasted over again!

static/js/publisher/metrics/graphs/activeDevices.js
Lukewh: That's what I thought, but for some reason my IDE complained!? I think I need to check the settings

Lukewh:
Being split in to separate PR's:

https://github.com/canonical-websites/snapcraft.io/pull/1050
and
[TO COME]

Lukewh:
Closing as it's massively out of date

PR #957, opened 2018-07-31
Add build.snapcraft.io homepage to snapcraft.io/build

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/919

- Added the content in a more familiar vanilla style

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/build
- Compare the page to https://build.snapcraft.io

### DIFF
--- a/static/sass/_snapcraft_custom-typography.scss
+++ b/static/sass/_snapcraft_custom-typography.scss
@@ -43,12 +43,6 @@
     @extend %paragraph-copy;
   }
 
-  // update to p-list is-ticked style
-  // using em instead of rem to position the tick to make it look good with bigger fonts
-  .is-ticked {
-    background-position: 0 .3em;
-  }
-
   // This can be removed when
   // https://github.com/vanilla-framework/vanilla-framework/issues/1627
   // is resolved.
--- a/static/sass/_snapcraft_p-build.scss
+++ b/static/sass/_snapcraft_p-build.scss
@@ -1,0 +1,15 @@
+@mixin snapcraft-p-build {
+  .p-build__github-button {
+    font-size: 1.25rem;
+    margin-top: $spv-inter--regular-scaleable;
+
+    img {
+      display: inline-block;
+      height: 1.5em;
+      margin-left: .5rem;
+      vertical-align: middle;
+    }
+  }
+}
+
+
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -180,3 +180,6 @@ img {
 
 @import 'snapcraft_release';
 @include snapcraft-release;
+
+@import 'snapcraft_p-build';
+@include snapcraft-p-build;
--- a/templates/_header.html
+++ b/templates/_header.html
@@ -32,7 +32,14 @@
         </a>
       </li>
       <li class="p-navigation__link" role="menuitem">
-        <a href="https://build.snapcraft.io">Build</a></li>
+        <a
+          {% if page_slug == 'build' %}
+            class="is-selected"
+          {% endif %}
+            href="/build">
+          Build
+        </a>
+      </li>
       <li class="p-navigation__link" role="menuitem"><a href="https://docs.snapcraft.io">Docs</a></li>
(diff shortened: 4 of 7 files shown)


therealjuan:
Probably the only comment I have will be to update the icons on the quotes:

matthewpaulthomas:
If I go to [the current Build page](https://build.snapcraft.io/), choose “Set up in minutes”, and auth to GitHub, I end up on a page where the header has exactly the same items in the same positions with the same font size as before. The only change is a deliberate one: showing that I’m now signed in.

But if I go to [the demo Build page](http://snapcraft.io-pr-957.run.demo.haus/build) and choose “Set up in minutes”, I end up on a page where the header has some items the same, some different, all of them in a different position, and all in a smaller font.

#921 requests to “Update build.snapcraft.io header to match snapcraft.io”. But doing that would fix the header on this page automatically. So, what’s the point of reimplementing this page? What’s the point of having one “Build” page on snapcraft.io, while all the other “Build” pages remain on build.snapcraft.io until it’s retired (at which point this page will be redundant anyway)?

(This might be easier to understand if #919 had a description. It’s linked to [snap-squad#640](https://github.com/CanonicalLtd/snap-squad/issues/640) which says “there will be no need to maintain 2 different headers and footers”, but the existence of #921 suggests that isn’t true.)

A separate problem is that if I return to this new Build page after signing in, it ignores the fact that I am signed in. Instead of saying “Hi Matthew Thomas, let’s take a look at your repos”, it invites me to “set up” as if I haven’t already done that (incidentally making [build.snapcraft.io#1083](https://github.com/canonical-websites/build.snapcraft.io/issues/1083) worse), and there’s no longer any “Sign in with GitHub” link. We snapcraft.io developers know that the link and the button do the same thing under the hood. But I don’t think it would be obvious to users that they should click a “set up” button just to sign in.

therealjuan:
> > If I go to the current Build page, choose “Set up in minutes”, and auth to GitHub, I end up on a page where the header has exactly the same items in the same positions with the same font size as before. The only change is a deliberate one: showing that I’m now signed in.
> 
> But if I go to the demo Build page and choose “Set up in minutes”, I end up on a page where the header has some items the same, some different, all of them in a different position, and all in a smaller font.
> 
> #921 requests to “Update build.snapcraft.io header to match snapcraft.io”. But doing that would fix the header on this page automatically. So, what’s the point of reimplementing this page? What’s the point of having one “Build” page on snapcraft.io, while all the other “Build” pages remain on build.snapcraft.io until it’s retired (at which point this page will be redundant anyway)?
> 
> (This might be easier to understand if #919 had a description. It’s linked to snap-squad#640 which says “there will be no need to maintain 2 different headers and footers”, but the existence of #921 suggests that isn’t true.)

The point is to have a similar CSS and JS being used on both places so if we update to a newer version of Vanilla the change does not need to be implemented in 2 different codebases. The problem with the header is why #921 exists, then we will fix the header using this same PR and we will have a look at the redirects so we can implement **the separate problem**.

Just give us some time and finish some briefs. Capicci? ;)

bartaz:
As long as BSI is separate project with separate codebase there will be always two codebases to update for header/footer consistency as long as we insist on BSI having the same header as snapcraft.io.

Especially that BSI uses GH as main authentication and it will not be an easy task to have it work with Store/snapcraft.io sign in and GH at the same time.

BSI doesn't even use Vanilla fully, only some specific components which makes the updates even harder and it's still on Vanilla 1.6.4 because of that. Before we migrate BSI to Vanilla 1.7 (or 1.8 now) it will be quite hard to maintain consistency.

lyubomir-popov:
@therealjuan attaching resized logos


[Archive.zip](https://github.com/canonical-websites/snapcraft.io/files/2256729/Archive.zip)

therealjuan:
Thanks @lyubomir-popov will upload to the assets site

therealjuan:
There are some redirect issues but I guess this PR only covers homepage. LGTM otherwise 👍

Lukewh:
@therealjuan I think we should release BSI and this stuff at the same time next week. When all issues in the epic are :+1:'d

therealjuan:
That's what I was imagining because it is 1/3 of the overall task 👍

Issue #959, opened 2018-08-01
TypeError: Cannot read property 'innerHTML' of null


Lukewh:
https://sentry.io/canonical-bm/snapcraftio/issues/627439445/

```
TypeError: Cannot read property 'innerHTML' of null
  at new ChannelMap (../public/snap-details/channelMap.js:18:5)
  at Object.channelMap (../public/snap-details/channelMap.js:414:5)
  at f.<anonymous> (/signal-desktop:589:28)
  at f.d (/static/js/modules/raven.min.js:2:4672)
  at f.context (/static/js/modules/raven.min.js:2:4476)
...
(1 additional frame(s) were not displayed)
```

PR #960, opened 2018-08-01
Further fixes for innerHTML and more logging

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/959

- If the channel map overlay isn't present at the start of execution, abandon with a message to sentry
- Start the channel map js on `DOMContentLoaded`
- Extra `try {} catch` around screenshots

# QA

Again hard to QA because we can't replicate the issue


bartaz:
In my QA the exception was not caught by Raven. But maybe it's because I was hacking it with breakpoints or because of demo server.

PR #973, opened 2018-08-03
Add undo release button

### LINKED ISSUE #967: Implement Undo for releases in the prototype

bartaz:
*Work in progress*

Fixes #967 

Done:
- adds undo 'X' button next to promoted release, so single release can be canceled
- adds tooltip with list of all revisions to be released next to confirm button

screen shot 2018-08-06 at 17 14 57

### DIFF
--- a/static/js/publisher/release/revisionsTable.js
+++ b/static/js/publisher/release/revisionsTable.js
@@ -271,9 +271,9 @@ export default class RevisionsTable extends Component {
     return (releasesCount > 0 &&
       <p>
         <span className="p-tooltip">
-          { releasesCount } revision{ releasesCount > 1 ? 's' : '' } to release
-          {' '}
           <i className="p-icon--question" />
+          {' '}
+          { releasesCount } revision{ releasesCount > 1 ? 's' : '' } to release
           <span className="p-tooltip__message" role="tooltip" id="default-tooltip">
             { Object.keys(releases).map(revId => {
               const release = releases[revId];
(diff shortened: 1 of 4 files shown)


therealjuan:
A few comments:

**Add more logic when promoting/de-promoting channels**
Let's say I promote from edge, all the way to stable:

image

But I decide to skip one channel using the X

image

Right now there is an action to promote to the next risk level but probably we will need to check that there is no currently a pending promotion on the immediate next risk level.

**Styling updates**
Right now the action buttons are quite close to the next architecture so maybe we can add more padding to the action buttons ( Arrow + X) to make it more obvious when the arch has a **–** 

image

**Add info icon to log**
There is a tooltip showing the summary of changes to do but to improve discoverability, we can include an info icon ("i") on the left-hand side to indicate users that they can mouse over to get more details or tap on mobile to get the tooltip.

therealjuan:
Some comments:

- Can we change the page title to `Releases and revision history for SNAP_NAME`
- Can we align on the left the tooltip icon like ` [?] X revisions to release` ?
- Can we add more spacing  between the ` [?] X revisions to release` and the action buttons `[Apply]` and `[Cancel]`?

therealjuan:
LGTM 👍

Lukewh:
@bartaz found an issue when promoting, check the i386 column:

screenshot_2018-08-07 releases and revision history for surl

I promoted 0.6 from edge, but the change didn't show up if I press the up arrow again, it just disappears. The tooltip seems correct though.

Issue #970, opened 2018-08-02
Redirect /logout to build logout and catch the second redirect


Lukewh:
On snapcraft.io/logout:

- Clear the snapcraft.io session
- Redirect to https://build.snapcraft.io/auth/logout to clear build's session
- BSI/auth/logout will then redirect BACK to snapcraft.io/logout so:
- capture based on referrer and redirect to the homepage

SO MANY REDIRECTS D:

PR #974, opened 2018-08-03
If no_redirect set on logout, don't logout of bsi and go straight to snapcraft.io homepage

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/970

Also relies on https://github.com/canonical-websites/build.snapcraft.io/pull/1129

- Added query parameter to `/logout` to prevent a redirect loop

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/login and login
- Visit http://0.0.0.0:8004/logout?no_redirect=true
- See the snapcraft.io homepage
- Visit http://0.0.0.0:8004/login and login
- Visit http://0.0.0.0:8004/logout
- See the build.snapcraft.io homepage (https://github.com/canonical-websites/build.snapcraft.io/pull/1129) will redirect back to the snapcraft.io homepage


webapp/login/views.py
```diff
@@ -72,6 +72,12 @@ def after_login(resp):
 
 @login.route('/logout')
 def logout():
+    no_redirect = flask.request.args.get('no_redirect')
+
     if authentication.is_authenticated(flask.session):
         authentication.empty_session(flask.session)
-    return flask.redirect('/')
+
+    if no_redirect == 'true':
```
bartaz: Not sure if we have any tests for logout, but it would be good to have one for this.

bartaz:
Works well, but I think we need to coordinate releasing this and BSI.

If we release this before update to BSI, logging out of snapcraft.io will end up at BSI home page (which would be confusing).

If we release BSI first, logging out of BSI will end up on snapcraft.io logout (home page?).

As far as I understand none of these will result in redirect loop by itself, so we are save on this side. Second case (BSI logout redirecting to snapcraft.io logout and home page, but not the other way around) is probably less confusing, so probably we should release BSI first and follow up with snapcraft.io release (especially that snapcraft deploy tends to be easier and quicker then BSI).

Lukewh:
Added a couple of basic tests and I agree @bartaz we definitely need to deploy BSI first and follow up immediately with snapcraft.io

webapp/login/views.py
```diff
@@ -72,6 +78,17 @@ def after_login(resp):
...
+        return flask.redirect('/')
+    else:
+        return flask.redirect(BSI_URL + '/auth/logout')
+
+
+@login.route('/login/bsi')
```
bartaz: I think this shouldn't be in the same PR as it would cause issues when deploying. See comment on: https://github.com/canonical-websites/build.snapcraft.io/pull/1129/files

bartaz:
Possibly should have `Don't merge` as well if we plan to merge/deploy it after BSI?

webapp/login/views.py
Lukewh: Agreed, moving to separate PR

webapp/login/views.py
```diff
@@ -72,6 +78,12 @@ def after_login(resp):
 
 @login.route('/logout')
 def logout():
+    no_redirect = flask.request.args.get('no_redirect')
```
tbille: `no_redirect = flask.request.args.get('no_redirect', default="false")`

Lukewh: done

webapp/login/views.py
Lukewh: Added some tests :dancer:

PR #991, opened 2018-08-06
Add caching layer to blog

tbille:
# Summary

Add cache layer to /blog and /article endpoint
This is a PR that will need discussion
I putted a timeout of 50sec

# QA

- `./run`
- http://0.0.0.0:8004/blog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,4 @@ env[23]/
 # Project specific ignores
 static/js/modules/*.js
 static/js/dist/*.js
+cache.tmp/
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -59,4 +59,5 @@ env[23]/
 
 # Project specific ignores
 static/js/modules/*.js
-static/js/dist/*.js
\ No newline at end of file
+static/js/dist/*.js
+cache.tmp/
```
nottrobin: Please add a newline, otherwise the `git ignore` command [from git-extras](https://github.com/tj/git-extras/blob/master/Commands.md#git-ignore) will fail.

PR #1001, opened 2018-08-08
Use releases v2 API instead of revision history

### LINKED ISSUE #968: Move to the new v2 API endpoint

bartaz:
Fixes #968 

### QA
- ./run or demo
- Go to releases page of some snaps
- Everything should work as before
- You can compare with public view to see that same revisions are shown as released

Also test locally with RELEASE_UI_ENABLED flag turned off. You may compare with public view

### DIFF
--- a/webapp/api/dashboard.py
+++ b/webapp/api/dashboard.py
@@ -25,6 +25,11 @@ DASHBOARD_API = os.getenv(
     'https://dashboard.snapcraft.io/dev/api/',
 )
 
+DASHBOARD_API_V2 = os.getenv(
+    'DASHBOARD_API_V2',
+    'https://dashboard.snapcraft.io/api/v2/'
+)
+
 SNAP_PUB_METRICS_URL = ''.join([
     DASHBOARD_API,
     'snaps/metrics',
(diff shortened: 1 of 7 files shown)


webapp/api/dashboard.py
```diff
@@ -71,6 +71,17 @@
...
 
 
+DASHBOARD_API_V2 = os.getenv(
+    'SNAPCRAFT_IO_API_V2',
+    'https://dashboard.snapcraft.io/api/v2/'
+)
+
+SNAP_RELEASE_HISTORY_URL = ''.join([
+    DASHBOARD_API_V2,
+    'snaps/{snap_name}/releases?page=1&size=100'
```
bartaz: By default page size is 500, but it seems that it results in time out for some bigger snaps (node, android-studio). Reducing size to 100 seems to help for some, but others still time out (node).

I'm not sure if it's an issue on our side (should we wait longer) or on store side.

Lukewh: I think it's probably our side. We're pretty tight on the timeouts

webapp/api/dashboard.py
```diff
@@ -71,6 +71,17 @@
 ])
 
 
+DASHBOARD_API_V2 = os.getenv(
+    'SNAPCRAFT_IO_API_V2',
```
tbille: 'DASHBOARD_API_V2'

bartaz: Ah, good catch - I renamed the var in code but not in env.

PR #1002, opened 2018-08-08
Hide public metrics on brand stores

Lukewh:
# Done

Don't show metrics on public brand store details pages.

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/core
- See the metrics
- Stop the app
- Change `.env` `WEBAPP=limenet`
- `./run`
- Visit http://0.0.0.0:8004/core
- Don't see the metrics

### DIFF
--- a/webapp/store/views.py
+++ b/webapp/store/views.py
@@ -249,6 +249,7 @@ def store_blueprint(store_query=None):
         """
 
         error_info = {}
+        status_code = 200
 
         try:
             details = api.get_snap_details(snap_name)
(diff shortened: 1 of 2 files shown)


webapp/store/views.py
```diff
@@ -17,7 +17,7 @@
 from urllib.parse import quote_plus
 
 
-def store_blueprint(store_query=None):
+def store_blueprint(app, store_query=None):
```
tbille: if you need to get some configuration you can:
```
conf = flask.current_app.config.get('CONFIG')
``

This can avoid weird behaviours (circular imports) and it keeps the blueprint "independent" from the main app

Lukewh: I was trying to find something like this, thanks

PR #1011, opened 2018-08-10
Move different flex grid styles into one pattern

bartaz:
Fixes #597 

Consolidates flex-box grid pattern use on snapcraft home page (for logos, linked OS logos, tools cards, featured snaps) into one pattern that can be possibly adapted and proposed to Vanilla.

Once this gets reviewed and merged we should propose it to Vanilla.

Done:
- created new `p-flex-grid` pattern for flex-box based grid
- `p-flex-grid--centred` can be used if grid items are meant to be centered instead of spread in whole width
- added `p-flex-grid__item` and `p-flex-grid__item--small` for 2 sizes 
- moved existing custom components that use flex-box grid to new `p-flex-grid` and items patterns

### QA

- ./run or http://snapcraft.io-pr-1011.run.demo.haus/
- go to snapcraft home page
- make sure all grid components look good in all screen sizes (for component list see https://github.com/canonicalltd/snap-squad/issues/597#issuecomment-412020662)
- go to snapcraft build page
- make sure OS logos look good on all screen sizes


static/sass/_patterns_logo-links.scss
```diff
@@ -1,11 +1,14 @@
 @mixin p-logo-link {
-
   .p-logo-link {
     border: 1px solid transparent;
     border-radius: 2px;
     display: block;
     margin-top: 0;
 
+    img {
```
anthonydillon: You should target elements in patterns. It coupled the HTML to closely to the pattern.

static/sass/_patterns_flex-grid.scss
```diff
@@ -0,0 +1,48 @@
+@mixin p-flex-grid {
```
anthonydillon: Nice pattern. I have a concern about the name. We should name patterns by the purpose and not the Styling they are applying. For example, in the future we may swap flex for grid or something new. Then the name of the pattern will become confusing.

bartaz: `flex` is kind of meant to be coming from 'flexibility' rather then 'flexbox' - meaning flexible grid, or adaptive grid. But any name suggestions welcome.

static/sass/_patterns_logo-links.scss
bartaz: It bothers me how many times I need to add `display: block` to images in Vanilla. And adding a specific pattern element class name just to do that seemed overkill, but I agree that selectors targeting elements should be avoided, so I guess it will need to be `p-logo-link__image`.

static/sass/_patterns_flex-grid.scss
anthonydillon: I understand but you can see how it could become confusing. `p-agile-grid`, `p-adaptive-grid`, `p-fliud-grid`?

PR #1018, opened 2018-08-15
Keep OS versions to avoid os aggregation on metrics

### LINKED ISSUE #1017: OS version missing in active devices metrics
Following [this post in the forum](https://forum.snapcraft.io/t/why-does-neon-16-04-show-as-neon-in-the-store/6868) some metrics for certain snaps are not displaying the version of an OS.

If we take the `kde-frameworks-5` snap as an example, there is only one reference to `neon` which currently has `16.04` and `18.04` versions but that version number is not represented on the metrics:
image

Looking at the public metrics page for that snap, the version number is referred to in the metrics table for users by OS:

image

tbille:
Fixes #1017


therealjuan:
LGTM

PR #1027, opened 2018-08-21
Add blog posts to snap details pages

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1025 and https://github.com/canonical-websites/snapcraft.io/issues/1030

# QA

- Pull the branch
- `./run`
- Open a few snaps from that page, some will have blog posts, others won't
- Give feedback

# Screenshots (and related issues)
## 1 post
screenshot_2018-08-21 install discord for linux linux apps in seconds snap store
Waste of space

## 2 posts
screenshot_2018-08-21 install skype for linux linux apps in seconds snap store
Not the worse

## Short description and screenshots
screenshot_2018-08-21 install zenkit for linux linux apps in seconds snap store
Description gets lost

### DIFF
--- a/templates/store/snap-details.html
+++ b/templates/store/snap-details.html
@@ -160,7 +160,7 @@
       </div>
       <div class="row u-equal-height">
         {% for article in blog_articles %}
-          {% include 'partials/blog-card.html' %}
+          {% include 'partials/blog-card--minimal.html' %}
         {% endfor %}
       </div>
     </div>
--- a/templates/store/snap-details.html
+++ b/templates/store/snap-details.html
@@ -148,7 +148,7 @@
     </div>
   </div>
 
-  {% if blog_articles %}
+  <div data-js="blog-articles" style="display: none;">
     <div class="row">
       <div class="col-12">
         <hr />
@@ -158,14 +158,13 @@
       <div class="row">
         <h4>Related blog posts</h4>
       </div>
-      <div class="row u-equal-height">
-        {% for article in blog_articles %}
-          {% include 'partials/blog-card--minimal.html' %}
-        {% endfor %}
+      <div class="row u-equal-height" data-js="blog-article-list" data-snap="{{ package_name }}">
       </div>
     </div>
-  {% endif %}
-
+  </div>
+  {% set article_title = {'rendered': '${title}'} %}
+  {% set article = {'slug': '${slug}', 'title': article_title} %}
+  <script type="text/template" id="blog-article-template">{% include 'partials/blog-card--minimal.html' %}</script>
   {% if countries or normalized_os %}
     <div class="row">
       <div class="col-12">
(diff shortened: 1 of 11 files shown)


therealjuan:
**TL/DR;** LGTM 👍 

We can probably think about how to encourage publishers to provide longer descriptions: Adding a word/letter counter on the store listing page or similar. Also, we can be defensive and provide a minimum height to the description element but not sure if that is a good idea.

The padding on the blog posts cards might do with some tweaking but I will create a separate issue to take care of this problem properly.

Would we allow publishers the ability to hide related blog posts on their snap listing pages? Maybe not an option for now as there is not much content on a listing page but something to consider once we start creating more content on listing pages.

evandandrea:
I'd drop the "by" line. It's too crowded. Possibly the images as well.

If the images stay, maybe move the whole thing beneath the graph and distro chart (though I agree with @matthewpaulthomas on levels of importance), to make it look less like screenshots for this app.

matthewpaulthomas:
On further thought, two more reasons to drop the post images:

- In articles of the form “N apps for doing X”, it’s highly likely the image will be a screenshot of a competitor. For example, [the PyCharm page](http://snapcraft.io-pr-1027.run.demo.haus/pycharm-community) shows a screenshot of Visual Studio Code. (A “hide related posts” option wouldn’t help here, because this possibility wouldn’t occur to vendors until it was too late, and because the option would also hide innocuous posts.)

- Conversely, where the image _is_ of the snap you’re looking at, it’s fairly likely to be one of the vendor’s provided screenshots, which would be visually repetitive. For example, [the Visual Studio Code page](http://snapcraft.io-pr-1027.run.demo.haus/vscode) shows two thumbnails of the same screenshot.

lyubomir-popov:
@Lukewh Not much to comment visually, just wondering - would it be a good idea to show dates? I'd find them useful in estimating how 'alive' the project is.

lyubomir-popov:
In writing the above, I assume the publisher actually writes these so a date would an indication of 'liveliness'. If it is us, then maybe that doesn't matter that much.

evandandrea:
@lyubomir-popov we write the posts.

evandandrea:
I'm +1 on this now. Looks great and behaves as expected.

matthewpaulthomas:
I like the idea of including dates, if only to make them look more recognisable as news posts. But it’s fine for now.

templates/store/snap-details.html
```diff
@@ -148,6 +148,23 @@ <h1 class="p-heading--two p-snap-heading__name">{{ snap_title }}</h1>
...
+    </div>
+    <div class="p-strip is-shallow">
+      <div class="row">
+        <h4>Related blog posts</h4>
+      </div>
+      <div class="row u-equal-height" data-js="blog-article-list" data-snap="{{ package_name }}">
+      </div>
+    </div>
+  </div>
+  {% set article_title = {'rendered': '${title}'} %}
```
bartaz: Is `article_title` used anywhere on it's own? Can't it be defined inside of `article` below?

bartaz: Maybe we could prepare this data in the view then? Both solutions are not perfect, I don't have a strong preference.

Lukewh: I thought we might want to use the partial for something else in the future so wanted to keep the variable definitions separate.

PR #1041, opened 2018-08-27
Broken link to Fingbox CS on the IoT page

caldav:
* Fingbox case study link was linking nowhere.


bartaz:
Code looks good, but can't QA due to #1042

Issue #911, opened 2018-07-19
View related articles from a series of blog posts


therealjuan:
There is an increasing need to create a playlist of blog posts that relate to each other so I can read the whole series of blog posts easily and continue to understand the complete story.

Lukewh:
This is dependent on the content being linked by tags - needs conversations with advocacy and marketing!

PR #1049, opened 2018-08-29
Blog series

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/911

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/blog/your-first-robot-a-beginners-guide-to-ros-and-ubuntu-core-1-5
- See a list of the other blogs in the series on the right-hand side
- Visit http://0.0.0.0:8004/vscode and ensure the related blog posts appear there still

### DIFF
--- a/static/js/public/public.js
+++ b/static/js/public/public.js
@@ -2,6 +2,13 @@ import map from './snap-details/map';
 import screenshots from './snap-details/screenshots';
 import channelMap from './snap-details/channelMap';
 import { storeCategories } from './store-categories';
-import { snapDetailsPosts } from './snap-details/blog-posts';
+import { snapDetailsPosts, seriesPosts } from './snap-details/blog-posts';
 
-export { map, screenshots, channelMap, storeCategories, snapDetailsPosts };
+export {
+  map,
+  screenshots,
+  channelMap,
+  storeCategories,
+  snapDetailsPosts,
+  seriesPosts
+};
--- a/static/js/public/snap-details/blog-posts.js
+++ b/static/js/public/snap-details/blog-posts.js
@@ -1,46 +1,107 @@
 import 'whatwg-fetch';
 
+class BlogPosts {
+  constructor(url, holderSelector, templateSelector) {
+    if (!url) {
+      throw new Error('`url` must be defined');
+    }
+    if (!holderSelector) {
+      throw new Error('`holderSelector` must be defined');
+    }
+    if (!templateSelector) {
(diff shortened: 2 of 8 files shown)


lyubomir-popov:
@Lukewh I have a couple of suggestions:

Currently:

image



After applied alterations:

image


Most importantly, things don’t look properly grouped. The title is in a strip of its own, with a marketing-y background, ansd the related articles appear almost as text accompanying  the image

So I got rid of the top banner, and added an extra column of white space in the aside
I would also wrap the image in a p-strip is-shallow
(not done above). Images from cms’s often get wrapped in <p>’s and we have a rule that brings p+p closer together - so the space between image and paragraph is too small.

lyubomir-popov:
@Lukewh I've added a -1 as I'm not aware of any ongoing discussions, and am assessing this as is. If there are reasons we want this out quickly, and have plans to iterate on it later, pls let me know and I'll add a plus one as the above issues are in no way critical.

lyubomir-popov:
@Lukewh @bartaz Also, there seem to be leftover overrides from before Vanilla 1.7/8:

image

matthewpaulthomas:
- The list contains the current post, so “Other posts” is incorrect, sorry, –1 for that. How about just “In this series”? (That would also keep the heading to one line.)

- Linking the post that you’re already on (besides the usual self-link risk of no-op clicking) makes it difficult to see where you are in the series. It would be better as just plain text, or even bold.

- Seeing where you are in the series makes it good that the links are visible at the top of the page. Besides that, though, their main use is to navigate to the next post once you’ve finished this one, right? And for that, they should be visible at the bottom of the page too, so you don’t need to scroll all the way back to the top when you’ve finished. (And so you don’t forget that the links were there at all.) Since the links have a column to themselves, this can be achieved by making them sticky (or at least semi-sticky, scrolling only if the list is taller than the viewport).

Lukewh:
Thanks @matthewpaulthomas, updated if you could take another look. The demo should update soon.

matthewpaulthomas:
Excellent, thanks @Lukewh.

One remaining glitch: If the list is taller than the viewport, you need to scroll all the way to the bottom of the page (well, the bottom of the `<section id="maincontent">`) before you get to see the last item in the list. But that’s quite unlikely, and I guess it’s hard to fix (you’d need to switch between making the top or bottom sticky depending on the scroll direction), so it needn’t block this change.

PR #1061, opened 2018-09-05
Use licenses from spdx github

### LINKED ISSUE #1031: [Backend] Give template list of licences

tbille:
# Summary

Fixes #1031
We are getting the licenses from this file to always get the last licenses updates: https://github.com/spdx/license-list-data/blob/master/json/licenses.json

- First commit is a little improvement for my tests.
- The file is fetched at build time (`yarn build`)

# QA

For now I am just giving the licenses to the frontend, so you can check that the array is well passed to the frontend (by just displaying it for now)

### DIFF
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -1,5 +1,6 @@
 import flask
 from webapp import authentication
+from webapp.helpers import get_licenses
 import webapp.metrics.helper as metrics_helper
 import webapp.metrics.metrics as metrics
 import webapp.api.dashboard as api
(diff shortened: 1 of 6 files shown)


webapp/publisher/snaps/views.py
```diff
@@ -215,6 +212,11 @@ def get_listing_snap(snap_name):
         m['url'] for m in snap_details['media']
         if m['type'] == 'screenshot']
 
+    try:
+        licenses = github_api.get_licenses()
```
bartaz: Do we really need to call API for licences every time we call for listing page? I guess that's not something that changes very often...

tbille: this endpoint is super cached so we won't make that many api calls

Lukewh:
QA bad

`wget not found.`

PR #1062, opened 2018-09-06
d3 version of account metrics

Lukewh:
# Done

Implement the metrics graph in d3.

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/{snap_name}/metrics
- Make sure the graph works as close to what's currently in production


bartaz:
These don't look 'close', is the Y axis inverted?

screen shot 2018-09-07 at 08 11 43

webapp/publisher/snaps/views.py
```diff
@@ -121,7 +121,8 @@ def publisher_snap_metrics(snap_name):
     metrics_query_json = metrics_helper.build_metrics_json(
         snap_id=details['snap_id'],
         installed_base=installed_base,
-        metric_period=metric_requested['int'],
+        # Magic -1 otherwise for 7 days we get 8, for 30 days we get 31?
+        metric_period=metric_requested['int'] - 1,
```
bartaz: Was it wrong before? Is this the issue with API?

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -1,200 +1,253 @@
...
+      <span class="p-tooltip__message" role="tooltip" style="display: block;">\
+        <span class="snapcraft-graph-tooltip__title">${tooltipTimeFormat(data.date)}</span>\
+        ${tooltipRows(data, currentHoverKey, colorScale)}\
+      </span>\
+    </div>`);
+};
+
+function drawGraph(activeDevices) {
+  // Basic svg setup
+  const holder = document.getElementById('activeDevices');
```
bartaz: While we are changing this it would be good not to hardcode ids into JS, just pass them to the function like we do in many other places.

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -1,200 +1,253 @@
...
+    tooltip.style('display', 'none');
+  }
+
+  showGraph(holder);
+}
+
+export default function activeDevices(activeDevices) {
+  drawGraph(activeDevices);
+
+  select(window).on('resize', () => {
```
bartaz: Would be good to debounce resize event to prevent redrawing while user moves the window. Previous version was doing this.

static/js/publisher/metrics/graphs/helpers.js
```diff
@@ -0,0 +1,53 @@
...
+    } else {
+      ticks[tick].classList.add('active');
+      text.setAttribute('x', '-22');
+      line.setAttribute('x2', '-16');
+    }
+  }
+}
+
+function singleLine(str) {
+  return str.replace(/\s\s+/g, '');
```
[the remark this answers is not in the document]
Lukewh: When using template literals, you get extra white space on new lines. So you get random extra lines in the output, making things look XL-padded.

Probably better off just using concat'd arrays for the templates, will change them to that and remove this function.

static/js/publisher/metrics/graphs/activeDevices.js
Lukewh: Good point, will fix

static/js/publisher/metrics/graphs/activeDevices.js
Lukewh: :+1:

Lukewh:
@bartaz Regarding the y-axis. The new version correctly implements new series starting from 0 and taking market share from the previous versions - as per the [spec](https://docs.google.com/document/d/1PQzECKRrVZ-y3W7DcdkfeZipBVr5Xl2mpMj5Vqhux2E/edit#heading=h.rtb5soqn673v). The old version incorrectly started from the top.

webapp/publisher/snaps/views.py
Lukewh: We're asking for today - the period length. Which is inclusive, so we get 1 extra point. Rather then doing this magic number, I've added it to the logic file

webapp/metrics/helper.py
```diff
@@ -39,10 +39,11 @@ def build_metrics_json(
 
     if metric_bucket == 'd':
         start = end - relativedelta.relativedelta(
-            days=metric_period)
+            days=metric_period - 1)
```
bartaz: Thanks for moving it into logic. Would be good to keep a comment what explains the reasoning of `-1` for future generations.

Lukewh: Added :+1:

bartaz:
It seems that versions in tooltip differ a bit:
screen shot 2018-09-07 at 11 09 23

Lukewh:
@bartaz the tooltip difference actually looks like a bug in the old code. It should only ever add 'Other' for item's that are below 0.1% of the total devices IF there are more than 10 versions, which in this case there isn't.

PR #1065, opened 2018-09-06
Add hook precommit

tbille:
# Summary

Added a lint-staged files: there are 2 jobs: one for prettier and one for black. You can have a look at the first commit
I applied black and prettier in the other commits

# QA

- `./run`
- Make sure it works

Love :heart:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,10 +16,10 @@ jobs:
       - checkout
       - run:
           name: Install dependencies
-          command: pip3 install flake8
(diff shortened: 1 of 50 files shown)


tbille:
@bartaz @Lukewh I removed my prettier commit for now, but kept the black commit

It will still have the configuration for prettier for the next commits, seems good?

tbille:
When we decide to apply prettier:  we need to apply this:
`prettier --list-different 'static/**/*.js'`
` prettier --list-different 'static/**/*.scss`

PR #1068, opened 2018-09-10
Don't show related posts if there are none

Lukewh:
# QA

- Pull the branch
- `./run`
- Go to a snap without a related blog post
- The heading 'Related blog posts' shouldn't appear

### DIFF
--- a/static/js/public/snap-details/blog-posts.js
+++ b/static/js/public/snap-details/blog-posts.js
@@ -36,7 +36,7 @@ class BlogPosts {
       then(response => response.json()).
       then(posts => {
         if (posts.length === 0) {
-          return;
+          return false;
         }
         const postsHTML = [];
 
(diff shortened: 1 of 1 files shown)


static/js/public/snap-details/blog-posts.js
```diff
@@ -36,7 +36,7 @@ class BlogPosts {
       then(response => response.json()).
```
bartaz: This way of writing chained calls is confusing. It looks like a separate function call. But as it's not related to this PR lets wait till we turn prettier checks. They should catch those I guess.

static/js/public/snap-details/blog-posts.js
```diff
@@ -77,8 +77,8 @@ function snapDetailsPosts(holderSelector, templateSelector, showOnSuccessSelecto
 
   blogPosts.path = snap;
 
-  blogPosts.fetch(function() {
```
bartaz: Maybe .fetch could just return a promise and not get callback as an argument, but again it's a code style comment not really related to this PR.

PR #1078, opened 2018-09-11
Redirect if user has no usernames

tbille:
# Summary

Redirect to login.ubuntu.com if username is not set

# QA

- Create an account without username
- `./run`
- Try to login
- You should be redirected to login.ubuntu.com


Lukewh:
Wow, is this really Toto that built this?

PR #1086, opened 2018-09-13
Add missing custom licenses

tbille:
# Summary 

Some custom licenses not part of spdx are missing (Proprietary for ex)
They are added trying to follow the pattern of this [file](https://github.com/spdx/license-list-data/blob/master/json/licenses.json)

# QA

- Print the licenses in the listing page
- `./run`
- http://127.0.0.1:8004/account/snaps/toto/listing
- Make sure the licenses added are in the array


Lukewh:
Should this not be done in a similar way to the other licenses? Even if we have a repo that we keep updated?

Issue #1072, opened 2018-09-11
Implement tag component for whitelist/blacklist


Lukewh:
As per the designs

Issue #784, opened 2018-06-20
Blacklist & Whitelist on publisher pages


therealjuan:
As a snap publisher I want to be able to black and white list different countries to distribute my snaps on relevant regions.

Users will be able to CRUD (create, read, update and delete) information about countries / regions where a snap is being able to be distributed.

This functionality is already present at dashboard.snapcraft.io

![image](https://user-images.githubusercontent.com/8167677/41662037-753b6ac2-74a0-11e8-83e6-c7959b7efdf8.png)

Greggless:
Dashboard allows a user to define both a blacklist _and_ a whitelist for their software. I don't think this makes sense. Hence I've made it a binary choice to black or whitelist. 

In Snapcraft a user will be able to 'turn on' distribution list, choosing whether to black or whitelist specific countries. The distribution list is switched off by default meaning the software is freely distributable to all countries

For `Blacklist` and `Whitelist` UI I would recommend using [this multiple search and select](https://semantic-ui.com/modules/dropdown.html#multiple-search-selection) pattern. 

### Distribution list - OFF
![image](https://user-images.githubusercontent.com/23459065/42037657-dff40572-7ae0-11e8-9064-b0a0eb5a8ee9.png)

### Distribution list - ON - Whitelist
![image](https://user-images.githubusercontent.com/23459065/42038356-a6b32192-7ae2-11e8-9e07-12136b0c30fa.png)


### Distribution list - ON - Blacklist - Drop down
![image](https://user-images.githubusercontent.com/23459065/42038078-f1271658-7ae1-11e8-8540-6e31836f5c97.png)

### Distribution list - ON - Blacklist - Drop down
![image](https://user-images.githubusercontent.com/23459065/42038233-53a6b02c-7ae2-11e8-8352-a94794148884.png)


### Distribution list - ON - Blacklist - Drop down - Selection made
![image](https://user-images.githubusercontent.com/23459065/42038197-413ab9ce-7ae2-11e8-9f00-986bf4ef7bfc.png)

therealjuan:
I believe that in some use cases it is easier to provide both controls to publishers as it might happen that you are operating with a certain technology or set of business rules that can be obsolete due to new legislation and that could be quite tricky to achieve with a simple binary choice.

On the other hand, what you propose can be achieved if we allow publishers to decide when a country is either black or whitelisted within the country list and is able to see a summary of white and blacklisted countries working as tag clouds as you envisioned.

Thoughts?

[...]

PR #1089, opened 2018-09-13
Add multiselect component and wire it up for country black/white lists

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1072
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1074
Fixes https://github.com/CanonicalLtd/snapcraft-design/issues/490

UX: https://github.com/canonical-websites/snapcraft.io/issues/784
Design: (tags UI) https://github.com/CanonicalLtd/snapcraft-design/blob/master/snapcraft.io/market%20/Snap%20details%20-%20Listing%20view%20-%20Tags%20UI%20%20REV02.jpg

## QA

- Pull the branch
- `./run`
- Click on a snap from your list of snaps
- Scroll down to the bottom of the 'Listing' page
- Change the checkbox from 'All territories' to 'Custom' under 'This snap should be available in:'
- Select either option and click on the input
- Try filtering countries by typing
- Use the mouse to scroll up and down the options
- Click on options to add them
- Use the arrow keys to scroll up and down the items
- Use the return key to add them
- Use backspace to remove your search term
- Use backspace when the search term is empty to delete the item's you've added, one by one
- Click off the element to lose focus
- Lose focus while a search term is types
- Reclick the element, the search drop-down should reappear with the term you previously searched for
- Do anything else you can think of
- ## edit fields on the Listing page to make sure that still works

# Marked as WIP until an API error is sorted - Saving works, but doesn't get served back... You can however check https://dashboard.snapcraft.io to see the changes have taken effect.


therealjuan:
What about this design from [this issue](https://github.com/canonical-websites/snapcraft.io/issues/784)?

image


We can have 3 exclusive radio button options showing the relevant tag component to simplify choices.

Lukewh:
@therealjuan "We have our solution
A tada fireworks boom" https://github.com/canonical-websites/snapcraft.io/issues/784#issuecomment-403034504

matthewpaulthomas:
Two bugs:

1\. Choose “Custom list”.
2\. Type “br”.
3\. Choose “Brazil”, either with keyboard or pointer.
What should happen: The field now contains only “Brazil” (since that’s what you typed “br” to find).
What actually happens: The field contains “Brazil”, followed by the no-longer-wanted “br”.

1\. Choose “Custom list”.
2\. Type “b”.
What should happen: The list begins “Bahamas”, “Bahrain”, “Bangladesh” — that is, territories starting with what you entered. (Or maybe territories containing a word starting with what you entered, in which case the list should begin “Antigua & Barbuda”, “Bahamas”, “Bahrain”, “Bangladesh”.)
What actually happens: The list begins “Albania”, “Antigua & Barbuda”, “Aruba”, “Azerbaijan”, etc — that is, territories with a “b” _anywhere_ in their name, which is not helpful in this situation.

> I also don't think it's worth creating an 'admin page' to add four related radio buttons.

The tab is destined to contain other stuff too. So maybe it’s better to create it anyway — even if it temporarily has a more specific name, like “Territories”.

static/sass/_patterns_form-multiselect.scss
```diff
@@ -0,0 +1,94 @@
+@mixin snapcraft-p-multiselect {
+  .p-multiselect {
+    @extend %vf-clearfix;
+    align-content: flex-start;
+    align-items: flex-start;
+    border: 1px solid $color-mid-light;
+    border-radius: .125rem;
+    box-shadow: inset 0 1px 1px rgba(0, 0, 0, .12);
```
matthewpaulthomas: This is the same as the current `box-shadow` for Vanilla text fields. But:
- If the Vanilla style changes in future, this will be left behind.
- It gets overridden anyway: http://snapcraft.io-pr-1089.run.demo.haus/static/css/styles.css has `p-multiselect, .p-multiselect .p-multiselect__options {box-shadow: 0 1px 5px 1px rgba(17, 17, 17, 0.2); }`, making these fields inconsistent with the rest of snapcraft.io.
snapcraft io-1089-outset

Lukewh:
I've moved things to an 'Admin' page - Private, Distribution territories and metrics. I'll also add licenses to this page.

I've also fixed the bugs discovered by @matthewpaulthomas, plus a few more.

rollup.config.js
```diff
@@ -116,4 +116,52 @@ export default [
       sourcemap: true
     }
   },
+  {
+    input: 'static/js/publisher/form/multiselect.js',
```
bartaz: Do we need it as separate file? This will make it bit harder when we finally migrate to webpack and also it uses react, so it means that users will download react multiple times (in each bundle that uses it).

bartaz:
### QA issues/questions

1. If I add some countries to 'Only these territories' list but then click on 'Exclude...' countries added to 'Only these' stay in place, which is confusing as the radio button is switched off, so in theory the list should be empty.

Even if we don't want to clear the value (to allow user to change their mind without loosing potentially long list of countries, we could show the multiselect as disabled to have visual clue that it's value is inactive (and won't be saved).

2. Save/Revert is on the bottom of Admin page while it's on top of Listing page.

Lukewh:
Renamed 'Admin' to 'Settings' as per https://github.com/CanonicalLtd/snapcraft-design/issues/490

rollup.config.js
Lukewh: Good point, updated to include it as part of 'publisher.js' and included the relevant bits in the config.

Lukewh:
@bartaz I've updated the inputs so that they're disabled when unavailable. I've also moved the save/revert to the top of the page, as in listings.

matthewpaulthomas:
It’s a bit unfortunate that with the pointer you can delete any territory you’ve entered, but with the keyboard you can delete only the last one. In macOS token fields, you can use arrow keys to move the insertion point between tokens, and delete either left or right. I guess this would be a lot of work to implement, because it would involve reimplementing a lot of other text field behaviour as well, rather than  just using a borderless native text field.

Simpler design issues:
- The “Metrics” controls are specifically about things that appear on the listing page, so they should remain in the “Listing” tab. (None of the designs in the linked issues suggest that they should be moved.)
- Hovering over the non-X part of a token, it darkens, suggesting that it will do something if you click it, but it doesn’t. If the tokens had a selected state, that would suggest you can type to replace the selection, but then if someone typed to replace the selection they’d expect the new text to appear in-place, which would introduce a lot of work as above. Soooo, probably better for it not to darken at all.

Bugs:

1\. Go to [/account/snaps/lukewh-test/settings](http://snapcraft.io-pr-1089.run.demo.haus/account/snaps/lukewh-test/settings)
2\. Choose “Custom list”
3\. Enter one territory, e.g. Afghanistan
4\. Look at the “Save” button, then click it
What you see: It’s greyed out and has the No cursor, but it’s clickable
What you should see: Whenever it’s clickable, it isn’t greyed out and has the normal cursor

1\. Go to [/account/snaps/lukewh-test/settings](http://snapcraft.io-pr-1089.run.demo.haus/account/snaps/lukewh-test/settings)
2\. Choose “Custom list”
3\. Enter one territory, e.g. Afghanistan
4\. Choose “Save”
What you see: “⊙ All territories”
What you should see: “⊙ Custom list” “⊙ Only these territories” “Afghanistan”

Lukewh:
> It’s a bit unfortunate that with the pointer you can delete any territory you’ve entered, but with the keyboard you can delete only the last one. In macOS token fields, you can use arrow keys to move the insertion point between tokens, and delete either left or right. I guess this would be a lot of work to implement, because it would involve reimplementing a lot of other text field behaviour as well, rather than just using a borderless native text field.
> 
> Simpler design issues:
> 
>     * The “Metrics” controls are specifically about things that appear on the listing page, so they should remain in the “Listing” tab. (None of the designs in the linked issues suggest that they should be moved.)
> 
>     * Hovering over the non-X part of a token, it darkens, suggesting that it will do something if you click it, but it doesn’t. If the tokens had a selected state, that would suggest you can type to replace the selection, but then if someone typed to replace the selection they’d expect the new text to appear in-place, which would introduce a lot of work as above. Soooo, probably better for it not to darken at all.
> 

I've removed the hover

> 
> Bugs:
> 
> 1. Go to [/account/snaps/lukewh-test/settings](http://snapcraft.io-pr-1089.run.demo.haus/account/snaps/lukewh-test/settings)
> 2. Choose “Custom list”
> 3. Enter one territory, e.g. Afghanistan
> 4. Look at the “Save” button, then click it
> What you see: It’s greyed out and has the No cursor, but it’s clickable
> What you should see: Whenever it’s clickable, it isn’t greyed out and has the normal cursor

This should be fixed

> 1. Go to [/account/snaps/lukewh-test/settings](http://snapcraft.io-pr-1089.run.demo.haus/account/snaps/lukewh-test/settings)
> 2. Choose “Custom list”
> 3. Enter one territory, e.g. Afghanistan
> 4. Choose “Save”
> What you see: “⊙ All territories”
> What you should see: “⊙ Custom list” “⊙ Only these territories” “Afghanistan”

As mentioned in the big text of the description of this PR:
"Marked as WIP until an API error is sorted - Saving works, but doesn't get served back... You can however check dashboard.snapcraft.io to see the changes have taken effect."

lyubomir-popov:
@Lukewh Is the online link updated? I don't see a checkbox All territories

static/js/libs/arrays.js
```diff
@@ -0,0 +1,23 @@
...
+  if (
+    oldArray === null || newArray === null ||
+    (oldArray.length !== newArray.length)) {
+    return false;
+  }
+
+  oldArray.sort();
+  newArray.sort();
+
+  for(let i = 0; i < oldArray.length; i++) {
```
bartaz: Is `sort` needed? If arrays are meant to be equal can't we just check elements one by one without sorting? Or does `arraysEqual` checks not for equality but just if arrays contain exactly the same elements (in any order?) - in which case the name of the function is confusing.

templates/publisher/settings.html
```diff
@@ -0,0 +1,220 @@
...
+        {% else %}
+          null
+        {% endif %}
+      );
+
+      document
+        .getElementById('settings-form')
+        .addEventListener(
+          'change',
+          snapcraft.publisher.settings.changeHandler
```
bartaz: Why this is separate from `initForm`? Does it handle different things that the form does? If this is not a market page why it uses `market.initForm`?

static/js/libs/arrays.js
```diff
@@ -0,0 +1,23 @@
...
+  if (oldArray === newArray) {
+    return true;
+  }
+  if (
+    oldArray === null || newArray === null ||
+    (oldArray.length !== newArray.length)) {
+    return false;
+  }
+
+  oldArray.sort();
```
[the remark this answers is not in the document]
Lukewh: Yeah so before the sorting the function checks "are they the same array?" if not is either of them null, or are the lengths different? Then is sorts both of them and checks each element in the array, the side effects or sorting shouldn't matter as after the sort it's checking on an item by item basis.

static/js/libs/arrays.js
Lukewh: The sorting is a catch-all in case there's some sorting elsewhere

templates/publisher/settings.html
Lukewh: I'm tempted to rename market.js to form.js... I wanted to reuse all the state logic you'd already written so I did, though it's true, this is not the market page.

static/js/libs/arrays.js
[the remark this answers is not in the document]
bartaz: By side effects I mean you are chaning the arrays you are 'checking'.

```js
a = [3, 2, 1];
b = [2, 1, 3];
arraysEqual(a, b); // true
a; // [1, 2, 3]
b; // [1, 2, 3]
```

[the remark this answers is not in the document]
bartaz: Unless this is expected and you want to make the arrays equal?

templates/publisher/settings.html
bartaz: If it's shared makes sense to rename it - would be less confusing.

static/js/publisher/form/multiselect.js
```diff
@@ -0,0 +1,376 @@
...
+                  key: results[highlighted].key
+                }
+              }
+            });
+          }
+          break;
+        case 'Backspace':
+          if (this.state.selected.length > 0 && this.state.searchTerm === '') {
+            this.removeItem({
+              target: {
```
bartaz: This feels a bit too hacky - building fake event object with just the fields `removeItem` uses...
Please create separate `removeItemClick(event)` and `removeItem(whatever, params, it, needs)`.

static/js/libs/arrays.js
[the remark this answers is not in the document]
Lukewh: Oh I see! I'll add a param to the function to indicate whether you want sorting or not

static/js/publisher/market/state.js
```diff
@@ -7,9 +10,18 @@ const allowedKeys = [
   'contact',
   'private',
   'public_metrics_enabled',
-  'public_metrics_blacklist'
+  'public_metrics_blacklist',
```
bartaz: If this file starts to be reused in different places I feel it may require some refactor.

It was developed for market page form and has a hardcoded list of market page inputs. Now we added other inputs here that we moved to a different page. So now we have a 'generic' form state library that contains list of fields from 2 different forms.

Maybe `state` could stay generic but it would receive `allowedKeys` and `transform` functions as parameters when created/initialized? This way shared functionality would stay here while form dependant list of the fields would be separated.

static/js/publisher/form/multiselect.js
```diff
@@ -0,0 +1,376 @@
...
+      showSearch: true,
+      highlightedOption: highlighted
+    });
+  }
+
+  /**
+   * Update the original input value and dispatch a change event to the input and form.
+   */
+  updateInputValue() {
+    this.props.input.setAttribute('value', this.state.selected.map(item => item.key).join(', '));
```
bartaz: Wouldn't `this.props.input.value  = ...` do the same thing? And additionaly it should also trigger change event. There really shouldn't be a need to trigger change event manually on a form.

static/js/publisher/form/multiselect.js
```diff
@@ -0,0 +1,376 @@
...
+        case 'ArrowUp':
+          if (highlighted > 0) {
+            highlighted -= 1;
+          }
+          break;
+        case 'Enter':
+          event.preventDefault();
+          if (highlighted >= 0) {
+            this.addItem({
+              target: {
```
bartaz: See the comment about `removeItem`. Please separate `addItem(key)` and `addItemClick(event)` functions not to duct tape fake event object.

static/js/publisher/settings.js
```diff
@@ -0,0 +1,40 @@
+function enableInput(type) {
```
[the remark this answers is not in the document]
Lukewh: I'll create a new issue for this... This PR has gone wildly out of control considering it was to add a simple blacklist/ whitelist functionality and we've ended up with an entirely new page.

static/js/libs/arrays.js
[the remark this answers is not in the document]
bartaz: @Lukewh if `arraysEquals` is meant only to check if 2 arrays contain same elements (and from its usage it seems it is) it should never change the arrays given as params. If sorting is required to allow checking for items in any order it should be done on clones of original arrays, so that objects given to the function would not be affected by it.

static/js/publisher/form/multiselect.js
Lukewh: aah the magical `new Event('change', { 'bubbles': true });` :+1: thanks

static/js/libs/arrays.js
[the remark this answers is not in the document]
Lukewh: of course, sorry baby brain.

templates/publisher/settings.html
Lukewh: Renamed.

The reason this is here is that is shows/hides the custom radio's based on what's selected. I didn't want to add yet more to the marketForm js.

I agree this should be refactored at a later date, now we have 2 separate forms using the same code.... Probably a good time to reactify it too.

static/js/publisher/market/state.js
Lukewh: See above re. refactor

static/js/publisher/settings.js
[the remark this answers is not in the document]
Lukewh: https://github.com/canonical-websites/snapcraft.io/issues/1125

static/js/libs/arrays.js
```diff
@@ -0,0 +1,26 @@
+function arraysEqual(oldArray, newArray) {
```
bartaz: Just add a comment explaining what the function does (especially to mention that arrays containing same elements in different order are considered equal).

static/js/publisher/form.js
```diff
@@ -1,6 +1,7 @@
-import { initSnapScreenshotsEdit } from './screenshots';
+import { initSnapScreenshotsEdit } from './market/screenshots';
```
bartaz: This file got quite confusing now. While it has some bits that are common to both forms it has also a lot of things for one or another - like unnecessary screenshots on admin page or unnecessary whitelist multiselect on market page.

But let's handle it in the refactor task.

static/js/publisher/form/multiselect.js
```diff
@@ -0,0 +1,355 @@
+import React  from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+
+class MultiSelect extends React.Component {
+  constructor(props) {
+    super(props);
+
+    // hide the original input
+    this.props.input.style.display = 'none';
```
bartaz: Ideally I would say that React components should not touch DOM directly by themselves. Things like that would be better off in some 'init' script that would read the data from input, hide it, and pass the data into React component.

The `init` function down below this class seems like good place to do it.

If it complicates things too much now I'm fine with leaving it for refactor PR which should clear up some things.

bartaz:
### QA

It seems that country dropdown is not focused/selected by default when the page first loads which makes it harder to enter something with keyboard only.

To reproduce:
1. go to settings page of a snap that doesn't have any custom countries set
2. click on 'custom' checkbox
3. start typing contry name
4. dropdown shows up but no country is selected, so even if only one country is in dropdown it cannot be added with `enter`
5. you have to push arrow down to focus/select show up in the dropdown

screen shot 2018-09-24 at 14 08 36

It would be good to have one value in dropdown selected by default. It seems to work this way after one item gets manually selected.

bartaz:
### QA

Tabbing out of multiselect dropdown doesn't make it look like it looses focus.

1. Go into multiselect dropdown
2. start typing
3. hit `Tab`
4. multiselect looses focus, but it still looks focused

screen shot 2018-09-24 at 14 23 31

If multiselect looses focus (either by mouse or keyboard) it should visually loose focus as well.

bartaz:
### QA

Can't get back to 'All teritories'.

1. Select some territories in 'Only these countries'
2. Save
3. Click 'All teritories'
4. Save
5. It says that changes were saved correctly, but nothing changes

screen shot 2018-09-24 at 14 33 46

bartaz:
### QA

Can't change from 'Only' territories to 'Exclude' territories

1. Add some territories to 'only' list
2. Save
3. Switch to 'Exclude' territories and add some
4. Save
5. Key 'extra' error

screen shot 2018-09-24 at 14 34 45

static/js/publisher/form/multiselect.js
Lukewh: Moved!

static/js/libs/arrays.js
Lukewh: Done

bartaz:
### QA

There is no snap name in browser page title:

screen shot 2018-09-25 at 15 31 55

bartaz:
### QA

KeyError: 'extra' when trying to change from 'Only these' to 'Exclude' with keyboard only.

- go to settings page
- set some 'only these territories'
- Save
- Only using keyboard tab into territories radio button, change to 'Exclude' territories
- Type some territories
- Shift-Tab to 'Save' button
- Save
- You get a key error, nothing is saved

screen shot 2018-09-25 at 15 36 22

therealjuan:
- Can we change `Custom list` to `Selected territories` ?
- Can we include some padding / margin between `All territories | Selected territories` and likewise between `[ ] Only for...` and `[ ] Excluse these...` with the super select tag component?
- Would it be much work to include a `Clear list` action on the top right of the text field with more than 2 or 3 values` [ ] Only these territories` (left) `Clear list` (right)?

*Note:* There is a territory called [Christmas Island](https://en.wikipedia.org/wiki/Christmas_Island)

You can go quite far with the component...
image

therealjuan:
LGTM

PR #1095, opened 2018-09-14
Logout on sso failure

### LINKED ISSUE #1090: Macaroon refresh crashes if SSO declines
### Expected behaviour
Changing my password in SSO should result in me being logged out of snapcraft.io rather than a crash.

1. Log into snapcraft.io.
2. Change SSO password.
3. Wait for snapcraft.io's macaroon to need a refresh.
4. Visit snapcraft.io, observe 500 from `get_refreshed_discharge`: `KeyError: 'discharge_macaroon'`.

SSO may decline to refresh a discharge macaroon, for example if the user's password has changed. `get_refreshed_discharge` crashes in this case, as the response from `https://login.ubuntu.com/api/v2/tokens/refresh` will be a 401 and won't have a `discharge_macaroon` key.

tbille:
# Summary

Fixes #1090
Fixes https://sentry.io/canonical-bm/snapcraftio/issues/683010199/?query=is:unresolved

Improve requests to get/recharge the macaroon
Improve exception handling
Use same pattern as other api services that we have

# QA

- Log into snapcraft.io.
- Change SSO password.
- Wait for snapcraft.io's macaroon to need a refresh.
- You should be redirected to `/logout`


webapp/api/sso.py
```diff
@@ -0,0 +1,47 @@
...
+def process_response(response):
+    try:
+        body = response.json()
+    except ValueError as decode_error:
+        api_error_exception = ApiResponseDecodeError(
+            "JSON decoding failed: {}".format(decode_error)
+        )
+        raise api_error_exception
+
+    if not response.ok:
```
wgrant: It's usually best to check that the response code is what you expect before you try to decode the response body, or you're going to get an uninformative JSON decode failure when you get a non-JSON error, and you'll have no idea what the response code was. response.ok is also probably not what you want, since that is true if the response is a redirect, for example.

webapp/login/views.py
```diff
@@ -26,7 +26,11 @@ def login_handler():
     if authentication.is_authenticated(flask.session):
         return flask.redirect(open_id.get_next_url())
 
-    root = authentication.request_macaroon()
+    try:
+        root = authentication.request_macaroon()
+    except Exception:
```
wgrant: This is extremely broad. It will, for example, log you out if there's a brief network glitch. Perhaps limit the redirect to when you get a 401 from SSO?

webapp/api/sso.py
tbille: makes sense
Done

webapp/login/views.py
tbille: Now redirecting to an error page on anything else than `401` otherwise logout

webapp/publisher/snaps/views.py
```diff
@@ -28,11 +28,19 @@
...
 def refresh_redirect(path):
-    macaroon_discharge = authentication.get_refreshed_discharge(
-        flask.session["macaroon_discharge"]
-    )
-    flask.session["macaroon_discharge"] = macaroon_discharge
+    try:
+        macaroon_discharge = authentication.get_refreshed_discharge(
+            flask.session["macaroon_discharge"]
+        )
+    except ApiResponseError as api_response_error:
```
Lukewh: You haven't imported `ApiResponseError` so it fails.

webapp/publisher/views.py
```diff
@@ -18,11 +18,19 @@
...
 def refresh_redirect(path):
-    macaroon_discharge = authentication.get_refreshed_discharge(
-        flask.session["macaroon_discharge"]
-    )
-    flask.session["macaroon_discharge"] = macaroon_discharge
+    try:
+        macaroon_discharge = authentication.get_refreshed_discharge(
+            flask.session["macaroon_discharge"]
+        )
+    except ApiResponseError as api_response_error:
```
Lukewh: Same again, no import.

tbille:
> Is there any way to generalise all of these try excepts? Seems like we have to update them all over the place if things change?

We can think about it in another PR

Lukewh:
> > Is there any way to generalise all of these try excepts? Seems like we have to update them all over the place if things change?
> 
> We can think about it in another PR

Yeah, I didn't mean here. Maybe for a christmas special of refactoto.

tbille:
@Lukewh a little bird told that a new season might arrive

PR #333, opened 2018-02-22
Add both Developer contact and website fields

Lukewh:
# Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/317
- Added both contact and website fields to market page
- Added both fields (if present) to snap details page

# QA

- Pull this branch
- `./run`
- Visit http://0.0.0.0:8004/account/snaps/SNAP_NAME/market
- Update the Developer contact and Developer website fields.
- Save your changes and ensure the change have persisted
- Go to http://0.0.0.0:8004/SNAP_NAME
- Make sure the Developer website and Developer contact link go to the correct places
- Repeat if you fancy it

# Note

Apologies if there are designs/ wireframes for any of this that I've ignored - point me in the right direction and I'll update the PR thanks!

# Screenshots

Snap market page:
screen shot 2018-02-27 at 13 38 01

Snap details page:
screen shot 2018-02-27 at 13 38 21

PR #1105, opened 2018-09-18
Fixed styling for global nav

therealjuan:
## QA

1. Pull the branch or open the demo link
2. Open `https://0.0.0.0:8004/` and `https://ubuntu.com` on 2 different tabs
3. Cycle between the 2 and notice no style difference

### DIFF
--- a/static/sass/_snapcraft_p-navigation.scss
+++ b/static/sass/_snapcraft_p-navigation.scss
@@ -2,8 +2,10 @@
   $navigation-border-color: rgba(255, 255, 255, .2);
   $navigation-hover-color: rgba(0, 0, 0, .92);
   $navigation-highlight-color: rgba(250, 90, 64, 1);
+  $navigation-background-color: rgba(51, 51, 51, 1);
 
   .p-navigation {
+    background: $navigation-background-color;
     z-index: 100; // appear over channel map
 
     // FIXME: workaround for https://github.com/vanilla-framework/vanilla-framework/issues/1892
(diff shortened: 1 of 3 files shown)


static/sass/_snapcraft_p-navigation.scss
```diff
@@ -2,8 +2,10 @@
   $navigation-border-color: rgba(255, 255, 255, .2);
   $navigation-hover-color: rgba(0, 0, 0, .92);
   $navigation-highlight-color: rgba(250, 90, 64, 1);
+  $navigation-background-color: rgba(51, 51, 51, 1);
```
Lukewh: Can we set `$nav-bg-color` to `#333` and remove the `background: $navigation-background-color` below?

`#333` is easier to understand than rgba and We should fall back to vanilla where we can. We're overriding a variable with a variable here - seems counter-intuitive.

bartaz:
I feel like spacing from ubuntu.com doesn't work as well in snapcraft header:

screen shot 2018-09-18 at 12 00 25

Canonical logo seems misaligned (too much left spacing) - because on ubuntu.com it's aligned to 'ubuntu' text in the logo. Here it uses the same spacing but it's just a random place above the snapcraft logo bird.

'We are hiring' is aligned to the right, but it looks misaligned compared to Developer account logo below it.

static/sass/_snapcraft_p-navigation.scss
therealjuan: LGTM!

PR #1107, opened 2018-09-18
Enable feature flag for canonical users

tbille:
# Summary

Add in session `features_enabled` for canonical users. This will help us enable feature flags only for canonical users.

# QA

- `./run`
- Login
- Check if the session has the feature_enabled object
- Try with a non canonical email address

### DIFF
--- a/webapp/login/views.py
+++ b/webapp/login/views.py
@@ -47,6 +47,13 @@ def after_login(resp):
     if not resp.nickname:
         return flask.redirect(LOGIN_URL)
 
+    if resp.email.endswith("ubuntu.com") or resp.email.endswith(
+        "canonical.com"
+    ):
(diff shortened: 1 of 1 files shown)


Lukewh:
"Check if the session has the feature_enabled object"

how?

tbille:
@Lukewh on any endpoint add the line `print(session['feature_enabled'])` and check if it's true or false (true for canonical users and false otherwise)

Lukewh:
> @Lukewh on any endpoint add the line `print(session['feature_enabled'])` and check if it's true or false (true for canonical users and false otherwise)

`name session is not defined`

bartaz:
We talked with store team that this may not be needed for now. We should just be more open to share demo server URLs with other teams for testing some new features.

PR #1121, opened 2018-09-19
Enable access log and remove sentry dependency

tbille:
# Summary

- enable access log on talisker.gunicorn
- Remove sentry dependency that was overwritting talisker"s configuration 

# QA

- `./run`
- Navigate in the website and make sure there is more info on the console

### DIFF
--- a/entrypoint
+++ b/entrypoint
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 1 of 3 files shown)


WillMoggridge:
Looks like the tests are filing missing `blinker` module that used to be installed with `raven[flask]`

nottrobin:
Great work @tbille! Thanks for your help @bloodearnest =)

PR #1139, opened 2018-09-24
Rollup upgrade

bartaz:
Upgrades `rollup` to latest version to fix issues with rollup watch.

### QA

- ./run build
- everything should build properly as before
- please note there will be some 'circular dependency' warnings that rollup didn't show before
- ./run exec yarn watch-js
- rollup watch should run and should properly update on change in any JS file that is part of any bundle

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "raven-js": "^3.24.1",
     "react": "^16.4.1",
     "react-dom": "^16.4.1",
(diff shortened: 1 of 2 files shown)


Lukewh:
Updating `static/js/publisher/metrics/graphs/helpers.js` rebuilds `publisher.js`on first change but not on subsequent changes still, for me.

PR #1141, opened 2018-09-25
Add close channel endpoint

### LINKED ISSUE #1132: [backend] Closing a channel

tbille:
# Summary

Fixes #1132 
Add close channel endpoint 
https://dashboard.snapcraft.io/docs/api/snap.html#close-a-channel-for-a-snap-package

# QA

It's written in the same way as the realease endpoint

@bartaz you can try with the frontend you write

```
POST /SNAP_NAME/release/close-channel
{
  info to send to the API
}
```

### DIFF
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -495,18 +495,18 @@ def post_close_channel(snap_name):
     if not data:
         return flask.jsonify({})
 
-    if data["id"]:
-        snap_id = data["id"]
-    else:
-        return flask.jsonify({})
-
-    if data["info"]:
-        json = "info"
-    else:
-        return flask.jsonify({})
+    try:
+        snap_id = api.get_snap_id(snap_name, flask.session)
+    except ApiResponseErrorList as api_response_error_list:
+        if api_response_error_list.status_code == 404:
+            return flask.abort(404, "No snap named {}".format(snap_name))
+        else:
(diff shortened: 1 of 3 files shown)


webapp/publisher/snaps/views.py
```diff
@@ -499,6 +499,37 @@ def post_release(snap_name):
...
+        response = api.post_close_channel(flask.session, snap_id, json)
+    except ApiResponseErrorList as api_response_error_list:
+        if api_response_error_list.status_code == 404:
+            return flask.abort(404, "No snap named {}".format(snap_name))
+        else:
+            return flask.jsonify(api_response_error_list.errors), 400
+    except ApiError as api_error:
+        return _handle_errors(api_error)
+
+    return flask.jsonify(response)
```
bartaz: It seems that the success response from API doesn't have `success: true` in data (in similar way that release API does). It would be very helpful if we could add that on our side for successful responses, so they are compatible with release API.

tbille: ok I can add this field. But we should be careful, this is something we should request to the snap store team instead, especially if we proxy the API calls like this.

bartaz: Yes, makes sense to propose it to store API. Especially that release endpoint has that.

Still would be good to add it ourselves before it's added in the store. Thanks!

PR #1151, opened 2018-09-27
Show extra metadata

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1131

## QA

- Pull the branch
- Visit http://0.0.0.0:8004/<snap_name>/settings
- See extra information in the sticky bar

### DIFF
--- a/static/sass/_snapcraft_market.scss
+++ b/static/sass/_snapcraft_market.scss
@@ -70,5 +70,9 @@
         margin-top: $sp-x-small !important;
       }
     }
+
+    .p-inline-list__item {
+      line-height: 2.3rem;
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


matthewpaulthomas:
These three bits of data fit snugly alongside the “Revert” and “Save” buttons.

Unfortunately, because there are horizontal lines immediately above and below, it looks very much as if the buttons apply just to those three bits of data, when they don’t.

snapcraft io-1151-metadata

Since the page is not short of space, I suggest laying out this data in key-value columns below the lower line, just like the rest of the data on the same page. (It doesn’t matter that for now it’s read-only.)

(And I know the status values aren’t touched by this PR, but for consistency with the other data, “Status: published” should be “Status: Published”.)

matthewpaulthomas:
Much better, thanks.

PR #1153, opened 2018-09-28
Licenses

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1033
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1034

# QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/<snap_name>/settings
- Change the licenses
- See the results appear on http://0.0.0.0:8004/<snap_name> with multiple licenses listed as `license OR license`
- Try adding an expression to the custom box (as on the chromium snap)
- Save
- See the results appear on the public page

### DIFF
--- a/templates/publisher/listing.html
+++ b/templates/publisher/listing.html
@@ -327,13 +327,25 @@
 
         <div class="row">
           <div class="col-2">
-            <label for="license">License</label>
+            <label>License</label>
           </div>
           <div class="col-8">
-            <div class="js-license p-multiselect__wrapper">
-              <input type="text" class="js-multiselect-input" name="license" value="{{ license }}" />
+            <input type="hidden" name="license" value="{{ license }}" />
+            <input type="radio" name="license-type" value="simple" id="license-simple-option" {% if license_type == 'simple' %}checked="checked"{% endif %} />
+            <label for="license-simple-option">Simple</label>
+            <div class="js-license p-multiselect__wrapper" {% if license_type != 'simple' %}style="display: none;"{% endif %} id="license-simple-input">
+              <input type="text" class="js-multiselect-input" name="license-simple" value="{% if license_type == 'simple' %}{{ license }}{% endif %}" />
               <div class="js-multiselect-holder"></div>
+              <p class="p-form-help-text">The license(s) under which you will release your snap.<br />Multiple licenses can be selected to indicate alternative choices.</p>
             </div>
+            <input type="radio" name="license-type" value="custom" id="license-custom-option" {% if license_type == 'custom' %}checked="checked"{% endif %} />
+            <label for="license-custom-option">Custom SPDX expression</label>
+            <textarea
+              type="text"
+              id="license-custom-input"
+              name="license-custom"
+              {% if license_type != 'custom' %}style="display: none;"{% endif %}
+            >{% if license_type == 'custom' %}{{ license }}{% endif %}</textarea>
           </div>
         </div>
       </section>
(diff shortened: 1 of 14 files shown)


tbille:
@Lukewh https://snapcraft.io/chromium

bartaz:
As licences are visible on public details page I would expect them to be part of listing page rather then 'Settings' page which seems to have more 'admin' or 'non-user-visible' settings.

webapp/publisher/snaps/views.py
```diff
@@ -235,6 +235,16 @@ def get_listing_snap(snap_name):
...
     ]
 
+    licenses = []
+    for license in get_licenses():
+        licenses.append({"key": license["licenseId"], "name": license["name"]})
+
+    license = snap_details["license"]
+    license_type = "custom"
+
+    if "AND" not in license.upper() and "WITH" not in license.upper():
```
bartaz: Not sure if this may be the case right now, but if any license name/code contains 'AND' or 'WITH' this check would fail.

Lukewh: hmm, there are some licenses that include "WITH", good catch. a bit hackey but changing this to `if " AND " ... and " WITH "...` should fix this potential issue

templates/publisher/listing.html
```diff
@@ -318,6 +318,36 @@
...
+        </div>
+
+        <div class="row">
+          <div class="col-2">
+            <label>License</label>
+          </div>
+          <div class="col-8">
+            <input type="hidden" name="license" value="{{ license }}" />
+            <input type="radio" name="license-type" value="simple" id="license-simple-option" {% if license_type == 'simple' %}checked="checked"{% endif %} />
+            <label for="license-simple-option">Simple</label>
```
bartaz: I'm not sure if 'Simple' is a good name for this. Do we need to name this checkbox at all?

screen shot 2018-10-02 at 13 05 17

templates/publisher/listing.html
```diff
@@ -318,6 +318,36 @@
...
+            <div class="js-license p-multiselect__wrapper" {% if license_type != 'simple' %}style="display: none;"{% endif %} id="license-simple-input">
+              <input type="text" class="js-multiselect-input" name="license-simple" value="{% if license_type == 'simple' %}{{ license }}{% endif %}" />
+              <div class="js-multiselect-holder"></div>
+              <p class="p-form-help-text">The license(s) under which you will release your snap.<br />Multiple licenses can be selected to indicate alternative choices.</p>
+            </div>
+            <input type="radio" name="license-type" value="custom" id="license-custom-option" {% if license_type == 'custom' %}checked="checked"{% endif %} />
+            <label for="license-custom-option">Custom SPDX expression</label>
+            <textarea
```
bartaz: Does it need to be a text area? Wouldn't a text input be enough?

screen shot 2018-10-02 at 13 05 29
Also it could be useful to add some hint text linking to what SPDX expressions are?

templates/publisher/listing.html
```diff
@@ -318,6 +318,36 @@
...
+              type="text"
+              id="license-custom-input"
+              name="license-custom"
+              {% if license_type != 'custom' %}style="display: none;"{% endif %}
+            >{% if license_type == 'custom' %}{{ license }}{% endif %}</textarea>
```
bartaz: It would be useful for custom expression to match the value of 'simple' input, not being cleared when 'simple' is chosen.

So user can start with 'simple', but extend it in custom field without starting from scratch. But it's fine to fix it later as separate issue.

bartaz:
Populating custom licence with initial value works, but would be nice to populate it on multiselect change as well. But can be done as improvement later.

templates/publisher/listing.html
```diff
@@ -350,13 +382,37 @@
...
+      'license': {{ license|tojson }}
     },
     {% if error_list %}
       {{ error_list|tojson}}
     {% else %}
       null
     {% endif %}
     );
+
+    var licenseRadioContent = [
```
bartaz: Would be nice to have this licence init code somewhere in the module instead of inline script. Would it work inside of `market.initForm`? I guess screenshots are initialized in there.

PR #1160, opened 2018-10-01
Header and footer update

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1157
Supercedes https://github.com/canonical-websites/snapcraft.io/pull/1149

## QA

- Pull branch
- `./run`
- Visit http://0.0.0.0:8004
- Ensure the global-nav and footer looks ok

### DIFF
--- a/static/sass/_global-nav.scss
+++ b/static/sass/_global-nav.scss
@@ -3,7 +3,8 @@
 
   .global-nav {
     align-items: center;
-    background: $color-navigation-background;
+    background: $color-dark;
+    border-bottom: 1px solid $color-navigation-background;
     display: flex;
     min-height: 2rem;
     position: relative;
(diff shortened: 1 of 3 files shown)


bartaz:
@Lukewh @lyubomir-popov Why the border addition?

Ubuntu.com doesn't seem to have a border on global nav, at seems it is not visible.

In here border is barely, but visible, especially when items in main nav are hovered or selected:
screen shot 2018-10-01 at 11 31 35

So it seems like the border color is different from global nav color, different from main nav color, different from main nav hover/select color.

Is this intended?

lyubomir-popov:
@bartaz good spot, I think you're right. What I was trying to do is prevent the merging of selected menu item and top strip. But we did this better on ubuntu.com and should just port it over from there, including the hover state colour.

Lukewh:
@bartaz removed the global-nav border

bartaz:
Looks good. But now it seems like header navigation colours were slightly changed here on snapcraft.io, which again leads to inconsistent navigation styles between snapcraft.io, docs and BSI.

lyubomir-popov:
@Lukewh the only thing I notice is that the top canonical bar is darker on snapcraft than on ubuntu.com

PR #1163, opened 2018-10-02
Add circuit breaker to the application

### LINKED ISSUE #1142: Allow maintenance mode to be set
### Expected behaviour
When the store is under maintenance, the snapcraft.io front-end should be able to be flipped into maintenance mode to a) help protect the upstream services b) advise users that their snaps are still there, even if they can't see them

### Specs

- _URL:_ https://snapcraft.io/store
- _Operating system:_ Ubuntu 18.04 LTS
- _Browser:_ Firefox 62.0.2 (64-bit)

tbille:
# Summary

Fixes #1142 
- Add circtuit breaker to the requests class. After a 100 failures the requests are blocked and returns an excpection `ApiCircuitBreaker` which when catched goes to a 503 page: 

screenshot_2018-10-02 maintenance mode enabled

# QA

- in `webapp/api/requests.py` modify `db_breaker = CircuitBreaker(fail_max=100)` to `db_breaker = CircuitBreaker(fail_max=5)`
- Disconnect from network
- `./run`
- reload 5 times page http://127.0.0.1:8004/toto
- you should be redirected after 5 times to the maintenance page :heart:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ gunicorn[gevent]
 humanize==0.5.1
 prometheus_client
 prometheus-flask-exporter==0.2.0
+pybreaker==0.4.4
 pycountry==17.9.23
 pymacaroons==0.12.0
 python-dateutil==2.6.1
--- a/webapp/api/requests.py
+++ b/webapp/api/requests.py
@@ -1,15 +1,16 @@
-# Core packages
 import os
 from urllib.parse import urlparse
 
-# Third-party packages
 import requests
-import requests_cache
+
 import prometheus_client
-
-# Local packages
-from webapp.api.exceptions import ApiTimeoutError, ApiConnectionError
-
+import requests_cache
+from pybreaker import CircuitBreaker, CircuitBreakerError
+from webapp.api.exceptions import (
+    ApiCircuitBreaker,
+    ApiConnectionError,
+    ApiTimeoutError,
+)
 
 timeout_counter = prometheus_client.Counter(
     "feed_timeouts", "A counter of timed out requests", ["domain"]
@@ -57,12 +58,15 @@ class BaseSession:
 
         headers = {"User-Agent": storefront_header}
         self.headers.update(headers)
+        self.api_breaker = CircuitBreaker(fail_max=5, reset_timeout=60)
 
     def request(self, method, url, **kwargs):
         domain = urlparse(url).netloc
 
         try:
-            request = super().request(method=method, url=url, **kwargs)
+            request = self.api_breaker.call(
+                super().request, method=method, url=url, **kwargs
+            )
         except requests.exceptions.Timeout:
             timeout_counter.labels(domain=domain).inc()
 
@@ -74,6 +78,10 @@ class BaseSession:
 
             raise ApiConnectionError(
                 "Failed to establish connection to {}.".format(url)
+            )
+        except CircuitBreakerError:
+            raise ApiCircuitBreaker(
+                "Requests are closed because of too many failures".format(url)
             )
 
         latency_histogram.labels(
--- a/templates/503.html
+++ b/templates/503.html
@@ -1,0 +1,16 @@
+{% extends webapp_config['LAYOUT'] %}
+
+{% block meta_title %}Maintenance mode enabled{% endblock %}
+
+{% block content %}
+<div class="p-strip">
+  <div class="row">
+    <div class="col-10">
+      <h1>We're experiencing issues or doing maintenance</h1>
+      <h2>Please try again later</h2>
+      <p>You can check the service status at <a href="https://status.snapcraft.io">https://status.snapcraft.io</a>.</p>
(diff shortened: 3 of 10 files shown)


Lukewh:
We should maybe say something a little less 'techy' and 'blamey'. The API might be working but in maintenance mode or only some endpoints might be down.

```
<h1>We're experiencing issues or doing maintenance</h1>
<h2>Please try again later</h2>
<p>You can check the service status at <a href="https://status.snapcraft.io">https://status.snapcraft.io</a>.</p>
<p>If the problem persists please check the Snapcraft <a href="https://forum.snapcraft.io">forum</a> or <a href="https://github.com/canonical-websites/snapcraft.io/issues/new">report an issue</a>.</p>
```

requirements.txt
```diff
@@ -7,6 +7,7 @@ gunicorn[gevent]
 humanize==0.5.1
 prometheus_client
 prometheus-flask-exporter==0.2.0
+pybreaker==0.4.4
```
nottrobin: Why did you decide on `pybreaker`? I thought you were going to use `circuitbreaker`? I know very little about this, but just from reading their PyPi pages, `circuitbreaker` looks like it is more mature and has more features.

webapp/api/requests.py
```diff
@@ -1,15 +1,18 @@
...
+import prometheus_client
+import requests_cache
+from pybreaker import CircuitBreaker, CircuitBreakerError
+from webapp.api.exceptions import (
+    ApiCircuitBreaker,
+    ApiConnectionError,
+    ApiTimeoutError,
+)
 
+db_breaker = CircuitBreaker(fail_max=100)
```
nottrobin: I think a failure threshold of `100` is insanely large. Too large to be of any practical use. Unless we have some indication that 5 or more consecutive failure are actually incredibly common? If so that would be very concerning.

I'd also suggest that for this circuit breaker to be useful, its failure threshold should be less than the `failureThreshold` of the Kubernetes `readinessProbe`. Otherwise pods will be brought out of rotation before the circuit breaker kicks in. We don't seem to have set that setting in [our configs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/), which means it must use the default `failureThreshold` of 3.

So maybe we should set this threshold to the default of 5, and then increase the Kubernetes threshold to a larger number? Or should we go the other way and set the circuit breaker to be super aggressive at a threshold of 2?

@jpmartinspt do you have any thoughts?

requirements.txt
tbille: I used this module because `circuitbreaker` looks more mature. But it actually doesn"t work. I played a bit arround with it and some features are not implemented or do not work... This module has more users that uses it and more devs on it.

webapp/api/requests.py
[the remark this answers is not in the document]
tbille: you are right this is a bad copy/paste

requirements.txt
nottrobin: @tbille good answer =)

webapp/api/requests.py
nottrobin: Good point, your right that simple connection failures shouldn't effect k8s.

But timeouts still might. The timeout in the app requests is what now? 2 seconds? 3? The timeout for the readinessprobe is 5. We could easily hit that if a route makes multiple API calls, or if the timeout isn't working properly for some reason, or if there are delays elsewhere in the app.

So I still think it's true that any failure solutions in the app itself should have lower thresholds than those at the Kubernetes layer, for robustness.

nottrobin: But 5 is fine, but I think we should up the k8s one.

nottrobin:
This appears to break after *any* 5 consecutive API calls fail, and will then return the error for *all* API calls. It even looks like it will break calls to dashboard.snapcraft.io, or admin.insights.ubuntu.com if api.snapcraft.io calls are failing, which is not at all good.

Also, pages that were gracefully handling connection failures, like `/store`, are no longer doing so. `/store` used to show you an empty search field in the absence of API data, but now it shows the error page instead.

So I think this needs 2 changes to pass QA:

- The breaker should check for consecutive failures of *the same API call*. If this is impossible, we could discuss whether it makes sense to break on consecutive failures of any API call to the same API hosts, but *at least* it should differentiate between API hosts.
- All pages that currently have graceful fallbacks for failing API calls should continue to do so.

Lukewh:
@nottrobin are those two points required for landing this branch? Could they be improvements to iterate on next iteration?

On the point regarding `/store` - if the API is down, why would we want to display an empty page with a search bar that won't work? That seems to say "this is kinda broken, and we might/ might not know about it" surely it's better to say "it's broken and we know it, bare with"? A better example would be `/blog`, but I don't think that should block this landing, as long as there's a follow up to improve it soon!

nottrobin:
:+1: confirmed, the 3 APIs (api.snapcraft.io, dashboard.snapcraft.io, admin.insights.ubuntu.com) now appear to circuit-break independently.

PR #1184, opened 2018-10-06
Use swiper for carousels

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1180
Fixes https://github.com/CanonicalLtd/snapcraft-design/issues/489

Partially supercedes https://github.com/canonical-websites/snapcraft.io/pull/1148

Use [Swiper](http://idangero.us/swiper) for the store categories carousels and snap details screenshots carousel, this means the two separate implementations of a carousel have been removed :+1:.

I copied/ pasted the CSS from Swipers `/src/components/core/core.less` file rather than including the CSS as an external file to keep the size down and remove unused styles.

I also included Swiper as ES modules in an attempt to keep the file size down (297kb using modules vs 412kb including the whole library).

Note: This PR does not include the new designs as they are yet to be signed off.

## QA

- Pull the branch
- `./run`
- Resize the browser and scroll through the category snaps
- Visit a snap with screenshots (Spotify)
- Scroll through the screenshots and resize the browser


static/js/public/snap-details/screenshots.js
```diff
@@ -1,5 +1,9 @@
 import lightbox from './../../publisher/market/lightbox';
 import { isMobile } from '../../libs/mobile';
+import { Swiper, Navigation } from 'swiper/dist/js/swiper.esm';
```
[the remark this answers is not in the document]
Lukewh: Just 'swiper' installs the entire library. The 'esm' version is the moduled version, so you can just include the modules you want.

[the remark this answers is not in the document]
bartaz: Cool, is it documented/recommended somewhere? Looks like something that can break if they change the file structure.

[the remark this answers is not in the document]
Lukewh: Yeah in the official docs: https://idangero.us/swiper/api/#custom-build

bartaz:
Not really a blocking issue, but quite visible sometimes:

Because it seems now it's swiper that scales the images, if image is larger then the size it will be later resized to by swiper it 'blinks' in a bigger size on page load before swiper script runs.

For example go to http://snapcraft.io-canonical-websites-pr-1184.run.demo.haus/spotify and reload the page (with cache disabled). Large images show up for a fraction of second before being scaled down by swiper.

Maybe we could limit the `max-height` of images or something with CSS so they don't jump in size until swiper loads?
Also, what happens if swiper/scripts don't load at all? Screenshots will be huge and unscrollable?

bartaz:
Yes, if script fails to load or swiper fails to run screenshots are huge and unscrollable:
screen shot 2018-10-08 at 16 05 02

Lukewh:
Good catch @bartaz updated

static/js/public/store-categories.js
```diff
@@ -10,10 +13,19 @@ function getCategory(holder) {
   const category = holder.dataset.category;
 
   // Write the html and init the carousel
-  const writeCategory = function(response) {
-    holder.innerHTML = response;
-    new Carousel(holder.querySelector('.p-carousel'));
-  };
+  const writeCategory = function() {
+    const _category = category;
```
bartaz: Why is this copy of `category` needed? Can't `category` used in places where `_category` is used now?

bartaz:
QA: Padding on screenshot make some things misaligned.

First of all 'fullscreen' icon doesn't have equal space from top and right:

screen shot 2018-10-09 at 09 56 38

Also, there is a padding from left side and it makes first screenshot not left aligned with the rest of the page:

screen shot 2018-10-09 at 09 56 50

static/sass/_patterns_carousel.scss
```diff
@@ -4,92 +4,184 @@
...
 
-      @media screen and (min-width: $breakpoint-large) {
-        right: -1.5rem;
-      }
+    @media screen and (min-width: $breakpoint-large) {
+      right: -1.5rem;
+    }
+  }
 
+  .swiper-container {
```
bartaz: If these are copied from swiper would be good to comment with a link to CSS file on GH or a path in project, so anyone knows where to find it.

static/sass/_patterns_maas_modal.scss
```diff
@@ -134,6 +134,7 @@ $color-modal-controls: rgba($color-x-dark, .2);
 
   .vbox-next {
     background-image: url('#{$assets-path}9716bbab-chevron.svg');
+    margin-right: 0 !important;
```
bartaz: Please add a comment explaining that it's here to override vanilla button margin.

static/js/public/store-categories.js
Lukewh: Good catch, it's not in a look or anything so doesn't need to be a closure :+1:

PR #1202, opened 2018-10-11
Add first snap flow tests

tbille:
Lol tests


bartaz:
Won't these break when we add any real content to the page?

tbille:
@bartaz isn't is the point of tests? :P

PR #1211, opened 2018-10-16
Removed notification bar on publisher pages

therealjuan:
## QA
Pull the branch or visit the demo link

1. Open snapcraft.io
2. Login
3. Make sure there is no notification appearing in regards to updates on T&Cs on the published snaps summary page, listing pages or account details


therealjuan:
Good point @anthonydillon so i just deleted the file and we can revert as you suggested if need be

PR #1215, opened 2018-10-17
Abort 503 when maintenance mode enabled

tbille:
# Summary

Add a way to switch on and off maintenance mode from a job that will enable it.

https://github.com/canonical-webteam/deployment-configs/pull/187

# QA

no qa for now, just starting a discussion

### DIFF
--- a/webapp/blog/views.py
+++ b/webapp/blog/views.py
@@ -1,3 +1,4 @@
+import os
 import flask
 
 import webapp.api.blog as api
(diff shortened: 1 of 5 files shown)


nottrobin:
Since, as discussed, this is hopefully a stopgap measure until we have a more capable caching & redundancy layer, could we err on the side of being more minimal with these maintenance mode flags?

Would you mind just adding `MAINTENANCE_MODE_PUBLISHER` for now? I don't quite understand why `MAINTENANCE_MODE_LOGIN` needs to be separate, couldn't they be handled by the same variable? Aren't they both using the same API?

Making use of the [`old_data_on_error` option in requests_cache](https://requests-cache.readthedocs.io/en/latest/api.html) should take care of both the Store and the Blog in almost all cases.

PR #1219, opened 2018-10-18
Add 50X error template

### LINKED ISSUE #1181: Create generic 5xx page

bartaz:
Fixes #1181

Adds a template to 500 and 504 errors.

### QA

- ./run or demo
- try to get Gateway timeout or internal server error (may require forcing some error locally)
- styled error should appear instead of plain HTML one

screen shot 2018-10-18 at 15 29 41

### DIFF
--- a/webapp/handlers.py
+++ b/webapp/handlers.py
@@ -52,9 +52,17 @@ def set_handlers(app):
 
         return flask.render_template("404.html", error=error.description), 404
 
+    @app.errorhandler(500)
+    def internal_error(error):
+        return flask.render_template("50X.html", error=error), 500
+
     @app.errorhandler(503)
     def service_unavailable(error):
         return flask.render_template("503.html"), 503
+
+    @app.errorhandler(504)
+    def gateway_timeout(error):
+        return flask.render_template("50X.html", error=error), 504
 
     # Global tasks for all requests
     # ===
--- a/templates/50X.html
+++ b/templates/50X.html
@@ -1,0 +1,17 @@
+{% extends webapp_config['LAYOUT'] %}
+
+{% block meta_title %}{{ error.name }}{% endblock %}
+
+{% block content %}
+<div class="p-strip">
+  <div class="row">
+    <div class="col-10">
+      <h1>{{ error.name }}</h1>
+      <h2>Please try again later</h2>
+      <p>{{ error.description }}</p>
+      <p>You can check the service status at <a href="https://status.snapcraft.io">https://status.snapcraft.io</a>.</p>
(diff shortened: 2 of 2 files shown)


webapp/handlers.py
```diff
@@ -52,10 +52,18 @@ def page_not_found(error):
 
         return flask.render_template("404.html", error=error.description), 404
 
+    @app.errorhandler(500)
```
tbille:
```suggestion
    @app.errorhandler(502)
    def internal_error(error):
        return flask.render_template("50X.html", error=error), 502
```

tbille: we don't return 500 errors, only 502 `Bad Gateway`

bartaz: Even if there is internal server error (typo in backend code, etc)?

bartaz: @tbille It is possible to get 500 (if something is messed in the code):

screen shot 2018-10-18 at 16 10 32

Sure, we will catch most of there during development or review/QA, but it is possible that some code would fail internally on prod.

bartaz: Can the same handler be used for multiple codes?

```python
@app.errorhandler(500)
@app.errorhandler(501)
@app.errorhandler(502)
def error_handler(error):
   return flask.render_template("50X.html", error=error), error.code
```

nottrobin: Yes you can, and please do. You should definitely catch `500` as well as `502` and any [other `5xx` error](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_errors) that might possibly happen. I would suggest you at least catch `500` to `505`.

Lukewh:
Can this get another review please @tbille @bartaz @nottrobin

webapp/handlers.py
```diff
@@ -52,6 +52,14 @@ def page_not_found(error):
...
+    @app.errorhandler(501)
+    @app.errorhandler(502)
+    @app.errorhandler(504)
+    @app.errorhandler(505)
+    def internal_error(error):
+        return flask.render_template("50X.html", error=error), error.code
```
nottrobin: You're expecting an `HTTPError` width `name` and `code` properties, but it's really quite likely to be any random Python error, like a `ValueError` or `DatabaseError`, which won't have these properties.

I suggest you use a function like this, which caters for any kind of error:

``` python3
    @app.errorhandler(500)
    @app.errorhandler(501)
    @app.errorhandler(502)
    @app.errorhandler(504)
    @app.errorhandler(505)
    def internal_error(error):
        error_name = getattr(error, "name", type(error).__name__)

        return_code = getattr(error, "code", 500)

        return (
            flask.render_template("50X.html", error_name=error_name),
            return_code,
        )
```

PR #1257, opened 2018-10-30
Filter revisions table based on arch clicked

### LINKED ISSUE #1225: Clickable unassigned revisions

bartaz:
Fixes #1225 

Adds clickable 'unassigned revisions' to release table, where users can choose to select revisions from popover with revisions filtered by architecture.

### QA

- ./run or https://snapcraft-io-canonical-websites-pr-1257.run.demo.haus/
- 'Unassigned revisions' row show be visible in releases table
- Click on any '+ Add revision'
- Popover should show up with latest revisions from given architecture
- You can select a revision from the list, it will show in the releases table, ready to be promoted to some channel

screen shot 2018-11-02 at 08 07 25


Lukewh:
There seems to be no styling when there are no releases

screenshot_2018-10-31 releases and revision history for test-lukewh

Lukewh:
The popover table could have a better ratio for the columns (on Firefox)

screenshot_2018-10-31 releases and revision history for lukewh-test

therealjuan:
Comments from review:

- Make table of revisions available expandable with `Show all` action link
- Add margins to top and bottom of channel map table
- Look at empty releases (separate design task)
- Unassigned revisions to be decoupled from the channel map table

bartaz:
@therealjuan Can you have a look at this. Functionality of filtered table is implemented, but I guess it could use some better visual clue to show what is currently 'selected' cell and where the chosen revision would appear. Especially when we later allow clicking on channel cells to show history, the list below the table needs some good visual clue that it's connected to the cell clicked.

static/js/publisher/release/releasesController.js
```diff
@@ -461,6 +490,20 @@ export default class ReleasesController extends Component {
       }
     );
 
+    let filteredRevisions = this.props.revisions;
+    let title = null;
+    let filters = this.state.revisionsFilters;
+
+    if (filters && filters.arch) {
+      title = "Latest revisions";
```
Lukewh: Could this be the initialisation value at line 494?

bartaz: No, if there are no filters we want title to be `null`, so `RevisionsList` will use its default value 'Available revisions'. But I'm thinking about removing title 'logic' from `RevisionsList` altogether, so it will hopefully make more sense in some of the next iterations of this code.

therealjuan:
Can we add some extra spacing between the `X` and `show available...`?

image

Otherwise LGTM 👍

PR #1263, opened 2018-11-01
Fsf ga snap

Lukewh:
Add a google analytics event for the snap name created during fsf.

### DIFF
--- a/static/js/public/first-snap-flow.js
+++ b/static/js/public/first-snap-flow.js
@@ -124,17 +124,6 @@ function push() {
     clearTimeout(timer);
 
     getSnapCount(data => {
-      if (data.snaps.length !== initialSnaps.length) {
-        const newSnaps = newArrayValue(initialSnaps, data.snaps);
-        if (newSnaps.length > 0 && typeof ga !== "undefined") {
-          ga("gtm1.send", {
-            hitType: "event",
-            eventCategory: "First Snap Flow",
-            eventAction: "Snap pushed",
-            eventLabel: `${newSnaps.join(",")}`
-          });
-        }
-      }
       if (initialCount === null) {
         initialCount = data.count;
         initialSnaps = data.snaps;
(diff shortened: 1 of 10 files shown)


static/js/public/first-snap-flow.js
```diff
@@ -44,60 +43,113 @@ function install(language) {
...
+      }
       if (initialCount === null) {
-        initialCount = count;
-      } else if (count !== initialCount) {
+        initialCount = data.count;
+        initialSnaps = data.snaps;
+      } else if (data.count !== initialCount) {
+        const newSnaps = newArrayValue(initialSnaps, data.snaps);
+
+        if (newSnaps.length > 0 && typeof ga !== "undefined") {
```
bartaz: Why is the same condition here as in the line 129? Seems like we send same ga event twice?

static/js/public/first-snap-flow.js
```diff
@@ -1,39 +1,38 @@
-/* globals ClipboardJS */
+/* globals ClipboardJS, ga */
 
-import 'whatwg-fetch';
+import "whatwg-fetch";
```
bartaz: Aaaa, prettier changes again...

static/js/public/first-snap-flow.js
Lukewh: Removed, was for my testing

static/js/public/first-snap-flow.js
Lukewh: :+1: this is the only branch that should be affected, so once these changes are made, we should be good

static/js/public/first-snap-flow.js
```diff
@@ -84,22 +84,61 @@ function getSnapCount(cb) {
   fetch("/snaps/api/snap-count")
     .then(r => r.json())
     .then(data => {
-      cb(data.count);
+      cb(data);
     });
 }
 
+function newArrayValue(arr1, arr2) {
```
bartaz: Not sure exactly if I read this function correctly, but could we name it more clearly to what it does? Something like `getArrayDiff` (if I get it right)?

Lukewh: :+1: I had a complete mind blank when naming this function, thanks. Will change it

PR #1264, opened 2018-11-02
Rename release endpoint to releases

tbille:
# Summary

Rename release to releases

# QA

- `./run`
- http://localhost:8004/toto/release should redirect to http://localhost:8004/toto/releases
- http://localhost:8004/toto/release updates should still works

### DIFF
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -458,6 +458,7 @@ def post_listing_snap(snap_name):
 
 
 @publisher_snaps.route("/account/snaps/<snap_name>/release")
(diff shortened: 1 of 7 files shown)


webapp/publisher/snaps/views.py
```diff
@@ -495,14 +496,15 @@ def get_release_history(snap_name):
 
 
 @publisher_snaps.route("/account/snaps/<snap_name>/release", methods=["POST"])
```
bartaz: Do we need redirect for POST? I don't think anyone would have that bookmarked or linked, and if we update internal URL it should be fine.

Of course, assuming we update our links to `/release` in HTML and API/POST URL in JavaScript.

PR #1281, opened 2018-11-07
Add global ga-tracking

Lukewh:
Fixes https://github.com/canonicalltd/snap-squad/issues/759

## Done

Added generic tracking as per @caldav spec.

## QA

- Pull the branch
- Add `console.log(category, from, to, label);` to `static/js/base/ga.js:24`
- `./run`
- Open http://0.0.0.0:8004 and open the console, enable "Persistent logs"
- Browse around the site, seeing that the correct 'events' are being triggered

### DIFF
--- a/static/js/base/ga.js
+++ b/static/js/base/ga.js
@@ -20,41 +20,47 @@ const events = {
   ".p-strip a": "content-link"
 };
 
-window.addEventListener("click", function(e) {
-  let target = e.target.closest("a");
-  if (!target) {
-    target = e.target.closest("button");
+function triggerEvent(category, from, to, label) {
+  if (dataLayer) {
+    dataLayer.push({
+      event: "GAEvent",
+      eventCategory: `${categoryPrefix}${category}`,
+      eventAction: `from:${origin} to:${to}`,
+      eventLabel: label,
+      eventValue: undefined
+    });
   }
+}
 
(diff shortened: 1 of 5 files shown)


static/js/base/ga.js
```diff
@@ -0,0 +1,60 @@
+/* global dataLayer */
+const origin = window.location.href;
+const categoryPrefix = "snapcraft.io-";
+
+const events = {
```
caldav: We should probably kill the current events on details pages since they are not sent to the combo tracker therefore don't allow for viewing cross site funnels. I'll adjust the existing reports depending on them.

So, my suggestion would be to track clicks on: 
`p-view-store-button` (action: `from:{origin} to:{snap://package-name}`)
`p-snap-install-buttons__versions` (action: `from:{origin} to:{the content of "data-control"}`)
`p-snap-install-buttons__install` (action: `from:{origin} to:{the content of "data-control"}`)

Thoughts?

Lukewh: Makes sense, should they be `cta-1` categories I guess?

caldav: Looking at it from a user perspective, cta-0 (primary call to action) for "p-snap-install-buttons__install" and cta-1 for the others.

caldav:
Final updates look good

PR #1286, opened 2018-11-08
Shows release history when channel/arch cell is clicked in release UI

### LINKED ISSUE #1266: Show channel release history

bartaz:
Fixes #1266 

Shows release history when channel/arch cell is clicked in release UI

### QA

- ./run or https://snapcraft-io-canonical-websites-pr-1286.run.demo.haus/
- click on any channel/arch cell
- history of releases for this channel in this architecture should show in table below

screen shot 2018-11-08 at 15 43 13

### DIFF
--- a/static/js/publisher/release/revisionsList.js
+++ b/static/js/publisher/release/revisionsList.js
@@ -11,7 +11,7 @@ export default class RevisionsList extends Component {
     this.props.selectRevision(revision);
   }
 
-  renderRows(revisions, showCheckboxes) {
+  renderRows(revisions, isHistory) {
     return revisions.map(revision => {
       const revisionDate = revision.release
         ? new Date(revision.release.when)
@@ -23,7 +23,7 @@ export default class RevisionsList extends Component {
       // disable revisions from the same architecture that already selected
       // but only if checkboxes are visible (not in channel history)
       const isDisabled =
-        showCheckboxes &&
+        isHistory &&
         !isSelected &&
         revision.architectures.some(
           arch =>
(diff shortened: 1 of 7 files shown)


therealjuan:
Probably we would like to show the current available release on a channel (i.e. in this case 9572 on the list):

image

It would be good to have a similar hover state with a cursor pointer on unassigned and release history of a channel:
image

bartaz:
@therealjuan I addressed the issues you mentioned.

Hover state added.

For the revisions/releases not showing in history - I removed a 100 size restriction from the API (so we get 500 past releases by default). This should cover most (if not all) current snaps to have at least currently released revisions showed properly. We should think about how to address it in the future.

bartaz:
'Waiting for push' button has hover state and cursor even that it cannot be clicked.

screen shot 2018-11-09 at 17 05 39

Could we make it disabled?

static/js/publisher/release/revisionsList.js
```diff
@@ -11,35 +11,46 @@ export default class RevisionsList extends Component {
...
+        ? new Date(revision.release.when)
+        : new Date(revision.created_at);
       const isSelected = this.props.selectedRevisions.includes(
         revision.revision
       );
+
+      // disable revisions from the same architecture that already selected
+      // but only if checkboxes are visible (not in channel history)
       const isDisabled =
+        showCheckboxes &&
```
Lukewh: is it worth using a different variable name that's more descriptive like "inChannelHistory" that both the checkboxes and this can use?

static/js/publisher/release/revisionsList.js
```diff
@@ -11,35 +11,46 @@ export default class RevisionsList extends Component {
...
-              id={`revision-check-${revision.revision}`}
-              onChange={this.revisionSelectChange.bind(this, revision)}
-            />
-            <label
-              className="u-no-margin--bottom"
-              htmlFor={`revision-check-${revision.revision}`}
-            >
-              {revision.revision}
-            </label>
+            {showCheckboxes ? (
```
Lukewh: again `inChannelHistory` or some such?

therealjuan:
Some comments:

- I was wondering if we can make the whole row of a revision clickable to toggle the checkbox from the row. Right now you can click the check box or revision number but since that is the only action you can do in the row it can be quite useful
- Maybe might be useful to showcase the proposed revision to release as part of the revision history and add it as a highlighted item. WDYT?
- Probably might need to work on some designs updated for the summary log to showcase other information more succintly (that's on me)

therealjuan:
Clicking the whole row makes a lot of sense. Probably we need to have another review with Lyubo once he is back to make sure that it makes sense.

For the remaining items, yes we can work them out as a separate task so you can get on with the visual updates.

Issue #1222, opened 2018-10-18
When creating a link to another post, one has to choose either the Ubuntu blog or the Snapcraft blog


kyrofa:
I have multiple blog series on both the Ubuntu blog as well as the Snapcraft blog. There is numerous inter-linking, there (a table of contents, or "as discussed in part1...", or "in the next part, we'll discus..."). The URLs of all the links are using Insights, since that's all that existed when they were written. What this means is that, if any of the links are clicked from the Snapcraft blog, one is taken to the Ubuntu blog instead of staying on the series on the Snapcraft blog.

### Expected behaviour

There should be a way to create a link to another blog post where, if the link is clicked from snapcraft.io/blog, the visitor should see the blog post in snapcraft.io/blog. If the link is clicked from the Ubuntu blog, the visitor should stay on the Ubuntu blog.

### Steps to reproduce the problem

Create a new post in wordpress that links to a blog post in the Ubuntu blog, but one that is also available in the Snapcraft blog. Publish that post on both the Ubuntu blog as well as Snapcraft. Note that if you click the link from the Snapcraft blog,  you're taken to the Ubuntu blog.

matthewpaulthomas:
Example of the current URL structures:
- [snapcraft.io/blog/your-first-robot-the-driver-4-5](https://snapcraft.io/blog/your-first-robot-the-driver-4-5)
- [blog.ubuntu.com/2018/03/09/your-first-robot-the-driver-4-5](https://blog.ubuntu.com/2018/03/09/your-first-robot-the-driver-4-5)

You could use relative links to other posts, keeping visitors on the same site, if:
- blog.ubuntu.com URLs stopped including the date hierarchy,
  `<a href="../your-first-robot-the-controller-3-5">`, OR
- snapcraft.io/blog URLs started including the date hierarchy,
  `<a href="../../../../2018/02/11/your-first-robot-the-controller-3-5">`.

The second would be quite difficult to remember and to type correctly. So if it was left up to authors, it would be a tradeoff: the value of including dates in URLs as a hint of an article’s currency, vs. increased difficulty of creating relative links between posts.

An alternative would be some sort of plugin that automatically rewrites link URLs when posts are republished.

Whether done manually or automatically, a corner case would be ensuring that you _don’t_ use relative links when linking to posts that, for whatever reason, happen to be only on one of the two sites.

kyrofa:
Thanks @matthewpaulthomas, that outlines the options nicely.

[...]

PR #1299, opened 2018-11-12
Redirect blog.ubuntu.com URL forms to the blog

nottrobin:
This will redirect URLs of the following forms:

- `/blog/{year}/{slug}`
- `/blog/{year}/{month}/{slug}`
- `/blog/{year}/{month}/{day}/{slug}`

To the canonical form of `/blog/{slug}`.

This is to provide parity in URL support with blog.ubuntu.com, to help with #1222.

QA
--

`./run` and go to any of:

- http://0.0.0.0:8004/blog/2018/03/09/your-first-robot-the-driver-4-5
- http://0.0.0.0:8004/blog/2018/03/your-first-robot-the-driver-4-5
- http://0.0.0.0:8004/blog/2018/your-first-robot-the-driver-4-5

Check you're correctly redirected to http://0.0.0.0:8004/blog/your-first-robot-the-driver-4-5

Also, check `./run test` passes.

### DIFF
--- a/webapp/blog/views.py
+++ b/webapp/blog/views.py
@@ -117,6 +117,16 @@ def feed():
     return flask.Response(right_title, mimetype="text/xml")
 
 
+@blog.route(
+    '/<regex("[0-9]{4}"):year>/<regex("[0-9]{2}"):month>/'
+    '<regex("[0-9]{2}"):day>/<slug>'
+)
+@blog.route('/<regex("[0-9]{4}"):year>/<regex("[0-9]{2}"):month>/<slug>')
+@blog.route('/<regex("[0-9]{4}"):year>/<slug>')
+def article_redirect(slug, year, month=None, day=None):
+    return flask.redirect(flask.url_for(".article", slug=slug))
+
+
 @blog.route("/<slug>")
 def article(slug):
     try:
--- a/tests/blog/tests_index.py
+++ b/tests/blog/tests_index.py
@@ -308,3 +308,31 @@ class BlogPage(TestCase):
         response = self.client.get("/blog/feed")
 
         assert response.status_code == 502
+
+    def test_article_redirects(self):
+        """
+        Check that blog articles can be requested with
+        year, month and day parts in the URL, which will
+        result in a redirect to the canonical URL, with just the slug
+        """
+
+        year_redirect = self.client.get("/blog/2005/article_slug")
+        month_redirect = self.client.get("/blog/2005/10/article_slug")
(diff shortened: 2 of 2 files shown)


webapp/blog/views.py
```diff
@@ -117,8 +117,14 @@ def feed():
     return flask.Response(right_title, mimetype="text/xml")
 
 
+@blog.route(
+    '/<regex("[0-9]{4}"):year>/<regex("[0-9]{2}"):month>/'
+    '<regex("[0-9]{2}"):day>/<slug>'
+)
+@blog.route('/<regex("[0-9]{4}"):year>/<regex("[0-9]{2}"):month>/<slug>')
+@blog.route('/<regex("[0-9]{4}"):year>/<slug>')
```
tbille: Can this go on it's own view that redirects directly to the slug. This will avoid a useless api call or logic treatement.

```python

@blog.route(
    '/<regex("[0-9]{4}"):year>/<regex("[0-9]{2}"):month>/'
    '<regex("[0-9]{2}"):day>/<slug>'
)
@blog.route('/<regex("[0-9]{4}"):year>/<regex("[0-9]{2}"):month>/<slug>')
@blog.route('/<regex("[0-9]{4}"):year>/<slug>')
def redirect_article(slug, year=None, month=None, day=None):
    return flask.redirect(flask.url_for(".article"), slug="{slug}".format(**locals()))
```

webapp/blog/views.py
```diff
@@ -129,6 +135,9 @@ def article(slug):
     if not articles:
         flask.abort(404, "Article not found")
 
+    if day or month or year:
+        return flask.redirect("/blog/{slug}".format(**locals()))
```
tbille: I try to avoid as much as I can to write urls out of routes functions `@app.route("/my-route")`  because it happened that we had to change our url structure. We had to update every redirect in the code.

To solve this flask gives us the possibility to redirect directly to a function. This avoids to have a strong dependency in the code with the urls. It looks like this:
```python
return flask.redirect(flask.url_for(".article"), slug="{slug}".format(**locals()))
```

webapp/blog/views.py
tbille: And you can add test to make sure the redirection redirects :smile:

nottrobin: :+1: Yep fair point. I'll do that.

webapp/blog/views.py
nottrobin: :+1:

nottrobin:
@tbille good points. I've addressed them both.

Issue #1309, opened 2018-11-14
Last point on active devices graph is not accessible


Lukewh:
See https://forum.snapcraft.io/t/store-metrics/8434/6 and https://forum.snapcraft.io/t/publisher-metrics-dont-add-up/8487/1

PR #1310, opened 2018-11-14
Ensure the last point on the graph is accessible

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1309

This PR offsets the hover date by -12 hours so the middle of a day is aligned with the tick.

## QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/<snap_name>/metrics
- Hover over the graph around the last tick
- You should be able to see the last available date (today - 1)


bartaz:
I don't see the 13th yet, but I guess stats are not yet available, stats for 12th are accessible in the tooltip.
Not sure how this fix relates to timezone etc discussions on forum, but it helps seeing stats from last day.

PR #1314, opened 2018-11-14
Update ES, set asset version number as variable

Lukewh:
Updated Spanish Instala to Instalar as per https://github.com/canonical-websites/snapcraft.io/commit/e7ab4d663cb6f927f75b8f697a8be93a1cf4af4b#r31297033.

Also set the download version number to be a variable for easier updating in the future.

## QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/<snap>/publicise
- Ensure the Spanish version has the correct text, both in the copy and the image


static/images/badges/es/snap-store-black.svg
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <svg width="182px" height="56px" viewBox="0 0 182 56" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-    <!-- Generator: Sketch 52.3 (67297) - http://www.bohemiancoding.com/sketch -->
+    <!-- Generator: Sketch 52.4 (67378) - http://www.bohemiancoding.com/sketch -->
```
bartaz: Probably not necessary in this PR (as likely affects all the badges), but seems like these are straight from Sketch and could be optimized with some tool like SVGOMG.

webapp/publisher/snaps/views.py
```diff
@@ -896,6 +896,7 @@ def get_publicise(snap_name):
         "snap_title": snap_details["title"],
         "snap_id": snap_details["snap_id"],
         "available": available_languages,
+        "download_version": "v1.1",
```
bartaz: Too bad we would need to update this manually every time. Hopefully not to often.

There seems to be a way to get latest version from API, but probably not worth implementing at this point.

Lukewh: Yeah I thought about that, I think we might update more frequently in the coming months, but updates will probably slow

PR #1320, opened 2018-11-15
Use cloudinary for blog post images

Lukewh:
@pmahnke helpfully [added a snippet](https://github.com/canonical-websites/blog.ubuntu.com/pull/416/files#diff-e8b4e4b3fe31ecb310658b8a0d0f0c8eR106) to the Ubuntu blog. Here I've copied/ pasted and modified it slightly.

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/blog
- Click on an article
- Open the network inspector and see how small the inline images are
- Compare to the same post and network inspector on snapcraft.io/blog

### DIFF
--- a/webapp/blog/logic.py
+++ b/webapp/blog/logic.py
@@ -16,7 +16,32 @@ def strip_excerpt(raw_html):
     return html.unescape(cleantext).replace("\n", "")
 
 
-def transform_article(article, featured_image=None, author=None):
+def replace_images_with_cloudinary(content):
+    """Prefixes images with cloudinary optimised URLs and adds srcset for
+    image scaling
+
+    :param content: The HTML string to convert
+
+    :returns: Update HTML string with converted images
+    """
+    cloudinary = (
+        "https://res.cloudinary.com/" "canonical/image/fetch/q_auto,f_auto,"
+    )
+    return re.sub(
+        r"img(.*)src=\"(.[^\"]*)\"",
+        r'img\1 src="{url}w_1120/\2"'
+        r'srcset="{url}w_750/\2 750w,'
+        r'{url}w_960/\2 960w, {url}w_1120/\2 1120w"'
+        r'sizes="(max-width: 375px) 560px,'
+        r"(max-width: 480px) 880px,"
+        r'1120px"'.format(url=cloudinary),
+        content,
+    )
+
+
+def transform_article(
+    article, featured_image=None, author=None, optimise_images=False
+):
     """Transform article to include featured image, human readable
     date and a stipped version of the excerpt
 
@@ -51,6 +76,15 @@ def transform_article(article, featured_image=None, author=None):
         # join it back up
         article["excerpt"]["raw"] = "".join(
             [raw_article_start, raw_article_end, " […]"]
+        )
+
+    if (
+        optimise_images
+        and "content" in article
+        and "rendered" in article["content"]
+    ):
+        article["content"]["rendered"] = replace_images_with_cloudinary(
+            article["content"]["rendered"]
         )
 
     return article
--- a/webapp/blog/logic.py
+++ b/webapp/blog/logic.py
@@ -27,9 +27,9 @@ def replace_images_with_cloudinary(content):
     cloudinary = "https://res.cloudinary.com/"
 
     urls = [
-        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_540/\g<url>",
-        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_1080/\g<url>",
-        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_1620/\g<url>",
+        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_650/\g<url>",
+        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_1300/\g<url>",
+        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_1950/\g<url>",
     ]
 
     image_match = (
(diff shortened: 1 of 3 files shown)


pmahnke:
FYI, I added a decode="async" based on web.dev as well

```
if post["content"]:
        CLOUDINARY = (
            "https://res.cloudinary.com/"
            "canonical/image/fetch/q_auto,f_auto,"
        )
        """
        Remove existing cloudinary urls
        """
        post["content"]["rendered"] = re.sub(
            r'img(.*)src="https://res.cloudinary.com/canonical'
            r'(.[^http]*)/http(.[^"]*)"',
            r'img\1 src="\3"',
            post["content"]["rendered"],
        )
        """
        Add cloudinary urls with a srcset
        """
        post["content"]["rendered"] = re.sub(
            r"img(.*) src=\"(.[^\"]*)\"",
            r'img\1 decoding="async" src="{url}w_560/\2"'
            r'srcset="{url}w_375/\2 375w,'
            r'{url}w_480/\2 480w, {url}w_560/\2 560w"'
            r'sizes="(max-width: 375px) 280px,'
            r"(max-width: 480px) 440px,"
            r'560px"'.format(url=CLOUDINARY),
            post["content"]["rendered"],
        )
```

webapp/blog/logic.py
```diff
@@ -16,7 +16,32 @@ def strip_excerpt(raw_html):
...
+    image scaling
+
+    :param content: The HTML string to convert
+
+    :returns: Update HTML string with converted images
+    """
+    cloudinary = (
+        "https://res.cloudinary.com/" "canonical/image/fetch/q_auto,f_auto,"
+    )
+    return re.sub(
```
tbille: This is really hard to read and understand just by looking at it like this...
Is there no other way to make it simplier

nottrobin: To try to make this more readable, how about something like:

```suggestion
    cloudinary = 'https://res.cloudinary.com/'

    urls=[
        cloudinary + 'canonical/image/fetch/q_auto,f_auto,w_1120/\g<url>',
        cloudinary + 'canonical/image/fetch/q_auto,f_auto,w_750/\g<url>',
        cloudinary + 'canonical/image/fetch/q_auto,f_auto,w_960/\g<url>',
        cloudinary + 'canonical/image/fetch/q_auto,f_auto,w_1120/\g<url>',
    ]
    
    image_match = r'<img(?P<prefix>[^>]*) src="(?P< url>[^"]+)"(?P<suffix>[^>]*)>'
    replacement = (
        '<img\g<prefix>'
        ' src="{urls[0]}"'
        ' srcset="{urls[1]} 750w, {urls[2]} 960w, {urls[3]} 1120w"'
        ' sizes="(max-width: 375px) 560px, (max-width: 480px) 880px, 1120px"'
        '\g<suffix>>'
    ).format(urls)

    return re.sub(image_match, replacement, content)
```

(I've not tested this code)

webapp/blog/logic.py
```diff
@@ -16,7 +16,32 @@ def strip_excerpt(raw_html):
...
+        r"img(.*)src=\"(.[^\"]*)\"",
+        r'img\1 src="{url}w_1120/\2"'
+        r'srcset="{url}w_750/\2 750w,'
+        r'{url}w_960/\2 960w, {url}w_1120/\2 1120w"'
+        r'sizes="(max-width: 375px) 560px,'
```
nottrobin: Can someone explain this sizes attribute to me? 1) Why don't the suggested sizes here (`560px`, `880px`) match up to the sizes of images we're requesting from cloudinary? 2) Why are these widths so much smaller than the suggested image size? 3) How can we make any "sizes" recommendations when we don't know anything about the image we're replacing? Shouldn't we just use `sizes="100vw"`?

webapp/blog/logic.py
Lukewh: I'll try your suggestion out locally nottrobin, if it works I'll accept it :+1:

webapp/blog/logic.py
Lukewh: Again I'll try your code and approve if it works!

With regards to the different sizes: It's a crude way to have sharper images even when optimised (pixels removed) by Cloudinary. I noticed some examples where the image that was sharp before became blurry - the most extreme of which was the infographic on https://snapcraft.io/blog/infographic-snaps-in-numbers.

I'm open to alternatives and I should probably study the documentation around `srcset` more closely

nottrobin: Yeah I think there's clever stuff you can do with pixel density.

Lukewh: :+1: will have a look, not sure there'll be something that fixes blurry images at 72dpi though :man_shrugging:

bartaz:
There is quite significant visual difference between images before and after. Especially if it's not a photo but a screenshot.

screen shot 2018-11-19 at 16 30 29

We seem to be requesting images in smaller size that we use to display them:
screen shot 2018-11-19 at 16 31 09

We use 540px wide image scaled up to 649px. There is no way it's not going to be blurry, especially on hi dpi screen.

Lukewh:
Just updated it @bartaz. When I was zoomed in it was using different sized images in Firefox. I've set the base to 650px now, thanks.

Lukewh:
@bartaz seems using the 2x, 3x syntax is old - the browser now figures it out for you :+1:. Removed and chrome works!

webapp/blog/logic.py
```diff
@@ -16,7 +16,40 @@ def strip_excerpt(raw_html):
...
+
+    image_match = (
+        r'<img(?P<prefix>[^>]*) src="(?P<url>[^"]+)"(?P<suffix>[^>]*)>'
+    )
+    replacement = (
+        "<img\g<prefix>"
+        f' decoding="async"'
+        f' src="{urls[0]}"'
+        f' srcset="{urls[0]} 650w, {urls[1]} 1300w, {urls[2]} 1950w"'
+        f' sizes="650px"'
```
bartaz: Do we need `1300w` or `1950w` is we have only `650px` in sizes?

This seems to be only for blog content, which we always have at 650px width, so we will never need bigger images. If anything, we could use smaller images on smaller screens.

Lukewh: 1300w and 1950w are for retina+ displays

I've added one smaller size for below 400px. The way the columns resize means that at ~400px the columns go 100% width so I don't think it's worth having a complicated srcset. wdyt?

bartaz: Ah, OK - I was debugging it incorrectly. Inspecting HTML Chrome was telling me 650px image is used (as this is img src), but looking at network panel I see that bigger is downloaded for retina.

Thanks for adding one smaller one. Works well.

PR #1328, opened 2018-11-16
Release UI - visual updates

### LINKED ISSUE #1308: Update styles of release table
Designs: https://github.com/canonicalltd/snapcraft-design/issues/552

### TODO:

- [x] improve grid/spacing as per design
- [x] remove borders from contextual menu button
- [x] smaller X cancel release button
- [x] highlight pending releases in 'yellowish'
- [x] fix size of the table when nothing is released yet
- [x] make sure it looks nice for 6 archs (/surl/releases)

bartaz:
Fixes #1308

Designs: https://github.com/canonicalltd/snapcraft-design/issues/552

Done:

- improved grid/spacing of releases table
- updated contextual menu buttons for channels
- smaller cancel release button [X]
- highlights pending releases in 'yellowish' colour
- makes sure it looks nice for 6 archs (no wrapping 'Add revision')

### QA

- ./run or https://snapcraft-io-canonical-websites-pr-1328.run.demo.haus/
- check if everything looks nice
- go to releases page of snap with 6 architectures (surl)
- everything should look ok

screen shot 2018-11-19 at 08 33 42

### DIFF
--- a/static/js/publisher/release/contextualMenu.js
+++ b/static/js/publisher/release/contextualMenu.js
@@ -54,10 +54,11 @@ export default class ContextualMenu extends Component {
   render() {
     const { position } = this.props;
     const menuClass = "p-contextual-menu" + (position ? `--${position}` : "");
+    const appearance = this.props.appearance || "neutral";
 
     return (
       <span
-        className={`p-promote-button p-button--neutral p-icon-button ${menuClass}`}
+        className={`p-promote-button p-button--${appearance} p-icon-button ${menuClass}`}
         onClick={this.dropdownButtonClick.bind(this)}
       >
         {this.renderIcon()}
@@ -71,6 +72,7 @@ export default class ContextualMenu extends Component {
 
 ContextualMenu.propTypes = {
   position: PropTypes.oneOf(["left", "center"]), // right is by default
+  appearance: PropTypes.oneOf(["base", "neutral"]),
   channel: PropTypes.string.isRequired,
   closeChannel: PropTypes.func.isRequired
 };
--- a/static/js/publisher/release/contextualMenu.js
+++ b/static/js/publisher/release/contextualMenu.js
@@ -55,10 +55,11 @@ export default class ContextualMenu extends Component {
     const { position } = this.props;
     const menuClass = "p-contextual-menu" + (position ? `--${position}` : "");
     const appearance = this.props.appearance || "neutral";
+    const className = this.props.className || "";
 
     return (
       <span
-        className={`p-promote-button p-button--${appearance} p-icon-button ${menuClass}`}
+        className={`p-promote-button p-button--${appearance} p-icon-button ${menuClass} ${className}`}
         onClick={this.dropdownButtonClick.bind(this)}
       >
         {this.renderIcon()}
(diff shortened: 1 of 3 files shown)


lyubomir-popov:
Nice and clean, thanks @bartaz! Just a couple of small things I spotted:

wrapping buttons on small screens:
image


can you add a context menu placeholder so promote buttons stay aligned?
image

• since channel / risk level is of known width, can we set the column to that? It takes up space that can be used by the version col. I also think we should consider ellipsizing rather than wrapping. @therealjuan what do you think?

• vertical-align:top. Never middle:

therealjuan:
> • since channel / risk level is of known width, can we set the column to that? It takes up space that can be used by the version col. I also think we should consider ellipsizing rather than wrapping. @therealjuan what do you think?

Problem is that this varies as there can be instances where you have a revision released in multiple channels (and even across tracks). Perhaps a solution here will be to use ellipsis in both cases?

lyubomir-popov:
@therealjuan I meant the track-name/risk-level pair - isn't that of known width? I.e.: "latest/stable".

therealjuan:
It is known for the `latest` track but it could happen that someone wants to label it `version-specially-implemented-for-someone` as it can be requested.

bartaz:
@lyubomir-popov @therealjuan I addressed the comments.

I made channel name column fixed width, so there may be a layout issue when some track has very long name, but that would have been even bigger issue before, as this column would never grow bigger to fit it anyway, it was only getting smaller on smaller screens (due to % width).

Now it's fixed width so it doesn't shrink on smaller screens. If we want to address the issue of long track name we would need to either use ellipsis (...) or break the line to have `track/` in one line and `risk` in the other. But let's do it separately as I don't think that it would be an issue for any of existing snaps ATM.

Lukewh:
Not a big deal but it's quite jarring when you mouse between cells as the tooltip is only triggered by the centre part of the tab/button

bartaz:
@Lukewh @therealjuan 

> Not a big deal but it's quite jarring when you mouse between cells as the tooltip is only triggered by the centre part of the tab/button.
As for the tooltips on table cells. I feel like there is almost no need for them now. It feels like they are repeating mostly what cells already says by itself.
Two cases when they may be needed - to explain devmode snaps (we do that in other places) and when version doesn't fit into cell (but we can't show tooltip only in that case as we cannot detect that in CSS).

> It also feels a bit strange that if I open the "Show x latest revisions" at the bottom of the page, this then closes if I click on a cell. They feel like 2 independent actions.

As much as they may feel separate actions (and panels) they are just different filtering of the same 'Available revisions'. Maybe we should consider making this part of the table more (like clicking on 'Available revisions column' to show revisions from all architectures.

lyubomir-popov:
@bartaz image
image
Shouldn't these be same width? (The one with the white space).

therealjuan:
> As much as they may feel separate actions (and panels) they are just different filtering of the same 'Available revisions'. Maybe we should consider making this part of the table more (like clicking on 'Available revisions column' to show revisions from all architectures.

This would make sense to simplify the content displayed as `show available revisions` could be confusing. Maybe we should tackle this on the next iteration?

bartaz:
@lyubomir-popov @therealjuan The width of 'Show all revisions' is separate issue to discuss. Because it's all archs it has additional 'Architecture' column, so it may feel busy in smaller width. Also currently it's not a 'tab' (it doesn't select any architecture), so it doesn't feel like it needs to be smaller width 'expanded' panel.

Of course for UX and design consistency it would be good to give it similar treatment.

As @therealjuan suggested we can address it next iteration as an enhancement not to postpone merging these improvements too much. There will always be some tweaks we can do to improve things.

lyubomir-popov:
@bartaz @therealjuan makes sense Bartek, adding +1.

therealjuan:
LGTM 👍

PR #1334, opened 2018-11-19
Optimise gifs

Lukewh:
This PR uses Cloudinary to convert animated gifs to videos, and a fallback static jpg image.

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/gitkraken
- Open the browsers Network logs and see how large the page is (disable cache etc)
- Do the same thing on https://snapcraft.io/gitkraken

### DIFF
--- a/static/js/public/snap-details/screenshots.js
+++ b/static/js/public/snap-details/screenshots.js
@@ -12,9 +12,9 @@ export default function initScreenshots(screenshotsId) {
     return;
   }
 
-  const images = Array.from(screenshotsEl.querySelectorAll("img")).map(
-    image => image.dataset.original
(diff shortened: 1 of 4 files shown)


bartaz:
Opening gif/video in the lightbox doesn't work well:

screen shot 2018-11-20 at 08 49 50

First I thought that image doesn't load at all but it seems it loads but very slow.

Also navigating in the clicking on arrows leads to error in console:

screen shot 2018-11-20 at 08 49 57

Lukewh:
@bartaz fixed that issue and also ensured that the lightbox uses the same video as in the page to reduce load.

Issue #1342, opened 2018-11-21
Release UI: Trigger devmode warnings after promoting devmode release/s


therealjuan:
Right now if we look at snaps with releases on `devmode` we are showcasing a global warning and a warning next to each release:

![image.png](https://images.zenhubusercontent.com/56b0a0916caf56c9d38c2685/dda32c96-a47a-4c64-9fff-f7e38af654fe)

To avoid the interface to look like the publisher has done something wrong (as they haven't done anything yet) to show the warning once a publisher tries to promote release/s set in `devmode` to a `stable` or `candidate` channel

There is a separate issue to raise which is to include additional text and context on the release UI so publishers understand what can be done and therefore understand the interface.

**(From snapju workshop)**

bartaz:
@therealjuan what do you mean by 'once a publisher tries to promote release/s set in devmode to a stable or candidate channel'?

It is not possible to promote devmode snap to these channels. They were not available to select or now (#1350) are disabled. So publishers can't 'try' to promote them.

We can of course change that and allow them to select these devmode revisions to stable/candidate, but should we do it just to show them warning that they can't?

Maybe tooltips that we have on devmode icons are enough, and we can just drop the top page warning? Or we can stop using warning notification style and use info (which will be less alarming)?
Also - having some custom devmode icon instead of Vanilla warning one could help.

matthewpaulthomas:
Now that `candidate` and `stable` appear as disabled items, rather than being hidden altogether, one possibility (as I suggested on IRC) would be to explain the situation when hovering or clicking on the disabled items. For example, in a box that floats alongside the menu.

bartaz:
@therealjuan Since it was reported we changed the icon of devmode and moved the notification inside the history panels. Is there anything else we want to do with this?

therealjuan:
Because of the recent changes in this prior its report, we can include the same tooltip on the disabled menu items when promoting channels. 

The content of the tooltip is `Revisions in with grade devel cannot be released to stable or candidate channels`

![image](https://user-images.githubusercontent.com/8167677/51022660-c95db380-157c-11e9-94e6-ed39a067961e.png)

PR #1350, opened 2018-11-22
Disable unavailable channels in promote menu

### LINKED ISSUE #1204: Disable unavailable risk levels rather than hiding them
#1195 hides the “candidate” and “stable” items from the promotion menu in a snap’s “Releases” page, if that snap revision uses `devmode`.

From my review:
> I think it’s a little mysterious to hide the levels completely, rather than just greying them out. Unless great clutter is at stake, usual practice is to grey out items if they can’t be used now but could be used in future. That’s exactly the situation here: when you migrate your snap out of `devmode`, you’ll be able to use these levels again.

Reply by @bartaz:
> Good point. If possible I'd like to address this separately, as currently Vanilla context menu doesn't seem to support 'disabled' items, so I would need to add that.

### LINKED ISSUE #1332: “Close latest/candidate” is ellipsized needlessly
As the publisher of a snap that has a “candidate” release:
1\. Go to the “Releases” page.
2\. Open the kebab menu.

What you see: “Close latest/candi…”

snapcraft io-releases-ellipsis

🚪✨🍬

What you should see: “Close latest/candidate”

bartaz:
Fixes #1204 
Fixes #1332

Disables unavailable channels in promote menu instead of hiding them.
Also, works around #1332 by using default contextual menu direction.

### QA
- ./run or demo
- try to promote any channel with devmode snaps in it
- menu should show stable and candidate channels, but they should be disabled
- open close menu of 'latest/candidate' - menu text should not be truncated
- if given channel can't be released to any other channel (for example beta with devmode snap) - menu button should be disabled (and not clickable)

screen shot 2018-11-22 at 12 40 14


therealjuan:
For the purpose of the issues marked as fixed this works but would be good to have a look at #1342 in a further iteration so we can remove unnecessary warning signs.

static/js/publisher/release/contextualMenu.js
```diff
@@ -44,37 +44,36 @@ export default class ContextualMenu extends Component {
...
+    return this.props.icon || <i className="p-icon--contextual-menu" />;
   }
 
   render() {
-    const { position } = this.props;
+    const { isDisabled, position } = this.props;
     const menuClass = "p-contextual-menu" + (position ? `--${position}` : "");
     const appearance = this.props.appearance || "neutral";
-    const className = this.props.className || "";
+    const className = `p-promote-button p-icon-button p-button--${appearance} ${
```
Lukewh: Maybe create as an array and move the ternary's to if's for readability?

static/js/publisher/release/promoteButton.js
```diff
@@ -1,46 +1,70 @@
...
   }
 
   renderItems() {
-    const { track } = this.props;
-
     return (
       <span className="p-contextual-menu__group">
         <span className="p-contextual-menu__item">Promote to:</span>
-        {this.props.targetRisks.map(targetRisk => {
+        {this.props.targetChannels.map(targetChannel => {
```
Lukewh: Could be moved into its own method that contains the logic?

static/js/publisher/release/contextualMenu.js
bartaz: Yeah... I hate building these long conditional class names as well. Probably worth looking into [classNames](https://github.com/JedWatson/classnames) module sooner then later.

static/js/publisher/release/promoteButton.js
bartaz: Good point, will do.

PR #1365, opened 2018-11-27
Use bases in the first snap flow

evandandrea:
Use the `base` keyword in the snapcraft.yaml files in the first snap flow. The use of bases makes snapcraft use a tightly controlled build environment inside multipass to create the snap, _much_ more reliable than relying on whatever is in the developer's host system.

/cc @sergiusens

### DIFF
--- a/webapp/first_snap/content/python/package.yaml
+++ b/webapp/first_snap/content/python/package.yaml
@@ -32,7 +32,8 @@ explain:
     - text: It’s best to start a snap with the confinement in warning mode, rather than strictly applied. This is indicated through the <code>devmode</code> keyword.
     - text: Once an app is working well in devmode, you can review confinement violations, add appropriate interfaces, and switch to strict confinement.
   base:
-    - text: Snaps cannot see the root filesystem on end user systems. This prevents conflict with other applications and increases security. The <code>base</code> keyword specifies a special kind of snap that will be mounted as the root filesystem for your application.
+    - text: In general, snaps cannot see the root filesystem on end user systems. This prevents conflict with other applications and increases security. However, applications still need some location to act as the root filesystem. They would also benefit from common libraries (e.g. libc) being in this root filesystem rather than bundled into each application.
+    - text: The base keyword specifies a special kind of snap that provides a minimal set of libraries common to most applications. It will be mounted as the root filesystem for your application.
     - code: |
         base: core18
     - text: The <code>core18</code> base is recommended.
--- a/webapp/first_snap/content/python/package.yaml
+++ b/webapp/first_snap/content/python/package.yaml
@@ -31,6 +31,11 @@ explain:
     - text: Snaps are containerised to ensure more predictable application behaviour and greater security.
     - text: It’s best to start a snap with the confinement in warning mode, rather than strictly applied. This is indicated through the <code>devmode</code> keyword.
     - text: Once an app is working well in devmode, you can review confinement violations, add appropriate interfaces, and switch to strict confinement.
+  base:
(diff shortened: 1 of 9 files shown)


igorljubuncic:
Commenting on:

Base

Snaps cannot see the root filesystem on end user systems. This prevents conflict with other applications and increases security. The base keyword specifies a special kind of snap that will be mounted as the root filesystem for your application.

base: core18

The core18 base is recommended.

=====

I find the explanation a little ambiguous.

- It should emphasize the isolation of the dev environment from the system.
- It should explain what types of dev libraries/tools will be used to build the snap.

webapp/first_snap/content/python/package.yaml
```diff
@@ -31,6 +31,11 @@ explain:
     - text: Snaps are containerised to ensure more predictable application behaviour and greater security.
     - text: It’s best to start a snap with the confinement in warning mode, rather than strictly applied. This is indicated through the <code>devmode</code> keyword.
     - text: Once an app is working well in devmode, you can review confinement violations, add appropriate interfaces, and switch to strict confinement.
+  base:
+    - text: Snaps cannot see the root filesystem on end user systems. This prevents conflict with other applications and increases security. The <code>base</code> keyword specifies a special kind of snap that will be mounted as the root filesystem for your application.
```
sergiusens: _Snaps cannot see the root filesystem on end user systems_ is technically incorrect if we consider `classic`. If this is intended however to be beginners text, then it seems conceptually understandable.

evandandrea: Exactly this. Maybe we should change it to, "In general, snaps cannot see the root filesystem on end user systems."

evandandrea:
@igorljubuncic @sergiusens how about:

In general, snaps cannot see the root filesystem on end user systems. This prevents conflict with other applications and increases security. The base keyword specifies a special kind of snap that will be mounted as the root filesystem for your application. When your application is run, it will first search library paths inside your snap then the library paths of the base.

```
base: core18
```

The core18 base is recommended.

igorljubuncic:
I find the third sentence confusing - the one that starts with "The base keyword". At this point users don't know much about snaps or their format, and mentioning the use of another snap can indeed be confusing.

Can we perhaps use 'build/dev package bundle' instead of snap?

Last sentence (added a comma snap, then).

Last sentence also - is this not more about when the app is being built rather than run? And then, after the necessary base libs are pulled into the snap, then the search will be based on the lib path?

evandandrea:
>Can we perhaps use 'build/dev package bundle' instead of snap?

That's not what a base snap is. A base snap _only_ contains the run-time libraries. It does not contain build-time assets.

>Last sentence also - is this not more about when the app is being built rather than run? And then, after the necessary base libs are pulled into the snap, then the search will be based on the lib path?

The base snap provides a minimal set of common libraries. It is used at run-time to provide a root filesystem and resolve needed common libraries (e.g. libc6) that are not present inside your snap.

So @igorljubuncic, @sergiusens how about this:

>In general, snaps cannot see the root filesystem on end user systems. This prevents conflict with other applications and increases security. However, applications still need some location to act as the root filesystem. They would also benefit from common libraries (e.g. libc) being in this root filesystem rather than bundled into each application.
>
>The base keyword specifies a special kind of snap that provides a minimal set of libraries common to most applications. It will be mounted as the root filesystem for your application. When your application is run, it will first search library paths inside your snap, then the library paths inside the base.

sergiusens:
I will remove "When your application is run, it will first search library paths inside your snap, then the library paths inside the base." it is too specific and not necessarily true even if it is the case for 99% of the people.

evandandrea:
@igorljubuncic gave a +1 on Slack. This is clear for merging.

PR #1375, opened 2018-11-29
Release UI - introduce Redux to keep the revisions list

bartaz:
First steps of moving release page state into Redux:
- adds redux and related libraries
- make babel/eslint compile object spread operator `{ ...state }` to make reducers nice
- moves revisions list data from components into Redux state

It's obviously not an ideal Redux setup currently, but it's a first step to start using it.

### QA

- install Redux devtools extension
- ./run or demo
- go Releases page
- open Redux devtools
- you should see actions initializing the data
- state should contain list of all revisions of the snap
- release some revision to a channel
- you should see "UPDATE_REVISIONS" action after successful release
- everything should work as before (try selecting stuff, promoting channels, releasing)

screen shot 2018-11-29 at 15 19 44

### DIFF
--- a/static/js/publisher/release/reducers/index.js
+++ b/static/js/publisher/release/reducers/index.js
@@ -1,0 +1,19 @@
+import { combineReducers } from "redux";
+
+const revisions = function(state = {}, action) {
+  switch (action.type) {
+    case "UPDATE_REVISIONS":
+      return {
+        ...state,
+        ...action.payload.revisions
+      };
+    default:
+      return state;
+  }
+};
+
+const releases = combineReducers({
+  revisions
+});
+
+export default releases;
--- a/static/js/publisher/release/releasesController.js
+++ b/static/js/publisher/release/releasesController.js
@@ -1,5 +1,6 @@
 import React, { Component, Fragment } from "react";
 import PropTypes from "prop-types";
+import { connect } from "react-redux";
 import "whatwg-fetch";
 
 import ReleasesTable from "./releasesTable";
@@ -17,13 +18,18 @@ import {
   getReleaseDataFromChannelMap
 } from "./releasesState";
 
-export default class ReleasesController extends Component {
+class ReleasesController extends Component {
   constructor(props) {
     super(props);
 
     // init channel data in revisions list
+    // TODO: should be done in reducers?
     const revisionsMap = getRevisionsMap(this.props.releasesData.revisions);
     initReleasesData(revisionsMap, this.props.releasesData.releases);
+
+    // init redux store
+    // TODO: should be done outside component as initial state?
+    this.props.updateRevisions(revisionsMap);
 
     const releasedChannels = getReleaseDataFromChannelMap(
       this.props.channelMapsList,
@@ -42,8 +48,6 @@ export default class ReleasesController extends Component {
       releasedChannels: releasedChannels,
       // list of releases returned by API
       releases: this.props.releasesData.releases,
-      // map of revisions (with revisionId as a key)
-      revisionsMap: revisionsMap,
       // list of all available tracks
       tracks: tracks,
       // list of architectures released to (or selected to be released to)
@@ -75,8 +79,8 @@ export default class ReleasesController extends Component {
     const revisionsMap = getRevisionsMap(releasesData.revisions);
     initReleasesData(revisionsMap, releasesData.releases);
 
+    this.props.updateRevisions(revisionsMap);
     this.setState({
-      revisionsMap,
       releases: releasesData.releases
     });
   }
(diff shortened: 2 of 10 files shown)


static/js/publisher/release/reducers/index.js
```diff
@@ -0,0 +1,19 @@
+import { combineReducers } from "redux";
+
+const revisions = function(state = {}, action) {
+  switch (action.type) {
+    case "UPDATE_REVISIONS":
```
Lukewh: Can we put actions as "constants" at the top of the file so we can export them to be used elsewhere?
```
const UPDATE_REVISIONS = "UPDATE_REVISIONS";
...
case UPDATE_REVISIONS:
...
export { releases as default, UPDATE_REVISONS }
```

The exports can get quite big, but it makes it easier to keep track (I find)

Lukewh: Just realised this is the reducers file, these constants should really be in the actions... So disregard for this PR.

static/js/publisher/release/releasesController.js
```diff
@@ -17,14 +18,19 @@ import {
...
 } from "./releasesState";
 
-export default class ReleasesController extends Component {
+class ReleasesController extends Component {
   constructor(props) {
     super(props);
 
     // init channel data in revisions list
+    // TODO: should be done in reducers?
     const revisionsMap = getRevisionsMap(this.props.releasesData.revisions);
```
Lukewh: I think (if I understand this correctly) I'd do this in the `mapStateToProps`

static/js/publisher/release/releasesController.js
```diff
@@ -17,14 +18,19 @@ import {
...
     initReleasesData(revisionsMap, this.props.releasesData.releases);
```
Lukewh: Same here

static/js/publisher/release/releasesController.js
```diff
@@ -17,14 +18,19 @@ import {
...
 
+    // init redux store
+    // TODO: should be done outside component as initial state?
+    this.props.updateRevisions(revisionsMap);
```
Lukewh: Or even in `componentWillMount`?

static/js/publisher/release/reducers/index.js
bartaz: Yes, that's somewhere on the way of future changes.

static/js/publisher/release/releasesController.js
bartaz: Ideally this step should be done before data gets into the component. Somewhere where 'initial state' is prepared. But also it's needed when the data is updated via `fetch`, so that's why I think about moving it to reducers in future.

So some new action "RELEASE_SUCCESS" will update the revisions list (and channels revision was released to). And it will be a state change performed by reducer.

static/js/publisher/release/releasesController.js
bartaz: It's basically transforming an array of revisions into a map (keyed by revision id), so I think reducer could do it (especially that we need to do it on successful release fetch as well). Similarly - updating list of channels for each revisions.

static/js/publisher/release/reducers/index.js
squidsoup: > Can we put actions as "constants" at the top of the file so we can export them to be used elsewhere?
> The exports can get quite big, but it makes it easier to keep track (I find)

+1, this does create more boilerplate (which redux in general suffers from a bit), but it is safer. You can even namespace your actions by functional parts of your app, to prevent conflicts if you're paranoid, although probably overkill.

e.g. `const UPDATE_REVISIONS = "release/UPDATE_REVISIONS";`

squidsoup: > Just realised this is the reducers file, these constants should really be in the actions... So disregard for this PR.

You can keep constants in their own module (`constants.js`), just to reduce clutter in actions.

static/js/publisher/release/releasesController.js
```diff
@@ -555,5 +560,33 @@ ReleasesController.propTypes = {
...
+  return {
+    revisions: state.revisions
+  };
+};
+
+const mapDispatchToProps = dispatch => {
+  return {
+    // TODO: extract to outside actions module
+    updateRevisions: revisions => {
+      dispatch({
```
squidsoup: Might be worth considering using action creators to decouple action payload from the component.

bartaz: Sure, slowly getting there!

static/js/publisher/release/releasesController.js
bartaz: We've been using reselect in BSI as well. Currently I have a bunch of non-memoized selector functions (https://github.com/canonical-websites/snapcraft.io/blob/master/static/js/publisher/release/selectors/index.js), but have in plans to use reselect to memoize them.

static/js/publisher/release/reducers/index.js
bartaz: We already have a constants.js where we keep some strings (like risk names, etc), so may be a good place to move action names as well. This will allow to import them from just one place rather then each action module separately.

PR #1376, opened 2018-11-29
Update channel map to incorporate default track + safest risk level + version

therealjuan:
1. Pull the branch or run the demo server
2. Open any snap detail page
3. Channel map actions should be `[ default_track/safest_risk_level  version_number ] [ Install ]` on every snap page

**NOTE:** Some snaps to test with `[node](https://snapcraft.io/node)` , `[skype](https://snapcraft.io/skype)`, `[blender](https://snapcraft.io/blender)`

🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

### DIFF
--- a/webapp/store/logic.py
+++ b/webapp/store/logic.py
@@ -294,7 +294,7 @@ def get_version(channel_map, track, risk):
     :param track: The track of the channel
     :param risk: The risk of the channel
 
-    :returns: The confinement
+    :returns: The version
     """
     for arch in channel_map:
         if arch in channel_map and track in channel_map[arch]:
--- a/webapp/store/logic.py
+++ b/webapp/store/logic.py
@@ -278,7 +278,7 @@ def get_confinement(channel_map, track, risk):
     :returns: The confinement
     """
     for arch in channel_map:
-        if arch in channel_map and track in channel_map[arch]:
+        if track in channel_map[arch]:
             releases = channel_map[arch][track]
             for release in releases:
                 if release["risk"] == risk:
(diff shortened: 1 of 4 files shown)


webapp/store/logic.py
```diff
@@ -285,3 +285,22 @@ def get_confinement(channel_map, track, risk):
...
+
+
+def get_version(channel_map, track, risk):
+    """Get the version for a channel
+
+    :param channel_map: Channel map list
+    :param track: The track of the channel
+    :param risk: The risk of the channel
+
+    :returns: The confinement
```
tbille: bad copy paste :P

webapp/store/logic.py
```diff
@@ -285,3 +285,22 @@ def get_confinement(channel_map, track, risk):
...
+    """Get the version for a channel
+
+    :param channel_map: Channel map list
+    :param track: The track of the channel
+    :param risk: The risk of the channel
+
+    :returns: The version
+    """
+    for arch in channel_map:
+        if arch in channel_map and track in channel_map[arch]:
```
bartaz: Isn't `if arch in channel_map` always true because you are iterating over `for arch in channel_map`?

therealjuan:
The underlying problem of what you describe is a lack of education towards basic concepts related to snaps somewhere in snapcraft.io: risk levels, channels... specially in cases where snaps have no stable releases. 

Let's open another issue to work on this as a `How it works` could be interested to include similar to what maas.io has to explain some concepts

webapp/store/logic.py
therealjuan: 👍 just updated

therealjuan: @tbille mentioned that these two functions to get confinement and get version can do with some refactoring

PR #1382, opened 2018-11-30
Remove useless character

tbille:
# Summary

Remove useless character in publicise page

# QA

Do we really need QA for this :smile:


bartaz:
Spacing seems broken. Some tags not closed properly? Or was it the unnecessary character that made it look good? :D

screen shot 2018-11-30 at 13 15 06

PR #1389, opened 2018-12-02
Adds “Past year” option

### LINKED ISSUE #1359: Show metrics older than 6 months
Following #589, we have further months of data available. Can you please add "one year" and "two years" options?

matthewpaulthomas:
Drops the “Past 3 months” option, leaving options for 7 days, 30 days, 6 months, and 1 year.

Fixes #1359.


bartaz:
Time axis gets really busy and hard to read with 1 year of data:
screen shot 2018-12-03 at 08 49 39

Also choosing "Past year" makes the data load for a longer while without any indication that something is happening. Some loading indication on top of the graph would be helpful.

Issue #1384, opened 2018-11-30
Release UI Redux TODO


bartaz:
Just a place to track progress on introducing Redux in release UI JS.

### TODO

- [ ] migrate other components/state ...
  - [x] `pendingCloses`
  - [ ] `isLoading`
  - [ ] `error`
  - [x] `tracks`
  - [x] `archs`
  - [x] `currentTrack`
- [ ] prepare initial state outside of ReleaseController component
- [ ] use `reselect` or similar for memoized selectors
- [ ] add `enzyme` for component testing
- [ ] add tests of UI components
- [x] move fetching of API responses to actions
- [ ] clean up state to only keep revisions data in one place (`revisions` state), not in channel map
- [ ] split large components into smaller ones with as little logic as possible (just render function with props)

### DONE
- [x] adds Redux and related dependencies (#1375)
- [x] make babel/eslint compile object spread operator { ...state } to make reducers nice (#1375)
- [x] move revisions list data from components into Redux state (#1375)
- [x] move actions to separate modules (#1386)
- [x] test actions and reducers (#1386)
- [x] create selectors for data used by ReleaseList component (#1404)
- [x] test selectors (#1404)
- [x] move all actions from ReleaseList component to redux actions (#1404)
- [x] move all the state from ReleaseList component to redux
  - [x] `revisionsMap` (#1375)
  - [x] `revisionsFilters` (#1393)
  - [x] `releases` (#1393)
  - [x] `selectedRevisions` (#1404)
  - [x] `pendingReleases` (#1412)
  - [x] `releasedChannels` (#1404)

PR #1404, opened 2018-12-04
Release UI redux - channel map

bartaz:
Part of redux migration #1384 

### Done:
- moved released channel map data into Redux
- removed passing unnecessary props (using data directly from redux store or selectors)
- fixed an issue that clicking on 'Show all revisions' was not closing the table

### QA
- ./run or https://snapcraft-io-canonical-websites-pr-1404.run.demo.haus/
- open Redux dev tools
- open unassigned revisions, select any revision
- selecting revision should appear as an action in redux devtools, state should change
- click around, release something, etc, everything should work the same
- make sure that you can close 'all revisions' table by clicking on 'Show all latest revisions' again

### DIFF
--- a/static/js/publisher/release/releasesController.js
+++ b/static/js/publisher/release/releasesController.js
@@ -46,7 +46,7 @@ class ReleasesController extends Component {
     const tracks = getTracksFromChannelMap(this.props.channelMapsList);
 
     this.state = {
-      // default to latest track
+      // use "latest" if default track is not defined
       currentTrack: this.props.options.defaultTrack || "latest",
       error: null,
       isLoading: false,
@@ -337,7 +337,7 @@ class ReleasesController extends Component {
     if (json.success) {
       if (json.closed_channels && json.closed_channels.length > 0) {
         json.closed_channels.forEach(channel => {
-          // make sure default channels get prefixed with 'latest'
+          // make sure channels without track name get prefixed with 'latest'
           if (channel.indexOf("/") === -1) {
             channel = `latest/${channel}`;
           }


static/js/publisher/release/releasesController.js
```diff
@@ -387,24 +335,16 @@ class ReleasesController extends Component {
...
-            if (channel.indexOf("/") === -1) {
-              channel = `latest/${channel}`;
-            }
-
-            delete releasedChannels[channel];
-          });
-        }
+      if (json.closed_channels && json.closed_channels.length > 0) {
+        json.closed_channels.forEach(channel => {
+          // make sure default channels get prefixed with 'latest'
```
Lukewh: not always true :s! Node for example the default isn't "latest"

We need to figure out a way to expose the 'default' throughout the app.

bartaz: @Lukewh I think (and hope) that this is a bit different meaning of 'default'. Why node is on different track by default in the store (they don't have latest) channel map API reports it fully as "10/stable" not as "stable".

The code here is for the case when just an risk name is used by API which I believe is only a case for latest track. So channel "beta" is equivalent of "latest/beta" always. But I will verify that.

bartaz: Yes. Looking at the channel map data for node it seems that channels in (default) track "10" are named fully as "10/stable" and channels in track "latest" use the short form of "stable".

```
"10": [{"channel": "10/stable" ...
"latest": [{"channel": "edge",  ...
```

bartaz: I'll update the comment not to use the word 'default' which indeed may be confused with other meaning of "default track" the snap uses.

PR #1405, opened 2018-12-05
Support markdown for description

### LINKED ISSUE #1388: Investigate Markdown limited support libraries
From a conversation posted in the forum of snapcraft.io, we would like to support the following text formatting on the Description field:

* Code (text blocks inside ` or ``` pairs)
* Lists (* Foo)
* Italics (_foo_)
* Bold (**foo**)
* Paragraph merging (consecutive lines are joined)
* Literal URLs auto-link https://foo.bar
* URLs with title [title for the link](https://foo.bar)

Would be good to research how this can be achieved from a back-end / front-end perspective

tbille:
# Summary

Fixes #1388 
Fixes https://github.com/canonicalltd/snap-squad/issues/776

Support markdown displaying in description page

Needs to be done:
- [x] sanitize code
- [x] add tests

# QA

- `./run`
- http://127.0.0.1:8004/toto
- check if markdown is well displayed in html
- try to modify the desciption of one of your snap and check if you have the markdown well displayed

make sure that only this list of markown elements works:

* Code (text blocks inside ` or ``` pairs)
* Lists (* Foo)
* Italics (_foo_)
* Bold (**foo**)
* Paragraph merging (consecutive lines are joined)
* Literal URLs auto-link https://foo.bar
* URLs with title [title for the link](https://foo.bar)

For each thing that is allowed (or not) please let me know I'll add it in the test suite

### DIFF
--- a/tests/tests_markdown_parser.py
+++ b/tests/tests_markdown_parser.py
@@ -1,0 +1,135 @@
+import unittest
+
+from webapp.markdown import parse_markdown_description
+
+
+class TestMarkdownParser(unittest.TestCase):
+    """This class tests the custom parser for the snap description. This
+    parser allows only a limited amount of tags. We want a lot of tests for
+    it to make sure on upgrades we don't loose the custom tags that we want
+    to keep.
+
+    List of approved markdown tag allowed:
+
+    * Code (text blocks inside ` or ``` pairs)
+    * Lists (* Foo)
+    * Italics (_foo_)
+    * Bold (**foo**)
+    * Paragraph merging (consecutive lines are joined)
+    * Literal URLs auto-link https://foo.bar
+    * URLs with title [title for the link](https://foo.bar)
+    """
+
+    def test_parse_title(self):
+        """Title conversion shouldn't work
+        """
+        markdown = "# title"
+        result = parse_markdown_description(markdown)
+        expected_result = "<p># title</p>\n"
+
+        assert result == expected_result
+
+    def test_parse_urls(self):
+        """Literal URLs auto-link https://foo.bar
+        """
+        markdown = "https://toto.space"
+        result = parse_markdown_description(markdown)
+        expected_result = (
+            '<p><a href="https://toto.space">https://toto.space</a></p>\n'
+        )
+
+        assert result == expected_result
+
+    def test_parse_urls_title(self):
+        """URLs with title [title for the link](https://foo.bar)
+        """
+        markdown = "[toto](https://toto.space)"
+        result = parse_markdown_description(markdown)
+        expected_result = '<p><a href="https://toto.space">toto</a></p>\n'
+
+        assert result == expected_result
+
+    def test_parse_italics(self):
+        """Italics (_foo_)
+        """
+        markdown = "_text_"
+        result = parse_markdown_description(markdown)
+        expected_result = "<p><em>text</em></p>\n"
+
+        assert result == expected_result
+
+    def test_parse_bold(self):
+        """Bold (**foo**)
+        """
+        markdown = "**text**"
+        result = parse_markdown_description(markdown)
+        expected_result = "<p><strong>text</strong></p>\n"
+
+        assert result == expected_result
+
+    def test_parse_paragraph_merging(self):
+        """Paragraph merging (consecutive lines are joined)
+        """
+        markdown = "this is\n a paragraph"
+        result = parse_markdown_description(markdown)
+        expected_result = "<p>this is\n a paragraph</p>\n"
+
+        assert result == expected_result
+
+    def test_parse_paragraph(self):
+        """Paragraphs
+        """
+        markdown = "paragraph 1\n\n paragraph 2"
+        result = parse_markdown_description(markdown)
+        expected_result = "<p>paragraph 1</p>\n<p>paragraph 2</p>\n"
+
+        assert result == expected_result
+
+    def test_parse_text(self):
+        """Text conversion works
+        """
+        markdown = "text"
+        result = parse_markdown_description(markdown)
+        expected_result = "<p>text</p>\n"
+
+        assert result == expected_result
+
+    def test_parse_code_block(self):
+        """Code (text blocks inside ` or ``` pairs)
+        """
+        markdown = "```code block```"
+        result = parse_markdown_description(markdown)
+        expected_result = "<p><code>code block</code></p>\n"
+
+        assert result == expected_result
+
+    def test_parse_code_line(self):
+        """Code (text blocks inside ` or ``` pairs)
+        """
+        markdown = "`code line`"
+        result = parse_markdown_description(markdown)
+        expected_result = "<p><code>code line</code></p>\n"
+
+        assert result == expected_result
+
+    def test_parse_list(self):
+        """Lists (* Foo)
+        """
+        markdown = "* item \n* item \n* item \n"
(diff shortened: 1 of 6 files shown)


therealjuan:
Here you have a few: [htop](https://snapcraft.io/htop), [slack-term](https://snapcraft.io/slack-term), [GitKraken](https://snapcraft.io/gitkraken), [Cacher](https://snapcraft.io/cacher)

We should include some help on the Listing page about which formatting could be added to the list... Is there any potential extension we could use to have the proposed markdown available as a WYSIWYG?

bartaz:
Again, empty paragraphs and broken custom 'lists':
https://snapcraft-io-canonical-websites-pr-1405.run.demo.haus/cacher

screen shot 2018-12-05 at 17 40 40

@therealjuan @Lukewh I think we need to decide what we do with possible broken styling.
We can try to implement some 'hacks' to make these legacy descriptions work or accept that we changed how they are displayed and expect maintainers to address the issues.
It would be good to announce the markdown support and maybe even contact the owners of problematic descriptions we see and ask them to update?

therealjuan:
Probably we might need this PR open so people from other teams can chip in and provide feedback. That is a great suggestion and we certainly need to contact snap publishers 👍

webapp/markdown.py
```diff
@@ -1,7 +1,38 @@
...
+from mistune import BlockGrammar, BlockLexer, Markdown, _pure_pattern
+import re
+
+
+class DescriptionGrammar(BlockGrammar):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.list_block = re.compile(
+            r"^( *)(?=[*+-•]|\d+\.)(([*+-•])?(?:\d+\.)?) [\s\S]+?"
```
bartaz: Can this be unit tested in some way?
We are adding a very specific custom functionality to markdown parser. This is a HUGE and complicated regexp that I'm not even trying to fully understand. It would be great to have some test cases showing what this thing is doing.

As far as I understand we are taking text input (description) and turning it into HTML. Can we test that particular text input (with different types of bullets) ends up in HTML we expect?

tests/tests_markdown_parser.py
```diff
@@ -0,0 +1,135 @@
...
+        """Code (text blocks inside ` or ``` pairs)
+        """
+        markdown = "`code line`"
+        result = parse_markdown_description(markdown)
+        expected_result = "<p><code>code line</code></p>\n"
+
+        assert result == expected_result
+
+    def test_parse_list(self):
+        """Lists (* Foo)
```
bartaz: Can we extend this test to cover our custom lists as well?

webapp/markdown.py
```diff
@@ -0,0 +1,54 @@
...
+            r"(?:"
+            r"\n+(?=\1?(?:[-*_•] *){3,}(?:\n+|$))"  # hrule
+            r"|\n+(?=%s)"  # def links
+            r"|\n+(?=%s)"  # def footnotes\
```
bartaz: These 2 lines seem identical - is this intended?

therealjuan:
We should have a think about how to style code blocks. Check out this example from [htop](https://snapcraft-io-canonical-websites-pr-1405.run.demo.haus/htop)

I'll add later on a snap description with some markdown :)

webapp/markdown.py
tbille: test added

tests/tests_markdown_parser.py
tbille: test added

webapp/markdown.py
tbille: I added a comment to explain where does it comes from :smile:

bartaz:
There is still an empty paragraph rendered in the beginning of description of every snap:

screen shot 2018-12-11 at 17 55 10

It causes margins to collapse and there is less spacing under heading then needed because of it.

It doesn't seem to be caused by any specific formating of particular snap. It happens in all snaps.
We should get rid of empty paragraphs in the description. If not all of them, at least this first one that affects all the snaps.

webapp/markdown.py
bartaz: Thanks!

webapp/markdown.py
bartaz: Thanks, that's helpful.

tbille:
@bartaz It is super weird because the parser is not generating this paragraph.... It seems Jinja2 is doing something weird.
This is the line we have in the template: `<p>{{ description | safe }}</p>`

If I transform it this way  `<div>{{ description | safe }}</div>` I don't have the problem anymore...

Is it a clean thing to do on your side to put the desciption into a `div` as a fix for our problem?

bartaz:
@tbille Good catch. Sure, it was issue in the template. We were putting `<p>`s inside `<p>`. Browser was closing first paragraph as empty.
It was working earlier because we were passing array of paragraphs to the template. So yes, turn it into a div. Or maybe we don't need this wrapping div at all.

bartaz:
Should Markdown images be allowed? They seem to be now:
https://snapcraft-io-canonical-websites-pr-1405.run.demo.haus/1140-grid-bartaz

therealjuan:
They shouldn't be allowed. Great GIF btw

tests/tests_markdown_parser.py
```diff
@@ -0,0 +1,155 @@
...
+        markdown = "1. item \n2. item \n3. item \n"
+        result = parse_markdown_description(markdown)
+        expected_result = (
+            "<ol>\n<li>item </li>\n<li>item </li>\n<li>item </li>\n</ol>\n"
+        )
+
+        assert result == expected_result
+
+    def test_parse_image_link(self):
+        """Image link is converted into a simple link
```
bartaz: Good we have this tested 👍

bartaz:
Link definitions/footnotes are not supported? Only inline links?

screen shot 2018-12-13 at 18 09 09

webapp/store/views.py
```diff
@@ -269,9 +273,8 @@ def snap_details(snap_name):
         if not details.get("channel-map"):
             flask.abort(404, "No snap named {}".format(snap_name))
 
-        formatted_paragraphs = logic.split_description_into_paragraphs(
-            details["snap"]["description"]
-        )
+        clean_description = bleach.clean(details["snap"]["description"])
+        formatted_description = parse_markdown_description(clean_description)
```
bartaz: Would it make sense to use `bleach.clean` on result of `parse_markdown_description`? This way we would be sanitizing HTML rather then Markdown.

bartaz: This may actually help getting rid of unwanted HTML (like ``), because even if Markdown produces image it will be removed by bleach.

tbille: @bartaz the problem is that the `clean` method escapes some tags that I don't want (`<p>` for ex) that are generated by the mardown parser...

bartaz: Ah, right. By default it only allows inline styling elements.

But, as we are so careful about what we allow and what not - we could easily create a custom config for bleach as well? To list what we allow from Markdown and what not?

tbille: I am not sure it's the right solution, I don't really want to add custom rules to every library that we use. It will be quite confusing in the future to maintain some custom stuff that we add...

tbille:
I am merging this to a feature branch that will need QA before proposing to master

PR #1406, opened 2018-12-05
Add flake8 linter to test suite

tbille:
# Summary

Add flake8 linter and fix linting problems

# QA

- Make sure circleCi tests pass
- `./run test`
- test should pass

### DIFF
--- a/tests/blog/tests_index.py
+++ b/tests/blog/tests_index.py
@@ -130,7 +130,8 @@ class BlogPage(TestCase):
     #     url = "".join([self.api_url, "/media/123"])
     #
     #     responses.add(
-    #         responses.GET, url, body=requests.exceptions.Timeout(), status=504
+    #         responses.GET, url, body=requests.exceptions.Timeout(),
+    #         status=504
     #     )
     #
     #     response = self.client.get("/blog")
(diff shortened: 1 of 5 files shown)


tests/blog/tests_index.py
```diff
@@ -130,7 +130,8 @@ def test_posts_timeout(self):
     #     url = "".join([self.api_url, "/media/123"])
     #
     #     responses.add(
-    #         responses.GET, url, body=requests.exceptions.Timeout(), status=504
+    #         responses.GET, url, body=requests.exceptions.Timeout(),
```
bartaz: Out of curiosity: why doesn't black catch this as well? Isn't it supposed to format stuff to fit in 80 chars?

webapp/blog/logic.py
```diff
@@ -40,7 +40,8 @@ def replace_images_with_cloudinary(content):
         "<img\g<prefix>"
         f' decoding="async"'
         f' src="{urls[1]}"'
-        f' srcset="{urls[0]} 350w, {urls[1]} 650w, {urls[2]} 1300w, {urls[3]} 1950w"'
+        f' srcset="{urls[0]} 350w, {urls[1]} 650w, {urls[2]} 1300w,'
```
bartaz: This long line is not in comment, so not sure why black ignores it. Anyway - doesn't matter that much.

PR #1407, opened 2018-12-05
Add counter for the number of requests on badge

### LINKED ISSUE #1394: Investigate how to get metrics on specific URLs / assets
In order to monitor how certain assets are requested overall, we would like to be able to get metrics on specific assets like snap store badges ....

tbille:
# Summary

Fixes #1394 
Add counter for the number of requests on badges for logged in users and not logged in users
If the requester has a session I don't increment the counter, increment the logged in counter

This will allow us to know how many users request the badge from snapcraft (logged in) and from outside (readme ...)

# QA

- `vim .env`
- add line `TALISKER_NETWORKS=172.17.0.1`
- `./run`
- http://0.0.0.0:8004/_status/metrics
- Should have a line: `badge_counter 0.0`
- Should have a line: `badge_logged_in_counter 0.0`
- In a private tab
- http://127.0.0.1:8004/static/images/badges/en/snap-store-black.svg
- GOTO http://0.0.0.0:8004/_status/metrics
- Should have a line: `badge_counter 1.0`
- Should have a line: `badge_logged_in_counter 0.0`
- http://127.0.0.1:8004/toto/publicise
- Should have a line: `badge_counter 1.0`
- Should have a line: `badge_logged_in_counter 14.0`
- This case is less going to happen, but when a user is logged in and access the badge, it will be counted as a logged in user, even from outside the website
- http://127.0.0.1:8004/static/images/badges/en/snap-store-black.svg
- Should have a line: `badge_counter 1.0`
- Should have a line: `badge_logged_in_counter 15.0`

### DIFF
--- a/tests/tests_badge_counter.py
+++ b/tests/tests_badge_counter.py
@@ -1,6 +1,6 @@
 from flask_testing import TestCase
 from webapp.app import create_app
-from webapp.handlers import badge_counter
+from webapp.handlers import badge_counter, badge_logged_in_counter
 
 
 class TestsBadgePrometheusCounter(TestCase):
@@ -9,6 +9,7 @@ class TestsBadgePrometheusCounter(TestCase):
     def setUp(self):
         self.endpoint_url = "/static/images/badges/en/snap-store-black.svg"
         badge_counter._value.set(0.0)
+        badge_logged_in_counter._value.set(0.0)
 
     def create_app(self):
         app = create_app(testing=True)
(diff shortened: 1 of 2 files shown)


bartaz:
Not sure if that's the behaviour we want.

1. Be logged in to snapcraft.io (have active session in browser)
2. In same window open a new tab with a badge image
3. It is not counted

It means that for developers who have snapcraft.io account and happen to be logged in to snapcraft.io in current browser session we will not count ANY badge requests. This will make the numbers wrong (showing too few requests compared to real load).

Can we somehow check the referrer of the request? Or is there maybe other way we can detect it?

In worst case we could add a query parameter to images we have on publicise page (`?internal=true`), so we can ignore these and count all the other.

bartaz:
LGTM.

While I think that it would be better to have more meaningful distinction between external badge requests (coming from outside of snapcraft.io) and internal (rather than current slightly inaccurate "logged in" vs "not logged in"), it's probably fine for our current need. We don't need accurate numbers, we just want to know the scale of requests we need to deal with.

PR #1414, opened 2018-12-07
Highlight current channel and arch in releases table

### LINKED ISSUE #1397: Highlight currently open channel and column

bartaz:
Fixes #1397 

Highlights currently open channel and architecture in releases table

### QA
- ./run or https://snapcraft-io-canonical-websites-pr-1414.run.demo.haus/
- click on released version to open release history
- currently open channel name and arch should be highlighted (in bold)
screen shot 2018-12-10 at 16 07 27

### DIFF
--- a/static/js/publisher/release/releasesTable.js
+++ b/static/js/publisher/release/releasesTable.js
@@ -139,14 +139,25 @@ class ReleasesTable extends Component {
       }
     }
 
+    const channelName =
+      risk === UNASSIGNED ? <em>Unreleased revisions</em> : channel;
+
+    const filteredChannel =
+      this.props.filters &&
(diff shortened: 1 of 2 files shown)


therealjuan:
It might be better if we can `font-weight:700` instead of `400` as it doesn't seem to be enough and that way matches the revision number being highlighted:

image

bartaz:
@therealjuan Yes, I've noticed that these 'bolds' seem to have different weight, but didn't want to go out of what Vanilla does. I'll have a look at overriding this.

therealjuan:
Another tiny change might be related to devmode (could be a separate issue). What about changing the warning icon to an info icon without colour instead as it becomes less dangerous?

Basically move from `/i\` to `[i]`

therealjuan:
Looks great!

PR #1279, opened 2018-11-06
Updated python flow

Lukewh:
## Done

Changes requested

## QA

- Pull the branch
- Visit http://0.0.0.0:8004/first-snap
- Follow the instructions and ensure everything works and looks correct

PR #1415, opened 2018-12-07
Bring back the MacOS instructions, now that snapcraft 3.0.1 is out

evandandrea:
Snapcraft is now working on MacOS without any fiddling, so let's re-enable the MacOS flow. I've fixed a few things along the way and cleaned up the text.

### DIFF
--- a/templates/first-snap/build.html
+++ b/templates/first-snap/build.html
@@ -11,7 +11,6 @@
       <div class="col-8">
         <ol class="p-list-step has-margin">
           {% for step in steps %}
-            {% if not loop.last %}
               <li class="p-list-step__item">
                 <h4 class="p-list-step__title">
                   <span class="p-list-step__bullet">{{ loop.index }}</span>
@@ -27,7 +26,6 @@
                   {% endif %}
                 {% endif %}
               </li>
-            {% endif %}
           {% endfor %}
           {% for step in steps %}
             {% if loop.last and step.warning %}
--- a/templates/first-snap/test.html
+++ b/templates/first-snap/test.html
@@ -17,24 +17,23 @@
                 {{ step.action|safe }}
               </h4>
               {% if step.command %}
-                {% set snippet_value = step.command %}
-                {% set snippet_id = loop.index %}
-                {% if snippet_value.count('\n') == 0 %}
-                  {% include "/partials/_code-snippet.html" %}
-                {% else %}
-                  {% include "/partials/_code-snippet-multi.html" %}
-                {% endif %}
+              {% set snippet_value = step.command %}
+              {% set snippet_id = loop.index %}
+              {% if snippet_value.count('\n') == 0 %}
+              {% include "/partials/_code-snippet.html" %}
+              {% else %}
+              {% include "/partials/_code-snippet-multi.html" %}
               {% endif %}
+              {% endif %}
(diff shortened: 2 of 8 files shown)


therealjuan:
LGTM but would be worth to look at why the coverage went down :)

templates/first-snap/build.html
```diff
@@ -11,7 +11,6 @@
       <div class="col-8">
         <ol class="p-list-step has-margin">
           {% for step in steps %}
-            {% if not loop.last %}
```
Lukewh: Surely this was here for a reason?
Do the other flows still work as they did before?

tbille:
@evandandrea can you rebase on master to fix the test coverage.
We are making upgrades on the tests we need to rebase often on PRs, it shouldn't be a problem soon :smile:

templates/first-snap/build.html
evandandrea: It's not clear to me what that reason is. Why skip the last item in the list of actions?

https://github.com/canonical-websites/snapcraft.io/pull/1279#issuecomment-436935795

Everything seems to work fine without it. Can you elaborate on its purpose?

templates/first-snap/test.html
```diff
@@ -17,24 +17,23 @@ <h4 class="p-list-step__title">
...
               </h4>
               {% if step.command %}
-                {% set snippet_value = step.command %}
-                {% set snippet_id = loop.index %}
-                {% if snippet_value.count('\n') == 0 %}
-                  {% include "/partials/_code-snippet.html" %}
-                {% else %}
-                  {% include "/partials/_code-snippet-multi.html" %}
-                {% endif %}
+              {% set snippet_value = step.command %}
```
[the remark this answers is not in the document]
evandandrea: Fixed in 8a4f688

evandandrea:
@tbille done

templates/first-snap/build.html
Lukewh: Ah, because the last item in the list was a warning, and the loop was making it so it was step "x" instead of being related to the step before.

See the python 'build' step with your change:
screenshot_2018-12-13 snapcraft - snaps are universal linux packages

This should probably be fixed outside of this PR

Lukewh: https://github.com/canonical-websites/snapcraft.io/issues/1439 to track this issue

sergiusens:
in snapcraft 3.1, installing snapcraft is all you will need as the installation of multipass is taken care of. With the approval of @Saviq I would prefer to install multipass from the homebrew cask instead of manually doing this for the time being

sergiusens:
side note, if I click on nodejs, I do not get instructions on how to setup whereas going down the go path I do, is that on purpose?

evandandrea:
>side note, if I click on nodejs, I do not get instructions on how to setup whereas going down the go path I do, is that on purpose?

Correct. We haven't made "first snap flows" of all the language guides yet.

evandandrea:
>in snapcraft 3.1, installing snapcraft is all you will need as the installation of multipass is taken care of. With the approval of @Saviq I would prefer to install multipass from the homebrew cask instead of manually doing this for the time being

We could do a follow-on PR to switch to the cask, unless you foresee problems with Multipass installed the old way?

PR #1431, opened 2018-12-12
Update styling of devmode not to look like warning

### LINKED ISSUE #1420: Release UI – Make devmode informative rather than an alert
In order to reduce the alertness of the Release UI interface when you have devmode revisions:

## Update warning style
image
When you have revisions in devmode in a released channel, there is a warning-style notification that can feel too intrusive as there is nothing wrong with what is included in the release UI. There are 3 different options we can do here:

- Update the general style so it uses a `[i]` icon instead of the warning
- Move the notification out of the top of the release map so it is shown only on the top of the expanded beta or edge channel containing devmode releases
- Include the content as part of a tooltip

## Icon style from warning to notification
Right now on release UI we show warning icons on devmode releases and they can look like the user has done something wrong where the goal should be to inform that the release is in devmode state.

We can use the `[i]` icon (`p-icon--information`) instead of a `/i\` warning icon.

bartaz:
Fixes #1420 

Changes icon of devmode revisions and moves notification into history panel.

### QA
- ./run or https://snapcraft-io-canonical-websites-pr-1431.run.demo.haus/
- go to Releases page of a snap with devmode revisions
- devmode icon should be "info" [i] instead of warning
- open history panel of some channel - if any of the revisions in history is in devmode notification should appear with more information about devmode
- if any revision in history panel is not in devmode no warning should be shown

screen shot 2018-12-12 at 11 27 14

### DIFF
--- a/static/js/publisher/release/components/releasesTableCell.js
+++ b/static/js/publisher/release/components/releasesTableCell.js
@@ -46,25 +46,33 @@ class ReleasesTableCell extends Component {
         </span>
         <span className="p-tooltip__message">
           {isPending && "Pending release of:"}
-          <div>
+
+          <div className="p-tooltip__group">
             Revision: <b>{revision.revision}</b>
-          </div>
-          <div>
+            <br />
             Version: <b>{revision.version}</b>
+            {isInDevmode(revision) && (
+              <Fragment>
+                <br />
+                {revision.confinement === "devmode" ? (
+                  <Fragment>
+                    Confinement: <b>devmode</b>
+                  </Fragment>
+                ) : (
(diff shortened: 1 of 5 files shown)


therealjuan:
In the case of a revision in `devmode` we might want to include the same tooltip as the one on the table:

image

image

Probably adding a tooltip to the disabled promote icon must be something we can add once we converge close and promote into one contextual menu with disabled actions.

therealjuan:
We discussed to add the following text to the cell tooltip:
`Revisions in devmode can't be promoted to stable or candidate channels`

PR #1434, opened 2018-12-12
Add link to listing page on own snaps

tbille:
# Summary

Add link to listing page if the snap belongs to you

# QA

- `./run`
- login
- Go on a snap that belongs to you
- Check if you have the link to the listing page under the title like this

image

- Go on  a snap that doen't belongs to you
- This link should exist

### DIFF
--- a/templates/store/snap-details.html
+++ b/templates/store/snap-details.html
@@ -55,10 +55,10 @@
           {% endif %}
           <div class="p-snap-heading__title">
             <h1 class="p-heading--two p-snap-heading__name">{{ snap_title }}</h1>
-            {% if is_users_snap %}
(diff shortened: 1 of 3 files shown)


therealjuan:
Great! but might be better to swap this content with the publisher name. i.e `You own this snap. [Edit snap information]`

tbille:
@therealjuan like this?

image

bartaz:
@therealjuan I'm not sure if removing "by [publisher]" is good. Because as publisher you will have no way to see how details page looks (with you publisher name) for other people when you are logged in. This may not seem much (just publisher name), but maybe they would notice that they want to change their displayed name or something.

So I'd be for adding the [Edit] link, but not removing any information from the page.

tbille:
@bartaz @therealjuan 

image

PR #1449, opened 2018-12-14
Add shared global nav

nottrobin:
But without the "logins" menu.

This will mirror the nav [just merged](https://github.com/canonical-websites/docs.snapcraft.io/pull/77) into docs.snapcraft.io.

Fixes https://github.com/canonicalltd/snap-squad/issues/793

QA
--

`./run`

Look at the global nav in all it's glory.


bartaz:
Global nav has different width then snapcraft nav:

screen shot 2018-12-14 at 11 50 37

bartaz:
On MacOS is you scroll up while being on top of the page (pulling the whole page down a bit in 'elastic scroll') you can pull out some global nav links hidden above the nav:

you can see a ghost of the global nav if you scroll up

bartaz:
When you open global nav 'Products' and scroll down the dark overlay appears but in only covers half of the page.

screen shot 2018-12-14 at 11 53 46

Lukewh:
Removes the link to "We're hiring". Would be good to have somewhere to put that link, rather than remove it completely.

bartaz:
Global nav renders it a very weird position on small screens:

screen shot 2018-12-14 at 12 01 46

Lukewh:
Strange horizontal rule to the right of the product list
screenshot_2018-12-14 snapcraft - snaps are universal linux packages

bartaz:
Whole page jumps down on page load when global nav bar is loaded/rendered.
Can't see this happening on ubuntu.com - is there something ubuntu.com does to prevent this?

nottrobin:
@bartaz for most of your points, could you check if you see the same issues on www.ubuntu.com? If so, can I suggest this is out-of-scope for this PR, and that this PR should be merged as is as long as the experience is as good www.ubuntu.com? If that's okay, then file bugs on https://github.com/canonical-webteam/global-nav for all these issues please.

bartaz:
@nottrobin Apart from 'scroll up on MacOS' these seem like snapcraft.io specific styling issues. They make the nav look broken, so I don't think we should merge.

Should we open these in global-nav repo?

bartaz:
@nottrobin For the dark overlay - I'm not sure how/why it works on other sites. A workaround I found to work on snapcraft.io is adding 'bottom: 0' to overlay styles (to ensure it starts from bottom, not relying on height):

screen shot 2018-12-14 at 13 00 27

bartaz:
Positioning on small screens:

It seems global nav on small screens positions itself on the bottom of the page (using 'position: absolute' - which is asking for the trouble). Not only it requires having proper margin on the bottom to accomodate for the navigation, it also assumes other styles on body will not break it. Which they are in snapcraft.io because of 'sticky footer'.
Because if 'sticky footer' we set height of `html` and `body` to 100% which then makes global-nav render in the bottom of the screen, not the bottom of the page.

This may be hard to fix on snapcraft.io side (other then removing sticky footer styles), so probably should be reported as global-nav issue.

nottrobin:
@bartaz how would you fix it in global nav? What's the better way to place the global-nav at the bottom of the screen with styles?

bartaz:
@nottrobin Not sure if this is doable (in a way that will always work) in styles only. As global nav is injected with JS into DOM, I would suggest injecting it to the end of the body on small screens and moving it to top of body on larger screens. (reported https://github.com/canonical-webteam/global-nav/issues/62)

nottrobin:
Okay, I've fixed the width (@bartaz) and move the `<script>` to the line @Lukewh suggested.

The other fixes seem more involved and will take longer.

@bartaz why is it that the `<body>` has `has-sticky-footer` class? What needs to be sticky? Could we explore other ways to achieve that stickiness?

bartaz:
@nottrobin sticky footer is used so that the footer appears on the bottom of the screen when there is very little content on the page. It started on snap details page (when we didn't have much content on them), but is still needed in other places (like account details):

screen shot 2018-12-14 at 13 48 54

We explored different ways to achieve this and this solution required least assumptions about markup, other styles, etc. It's even part of Vanilla now https://github.com/vanilla-framework/vanilla-framework/blob/develop/scss/_utilities_sticky-footer.scss

I'm not sure if it would be easy to find other way to have both sticky footer and correct positioning of global nav.

nottrobin:
Well this is a sticky situation.

bartaz:
@nottrobin I think I found a possible source of some issues (like strange border positioning).
We have some our custom 'global-nav' styles in snapcraft: https://github.com/canonical-websites/snapcraft.io/blob/master/static/sass/_global-nav.scss

We used them for our previous "global nav" with hiring link. Now that this is replaces this file can probably be removed.

nottrobin:
@Lukewh, @bartaz I believe all issues are now fixed. Please re-review? And maybe @therealjuan too?

bartaz:
@nottrobin Thanks. You didn't need to delete coverage comments. codecov updates it on pushes.

templates/_layout.html
```diff
@@ -47,6 +47,9 @@
 
     {% include "_header.html" %}
 
+    <script src="/static/js/modules/global-nav.js"></script>
```
bartaz: Does this need to be exactly here in HTML?

Ideally it should be moved to the bottom of the template where other scripts are - especially after raven is configured, so in case of any errors they would be reported to sentry.
Or even better integrated as a part of `base.js` not to be loaded separately.

If it is meant to run immediately then it probably should be called before `_header.html` is loaded (together with GA), not after it.

therealjuan:
A few things:

* Can we include the we're hiring link in the footer for the time being? We probably might need to add it somewhere else but would be better not to remove it completely

* The ghost navigations still persists with elastic scroll

screen shot 2018-12-14 at 11 55 38

* The mobile global nav in mobile has a weird layout and the copydoc warning moves inside it

image

*Is the global nav already added to build.snapcraft.io?

bartaz:
> The ghost navigations still persists with elastic scroll

This is a bug with global-nav (https://github.com/canonical-webteam/global-nav/issues/61) and hopefully will be addressed there. I don't think we should block this PR on that.

> The mobile global nav in mobile has a weird layout and the copydoc warning moves inside it

That's a clash of styling between the nav and browser plugin. So hopefully not a huge issue as the plugin is used by Canonical internally on desktop (not small screens).
I probably need to update the plugin to make it possible to hide the edit link.

templates/_layout.html
nottrobin: > Does this need to be exactly here in HTML?

I moved it up so as to avoid the jumping of the page that you pointed out. The way to get the most consistent experience for the user is to run `canonicalGlobalNav.createNav` as high up in the page as possible - this way the global nav will always display, so nothing will be surprising.

Another other option would to write some custom styles to push the page down by default, sort of leave a placeholder for the global nav to be loaded into, and then load the global nav JS at the bottom of the page to load when the browser gets around to it. There are downsides to this approach:

- Although there'd be no "jumping", the user would still see a flicker as the global nav appears, which would lead to a slight feeling of the page not being entirely solid
- If the global nav failed to load for some reason (e.g. connection issues, JS error, the user has JS disabled) there'd be a weird gap above the header
- It takes control away from the global nav - if we wanted to, for example, make the global nav taller, then we'd have to remember to update snapcraft.io codebase

In general I like the idea of JS loading as soon as it's needed, so anything that is displayed immediately should be run ASAP. As you pointed out already, pages that visibly change after loading create a pretty shaky user experience. So I think I'll leave it at the top if that's okay with you Bartek.

> If it is meant to run immediately then it probably should be called before _header.html is loaded (together with GA), not after it.

I've moved it above the `_header`.

bartaz: Sure, if having it up the page solves the loading jumping it's fine to be there. Just wanted to make it consistently before header (which probably was the intention).

nottrobin:
@therealjuan I've added the "We're hiring" link to the footer:

image

therealjuan:
Ok, I'll probably discuss with the relevant teams involved as we might want to promote the positions in the homepage somehow.

therealjuan:
Just created the new issue for build:
 https://github.com/canonical-websites/build.snapcraft.io/issues/1183

nottrobin:
@therealjuan yeah. build.snapcraft.io doesn't currently have a global nav - and the whole header looks fairly different from snapcraft.io. So I don't think this going live should be blocked on https://github.com/canonical-websites/build.snapcraft.io/issues/1183.

therealjuan:
Sure but we need to track the work :D

Issue #1462, opened 2018-12-18
TypeError: e.target.closest is not a function


Lukewh:
https://sentry.is.canonical.com/canonical/snapcraftio/issues/205/

```
TypeError: e.target.closest is not a function
  at None (./static/js/base/ga.js:37:18)
  at d (/static/js/modules/raven.min.js:2:4670)
```

PR #1463, opened 2018-12-18
GA event fix

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1462

### DIFF
--- a/static/js/base/ga.js
+++ b/static/js/base/ga.js
@@ -34,7 +34,12 @@ function triggerEvent(category, from, to, label) {
 
 if (dataLayer) {
   window.addEventListener("click", function(e) {
-    let target = e.target.closest("a");
+    let target = e.target;
+    if (!target || !e.target.closest) {
+      return;
+    }
+
+    target = e.target.closest("a");
     if (!target) {
       target = e.target.closest("button");
     }


static/js/base/ga.js
```diff
@@ -34,7 +34,12 @@ function triggerEvent(category, from, to, label) {
 
 if (dataLayer) {
   window.addEventListener("click", function(e) {
-    let target = e.target.closest("a");
+    let target = e.target;
+    if (!target || !e.target.closest) {
```
bartaz: We have a polyfill for `.closest` so not sure what that would fail.

This seems more like a workaround to avoid sentry errors, but if we are fine with dropping some GA events silently, then it's probably fine.

Lukewh: I was thinking the issue is actually the 'target' that's missing for some reason, though I'm not sure how, because the browsers it's happening on don't need the polyfill.

bartaz: If target was undefined error would probably say something like "cannot call `closest` of undefined" I guess.
Maybe target is defined but is not a proper DOM node for some reason? Or is a 'document' or something that doesn't have `closest`? Anyway...

Current fix will do the trick - errors will stop appearing, GA event is not called anyway if it errors. So if we are fine with that, it's good to go. Not sure if we want to spend time investigating more.

Lukewh: I guess the click could be on the `document` if someone is clicking randomly on the page? Or a `.textNode`? I think if we see a significant drop in GA events we should investigate further.

PR #1478, opened 2018-12-19
Filter unreleased revisions

### LINKED ISSUE #1473: Release UI – Filter unassigned revisions
To help publishers reduce the amount of clicks and selection to do with revisions relevant for them, we will add a contextual menu on unassigned revisions to help with choices.

**Contextual menu options**

image

Filter revisions
* **Recent revisions** – Revisions with most recent unassigned version not released and up to a week old will be shown here.
* **Unreleased revisions** – Revisions not released ever to any channel will appear in this list.
* **All revisions** – All revisions released or not from all tracks will be shown in the list.

This panel will work similarly to a contextual menu but will include a description tagline below the title.

bartaz:
Part of #1473

First part of work on selecting available revisions.
- adds simple dropdown with list of filters (Recent, Unreleased, All)
- choosing one of the then filters the list of available revisions that can be selected
- as 'All' revisions filter are now part of the dropdown in releases table, 'Show all revisions...' is removed from below the table

Demo: https://snapcraft-io-canonical-websites-pr-1478.run.demo.haus/

### QA
- ./run or https://snapcraft-io-canonical-websites-pr-1478.run.demo.haus/
- 'Unassigned' row should be separated from the rest of the page with dropdown available
- you should be able to select filter to revisions (Recent, Unreleased and All)
- list of releases available to select should change when filter is changed

More functionality and visual updates will be added as follow up.

screen shot 2018-12-20 at 11 24 25


therealjuan:
Looking good, might be a good idea to drop show latest revisions from the bottom as we have all now in the dropdown. Not sure, if adding the number of revisions in each option will be useful. The only use case I can think of is to be aware of changes on revisions when you push as a reference point.

static/js/publisher/release/actions/availableSelect.js
```diff
@@ -0,0 +1,8 @@
+export const SET_AVAILABLE_SELECT = "SET_AVAILABLE_SELECT";
```
Lukewh: Can we call these `*_AVAILABLE_REVISIONS_*` and `*AvailableRevisions*` throughout? `AVAILABLE` isn't very descriptive at a glance IMO.

bartaz: Ok, good point. I didn't want to make the variable names too long, but I guess readability is more important. Fixed.

PR #1479, opened 2018-12-19
Self-serve categories

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1458

## Done
- Added select boxes and interaction as described in https://docs.google.com/document/d/1Teovi34AVAkHtFdO9agsILqS96QVmaDqZ9DMKyFs48U/edit
- Updated the backend to transform the snap categories to avoid reserved jinja2 key

## QA
- Pull the branch
- `./run`
- Select a snap
- You should see a new `Category:` section
- Play around with the categories, adding categories and secondary categories
- You should not be able to select the same category for both the primary and secondary categories
- When you're happy click "Save"
- Visit http://0.0.0.0:8004/search?category=<category_name> for the category(s) you selected
- If you go to the last page you should see your snap (may require patience and refreshes)
- **WHEN DONE, REMOVE THE CATEGORIES FROM TEST SNAPS AS THEY WILL DISPLAY ON PRODUCTION**

### DIFF
--- a/static/js/publisher/form.js
+++ b/static/js/publisher/form.js
@@ -114,14 +114,15 @@ function initForm(config, initialState, errors) {
   let state = JSON.parse(JSON.stringify(initialState));
 
   const stateInput = document.createElement("input");
-  stateInput.type = "hidden";
   stateInput.name = "state";
+  stateInput.value = "";
 
   formEl.appendChild(stateInput);
 
   const diffInput = document.createElement("input");
   diffInput.type = "hidden";
   diffInput.name = "changes";
+  diffInput.value = "";
 
   formEl.appendChild(diffInput);
 
--- a/static/js/publisher/form.js
+++ b/static/js/publisher/form.js
@@ -114,6 +114,7 @@ function initForm(config, initialState, errors) {
   let state = JSON.parse(JSON.stringify(initialState));
 
   const stateInput = document.createElement("input");
+  stateInput.type = "hidden";
   stateInput.name = "state";
   stateInput.value = "";
 
--- a/static/js/publisher/form.js
+++ b/static/js/publisher/form.js
@@ -205,7 +205,7 @@ function initForm(config, initialState, errors) {
     if (formEl["license"]) {
       license(formEl);
     }
-    if (formEl["primary_category"]) {
+    if (formEl.elements["primary_category"]) {
       categories(formEl);
     }
 
--- a/static/js/publisher/form.js
+++ b/static/js/publisher/form.js
@@ -3,6 +3,7 @@ import { updateState, diffState } from "./state";
 import { publicMetrics } from "./market/publicMetrics";
 import { whitelistBlacklist } from "./market/whitelistBlacklist";
 import { initLicenses, license } from "./market/license";
+import { categories } from "./market/categories";
 
 // https://gist.github.com/dperini/729294
 // Luke 07-06-2018 made the protocol optional
@@ -202,6 +203,9 @@ function initForm(config, initialState, errors) {
     }
     if (formEl["license"]) {
       license(formEl);
+    }
+    if (formEl["primary_category"]) {
+      categories(formEl);
     }
 
     let formData = new FormData(formEl);
--- a/static/js/publisher/form.test.js
+++ b/static/js/publisher/form.test.js
@@ -40,11 +40,15 @@ describe("initForm", () => {
   let submitButton;
   let revertButton;
   let titleInput;
+  let summaryInput;
+  let descriptionInput;
+  let websiteInput;
+  let contactInput;
   let primaryCategoryInput;
   let secondaryCategoryInput;
   let categoriesInput;
 
-  function setupForm(config) {
+  function setupForm(config, initialState) {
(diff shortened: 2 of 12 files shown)


therealjuan:
Works like a charm! 👍

tbille:
@Lukewh 

The problem is that the API doesn't have a notion of Primary or Secondary but only alphabetical...

To reproduce the problem:

- https://snapcraft-io-canonical-websites-pr-1479.run.demo.haus/toto/listing

```
Primary: Social networking
Secondary: Productivity
```

- Save and reload
- The categories are swapped

tbille:
QA fun:

```
Primary: Productivity
Second: Social networking
```

Remove second category, and replace first category with the deleted category.

We get an error: `Category names not found: {''} `

static/js/publisher/form.js
```diff
@@ -113,14 +114,15 @@ function initForm(config, initialState, errors) {
   let state = JSON.parse(JSON.stringify(initialState));
 
   const stateInput = document.createElement("input");
-  stateInput.type = "hidden";
```
bartaz: Was it removed for debugging?

screen shot 2018-12-20 at 17 20 33

Lukewh: ahh good catch, didn't mean to remove it!

static/js/publisher/form.test.js
```diff
@@ -34,3 +35,236 @@ describe("initSnapIconEdit", () => {
...
+
+    beforeEach(() => {
+      setupForm(config, initialState);
+    });
+
+    afterEach(() => {
+      form.parentNode.removeChild(form);
+    });
+
+    test("creates state input", () => {
```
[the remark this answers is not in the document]
Lukewh: I try and avoid "should" to be a more definitive statement. like "form when initialized creates a state input" :man_shrugging:

static/js/publisher/form.test.js
```diff
@@ -34,3 +35,236 @@ describe("initSnapIconEdit", () => {
...
+    form.appendChild(websiteInput);
+    form.appendChild(contactInput);
+
+    document.body.appendChild(form);
+
+    categories.categories = jest.fn();
+    market.initForm(config, initialState, undefined);
+  }
+
+  describe("simple", () => {
```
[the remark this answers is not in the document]
Lukewh: badly named?

I was just hacking in some tests so codecov would chill out :upside_down_face:

static/js/publisher/form.test.js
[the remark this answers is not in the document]
Lukewh: https://github.com/canonical-websites/snapcraft.io/issues/1482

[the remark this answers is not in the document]
bartaz: It's not like we have any guidelines for that, so I don't mind. Would be good to have some consistency around that. Anyway, better to have some tests inconsistently named then none, so thanks for taking a time to add those! 👍

static/js/publisher/form.test.js
```diff
@@ -34,3 +35,236 @@ describe("initSnapIconEdit", () => {
...
+    form.appendChild(primaryCategoryInput);
+    form.appendChild(secondaryCategoryInput);
+    form.appendChild(summaryInput);
+    form.appendChild(descriptionInput);
+    form.appendChild(websiteInput);
+    form.appendChild(contactInput);
+
+    document.body.appendChild(form);
+
+    categories.categories = jest.fn();
```
bartaz: I'm not sure why this works at all.
We are importing a module, changing its contents to a mock and expecting that it will be mocked when imported by another module (form) that we imported even earlier.

The fact that it works means that `categories` is now mocked for anything that would try to import it. So in theory it could affect other tests that import `categories`. But for some reason it doesn't. Maybe it's because of the order tests are run in, maybe because of a way `jest` runs tests in paralel and they don't share `categories` import between tests but it is shared within the test here? - I'm not sure.

Long story short - I think it's working by accident and it's possible it may have unexpected side effects (may modify `categories` for other tests).
Mocking imports is unfortunately hard - we used https://github.com/thlorenz/proxyquire for that in BSI.

I think I'd rather remove this mock and test before it causes any trouble :)

templates/publisher/listing.html
```diff
@@ -140,6 +140,93 @@
           </div>
         </div>
 
+        <div class="row">
+          <div class="col-2">
+            <label>Category:</label>
+          </div>
+          <div class="col-4">
+            <input type="hidden" name="categories"
```
bartaz: Can't it be handled in the `state` directly? Do we need this additional hidden input for categories data?
We are pushing exactly the same data into state via JS couple of lines below in the template.

Lukewh: :+1: good point https://github.com/canonical-websites/snapcraft.io/issues/1483

static/js/publisher/form.test.js
Lukewh: https://github.com/canonical-websites/snapcraft.io/issues/1484

Lukewh: Thanks

PR #1485, opened 2018-12-21
Autoselect available revisions

bartaz:
Part of #1473

Autoselects latest revisions when available revisions dropdown value is changed

### QA
- run or demo
- 'Recent' available revisions should be selected by default (if there are any)
- if there are any recent revisions they will be selected by default (show in the table as 'orange')
- Recent option should select revisions only from one most recent version (not from all architectures)
- If there are no revisions in Recent or Unreleased these options should be disabled
- change Recent to 'Unreleased' or 'All'
- selected revisions should be updated based on the choice (note that for Recent and Unrelased selection will be the same, because Recent is subset of Unreleased)

Couple test snaps:
- snap with some recent revisions (not in all architectures): [nextcloud](https://snapcraft-io-canonical-websites-pr-1485.run.demo.haus/nextcloud/releases)
- snap with some unreleased revisions: [skype](https://snapcraft-io-canonical-websites-pr-1485.run.demo.haus/skype/releases)
- snap with just 'All' revisions (all are released): [slack](https://snapcraft-io-canonical-websites-pr-1485.run.demo.haus/slack/releases)

screen shot 2018-12-21 at 14 57 50

### DIFF
--- a/static/js/publisher/release/actions/availableRevisionsSelect.js
+++ b/static/js/publisher/release/actions/availableRevisionsSelect.js
@@ -1,8 +1,30 @@
+import { getArchsFromRevisionsMap } from "../releasesState";
+import { selectRevision } from "./channelMap";
+import { getSelectedAvailableRevisionsForArch } from "../selectors";
+
 export const SET_AVAILABLE_REVISIONS_SELECT = "SET_AVAILABLE_REVISIONS_SELECT";
 
 export function setAvailableRevisionsSelect(value) {
   return {
     type: SET_AVAILABLE_REVISIONS_SELECT,
     payload: { value }
+  };
+}
+
+export function selectAvailableRevisions(value) {
+  return (dispatch, getState) => {
+    dispatch(setAvailableRevisionsSelect(value));
+
+    // for each architecture
+    const archs = getArchsFromRevisionsMap(getState().revisions);
+    // get latest revision to select
+    archs.forEach(arch => {
+      const revisions = getSelectedAvailableRevisionsForArch(getState(), arch);
+
+      const revisionToSelect = revisions[0];
+      if (revisionToSelect) {
+        dispatch(selectRevision(revisionToSelect));
+      }
+    });
   };
 }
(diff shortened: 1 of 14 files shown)


bartaz:
- make "Recent" show only one version - Recent unreleased version
- unselect when switching to other filter
- disable options that are empty
- add 'title' to buttons to have browser tooltip

therealjuan:
The only piece of feedback is perhaps the tooltip on the releases table might say something like `Promote to other channels`

static/js/publisher/release/actions/availableRevisionsSelect.js
```diff
@@ -6,3 +14,34 @@ export function setAvailableRevisionsSelect(value) {
...
+
+export function selectAvailableRevisions(value) {
+  return (dispatch, getState) => {
+    dispatch(setAvailableRevisionsSelect(value));
+    dispatch(clearSelectedRevisions());
+
+    const state = getState();
+
+    // for each architecture
+    const archs = getArchitectures(getState());
```
Lukewh: pass `state` through rather then the `getState` func?

static/js/publisher/release/actions/availableRevisionsSelect.js
```diff
@@ -6,3 +14,34 @@ export function setAvailableRevisionsSelect(value) {
...
+    let revisionsFilter = () => true;
+
+    // for Recent select only revisions from most recent uploaded version
+    if (value === AVAILABLE_REVISIONS_SELECT_RECENT) {
+      // find most recent version
+      const recentVersion = getSelectedAvailableRevisions(state)[0].version;
```
Lukewh: A shallow look at `getSelectedAvailableRevisions` it seems `[0]` *could* be `undefined`?

bartaz: Good catch. Shouldn't happen in practice (as you cannot select option with 0 revisions), but it is indeed a possible state. Will fix that. Thx.

static/js/publisher/release/actions/availableRevisionsSelect.js
bartaz: Ah, missed that one.

PR #1493, opened 2018-12-21
Show categories as annotations on metrics

Lukewh:
## QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/<snap_name>/metrics or https://snapcraft-io-canonical-websites-pr-1493.run.demo.haus/
  (example https://snapcraft-io-canonical-websites-pr-1493.run.demo.haus/slack/metrics)
- Change the period of time a few times

### DIFF
--- a/static/sass/_snapcraft_metrics.scss
+++ b/static/sass/_snapcraft_metrics.scss
@@ -46,17 +46,4 @@
       pointer-events: none;
     }
   }
-
-  .snapcraft-metrics__active-devices {
-    &::after {
-      background: linear-gradient(90deg, $transparent-white, $opaque-white);
-      bottom: 0;
-      content: " ";
-      display: block;
-      position: absolute;
-      right: 0;
-      top: 0;
-      width: 1.5rem;
-    }
-  }
 }
--- a/static/sass/_snapcraft_metrics.scss
+++ b/static/sass/_snapcraft_metrics.scss
@@ -1,4 +1,7 @@
 @mixin snapcraft-metrics {
+  $opaque-white: rgba(255, 255, 255, 1);
+  $transparent-white: rgba(255, 255, 255, 0);
+
   .snapcraft-metrics__graph {
     opacity: 0;
     position: relative;
@@ -41,6 +44,19 @@
     svg {
       opacity: .4;
       pointer-events: none;
+    }
+  }
+
+  .snapcraft-metrics__active-devices {
+    &::after {
+      background: linear-gradient(90deg, $transparent-white, $opaque-white);
+      bottom: 0;
+      content: " ";
+      display: block;
+      position: absolute;
+      right: 0;
+      top: 0;
+      width: 1.5rem;
     }
   }
 }
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -3,8 +3,6 @@ from json import loads
 import flask
 
 import bleach
-import calendar
-import datetime
 import pycountry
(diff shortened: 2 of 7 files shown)


therealjuan:
This is amazing! Happy new year!

How about adding a ⭐️instead of featured with a tooltip stating `Your snap was featured on DD MMM 201Y`

therealjuan:
What about...

a) ... adding a tooltip stating `Your snap was featured on DD MMM 201Y` ?
b) ... add the category friendly name instead of the category name from the Store and the date in a tooltip like a) ?
c) ... return to a) and b), rinse and repeat?

Lukewh:
@therealjuan adding to the current tooltip, or adding another tooltip?

therealjuan:
Probably better a new tooltip on the `[star] Featured` label

bartaz:
I'm not sure about usefulness of showing 'past' labels like this:

screenshot 2019-02-19 at 15 23 43

At some point in time graphs will all look like this with categories long in the past.
It would maybe be useful to show a label if it was in next period (so it would show up if I choose next period from the dropdown). But showing 'productivity' on a 7d graph when it was set 2 years ago doesn't feel useful or informative.

therealjuan:
👍 on @bartaz

therealjuan:
me > -1

webapp/publisher/snaps/views.py
```diff
@@ -202,23 +204,32 @@ def publisher_snap_metrics(snap_name):
...
+                "name": category["name"],
+            }
 
-        single_series["values"][index_of_date] = 1
+            single_series["values"][index_of_date] = 1
 
-        annotations["series"].append(single_series)
+            annotations["series"].append(single_series)
+        else:
+            new_date = datetime.datetime.strptime(date, "%Y-%m-%d")
```
bartaz: I think we have some logic function to format date that is used somewhere (snap details?). Maybe we could reuse it.

Lukewh: Good catch, thanks @bartaz

therealjuan:
Do we know when a snap is not featured or not part of a category?

* 🗂️`Added to {Category name} in June 2018 / Removed in September 2018`
* ⭐`Featured snap since August 2018 `
* ⭐`Featured snap between August 2018 and September 2018 `

static/sass/_snapcraft_metrics.scss
```diff
@@ -43,4 +46,17 @@
       pointer-events: none;
     }
   }
+
+  .snapcraft-metrics__active-devices {
```
bartaz: Is this used anywhere?

webapp/publisher/snaps/views.py
```diff
@@ -202,6 +202,33 @@ def publisher_snap_metrics(snap_name):
...
+
+        if date not in annotations["buckets"]:
+            annotations["buckets"].append(date)
+
+        index_of_date = annotations["buckets"].index(date)
+
+        single_series = {
+            "values": [0] * (len(annotations)),
+            "name": category["name"],
+            "display_name": category["name"].capitalize().replace("-", " "),
```
bartaz: Not for this PR, but we need a common method to do that. It's already scattered across couple of places both in store and publisher parts.

static/sass/_snapcraft_metrics.scss
Lukewh: Ah, now I've made the names go to the left of the | it's not needed, removing thanks

PR #1497, opened 2019-01-07
Show descriptions in available revisions menu dropdown

### LINKED ISSUE #1473: Release UI – Filter unassigned revisions

bartaz:
Fixes #1473

Adds descriptions to "Available revisions" dropdown menu

### QA

- ./run or https://snapcraft-io-canonical-websites-pr-1497.run.demo.haus/
- go to Releases page 
- Open 'Available' menu
- descriptions should be visible

screenshot 2019-01-08 at 12 18 31

### DIFF
--- a/static/js/publisher/release/components/availableRevisionsMenu.js
+++ b/static/js/publisher/release/components/availableRevisionsMenu.js
@@ -14,9 +14,18 @@ import {
 } from "../constants";
 
 const menuLabels = {
-  [AVAILABLE_REVISIONS_SELECT_RECENT]: "Recent",
-  [AVAILABLE_REVISIONS_SELECT_UNRELEASED]: "Unreleased",
-  [AVAILABLE_REVISIONS_SELECT_ALL]: "All"
+  [AVAILABLE_REVISIONS_SELECT_RECENT]: {
+    label: "Recent",
+    description: "Revisions from past week not released to any channel"
+  },
+  [AVAILABLE_REVISIONS_SELECT_UNRELEASED]: {
+    label: "Unreleased",
+    description: "Revisions not released to any channel"
+  },
+  [AVAILABLE_REVISIONS_SELECT_ALL]: {
+    label: "All",
+    description: "All uploaded revisions"
+  }
 };
 
 export class AvailableRevisionsMenu extends Component {
@@ -59,7 +68,11 @@ export class AvailableRevisionsMenu extends Component {
         href="#"
         onClick={!isDisabled ? this.itemClick.bind(this, item) : undefined}
       >
-        {menuLabels[item]} <span className="u-float--right">({count})</span>
+        {menuLabels[item].label}{" "}
+        <span className="u-float--right">({count})</span>
+        <span className="p-contextual-menu__description">
+          {menuLabels[item].description}
+        </span>
       </a>
     );
   }
@@ -77,10 +90,11 @@ export class AvailableRevisionsMenu extends Component {
   render() {
     return (
       <ContextualMenu
-        label={menuLabels[this.props.value]}
+        label={menuLabels[this.props.value].label}
         className="p-select-button"
         ref={this.setMenuRef}
         position="left"
+        isWide={true}
       >
         {this.renderItems()}
       </ContextualMenu>
--- a/static/js/publisher/release/components/availableRevisionsMenu.js
+++ b/static/js/publisher/release/components/availableRevisionsMenu.js
@@ -16,15 +16,15 @@ import {
 const menuLabels = {
   [AVAILABLE_REVISIONS_SELECT_RECENT]: {
     label: "Recent",
-    description: "Revisions from past week not released to any channel"
+    description:
+      "Revisions from past week not released in any channel with most recent version"
   },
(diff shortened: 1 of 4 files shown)


therealjuan:
A few things to discuss

### Shall we update  copy to ... ?
```
Recent – Any revision from past week not released in any channel with most recent unassigned version
Not released – Revisions not released to any channel
All – All uploaded revisions released or not
```

### Close panel when interacting with other elements
Just noticed that when you open the panel and click on any of the arch + channel cells, the panel is still open so probably we would need to close it:
recent panel

### Similar style for selected revisions to promote

Wondering if we should make these two different use cases: when you promote from one chanenl to another channel and when you select a revision to be promoted from the revisions list

image

image

bartaz:
@therealjuan 

> Close panel when interacting with other elements
> Similar style for selected revisions to promote

These are not related to this PR, can we have them as separate issues?

> Recent – Any revision from past week not released in any channel with most recent unassigned version

Would 'with most recent version' be enough? Seems like 'unassigned' is not necessary - these are already 'not released' revisions.

> All – All uploaded revisions released or not

'All uploaded revisions' seems to mean exactly the same as 'All uploaded revisions released or not' :)
Unless we really want to emphasise that.

matthewpaulthomas:
- Clicking a disabled menu item scrolls to the top of the page, and tacks `#` onto the URL (I guess those are the same bug).
- The Promote button is nearly, but not quite, lined up with the ones above it. Could it line up exactly?
- It seems a wee bit overprotective to disable a filter that happens to display zero revisions right now. (I guess that not doing this would fix the top-scrolling problem, too.)

I haven’t used this page in real life, so I’m ignorant about what’s part of this PR and what isn’t. That said, I haven’t been able to work out:
- What the orange tint means.
- Why clicking “+ Add revision” doesn’t let me add a revision. (It just reveals an empty table.)
- Why clicking any cell in “Revisions available to promote” greys out the whole of the “Releases available to install” table.
- Why, despite the table being greyed out, every control in it is still clickable.

therealjuan:
> What the orange tint means.
Revision to promote aka highlighted

> Why clicking “+ Add revision” doesn’t let me add a revision. (It just reveals an empty table.)
Becuase you open the table before adding a revision

> Why clicking any cell in “Revisions available to promote” greys out the whole of the “Releases available to install” table.
To focus the interface on the elements that are important at that stage in time

> Why, despite the table being greyed out, every control in it is still clickable.
Because we are not disabling and emphasising, we are just reducing opacity

bartaz:
>> Why clicking any cell in “Revisions available to promote” greys out the whole of the “Releases available to install” table.
>
> To highlight on the interface on the elements that are important at that stage in time for the publisher
>
>> Why, despite the table being greyed out, every control in it is still clickable.
>
>Because we are not disabling and emphasising on the interface.

This is a remainder of the feature we had before in 'All revisions' table with all architectures in one list - when revision was selected we faded out revisions from the same architecture to emphasize other architectures.
Now it's totally unnecessary as we never show multiple architectures at once.

I think we should remove this fading out - possibly together with introducing 'yellowish' highlight to selected revision (as suggested earlier by @therealjuan)

bartaz:
@matthewpaulthomas 

> The Promote button is nearly, but not quite, lined up with the ones above it. Could it line up exactly?

screenshot 2019-01-07 at 16 37 46

They seem to be aligned perfectly (tested on Mac in Chrome and Firefox).

> It seems a wee bit overprotective to disable a filter that happens to display zero revisions right now. 

Possibly it is, but why should we keep an option enabled if it doesn't do anything?

> Clicking a disabled menu item scrolls to the top of the page, and tacks # onto the URL (I guess those are the same bug).

Yes, good catch. Even that an option is disabled it's still a link (with empty href no functionality).

therealjuan:
> These are not related to this PR, can we have them as separate issues?

https://github.com/canonical-websites/snapcraft.io/issues/1498
https://github.com/canonical-websites/snapcraft.io/issues/1499

therealjuan:
LGTM! Thanks 👍

static/sass/styles.scss
```diff
@@ -45,161 +45,196 @@ $color-navigation-background: #252525;
...
 
-@import 'snapcraft_p-code-snippet';
+@import "snapcraft_p-code-snippet";
 @include snapcraft-p-code-snippet;
 
-@import 'snapcraft-publicise';
+@import "snapcraft-publicise";
 @include snapcraft-publicise;
+
+$baseline-color: #f00 !default;
```
Lukewh: What's this stuff for? Maybe add a comment as we should avoid putting stuff in the main `styles.scss`

bartaz: Oh, that was for debugging the baseline spacing. Shouldn't be commited. Will remove it.

PR #1502, opened 2019-01-09
Select revisions with same version in other archs

### LINKED ISSUE #1472: Release UI: Select similar version in other architectures
Once a publisher selects a revision in an architecture we can prompt them to select automatically for them revisions with the same version across all other architectures that include revisions with the same version:

image

* By hovering on `3 other architectures` there will be a tooltip displaying information about the different architectures that include that version
* Version number should be highlighted somehow as we do when opening a release history
* The main call to action on the panel will be `Assign across architectures`

bartaz:
Fixes #1472

Adds possibility to select revisions with same version as currently selected in other architectures.

### QA

- ./run or https://snapcraft-io-canonical-websites-pr-1502.run.demo.haus/
- go to releases of any snap with multiple architectures
- open panel with available revisions
- check any revision that is also available in other architectures
- box should appear with information about same version in other architectures
- click on a button to select all revisions with same version in other archs

screenshot 2019-01-09 at 12 47 30

### DIFF
--- a/static/js/publisher/release/actions/availableRevisionsSelect.js
+++ b/static/js/publisher/release/actions/availableRevisionsSelect.js
@@ -2,8 +2,8 @@ import { AVAILABLE_REVISIONS_SELECT_RECENT } from "../constants";
 import { selectRevision, clearSelectedRevisions } from "./channelMap";
 import {
   getArchitectures,
-  getSelectedAvailableRevisions,
-  getSelectedAvailableRevisionsForArch
+  getFilteredAvailableRevisions,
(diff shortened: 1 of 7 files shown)


Lukewh:
It seems that when there are multiple versions with the same revision but only for 1 arch the notification is shown. Not sure if this is intentional or it should only show different archs available?

screenshot_2019-01-09 releases and revision history for surl
(the 1 other architecture is amd64)

therealjuan:
What Luke said! 👍

PR #1503, opened 2019-01-09
Improve the layout of videos on snap details pages

Lukewh:
## Done
- Implemented sidebar style screenshot list with video to the left
- No screenshots = video centered
- No video = original carousel
- Mobile = screenshot sidebar drops down to a horizontal, wrapping list below the video
- asciinema has background colour for narrower "videos"

## QA
- Pull the branch
- `./run`
- Visit a publisher listing page
- Add screenshots, no video
- View the public page on both desktop and mobile sizes
- Remove all screenshots
- Add a video
- View the public page on both desktop and mobile sizes
- Add 1 screenshot
- View the public page on both desktop and mobile sizes
- Add additional screenshots and check the public page

- Add and remove screenshots and different sources of video - asciinema, youtube and vimeo

## Screenshots
**Not sure why there are borders around the videos in the screenshots, they're fine on my screen - maybe something to do with the resolution + zoom?**
screenshot_2019-01-09 install test-lukewh for linux using the snap store snapcraft
screenshot_2019-01-09 install test-lukewh for linux using the snap store snapcraft 1
screenshot_2019-01-09 install test-lukewh-2 for linux using the snap store snapcraft
screenshot_2019-01-09 install test-lukewh-2 for linux using the snap store snapcraft 1
screenshot_2019-01-09 install lukotron testing for linux using the snap store snapcraft
screenshot_2019-01-09 install lukotron testing for linux using the snap store snapcraft 1


bartaz:
When there is just a video it is not centered (renders misaligned to the right side) - happens to all media types:

https://snapcraft-io-canonical-websites-pr-1503.run.demo.haus/1140-grid-bartaz

screenshot 2019-01-09 at 16 57 47

screenshot 2019-01-09 at 16 55 32
screenshot 2019-01-09 at 16 54 59

templates/partials/_carousel-videos-layout.html
```diff
@@ -0,0 +1,16 @@
+<div class="p-carousel p-carousel--videos u-equal-height">
```
bartaz: It's a bit misleading that CSS class is called `p-carousel` when in case of layout with video we don't have any carousel at all.

static/sass/_patterns_carousel.scss
```diff
@@ -25,58 +26,106 @@
       text-align: center;
 
       img,
-      video,
-      iframe {
+      video {
         display: block;
-        margin-bottom: 0;
+        margin-bottom: 0 !important;
```
bartaz: Why `!important`? If this is required to override something (third party styles?) we should have a comment here about it.

static/sass/_patterns_carousel.scss
```diff
@@ -25,58 +26,106 @@
...
+        .p-carousel__item-screenshot img,
+        .p-carousel__item-screenshot video {
+          align-self: center;
+          width: 100%;
+        }
+
+        @media screen and (max-width: $breakpoint-medium) {
+          flex-direction: row;
+          flex-wrap: wrap;
+          justify-content: flex-start !important;
```
bartaz: Is this an override? In line 42 it's `flex-start` as well, so is this needed at all?

static/sass/_patterns_carousel.scss
```diff
@@ -25,58 +26,106 @@
...
+          margin-bottom: -$spv-inter--regular;
+          margin-right: -$sph-inter;
+          margin-top: $spv-inter--regular;
+
+          .p-carousel__item--screenshot {
+            margin-bottom: $spv-inter--regular !important;
```
bartaz: So many `!important`s... Is there something that changes the margin, other styles get applied here as well?

bartaz:
Possibly not related to this PR, but in Chrome vimeo throws a JS error on autoplay:

screenshot 2019-01-10 at 09 15 08

bartaz:
There are quite a lot of `!important` in CSS. Usage of `!important` could be accepted if it is required to override some third party styles, but it such cases it should be clearly commented why it is used.

But it seems that in some cases it's our styles that we try to override with `!important` and we should avoid it. Using `!important` may lead maintenance issues later. If `!important` is needed to override some of our own styles it's likely the sign that we need to restructure the CSS of given component.

templates/partials/_carousel-videos-layout.html
Lukewh: Updated!

static/sass/_patterns_carousel.scss
Lukewh: Updated!

static/sass/_patterns_carousel.scss
Lukewh: Updated!

static/sass/_patterns_carousel.scss
Lukewh: Updated :D

static/sass/_snapcraft_snap-details.scss
```diff
@@ -1,9 +1,91 @@
...
+          margin-bottom: 0;
+        }
+      }
+    }
+
+    iframe {
+      margin-bottom: 0;
+    }
+
+    .asciicast {
```
bartaz: What element has this class? Is it ours or rendered by asciinema? I can't find in in a DOM (even when I have asciinema embedded).

Lukewh: Yeah it's the embedded asciinema holder that gets added just after the script tag (has the id `asciicast-container-{id}`

bartaz: Ah, found it now. I had a typo in selector. 'Asciinema' etc are so "human friendly" names... 😄

therealjuan:
LGTM thanks!

PR #1505, opened 2019-01-10
Highlight selected revisions in same colour as pending revisions

### LINKED ISSUE #1498: Release UI: Similar style for selected revisions to promote
Wondering if we should make these two different use cases: when you promote from one chanenl to another channel and when you select a revision to be promoted from the revisions list

image

image

bartaz:
Fixes #1498 

Use the same highlight style for selected revisions as for pending revisions.
Also removes now unnecessary disabled style.

### QA

- ./run or  https://snapcraft-io-canonical-websites-pr-1505.run.demo.haus/
- go to Releases page for any snap
- open a panel with available revisions
- click on any revision to select it
- it should appear with orange background

screenshot 2019-01-10 at 12 24 58

### DIFF
--- a/static/js/publisher/release/revisionsList.js
+++ b/static/js/publisher/release/revisionsList.js
@@ -42,19 +42,10 @@ class RevisionsList extends Component {
       : new Date(revision.created_at);
     const isSelected = this.props.selectedRevisions.includes(revision.revision);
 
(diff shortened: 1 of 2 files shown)


therealjuan:
LGTM!

PR #1507, opened 2019-01-10
Improve code styling on descriptions

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1460
Simple improvement for code and code block styling within descriptions

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/lukewh-test (which uses the copy from /htop as it *could* appear)

### DIFF
--- a/static/sass/_snapcraft_snap-details.scss
+++ b/static/sass/_snapcraft_snap-details.scss
@@ -1,30 +1,4 @@
 @mixin snapcraft-snap-details {
-
-  .snapcraft-snap-details__description {
-    code {
-      background: $color-mid-dark;
-      border-radius: $border-radius;
-      color: $color-x-light;
-      padding: .12rem .25rem;
-    }
-
-    pre {
-      background: $color-mid-dark;
-      border: 0;
-      border-radius: $border-radius;
-      color: $color-x-light;
-      // For some reason Vanilla treats pre padding/margin differently
-      // than `p`'s this remedies it for better flow
-      margin-bottom: 1.1rem;
-      padding-top: .4rem;
-
-      code {
-        border-radius: 0;
-        padding: 0;
-      }
-    }
-  }
-
   @media screen and (max-width: $breakpoint-small) {
     .snapcraft-snap-details {
       &__snippet {
--- a/static/sass/_snapcraft_snap-details.scss
+++ b/static/sass/_snapcraft_snap-details.scss
@@ -1,4 +1,30 @@
 @mixin snapcraft-snap-details {
+
+  .snapcraft-snap-details__description {
+    code {
+      background: $color-mid-dark;
+      border-radius: $border-radius;
+      color: $color-x-light;
+      padding: .12rem .25rem;
+    }
+
+    pre {
+      background: $color-mid-dark;
+      border: 0;
+      border-radius: $border-radius;
+      color: $color-x-light;
+      // For some reason Vanilla treats pre padding/margin differently
+      // than `p`'s this remedies it for better flow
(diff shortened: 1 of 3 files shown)


bartaz:
Why are we introducing yet another style for code?

screenshot 2019-01-11 at 09 40 52

Vanilla has a code block pattern:
screenshot 2019-01-11 at 09 41 54

In Snapcraft we already have code in other places that is not a Vanilla code block.
On home page first snap flow we have coloured .yaml code:
screenshot 2019-01-11 at 09 43 54
In other parts of first snap flow we use similar pattern, but not coloured:
screenshot 2019-01-11 at 09 45 18

That's a lot of custom styles for a code block.
If we need a custom style that is not Vanilla we should at least try to be consistent within Snapcraft.

therealjuan:
Perhaps we can have a quick review from @lyubomir-popov to decide on one style?

lyubomir-popov:
@therealjuan @Lukewh @bartaz Please use the vanilla code block / code snippet everywhere.

therealjuan:
Any particular reason why we can't colour syntax within a code block to help readability?

lyubomir-popov:
@therealjuan good point. A closed issue on vanilla points to a recommended [solution](https://github.com/canonical-webteam-archive/vanilla-docs-theme/issues/32) - [discussion](https://github.com/vanilla-framework/vanilla-framework/issues/817).  @Lukewh would that work here?

therealjuan:
Definitively does not look like something that will be used enough to consider its 4 mb. Perhaps we can use the vanilla code block and add a way to highlight something, it could either be by using bold or the green highlight which is already implemented?

lyubomir-popov:
Makes sense.  @Lukewh what is the file size of this dependancy? (You say it increases the size to 4MB). Asking so I can reopen the original issue and comment that its size is prohibitive and we need another solution.

lyubomir-popov:
As agreed on irc with @therealjuan let's leave syntax highlighting out until we have a dependancy that matches our size requirements. @Lukewh please advise on what that file size limit should be so I can reopen the original thread.

Lukewh:
This discussion is still irrelevant to this PR IMO. The example shown is of YAML. Why would an end user need to copy and paste YAML from the marketing page of a snap?

How many times have you needed to copy and paste some YAML from the store page when installing an app from the Apple Appstore, or Google Play?

Server software might want to direct users in how to use the app. Let's take plex as an example, on [their github page for the docker version](https://github.com/plexinc/pms-docker) they have code blocks - it's not highlighted because there's no way github can know the semantics of plex or docker!

The filesize limit should be "as low as possible while implementing the features we need".

lyubomir-popov:
@Lukewh I don't mind lack of highlighting if that is not a requirement. Sounds like this is the case here.

My only 2 points were:

1)  if we use syntax highlighting _anywhere_ on snapcraft.io, it should match and update (as in, when the theme used in the dependency gets updated) in sync with any other syntax highlighted block on other canonical websites.

2) If we have a valid reason why we can't use that recommended solution in the linked WG [discussion](https://github.com/vanilla-framework/vanilla-framework/issues/817), then I would like to raise a Vanilla issue, so that eventually we get a solution that works.

bartaz:
This PR is only about adding code style to snap details page and it should remain so - without any highlighting.

The only place where we "highlight" code is yaml examples on the home page. It was simply manually made in HTML - any other automatic method would be an overkill for that.

If we want to update the style of code examples in first snap flow (home page and tutorials) it should be raised in separate issue.

PR #1508, opened 2019-01-14
Show available stores on snap registration page

Lukewh:
Fixes https://github.com/canonicalltd/snap-squad/issues/840
Fixes https://github.com/canonicalltd/snap-squad/issues/841
Fixes https://github.com/canonicalltd/snap-squad/issues/842

# QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/register-snap
- If you don't have access to any stores, you should see the original page
- If you do have stores, you should see a dropdown that contains the available store, including the Global store

### No stores
screenshot_2019-01-14 snapcraft - snaps are universal linux packages 1

### Stores
screenshot_2019-01-14 snapcraft - snaps are universal linux packages

### DIFF
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -639,6 +639,10 @@ def redirect_get_register_name():
 @publisher_snaps.route("/register-snap")
 @login_required
 def get_register_name():
+    user = api.get_account(flask.session)
+
+    available_stores = logic.filter_available_stores(user["stores"])
+
     snap_name = flask.request.args.get("snap_name", default="", type=str)
     is_private_str = flask.request.args.get(
         "is_private", default="False", type=str
@@ -665,6 +669,7 @@ def get_register_name():
         "is_private": is_private,
         "conflict": conflict,
         "already_owned": already_owned,
+        "available_stores": available_stores,
     }
     return flask.render_template("publisher/register-snap.html", **context)
 
@@ -695,6 +700,10 @@ def post_register_name():
             registrant_comment=registrant_comment,
         )
     except ApiResponseErrorList as api_response_error_list:
+        user = api.get_account(flask.session)
+
+        available_stores = logic.filter_available_stores(user["stores"])
+
         if api_response_error_list.status_code == 409:
             for error in api_response_error_list.errors:
                 if error["code"] == "already_claimed":
@@ -723,6 +732,7 @@ def post_register_name():
         context = {
             "snap_name": snap_name,
             "is_private": is_private,
+            "available_stores": available_stores,
             "errors": api_response_error_list.errors,
         }
 
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -639,15 +639,14 @@ def redirect_get_register_name():
 @publisher_snaps.route("/register-snap")
 @login_required
 def get_register_name():
-    user = api.get_account(flask.session)
+    try:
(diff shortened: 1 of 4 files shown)


webapp/publisher/snaps/views.py
```diff
@@ -677,6 +690,18 @@ def redirect_post_register_name():
...
+        # We don't use the data from this endpoint.
+        # It is mostly used to make sure the user has signed
+        # the terms and conditions.
+        user = api.get_account(flask.session)
+    except ApiResponseErrorList as api_response_error_list:
+        return _handle_error_list(api_response_error_list.errors)
+    except ApiError as api_error:
+        return _handle_errors(api_error)
+
+    available_stores = logic.filter_available_stores(user["stores"])
```
tbille: No need to run this function if the request works. Put it after line 707

webapp/publisher/snaps/views.py
```diff
@@ -639,6 +639,18 @@ def redirect_get_register_name():
 @publisher_snaps.route("/register-snap")
 @login_required
 def get_register_name():
+    try:
+        # We don't use the data from this endpoint.
+        # It is mostly used to make sure the user has signed
+        # the terms and conditions.
```
tbille: this is not really true you use at this line: `available_stores = logic.filter_available_stores(user["stores"])`

webapp/publisher/snaps/views.py
```diff
@@ -677,6 +690,18 @@ def redirect_post_register_name():
 @publisher_snaps.route("/register-snap", methods=["POST"])
 @login_required
 def post_register_name():
+    try:
+        # We don't use the data from this endpoint.
+        # It is mostly used to make sure the user has signed
+        # the terms and conditions.
```
tbille: Can you give me a use case that would make it fail?
This is handled at login. When the user logins he should be redirected before even getting into this function thanks to the annotation `@login_required`

webapp/publisher/snaps/views.py
Lukewh: :+1:

webapp/publisher/snaps/views.py
Lukewh: :+1:

webapp/publisher/snaps/views.py
Lukewh: :+1:

therealjuan:

webapp/publisher/snaps/views.py
```diff
@@ -639,6 +639,10 @@ def redirect_get_register_name():
 @publisher_snaps.route("/register-snap")
 @login_required
 def get_register_name():
+    user = api.get_account(flask.session)
```
tbille: please suround with `try...expect`

webapp/publisher/snaps/views.py
```diff
@@ -695,6 +700,10 @@ def post_register_name():
             registrant_comment=registrant_comment,
         )
     except ApiResponseErrorList as api_response_error_list:
+        user = api.get_account(flask.session)
```
tbille: please suround with `try...expect`

webapp/publisher/snaps/views.py
Lukewh: I went too far the other way!

webapp/publisher/snaps/views.py
Lukewh: I went too far the other way!

PR #1515, opened 2019-01-16
Only show revision version selector in available revisions panel

bartaz:
Fixes an issue where same version selector was shown in history panel of channels, when it should be only visible with available revisions.

Reported by @therealjuan via IRC:

selector-bug

### Code review

Diff looks bigger because of indent whitespace changes. Review with 'hide whitespace' setting to reveal true diff:

screenshot 2019-01-17 at 09 09 38

### QA
- ./run or demo
- go to Release page of any snap with multiple architectures
- open available revisions and select some revision (where same version is available in other architectures)
- a box should appear telling that same version is available in other architectures
- open history panel of some channel (beta, etc) of the same architecture
- a box about same version should not be visible there

### DIFF
--- a/static/js/publisher/release/revisionsList.js
+++ b/static/js/publisher/release/revisionsList.js
@@ -259,31 +259,32 @@ class RevisionsList extends Component {
             .
           </Notification>
         )}
-        {selectedVersionRevisions.length > 0 && (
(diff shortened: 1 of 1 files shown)


therealjuan:

PR #1519, opened 2019-01-17
Available markdown

Lukewh:
It's nice to show the markdown available :).

# QA
- Pull the branch
- `./run`
- Scroll down to the description
- Click the link below the textarea to see the available markdown

### DIFF
--- a/static/js/publisher/market/markdown.js
+++ b/static/js/publisher/market/markdown.js
@@ -5,5 +5,10 @@ export default function() {
   markdownLink.addEventListener("click", e => {
     e.preventDefault();
     markdownInfo.classList.toggle("u-hide");
+    if (markdownInfo.classList.contains("u-hide")) {
+      markdownLink.innerHTML = markdownLink.innerHTML.replace("Hide", "Show");
+    } else {
+      markdownLink.innerHTML = markdownLink.innerHTML.replace("Show", "Hide");
+    }
   });
 }
--- a/static/js/publisher/market/markdown.js
+++ b/static/js/publisher/market/markdown.js
@@ -2,12 +2,8 @@ export default function() {
   const markdownLink = document.querySelector(".js-toggle-markdown");
   const markdownInfo = document.querySelector(".js-markdown");
 
-  markdownLink.addEventListener("click", function(e) {
+  markdownLink.addEventListener("click", e => {
     e.preventDefault();
-    if (markdownInfo.classList.contains("u-hide")) {
-      markdownInfo.classList.remove("u-hide");
-    } else {
-      markdownInfo.classList.add("u-hide");
-    }
+    markdownInfo.classList.toggle("u-hide");
   });
 }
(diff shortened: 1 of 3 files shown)


static/js/publisher/market/markdown.js
```diff
@@ -0,0 +1,13 @@
+export default function() {
+  const markdownLink = document.querySelector(".js-toggle-markdown");
+  const markdownInfo = document.querySelector(".js-markdown");
+
+  markdownLink.addEventListener("click", function(e) {
```
hatched: This can be an arrow function

static/js/publisher/market/markdown.js
```diff
@@ -0,0 +1,13 @@
...
+    e.preventDefault();
+    if (markdownInfo.classList.contains("u-hide")) {
```
hatched: This is no longer necessary as `classList` has a `toggle` method.

https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

therealjuan:
A few things:

* Add top margin to `Supported markdwon syntax`
* Change to `Show supported markdown syntax | hide markdown syntax`
* Showcase the syntax like:

templates/publisher/listing.html
```diff
@@ -257,6 +257,35 @@
...
+              </div>
+              <div class="col-4">
+                <p><small>
+                  <b>Bold:</b> <code>**Foo**</code>
+                </small></p>
+                <p><small>
+                  <b>Literal URLs:</b> <code>https://foo.bar</code>
+                </small></p>
+                <p><small>
+                  <b>URLs with title:</b> <code>[foo bar](https://foo.bar)</code>
```
bartaz: Why not just call them "links"?

static/js/publisher/market/markdown.js
```diff
@@ -0,0 +1,14 @@
...
+  const markdownLink = document.querySelector(".js-toggle-markdown");
+  const markdownInfo = document.querySelector(".js-markdown");
+
+  markdownLink.addEventListener("click", e => {
+    e.preventDefault();
+    markdownInfo.classList.toggle("u-hide");
+    if (markdownInfo.classList.contains("u-hide")) {
+      markdownLink.innerHTML = markdownLink.innerHTML.replace("Hide", "Show");
+    } else {
+      markdownLink.innerHTML = markdownLink.innerHTML.replace("Show", "Hide");
```
bartaz: This solution heavily relies on the HTML content and the structure of the link label. Someone later may change the "Show supported markdown syntax" to something else and this will silently stop working.

Would be better if this would set whole link label to "Show supported ..." / "Hide supported ...", so at least when searching project for "Show supported markdown" it would show up.

Or - ideally - don't include it in HTML template at all, just render in JS. But that would be better if the page was rendering in React. I won't mind if we skip this here for sake of this PR.

bartaz:
Syntax help seems weirdly ordered (when displayed in 2 columns):

screenshot 2019-01-18 at 09 46 59

It starts with code (is it most common and important)?

Text formatting (bold and italics) are on completely opposite sides (bottom left / top right) - not one under another, not one next to other - I would expect to see them close.

PR #1520, opened 2019-01-18
 Markdown support for snap description

Lukewh:
Add markdown support for description

## QA
- Pull the branch
- `./run`
- Edit the description with some markdown, support and unsupported
- Save your changes and visit http://0.0.0.0:8004/<snap_name>
- Ensure the markdown looks and works as expected

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ Flask-WTF==0.14.2
 humanize==0.5.1
 Markdown==3.0.1
 prometheus_client==0.3.1
+mistune==0.8.4
 pybreaker==0.4.4
 pycountry==17.9.23
 pymacaroons==0.12.0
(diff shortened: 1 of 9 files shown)


evandandrea:
Hold fire on deployment until we get a +1 on the plan from @cprov

Lukewh:
There seems to be an issue that any sentence that starts with a single character is transformed to a list.

`A powerful...` becomes:
- powerful...

This must be an issue with the custom regex list parser that was added

nottrobin:
@tbille so, in a `[...]` range, the dash character `-` denotes a range - as in `[a-z]`, `[0-9]`. I believe what's happening here is that it's treating `[+-•]` as a range, so it's probably interpreting that to mean any character codes between those of `+` and `•`.

And so, if you want a `-` to be interpreted literally, you should either escape it (`[*+\-•]`) or put it at the beginning or the end of the list. Escaping it is probably a clearer indication of your intent.

PR #1534, opened 2019-01-25
Use entire word for months names

tbille:
# Summary 

Use entire month name and not abreviation: https://github.com/canonical-webteam/practices/blob/master/content/copy-reviews.md

# QA

- `./run`
- http://127.0.0.1:8004/toto
- Check that dates in dropdown for channel map are using the correct format (Entire month name)
- Same for last updated date

# Issue

Problem is that the dropdown is to small for month with "big" names:

screenshot_2019-01-25 install toto for linux using the snap store snapcraft

### DIFF
--- a/webapp/store/logic.py
+++ b/webapp/store/logic.py
@@ -160,7 +160,7 @@ def convert_date(date_to_convert):
     date_parsed = parser.parse(date_to_convert).replace(tzinfo=None)
     delta = datetime.datetime.now() - datetime.timedelta(days=1)
     if delta < date_parsed:
-        return humanize.naturalday(date_parsed)
+        return humanize.naturalday(date_parsed).title()
     else:
         return date_parsed.strftime("%-d %B %Y")
 
--- a/tests/store/tests_public_logic.py
+++ b/tests/store/tests_public_logic.py
@@ -298,4 +298,4 @@ class StoreLogicTest(unittest.TestCase):
         date_test = datetime.datetime.now().strftime("%Y-%m-%d")
         result = logic.convert_date(date_test)
 
-        self.assertEqual(result, "today")
+        self.assertEqual(result, "Today")


webapp/store/logic.py
```diff
@@ -154,7 +154,7 @@ def convert_date(date_to_convert):
     :returns: Readable date
     """
     date_parsed = parser.parse(date_to_convert)
-    return date_parsed.strftime("%b %d %Y")
+    return date_parsed.strftime("%B %d %Y")
```
nottrobin: This should be:

```suggestion
    return date_parsed.strftime("%d %B %Y")
```

https://wiki.canonical.com/Marketing/Style_guide#Dates

> Single day: 1 January 2013

nottrobin:
One simply way to help with the issue you mention is to change this line:

https://github.com/canonical-websites/snapcraft.io/blob/fe2616fbecad4c3680eaca53dca1de929fca8dc0/templates/store/snap-details/_channel_map.html#L99

To `width="30%"`. This gives the "published" line enough space in the most common sizes for that box. Could you do this?

More thorough fixes could include:

- Making the overlay box a bit bigger
- Using JavaScript to intelligently adjust the widths of the columns in that table depending on its content

But those seem like overkill for now.

tbille:
Updated with your suggestion @nottrobin 
@Lukewh @therealjuan what do you think about it?

tbille:
I am adding this to this PR to see the marketing requirements: https://wiki.canonical.com/Marketing/Style_guide#Dates

Lukewh:
screenshot_2019-01-28 install nextcloud for linux using the snap store snapcraft

Seems that single digits are 0 padded - this seems to go against the prescribed formatting

therealjuan:
The original issue suggested to have `x days ago` when the published date is less than a week and display the date in the prescribed formatting when the date is older than a week.

Shrinking the other elements from the channel map to increase the date width, has an impact on versions which could become lenghty strings so the best solution would be to increase the window width to `700px` from the original `600px`.

tbille:
New Design:

screenshot_2019-01-28 install nextcloud for linux using the snap store snapcraft

PR #1539, opened 2019-01-29
Make sticky element visually different on edit listing page

### LINKED ISSUE #635: Publisher pages: Make snap info action bar visually different when sticky
1. Visit a Snap information editing page of a Snap in your account (http://snapcraft.io/account/snaps/{%YOUR SNAP NAME%} )
2. Scroll down the page

### Current behaviour
The action bar with the Save and Apply actions looks identical to other page layout elements on screen and can visually confuse users.

### Expected behaviour
To distinguish visually when the snap information action bar. i.e Drop shadow, thick border at the bottom...

This is an example using `boxshadow: 1px 1px 10px rgba(0,0,0,0.2)`

image

therealjuan:
Fixes #635

## QA
2. Log in to your Developer account
3. Visit the /listing of any snap you own
4. Scroll up and down and notice there is a box shadow on the sticky element in the page

### DIFF
--- a/static/js/libs/throttle.js
+++ b/static/js/libs/throttle.js
@@ -1,0 +1,15 @@
+/**
+ * Throttle
+ * @param {Function} func Function to run.
+ * @param {Number} wait Time to wait between tries.
+ */
+
+export default function throttle(func, wait) {
+  let time = Date.now();
+  return function() {
+    if (time + wait - Date.now() < 0) {
+      func();
+      time = Date.now();
+    }
+  };
+}
--- a/static/js/publisher/market/stickyListingBar.js
+++ b/static/js/publisher/market/stickyListingBar.js
@@ -1,20 +1,13 @@
 export default function() {
-  const stickyBar = document.querySelector(".snapcraft-p-sticky");
-  const listings = document.querySelector(
-    ".u-float--left.p-heading--three.u-no-margin--bottom"
-  );
+  const stickyBar = document.querySelector("#store-listing-notification");
+  const listingsTitle = document.querySelector("#snap-name-listing");
 
-  const observer = new IntersectionObserver(entries => {
-    entries.forEach(entry => {
-      if (entry.isIntersecting) {
-        stickyBar.classList.remove("sticky-shadow");
-      } else {
-        stickyBar.classList.add("sticky-shadow");
-      }
-    });
+  const observer = new IntersectionObserver(entry => {
+    if (entry[0].isIntersecting) {
+      stickyBar.classList.remove("sticky-shadow");
+    } else {
+      stickyBar.classList.add("sticky-shadow");
+    }
   });
-
-  listings.forEach(listing => {
-    observer.observe(listing);
-  });
+  observer.observe(listingsTitle);
 }
--- a/static/js/publisher/market/stickyListingBar.js
+++ b/static/js/publisher/market/stickyListingBar.js
@@ -1,0 +1,20 @@
+export default function() {
+  const stickyBar = document.querySelector(".snapcraft-p-sticky");
+  const listings = document.querySelector(
+    ".u-float--left.p-heading--three.u-no-margin--bottom"
+  );
+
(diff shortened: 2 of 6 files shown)


static/js/publisher/market/stickyListingBar.js
```diff
@@ -0,0 +1,20 @@
+export default function() {
+  const stickyBar = document.querySelector(".snapcraft-p-sticky");
```
hatched: These could, in theory, match multiple elements. I just want to make sure that was intentional.

static/sass/_snapcraft_market.scss
```diff
@@ -56,15 +56,19 @@
...
     top: 0;
     z-index: 1;
 
     & .row {
+      padding-bottom: $sp-medium;
       padding-top: $sp-medium;
     }
 
+    &.sticky-shadow {
+      box-shadow: 1px 1px 10px rgba(0, 0, 0, .2);
```
anthonydillon: That is the rationale for split from Vanillas tabbed component?

static/js/publisher/market/stickyListingBar.js
```diff
@@ -0,0 +1,20 @@
...
+  const listings = document.querySelector(
+    ".u-float--left.p-heading--three.u-no-margin--bottom"
+  );
+
+  const observer = new IntersectionObserver(entries => {
```
bartaz: `IntersectionObserver` has quite limited browser support. It's not supported by IE, Safari possibly others. This means it will fail with an error in these browsers.

static/js/publisher/market/stickyListingBar.js
```diff
@@ -0,0 +1,20 @@
+export default function() {
+  const stickyBar = document.querySelector(".snapcraft-p-sticky");
+  const listings = document.querySelector(
+    ".u-float--left.p-heading--three.u-no-margin--bottom"
```
bartaz: This is overly specific selector that can easily change if we change something in HTML.
If we want to identify listings element we should add an `id` to it or some `js-listings` class that is specific for this use case.

static/sass/_snapcraft_market.scss
therealjuan: This does not change the styling from the tabs, it changes a sticky element that we have on the listing page:

image

image

Right now when it floats it can feel like you are in the same part of the page without context

static/js/publisher/market/stickyListingBar.js
therealjuan: Just added an `eventListener` on scroll for cross-browser support

static/js/publisher/market/stickyListingBar.js
therealjuan: This has been removed completely

static/js/publisher/market/stickyListingBar.js
```diff
@@ -0,0 +1,10 @@
+export default function() {
+  document.addEventListener("scroll", function() {
```
bartaz: Please use `debounce` to limit number of calls of `scroll` event.

See example of it in: https://github.com/canonical-websites/snapcraft.io/blob/master/static/js/public/fsf-language-select.js#L59-L60

static/js/libs/throttle.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Throttle
+ * @param {Function} func Function to run.
+ * @param {Number} wait Time to wait between tries.
+ */
+
+export default function throttle(func, wait) {
+  let time = Date.now();
```
bartaz: `time` shouldn't be set to `Date.now()` by default in `throttle`. It only makes sense it `func` is called immediately (which this throttle doesn't do by itself).

```suggestion
  let time;
```

therealjuan: 👍

bartaz: Or we can introduce `immediate` parameter - similar to our `debounce` implementation to set up throttle and immediately call function.

PR #1540, opened 2019-01-30
Move snap listing edit link to notification

### LINKED ISSUE #1453: New [Edit] link on snap details page is misleading
When you are signed in to snapcraft.io and visit snap details page of any snap you own you see a little Edit link next to your name (introduced in #1434)

It looks a bit misleading - it looks like it's meant to edit publisher name when it's the one thing that you cannot edit in snapcraft.io directly. In any way it doesn't suggest that you are about to edit contents of the whole page.

### Expected behaviour

'Edit' link should be positioned and maybe labeled in a way that will better suggest that it's meant for editing contents of the snap listing.

Maybe a notification or a 'bar' over the top of the page?

bartaz:
Fixes #1453

Moves snap details edit link into a notification

### QA

- ./run or https://snapcraft-io-canonical-websites-pr-1540.run.demo.haus/
- sign in and go to snap details page of snap you published
- notification should be visible with link to edit listing page

screenshot 2019-01-30 at 12 33 46

### DIFF
--- a/templates/store/snap-details.html
+++ b/templates/store/snap-details.html
@@ -62,9 +62,6 @@
               <span class="p-tooltip__message u-align--center" role="tooltip" id="{{ package_name }}-tooltip">Verified account</span>
             </span>
             {% endif %}
-            {% if is_users_snap %}
-              <a href="/{{ package_name }}/listing">Edit</a>
-            {% endif %}
           </p>
         </div>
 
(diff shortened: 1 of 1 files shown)


templates/store/snap-details.html
```diff
@@ -45,6 +45,7 @@
 
 {% block content %}
   <div class="p-strip--light is-shallow snapcraft-banner-background">
+
```
tbille: hello line

bartaz: Ah, leftover from moving stuff around.

PR #1545, opened 2019-01-31
Preview listing page

Lukewh:
Fixes https://github.com/CanonicalLtd/snap-squad/issues/852
Fixes https://github.com/CanonicalLtd/snap-squad/issues/853
Fixes https://github.com/CanonicalLtd/snap-squad/issues/854

## QA
- Pull the branch
- `./run`
- Click the preview button
- Switch back to the "listing" tab, make some changes, without saving
- Take a look at the "preview" tab and see the changes
- Try adding "?preview-true" to the url of a snap you don't have access to/ haven't visited the listing page for
- See an error notification


static/js/publisher/market/screenshots.test.js
```diff
@@ -28,7 +28,7 @@ describe("templates", () => {
 
   describe("screenshot template", () => {
     it("should render image with screenshot url", () => {
-      const screenshot = { url: "test-screenshot.png" };
+      const screenshot = { url: "http://localhost/test-screenshot.png" };
```
Lukewh: It's worrying that this didn't break the tests!?

static/js/public/preview.js
```diff
@@ -0,0 +1,300 @@
+import Remarkable from "remarkable";
```
bartaz: I'm not sure if `/public` is a good place for it.

Yes, preview is called on a "public" details page, but it is only used by publishers. We are adding unnecessary code (not only ours but also markdown library) to a public JS file where it will not be needed at all for most of the visitors.

Possibly a `publisher` bundle would be a better place for this.
If user is previewing they are already signed in to publisher part, so this file is already downloaded. We just need to include it on details page (only in preview mode).

static/js/publisher/form.js
```diff
@@ -230,6 +230,16 @@ function initForm(config, initialState, errors) {
...
+    updateLocalStorage();
+  }
+
+  function updateLocalStorage() {
+    if (!window.localStorage) {
+      document.querySelector(".js-listing-preview").classList.add("u-hide");
+      return;
+    }
+    const key = state["snap_name"];
+    window.localStorage.setItem(key, JSON.stringify(state));
```
bartaz: Great that passing data to preview can be as simple as that 👍

static/js/public/preview.js
```diff
@@ -0,0 +1,300 @@
...
+]);
+
+// For the different elements we might need to change different properties
+const functionMap = {
+  title: "innerHTML",
+  summary: "innerHTML",
+  description: "innerHTML",
+  website: "href",
+  contact: "href",
+  screenshots: "appendChild",
```
bartaz: Screenshots seem to be treated separately anyway (as a function rather then property), so why not getting it out of here and just handling where they are handled (without functionMap)?

bartaz: Another approach (thay may need a bit more work but may be more flexible and scallable for future changes) would be to create "functionMap" as something like:

```js
const functionMap = {
   title: (el, content) => el.innerHTML = content,
   website: (el, content) => el.href = content,
   //...
}
```

This way screenshots may have their own dedicated (more complex) function.

But I'm not sure how much it would actually simplify things from current state.

Lukewh:
Ahh that banner wasn't there when I started, must have come in through the rebase! Thanks @bartaz

static/js/public/preview.js
Lukewh: Good point. I originally did it as part of the publisher bundle, and moved it for reasons I don't remember... The issue with testing perhaps? I'll move it back and see what happens. We might need to pair on getting jest working with *swiper* though @bartaz, I hit a wall.

static/js/public/preview.js
Lukewh: Both good comments, I started the functionMap before getting to the complexity of screenshots, so well worth a refactor to neaten things up.

bartaz:
One more thing that may be useful (@therealjuan ?) - it would be good to have some way of showing that the page is just a preview - some banner, strip or notification on the top?

Lukewh:
@bartaz see the original description "To think about" section :+1:

bartaz:
@Lukewh I get some weird 'Remarkable' error on https://snapcraft-io-canonical-websites-pr-1545.run.demo.haus/1140-grid-bartaz/listing preview.

Preview 'bar' doesn't show up at all, and remarkable throws an error:

screenshot 2019-02-04 at 11 44 19

bartaz:
'Revert' and 'Save' don't work if there are no changes in the form.

1. Go to listing page of any snap
2. Click 'Preview' without changing anything
3. Click on 'Revert' on preview page - it will start spinning forever because there is nothing to revert
4. Click on 'Save' on preview page - it will start spinning forever because there is nothing to Save (Save is disabled on listing page)

screenshot 2019-02-04 at 11 49 10

Also the Revert/Save functionality on Preview page relies on listing page being open in other tab.
If user closes the listing page Revert/Save stop working (spin forever).
Not sure how easy it would be to implement, but maybe we could close preview when listing page is closed (or navigated from)?

Lukewh:
@bartaz I've changed markdown library and updated the state. Let me know what you think. It's not perfect - there's some funky business around how the storage events work that I can't figure out.

I'll rebase and update the PR to include the categories now too :+1:!

static/js/publisher/form.test.js
```diff
@@ -142,7 +148,7 @@ describe("initForm", () => {
     market.initForm(config, initialState, undefined);
   }
 
-  describe("simple", () => {
+  describe("", () => {
```
bartaz: Why empty label? Could it be `describe("form")` or something?
```suggestion
  describe("form", () => {
```

therealjuan:
UXLGTM

bartaz:
@Lukewh @therealjuan I just realised we have a quite big issue with current preview implementation.

Because of the way we base preview on the snap details page it's only going to work on snaps that have details page - meaning already published PUBLIC snaps.

If your snap is not public yet preview page just 404. Which is quite a serious issue. Once the preview feature is in users will likely expect it to work for all snaps - not just public ones. Maybe even more important for not published ones - so they can preview the page before it gets public.

therealjuan:
How about...

* Show the preview as disabled with a tooltip stating: `You can only preview snaps that have been already published`
* Implement the preview of non-published snaps as a follow-up issue

bartaz:
"Follow up issue" may in worst case mean completely different implementation if snap details page is not available.

Unless we can somehow render snap details preview page (not 404 it) even for private snaps - user is signed in, we have data for preview - we just need to make backend not 404 snap detail page if it is for preview (hopefully)?

bartaz:
@Lukewh 

There is still small issue with Saving unchanged preview.
Listing page 'Save' reacts to changes of the state so when state is changed and the change is then removed Save gets disabled.
But Preview page still can be attempted to be saved in unchanged state - leading to an error.

Simply:
1. make a change to listing page (add something to title?)
2. open preview
3. take change back on listing page (make title back the same)
4. Save on listing is disabled
5. Save on preview is still enabled, click it
6. It just spins until it fails with an error

Lukewh:
@bartaz I've hopefully fixed the issue you described with the state and the buttons.

I've also disabled the Preview button and added a tooltip as per @therealjuan comment.

I'll look into the 404 issue at the beginning of next week. I too fear that I might need to implement it how I was going to originally - with complicated template spaghetti.

bartaz:
@Lukewh Live preview that you implemented works great (and it's cool that we could quite easily pass the state to the preview and make it work).

But if we hit a dead end, maybe we need to reconsider. Instead of reimplementing a whole lot of stuff in JS to make it work maybe we should let backend render it via template.
We could POST the state to some /preview endpoint and it would render details page with the data from it instead from API. But this will make it a 'one time' preview, not refreshing to changes in different tab. And, obviously, its completely different implementation from scratch.

So hopefully we could make the JS preview work for good.

Lukewh:
@bartaz I was thinking of using the same code, but doing exactly that. So the initial preview load is from the POST but still use the state stuff to live update. That way we could use the same snap-details.html template. What do you think?

bartaz:
@Lukewh best of both worlds. 👍

bartaz:
Works well now, thanks for the updates.

I'm not sure what should we do with "preview data" and UX of the parts that don't work in preview.

Install and channel map buttons are clickable, but don't do anything.
Would be nice if we could disable them, or maybe show a tooltip on hover that in preview they are not accessible.
Same for 'Report snap' maybe? Now report snap is fully functional on preview page.

But my biggest concern is map/OS charts with preview data. The fake data that is used currently just looks too real - it shows a bunch of countries highlighted on a map and a long list of OSes. This may be easily confused with real data.

screenshot 2019-02-15 at 13 06 02


If we want to show something there it should clearly be marked as preview.
Maybe a grayed out map (with no countries highlighted), OS list with just a couple clearly fake names?

Thoughts @therealjuan @Lukewh

templates/publisher/listing.html
```diff
@@ -27,6 +27,10 @@
           </div>
           <div class="col-5">
             <div class="u-align--right u-clearfix">
+              <button class="js-listing-preview p-button--base u-no-margin--bottom p-tooltip p-tooltip--btm-center" form="preview-form" area-describedby="preview-tooltip">
```
bartaz: TIL about `form` on a `button`.
And I was wondering what kind of magic submits the preview... ;)

static/js/publisher/form.js
```diff
@@ -381,6 +407,14 @@ function initForm(config, initialState, errors) {
     updateFormState();
   });
 
+  const openPreview = () => {
+    const previewForm = document.getElementById("preview-form");
+    const stateInput = previewForm.elements.state;
+    stateInput.value = JSON.stringify(state);
+  };
+
+  previewButton.addEventListener("click", openPreview);
```
bartaz: Maybe it would be safer to do the handling on `submit` event on preview form rather than click?
The same button does the submit of the form we are modifying, so it kind of relies on the fact that click event is (hopefully) always triggered before submit.
But I guess we could just listen to submit and update the state input there?

therealjuan:
How about set opacity to 0.3 on the map with a text on top stating that is `This content is being displayed just for preview purposes` for the map. For the channel map / install button we can disabled buttons and show a tooltip with the same text.

static/js/publisher/market/storageCommands.js
```diff
@@ -0,0 +1,30 @@
...
+  formEl,
+  snap_name,
+  ignoreChangesOnUnload,
+  context = window
+) {
+  const key = `${snap_name}-command`;
+  if (e.key === key) {
+    switch (e.newValue) {
+      case "edit":
+        context.localStorage.removeItem(key);
```
bartaz: As it seems it is always removed from localStorage could it be just moved out from `switch`?

templates/publisher/listing.html
```diff
@@ -517,12 +525,13 @@
...
       ]
     }, {
+      snap_name: {{ snap_name|tojson }},
       title: {{ snap_title|tojson }},
       summary: {{ summary|tojson }},
       description: {{ description|tojson }},
       website: {{ website|tojson }},
       contact: {{ contact|tojson }},
-      categories: "{% if snap_categories.categories[0] %}{% for category in snap_categories.categories %}{% if loop.index > 1 %}, {% endif %}{{ category.name }}{% endfor %}{% endif %}",
+      categories: {{ snap_categories.categories|tojson }},
```
bartaz: Is it supposed to support categories that are not merged yet?
I believe that branch uses `categories` in the context of template.

static/js/publisher/form.js
Lukewh: Yeah, good catch. Originally I was submitting a JS generated form with JS so needed to add the event to the button, but now it's a proper form it makes sense to be on the form "submit" :+1

therealjuan:
Looks good to me! Tested in chrome now

bartaz:
When metrics charts are disabled on listing page preview shows double `<hr>`:

screenshot 2019-02-15 at 15 07 50


There are also small issues with clicking on different stuff on the page:
Clicking on channel map and install buttons does nothing.
Clicking on screenshots does nothing.
Clicking on report snap works.

PR #1548, opened 2019-02-01
Removed google plus

therealjuan:
2. Check on every page there is no Google Plus in the footer
3. #winning

### DIFF
--- a/static/sass/_patterns_social-icons.scss
+++ b/static/sass/_patterns_social-icons.scss
@@ -4,6 +4,7 @@
     background-image: url('#{$assets-path}6bfc3d6a-build.snapcraft.social.optimize.svg');
     background-repeat: no-repeat;
     background-size: 160px 80px; // 2x4 icons 40x40px each (in 2 rows for regular and hover state)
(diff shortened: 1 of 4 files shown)


bartaz:
Icons seems misaligned now (before they were taking space up to the right side of page):

screenshot 2019-02-01 at 11 59 51

We may need to narrow down the wrapper of the icons now that there is less of them.
Probably going to col-9/col-3 from col-8/col-4 should be enough.

static/sass/_patterns_social-icons.scss
```diff
@@ -4,6 +4,7 @@
     background-image: url('#{$assets-path}6bfc3d6a-build.snapcraft.social.optimize.svg');
     background-repeat: no-repeat;
     background-size: 160px 80px; // 2x4 icons 40x40px each (in 2 rows for regular and hover state)
+    color: transparent;
```
bartaz: @therealjuan  Why is this needed?

matthewpaulthomas:
LGTM

PR #1552, opened 2019-02-04
Updated heading

### LINKED ISSUE #1551: “Measure growth and retention” heading is inaccurate
[The snapcraft.io front page](https://snapcraft.io/) has a section titled “Measure growth and retention”:

snapcraft io-measure-growth-and-retention

Unfortunately, snapcraft.io does not let you measure retention. This is because the retention chart has not been implemented yet ([snapcraft-design#126](https://github.com/CanonicalLtd/snapcraft-design/issues/126)) — nor even the “Installs” chart, with which you could potentially calculate retention manually.

For example, with the current charts there is no way to tell the difference between 100% retention and 10% growth, or 90% retention and 22% growth: the “Weekly active devices” chart would look practically identical either way.

Until a way of measuring retention is implemented, the easiest way to fix this would be to change the “Measure growth and retention” heading to something like “Measure user growth” or “Measure growth of your audience”.

therealjuan:
Fixes #1551 

## QA

2. Check the heading in "Measure growth and retention" is now "Measure use growth"
3. #winning

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -177,7 +177,7 @@
 <section class="p-strip">
   <div class="row">
     <div class="col-6">
-      <h3 class="p-heading--two">Measure growth and retention</h3>
+      <h3 class="p-heading--two">Measure user growth</h3>
       <p>Make data-driven decisions with active install metrics. Watch as automatic updates migrate users to your latest release. Understand your audience with demographic breakdowns.</p>
     </div>
     <div class="col-6">


bartaz:
@therealjuan Possibly should be also updated in copy doc: https://docs.google.com/document/d/1_uwPkHmAgV6_uE9ISqIgXjZGJxXPBD90vdeO95h7ikE/edit

therealjuan:
I was about to do that...

PR #1570, opened 2019-02-07
Add list of categories to snap details page

### LINKED ISSUE #1496: Include self-serve categories on detail pages
Now that publishers can include categories as part of the listing publisher page, we should showcase the selection somewhere on detail pages.

Most obvious candidate is next to the publisher username. On hover, the category will get underlined linking to the relevant category store page.

bartaz:
Fixes #1496 

Adds links to categories to snap details pages (under snap name)

### QA

- ./run or demo
- go to any snap with category (for example via Store page)
- list of categories of this snap should be linked under snap name
- go to publisher listing page of any snap you own
- add it to two categories, Save
- see if both categories show on snap details page
- remember to remove any testing categories from test snaps!

screenshot 2019-02-07 at 12 52 33

### DIFF
--- a/webapp/store/views.py
+++ b/webapp/store/views.py
@@ -376,10 +376,14 @@ def store_blueprint(store_query=None, testing=False):
             ):
                 is_users_snap = True
 
+        # build list of categories of a snap
+        categories = logic.get_snap_categories(details["snap"]["categories"])
+
         context = {
             # Data direct from details API
             "snap_title": details["snap"]["title"],
             "package_name": details["name"],
+            "categories": categories,
             "icon_url": icons[0] if icons else None,
             "version": last_version,
             "license": details["snap"]["license"],
--- a/templates/store/snap-details.html
+++ b/templates/store/snap-details.html
@@ -75,15 +75,20 @@
         {% endif %}
         <div class="p-snap-heading__title">
           <h1 class="p-heading--two p-snap-heading__name" data-live="title">{{ snap_title }}</h1>
-          <p class="p-snap-heading__publisher">
-            by {{ display_name(publisher, username) }}
-            {% if developer_validation and developer_validation == VERIFIED_PUBLISHER %}
-            <span class="p-verified p-tooltip p-tooltip--top-center" aria-describedby="{{ package_name }}-tooltip">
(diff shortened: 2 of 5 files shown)


webapp/store/views.py
```diff
@@ -384,10 +384,39 @@ def snap_details(snap_name):
             ):
                 is_users_snap = True
 
+        # build list of categories of the snap
+        # get all categories with their names
+        try:
```
bartaz: @tbille 

Because snap info API only returns 'slugs' of the categories (not full names) I need to call categories API as well and build a proper list of categories for the view.

Not sure if it's good to keep such logic here in view or better to move it out somewhere?

therealjuan:

evandandrea:
👍 looks great with two categories too:
https://snapcraft-io-canonical-websites-pr-1570.run.demo.haus/slack

webapp/store/views.py
Lukewh: Might be worth - at some point - doing something similar to the licenses? Categories won't change that often so we can store a static file on app launch. Every time we deploy it'll be updated.

Not suggesting for this PR, just an idea for future.

webapp/store/views.py
```diff
@@ -384,10 +384,39 @@ def snap_details(snap_name):
...
+
+        categories = []
+        for category in snap_categories:
+            if category["name"] in category_names:
+                categories.append(
+                    {
+                        "slug": category["name"],
+                        "name": category_names[category["name"]],
+                    }
+                )
```
tbille: all this makes me thing of code that we already wrote on the logic file.

webapp/store/views.py
```diff
@@ -384,10 +384,39 @@ def snap_details(snap_name):
             ):
                 is_users_snap = True
 
+        # build list of categories of the snap
+        # get all categories with their names
+        try:
+            categories_results = api.get_categories()
+        except ApiError as api_error:
+            categories_results = []
+            status_code, error_info = _handle_errors(api_error)
```
tbille: this can be removed

tbille: just the line, we don't really care if we don't get an answer from the server

PR #1562, opened 2019-02-05
🚧 [Proof of concept] - report snap POST to google sheet

bartaz:
Fixes https://github.com/canonicalltd/snap-squad/issues/887

Proof of concept of posting report snap form into Google Sheet

### How to use
1. go to snap details page of any snap
2. find a testing "Report this snap" form
3. enter some details and submit (dummy success JSON should show up)
4. Go to [test report snap sheet](https://docs.google.com/spreadsheets/d/1CBGPSFnThb74NizuhSroV5T9y2pin72-EdBirxmz0rE/edit?usp=sharing) - reported snap should be in the bottom of the list

screenshot 2019-02-05 at 10 31 33

PR #1572, opened 2019-02-07
Add report endpoint

tbille:
# Summary

Add post endpoint to report snaps
Snaps reported will be pushed to a spreadsheet, shared with advocacy)

# QA

- `vim .env.local`
- Add: `SPREADSHEET_REPORT_SNAP=https://script.google.com/macros/s/AKfycbyLyPn-nVfVy5aLfaR7dqRhZngGSjKYpAm7PylOSi9yKSyxqUc/exec`
- `./run`
- Go on a snap and report snap (ask @bartaz for simple view to do it)

### DIFF
--- a/tests/store/tests_post_report.py
+++ b/tests/store/tests_post_report.py
@@ -1,0 +1,108 @@
+import requests
+
+import responses
+from flask_testing import TestCase
+from webapp.api import google
+from webapp.app import create_app
+
+
+class PostReportTest(TestCase):
+    render_templates = False
+
+    def setUp(self):
+        self.snap_name = "toto"
+        self.api_url = "https://url.to.spreadsheet/"
+        self.endpoint_url = "/" + self.snap_name + "/report"
+
+    def create_app(self):
+        google.SPREADSHEET_REPORT_SNAP = "https://url.to.spreadsheet/"
+
+        app = create_app(testing=True)
+        app.secret_key = "secret_key"
+        app.config["WTF_CSRF_METHODS"] = []
+
+        return app
+
+    def _check_calls(self):
+        assert len(responses.calls) == 1
+        called = responses.calls[0]
+        assert called.request.url == self.api_url
+
+    def test_no_spreadsheet_url(self):
+        google.SPREADSHEET_REPORT_SNAP = ""
+        response = self.client.post(self.endpoint_url)
+
+        assert len(responses.calls) == 0
+        assert response.status_code == 200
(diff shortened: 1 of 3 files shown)


therealjuan:
Sure, might be good to have a proper chat with @evandandrea about this and understand your concerns.

nottrobin:
Why @evandandrea? Has no-one talked to the Store team about it?

Currently, snapcraft.io is a stateless application. It is purely a presentational layer on top of the Snap Store API, delegating everything there.

Adding one new endpoint to save data into a random Google doc has a number of issues:

- The Google Docs API is now a core piece of our infrastructure. This new endpoint could easily hang or break and bring down the entire webapp with it.
- There may well be security concerns with sending and storing this data in Google docs (a question for the Store team)
- This endpoint is very hard to test effectively, as it works completely differently to everything else.
- As the feature won't be in the Store, adding the ability to report snaps through other avenues (e.g. the snap CLI) will be very tricky.
- It will be the only function that bypasses the Store, and the Store team completely, which is organisationally confusing

I'd be curious to know whether @cprov etc. support this idea.

cprov:
@nottrobin You have valid concerns about adding Gdocs API as a new infrastructure dependency, which requires additional safety measurements (circuit breaker, metrics and monitoring) additionally to keep the good testing coverage levels of the application.

However the argument for "everything should be implemented in the Store" doesn't reflect reality, it has been a while since the Storefront is successfully combining information from other systems (WP for articles, Discourse for docs, NPS and other events in Marketo) with the Store data.

I am not particularly fond of pumping this data into a Gdoc, but I see that as a valid experiment/PoC, once the implementation addresses the sustainability points you've mentioned.

therealjuan:
I just added a meet on Monday to discuss, so we can go through all the details.

bartaz:
The implementation here originates from my ['proof of concept'](https://github.com/canonical-websites/snapcraft.io/pull/1562) based on original requirement to save this data in Google Sheet.

Possibly more rigid solution (still keeping data in Google Sheet) would be to use Google Sheets API directly instead of simply pushing the data via a public form. But is is going to complicate the implementation on our side.
https://developers.google.com/sheets/api/quickstart/python

If we want to avoid adding Google tools as additional dependencies we need to decide where would we send/store the report data. If not Google maybe Marketo would be a good candidate - as this seems to be a target of most of our 'actionable' forms.

tbille:
As discussed this will be done in the frontend to avoid adding new dependency on the backend.

Issue #1475, opened 2018-12-18
Ordering of channels can change in hover on /metrics


evandandrea:
### Expected behaviour

Channels always go in order: stable, candidate, beta, edge.

### Steps to reproduce the problem

Hover over the graph and move from right to left, notice how the order of the channels changes.

![screenshot 2018-12-18 at 17 39 02](https://user-images.githubusercontent.com/1523179/50172066-0d49f980-02ec-11e9-8741-7bc5c3597b5c.png) ![screenshot 2018-12-18 at 17 38 52](https://user-images.githubusercontent.com/1523179/50172067-0d49f980-02ec-11e9-8bd5-bb83be8c0270.png)

PR #1581, opened 2019-02-08
Order channels by track/risk/branch

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1475

It also updates a few things.
- Added a channels module for sorting arbitrary channel strings (will update the channel map to use this at some point).
- Added "defaultTrack" to metrics page (currently hardcoded for [node](https://snapcraft.io/node)
- Moved the hardcoded logic for node to the helpers module.
- Made the tooltip appear above the site navigation

## QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/<snap_name>/metrics?active-devices=channel
- Hover over the graph and move about - the order of channels should only change if there's no data for that channel on that day.

### DIFF
--- a/static/js/libs/channels.js
+++ b/static/js/libs/channels.js
@@ -120,14 +120,14 @@ function createChannelTree(channelList) {
 function sortAlphaNum(list, hoistValue) {
   let numbers = [];
   let strings = [];
-  let latest = [];
+  let hoistList = [];
   list.forEach(item => {
     // numbers are defined by any string starting any of the following patterns:
     //   just a number – 1,2,3,4,
     //   numbers on the left in a pattern – 2018.3 , 1.1, 1.1.23 ...
     //   or numbers on the left with strings at the end – 1.1-hotfix
     if (hoistValue && item === hoistValue) {
-      latest.push(item);
+      hoistList.push(item);
     } else if (isNaN(parseInt(item.substr(0, 1)))) {
       strings.push(item);
     } else {
@@ -137,21 +137,20 @@ function sortAlphaNum(list, hoistValue) {
 
   // Ignore case
   strings.sort(function(a, b) {
-    return a[0].toLowerCase().localeCompare(b[0].toLowerCase());
+    return a.toLowerCase().localeCompare(b.toLowerCase());
   });
 
-  strings = latest.concat(strings);
+  strings = hoistList.concat(strings);
 
   // Sort numbers (that are actually strings)
   numbers.sort((a, b) => {
-    return b[0].localeCompare(a[0], undefined, {
+    return b.localeCompare(a, undefined, {
       numeric: true,
       sensitivity: "base"
     });
   });
 
   // Join the arrays together again
-
   return strings.concat(numbers);
 }
 
--- a/static/js/libs/channels.js
+++ b/static/js/libs/channels.js
@@ -1,0 +1,262 @@
+const RISKS = ["stable", "candidate", "beta", "edge"];
+
+/**
+ * Parse a channel string into an object
+ * @param {string} channelString
+ * @param {{defaultTrack: string}} options
+ * @returns {{format: {risk: boolean, track: boolean, branch: boolean}, risk: *, track: (string|*), branch: (string|*)}}
+ */
+function parseChannel(channelString, options) {
+  const format = {
+    track: false,
+    risk: false,
+    branch: false
+  };
+  const channelArr = ["latest", undefined, "_base"];
+
+  if (options) {
+    if (options.defaultTrack) {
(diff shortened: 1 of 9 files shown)


bartaz:
Works well.

One thing I noticed, that is a bit related, but I'm not sure if we can fix this - is the ordering of sections on the graph itself.

Currently it's different from the order on the tooltip which is a bit confusing - especially that the 'bold' on tooltip that highlights current channel is not very visible.

Hovering over first top section (green) - first 10/stable channel is highlighted:
screenshot 2019-02-11 at 09 17 15

Hovering over second from top (red) - way below 11/stable is highlighted:
screenshot 2019-02-11 at 09 17 22


Should the order on the graph be the same as in the tooltip? Can we even control it?

static/js/libs/channels.js
```diff
@@ -0,0 +1,262 @@
...
+ * @param {{defaultTrack: string}} options
+ * @returns {{format: {risk: boolean, track: boolean, branch: boolean}, risk: *, track: (string|*), branch: (string|*)}}
+ */
+function parseChannel(channelString, options) {
+  const format = {
+    track: false,
+    risk: false,
+    branch: false
+  };
+  const channelArr = ["latest", undefined, "_base"];
```
bartaz: Is there any reason for doing this in array instead of 3 variables (`track`, `risk`, `branch`)?

Also isn't the default branch empty? If we start using "_base" we may end up detecting a branch when there isn't any. But maybe as we approach adding support for branches in more elements of UI we should just start using "_base" as default. If that's the case it would be good to have "_base" (but "latest" as well) defined somewhere as consts and reused instead of being copied.

static/js/libs/channels.test.js
```diff
@@ -0,0 +1,533 @@
...
+          }
+        }
+      });
+    });
+  });
+});
+
+describe("sortAlphaNum", () => {
+  it("sorts numbers", () => {
+    expect(sortAlphaNum(["0", "5", "2", "6", "7", "1", "8"])).toEqual([
```
bartaz: I guess this should work, but would be good to test how "10" is sorted against smaller numbers (to check if it's not sorted alphabetically).

static/js/libs/channels.js
```diff
@@ -0,0 +1,262 @@
...
+ * defaultTrack/beta
+ * defaultTrack/beta/branches
+ * defaultTrack/edge
+ * defaultTrack/edge/branches
+ * track/stable
+ * track/stable/branches
+ * ...
+ *
+ * @param {[string]} channels An array of "risk" / "track/risk" / "track/risk/branch"
+ * @param {{defaultTrack: string}} options
```
bartaz: If there is no other option then `defaultTrack` maybe it just could be a `defaultTrack` param?

Lukewh:
@bartaz thinking about your ordering comment some more:

We could alter the order to match but the current implementation ensures that any new channel/ version/ os appears from the bottom to take a share of those above. Changing the order in the specific case you've shown would work (as they're all established channels) but if a new "latest/stable/test" was released, where should it appear?

I don't think it's worth the extra effort/ code/ calculations, but I could easily be wrong :D

static/js/libs/channels.js
Lukewh: I wanted to make it generic for future and be more descriptive in the code that calls it. I find:
`sortChannels([], "randomString")` harder to parse than `sortChannels([], { defaultTrack: "randomString" })` when scanning code... I'm open to arguments against though!

bartaz:
@Lukewh Sure, never mind then. This issue is about order in the tooltip anyway.

static/js/libs/channels.js
Lukewh: So I introduced the "_base" as a concept to have both a common format for all track/risks, regardless of whether they have branches and also to help with parsing in other parts in this file, for example reducing/ sorting `["stable/test", "stable"]` was producing `["stable/test"]` so stripping the "base" stable channel.

Setting "_base" hopefully makes it clear for anyone calling this code in the future (for another part of the site)

static/js/libs/channels.js
bartaz: Not sure if "randomString" would be used in such way directly in code. I would guess in most cases default track would be passed as a variable, so it would be something more like `sortChannels([...], defaultTrack)` which would be more clear.

But, I agree that unnamed option params may be confusing (especially when passing booleans), so I'm not against for sure.

In this case I just wondered if there is any reason to pass it via object. And because it seems like it's used like that consistently in at least 2 functions I'm fine with leaving it as it is.
Code is clear to read, and works, so not worth spending time changing it for little benefit.

PR #1583, opened 2019-02-11
FSF Node added

igorljubuncic:
Added workflow for node.

### DIFF
--- a/webapp/first_snap/content/node/package.yaml
+++ b/webapp/first_snap/content/node/package.yaml
@@ -48,8 +48,7 @@ explain:
           wethr:
             plugin: nodejs
             source: .
-    - text: The <code>nodejs</code> plugin builds upon the work you’ve already done to describe your app’s dependencies in your <code>package.json</code>. It will automatically include these in your snap.
-    - text: It will also include Node 6. This can be changed with the <code>node-engine</code> keyword.
+    - text: The <code>nodejs</code> plugin builds upon the work you’ve already done to describe your app’s dependencies in your <code>package.json</code>. It will automatically include these in your snap, including the LTS version of Node. This can be changed with the <code>nodejs-version</code> keyword.
   apps:
     - text: Apps are the commands you want to expose to users and any background services your application provides.
     - code: |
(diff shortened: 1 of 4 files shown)


webapp/first_snap/content/node/package.yaml
```diff
@@ -0,0 +1,62 @@
...
+  parts:
+    - text: |
+        Parts define what sources are needed to assemble your app. Parts can be anything: programs, libraries, or other needed assets.
+    - code: |
+        parts:
+          wethr:
+            plugin: nodejs
+            source: .
+    - text: The <code>nodejs</code> plugin builds upon the work you’ve already done to describe your app’s dependencies in your <code>package.json</code>. It will automatically include these in your snap.
+    - text: It will also include Node 6. This can be changed with the <code>node-engine</code> keyword.
```
evandandrea: "It will also include the LTS version of Node."

The `node-engine` option does not exist. Perhaps you need `nodejs-version`?
```
$ snapcraft help nodejs
The nodejs plugin is useful for node/npm based parts.

The plugin uses node to install dependencies from `package.json`. It
also sets up binaries defined in `package.json` into the `PATH`.

This plugin uses the common plugin keywords as well as those for "sources".
For more information check the 'plugins' topic for the former and the
'sources' topic for the latter.

Additionally, this plugin uses the following plugin-specific keywords:

    - nodejs-version:
      (string)
      The version of nodejs you want the snap to run on.
      This includes npm, as would be downloaded from https://nodejs.org
      Defaults to the current LTS release.

    - nodejs-package-manager
      (string; default: yarn)
      The language package manager to use to drive installation
      of node packages. Can be either `npm` or `yarn` (default).

    - nodejs-yarn-version:
      (string)
      Applicable when using yarn. Defaults to the latest if not set.
```

evandandrea:
Copy looks good, just testing the snapcraft.yaml changes (and committing to snapcraft-docs) now…

evandandrea:
The new snapcraft.yaml works.

bartaz:
@igorljubuncic Codecov is blocking merging this PR. Could you rebase your PR on top of fresh master?

PR #1584, opened 2019-02-12
FSF Rust

igorljubuncic:
Adding FSF guide for Rust.


templates/home/_fsf_rust.html
```diff
@@ -10,8 +10,45 @@ <h4>Why are snaps good for Rust projects?</h4>
...
+<b>description</b>: |
+  Parity's goal is to be the fastest, lightest, and 
+  most secure Ethereum client. We are developing 
+  Parity using the sophisticated and cutting-edge
+  Rust programming language. Parity is licensed 
+  under the GPLv3, and can be used for all your 
+  Ethereum needs.
+
+<b>confinement</b>: devmode
+<b>base</b>: core17
```
evandandrea: Typo

templates/home/_fsf_rust.html
```diff
@@ -10,8 +10,45 @@ <h4>Why are snaps good for Rust projects?</h4>
     </ul>
```
evandandrea: The above needs updating. Please use the Python flow as a guide and add a paragraph about how snaps compare to / fit alongside existing rust tooling, i.e. cargo.

templates/home/_fsf_rust.html
```diff
@@ -10,8 +10,45 @@ <h4>Why are snaps good for Rust projects?</h4>
...
+  Ethereum needs.
+
+<b>confinement</b>: devmode
+<b>base</b>: core17
+
+<b>parts</b>:
+  <b>parity</b>:
+    <b>source</b>: .
+    <b>plugin</b>: rust
+    <b>build-attributes</b>: [no-system-libraries]
```
evandandrea: no-system-libraries does not exist when using bases.

templates/home/_fsf_rust.html
```diff
@@ -10,8 +10,45 @@ <h4>Why are snaps good for Rust projects?</h4>
...
+    <b>stage-packages</b>:
+      - libssl1.0.0
+      - libudev1
+      - libstdc++6
+
+<b>apps</b>:
+  <b>parity</b>:
+    <b>command</b>: parity</pre></code>
+  </div>
+
```
evandandrea: The snapcraft.yaml fails to build (even after correcting `base:` and dropping `no-system-libraries`:
```
error: failed to load source for a dependency on `parity-ui-old-precompiled`

Caused by:
  Unable to update https://github.com/paritytech/js-precompiled.git?branch=v1#f3fa5e5e

Caused by:
  revspec 'f3fa5e5efb0ed8ada6cd091694fcdaa921e6720f' not found; class=Reference (4); code=NotFound (-3)
Failed to run '/root/.cargo/bin/cargo +stable fetch --manifest-path /root/parts/parity/src/Cargo.toml' for 'parity': Exited with code 101.
Verify that the part is using the correct parameters and try again.
Run the same command again with --debug to shell into the environment if you wish to introspect this failure.
An error occurred when trying to execute 'sudo -i snapcraft snap' with 'multipass': returned exit code 2.
```

webapp/first_snap/content/rust/package.yaml
```diff
@@ -0,0 +1,74 @@
...
+        parts:
+          parity:
+            source: .
+            plugin: rust
+            build-attributes: [no-system-libraries]
+            build-packages:
+              - libudev-dev
+              - libssl-dev
+              - make
+              - pkg-config
```
evandandrea: You're missing the stage packages.

igorljubuncic:
This example does not build (including the original up-to-date repo used for the fork). I am replacing with a different example and will resubmit with relevant changes (including a different yaml).

templates/home/_fsf_rust.html
```diff
@@ -10,8 +10,41 @@ <h4>Why are snaps good for Rust projects?</h4>
...
+<b>parts</b>:
+  <b>xsv</b>:
+    <b>plugin</b>: rust
+    <b>source</b>: .
+
+<b>apps</b>:
+  <b>xsv</b>:
+    <b>command</b>: bin/xsv
+    <b>plugs</b>:
```
evandandrea: Don't specify plugs. We won't cover strict confinement here and they're meaningless in devmode.

webapp/first_snap/content/rust/package.yaml
```diff
@@ -0,0 +1,65 @@
...
+        Parts define what sources are needed to assemble your app. Parts can be anything: programs, libraries, or other needed assets.
+    - code: |
+        parts:
+          xsv:
+            plugin: rust
+            source: .
+    - text: |
+        Parts define how to build your app. Parts can be anything: programs, libraries, or other assets needed to create and run your application. In this case we have one: the xsv source code. In other cases these can point to local directories, remote git repositories, or tarballs.
+    - text: |
+        This example will also bundle the current stable release of Rust in the snap using <code>Rustup</code>. You can define the exact version of Rust with the optional <code>rust-revision</code> keyword.
```
evandandrea: Rust is a compiled language. You don't bundle Rust. The `rust-revision` (still worth mentioning) option specifies the version of the Rust compiler to use.

templates/home/_fsf_rust.html
```diff
@@ -4,14 +4,44 @@ <h4>Why are snaps good for Rust projects?</h4>
     <ul>
       <li>Bundle all the runtime requirements, with support for multiple architectures.</li>
       <li>Simplify installation instructions, regardless of distribution, to snap install myrustapp.</li>
-      <li>Directly control the delivery of automatic application updates.</li>
       <li>Extremely simple creation of services.</li>
       <li>No complex installation shell scripts to maintain.</li>
     </ul>
+	<p>Programming in Rust can sometimes be difficult, especially when one of the crates in your project changes. Snaps let you bundle all the necessary dependencies without having to worry about breakages.</p>
```
evandandrea: >"Programming in Rust can sometimes be difficult, especially when one of the crates in your project changes."

Why do changes to crates make programming in Rust difficult?

>"Snaps let you bundle all the necessary dependencies without having to worry about breakages."

Do Rust applications typically use C library dependencies?

igorljubuncic: From what I'm able to discern, updating a dependency in cargo.toml from one version to another can lead to the crate being broken (in some cases), and this cannot necessarily be easily fixed unless you maintain all the crates.

evandandrea: How do snaps improve upon that?

igorljubuncic: I have pushed a revised template file. At the moment, I do not have any additional information to add on the Rust-specific advantages to the landing page of the Rust FSF guide. I will study this topic some more.

The snap builds and installs correctly (based on one of Alan's Top 100 projects).

webapp/first_snap/content/rust/package.yaml
```diff
@@ -0,0 +1,62 @@
...
+  parts:
+    - text: |
+        Parts define what sources are needed to assemble your app. Parts can be anything: programs, libraries, or other needed assets. In this case we have one: the xsv source code. In other cases these can point to local directories, remote git repositories, or tarballs.
+    - code: |
+        parts:
+          xsv:
+            plugin: rust
+            source: .
+    - text: |
+        You can define the exact version of Rust with the optional <code>rust-revision</code> keyword. Dependencies from your <code>Cargo.toml</code> will also be bundled in the snap.
```
evandandrea: >Dependencies from your <code>Cargo.toml</code> will also be bundled in the snap.

If you build and published a rust app outside of snaps, would the dependencies from cargo be included in your binary?

templates/home/_fsf_rust.html
evandandrea: Have a look at the Go text. They're similar languages from a project management perspective.


> * Easy to discover and install by millions using the Snap Store or command-line every day
> * Automatically updated to the latest stable version of your app
> * Revert to the previous version if an update fails, preserving data
>
> Programming in Go makes it easy to create a zip of your app that runs across Linux, without dependencies. However, end user discovery and update management remain a challenge. Snaps fill this gap, letting you distribute a Go app in an app store experience for end users.

Lukewh:
@igorljubuncic @evandandrea what's the state of this PR? Would be good to get it sorted and merged, it's starting to get ... :sunglasses: rusty :sunglasses: ....

webapp/first_snap/content/rust/package.yaml
evandandrea: This issue still needs to be addressed. As I understand it, cargo dependencies are statically linked.

* https://stackoverflow.com/a/29008355
* https://doc.rust-lang.org/cargo/guide/dependencies.html
* https://doc.rust-lang.org/reference/linkage.html
* https://www.reddit.com/r/rust/comments/8s8tv7/dynamic_linking_in_rust/

igorljubuncic:
Both static and dynamic linking is possible. I will slightly rephrase this, as the term bundling can be interpreted in different ways.

PR #1588, opened 2019-02-13
Fix linter with latest version on flake8

tbille:
# Summary 

Since the upgrade of `pycodestyle` to version 2.5.0 flake8 linter has become more strict on certain invalid characters for example: `\g`

> \g is in fact an invalid escape sequence.  In python3.8+ it will be a syntax error (you can enable it earlier by escalating warnings to errors)

Source: https://gitlab.com/pycqa/flake8/issues/474

# QA

- Make sure when accessing blog images are still pulled from cloudinary
- Make sure that on /blog/feed the links are snapcraft link (localhost on local machine)
- Make sure linter passes

love

### DIFF
--- a/webapp/blog/logic.py
+++ b/webapp/blog/logic.py
@@ -27,23 +27,23 @@ def replace_images_with_cloudinary(content):
     cloudinary = "https://res.cloudinary.com/"
 
     urls = [
-        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_350/\g<url>",
-        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_650/\g<url>",
-        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_1300/\g<url>",
-        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_1950/\g<url>",
+        cloudinary + r"canonical/image/fetch/q_auto,f_auto,w_350/\g<url>",
+        cloudinary + r"canonical/image/fetch/q_auto,f_auto,w_650/\g<url>",
+        cloudinary + r"canonical/image/fetch/q_auto,f_auto,w_1300/\g<url>",
+        cloudinary + r"canonical/image/fetch/q_auto,f_auto,w_1950/\g<url>",
     ]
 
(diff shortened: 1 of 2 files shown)


webapp/blog/logic.py
```diff
@@ -108,7 +108,7 @@ def change_url(feed, host):
     :returns: A string with converted urls
     """
     url_regex = re.compile(
-        "https:\/\/admin.insights.ubuntu.com(\/\d{4}\/\d{2}\/\d{2})?"
+        "https:\\/\\/admin.insights.ubuntu.com(\\/\\d{4}\\/\\d{2}\\/\\d{2})?"
```
nottrobin: Why not use raw strings?

```suggestion
        "https:\/\/admin.insights.ubuntu.com(\/\d{4}\/\d{2}\/\d{2})?"
```

Also, are you absolutely sure slashes even need escaping? I don't see why they would.
``` bash
robin@xps:~/git/juju-docs (master)$ ipython3 
impPython 3.6.6 (default, Sep 12 2018, 18:26:19) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import re

In [2]: re.match("/hello/", "/hello/world")
Out[2]: <_sre.SRE_Match object; span=(0, 7), match='/hello/'>
```

webapp/blog/logic.py
```diff
@@ -27,23 +27,23 @@ def replace_images_with_cloudinary(content):
...
+        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_1300/\\g<url>",
+        cloudinary + "canonical/image/fetch/q_auto,f_auto,w_1950/\\g<url>",
     ]
 
     image_match = (
         r'<img(?P<prefix>[^>]*) src="(?P<url>[^"]+)"(?P<suffix>[^>]*)>'
     )
     replacement = (
-        "<img\g<prefix>"
+        "<img\\g<prefix>"
```
nottrobin:
```suggestion
        r"<img\g<prefix>"
```

PR #1592, opened 2019-02-15
first snap flow: fix Base appearing when the language doesn't use them

evandandrea:
screenshot 2019-02-07 at 18 30 01


Lukewh:
Worst QA instructions evaaaar

PR #1614, opened 2019-02-18
Remove unnecessary print call

bartaz:


Lukewh:
Thanks @bartaz for fixing my mistakes :D

Issue #1416, opened 2018-12-07
Show search results count on /search


Greggless:
### Expected behaviour
When searching for a snap, it would be nice to know how many results there are (we found X snaps for XXX) and how many pages there will be to go through

PR #1579, opened 2019-02-08
🚧 Publisher pages WIP

bartaz:
Fixes https://github.com/CanonicalLtd/snap-squad/issues/885

Publisher pages template.

KDE WIP:
https://snapcraft-io-canonical-websites-pr-1579.run.demo.haus/publisher/kde

Minimal example:
https://snapcraft-io-canonical-websites-pr-1579.run.demo.haus/publisher/bartaz

screenshot 2019-02-11 at 14 51 19

Bigger example:
https://snapcraft-io-canonical-websites-pr-1579.run.demo.haus/publisher/canonical

screenshot 2019-02-13 at 12 54 22

PR #1619, opened 2019-02-20
Search page improvements

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1416

- Add total number of snaps found in search.
- Make the snaps list a grid.
- Align pagination better.

Next PR: use new Vanilla pagination https://user-images.githubusercontent.com/17748020/44727201-67227b80-aad1-11e8-9f26-9596a40f665b.png https://github.com/vanilla-framework/vanilla-framework/pull/2007/files

## QA

- Pull the branch
- `./run`
- Search for some things :)


bartaz:
Not sure how much control we have over number of items per page (I believe store was doing some filtering so number may vary per page?), but it looks a bit weird when we show items in a grid and there are some 'leftovers' in last row, but there is still next page to look at.

screenshot 2019-02-20 at 10 59 18

When I see one item in last row I think it's just a last result, I don't expect to see the pagination and next page.

templates/store/_category-partial.html
```diff
@@ -3,30 +3,8 @@
...
-                      <img src="https://assets.ubuntu.com/v1/75654c90-rosette.svg" />
-                      <span class="p-tooltip__message u-align--center" role="tooltip" id="{{ snap.package_name }}-tooltip">Verified account</span>
-                    </span>
-                  {% endif %}
-                </p>
-              </div>
-            </div>
-          </a>
+          {% set snap_no_margin = True %}
+          {% include "store/_single_snap.html" %}
```
bartaz: This looks familiar :)
https://github.com/canonical-websites/snapcraft.io/pull/1579/files#diff-b0189eeedbb624082216d645ba7400a6

But I guess that's what happens when branch sits unmerged for too long. Will update mine once this lands.

Lukewh: I like yours better - done

therealjuan:

kwm14:
## Design review
Looking good @Lukewh just a couple of small comments 😉 

- [ ] Increase margin bottom on p-media-object--snap to `2rem` as currently the snaps are equal distance apart from snap name and icon, so it's hard to know which is related to what (if you get what I mean 😕 

screenshot 2019-02-20 at 15 43 03

- [ ] Pagination misaligned on mobile
- [ ] Are we able to have `2-col` on mobile as it becomes a long scrollable list as `1-col`?

screenshot 2019-02-20 at 15 44 26

kwm14:
Niiice. Updates LGTM 👍 💯

static/sass/_patterns_media-object--snap.scss
```diff
@@ -5,7 +5,7 @@
 
   .p-media-object--snap {
     display: block;
-    margin-bottom: $spv-inter--regular;
+    margin-bottom: $spv-inter--regular * 2;
```
bartaz: Are we sure it won't cause spacing issues in other places where this is used (like store categories)?

static/sass/_patterns_fluid-grid.scss
```diff
@@ -16,6 +16,8 @@
     justify-content: start;
 
     .p-fluid-grid__item {
+      width: 50%;
```
bartaz: Why adding this, what does that change?

static/sass/_patterns_media-object--snap.scss
Lukewh: As long as you set the no-margin-bottom config to the template we should be good... I imagine we'll either want no-margin (handled by the surrounding elements) or this margin (for snap wrapping). Not sure of another way to make this work without it.

static/sass/_patterns_fluid-grid.scss
Lukewh: This changes the smallest size so it's still col-2, rather than 1-col. Even on mobile it looks better in 2 cols.

static/sass/_patterns_media-object--snap.scss
bartaz: I was thinking that maybe margin should be on grid item, not here, but I'm not sure.
As for now it works, so let's leave it as it is.

PR #1621, opened 2019-02-20
Add embeddable cards to Publicise snap publisher page

### LINKED ISSUE #1601: Add embeddable card settings to publisher Market page

bartaz:
Fixes #1601

Adds basic embeddable cards to Publicise page.

### QA

- ./run or https://snapcraft-io-canonical-websites-pr-1621.run.demo.haus/
- go to Publicise page of any snap
- you should see side navigation on the left side
- go to "Embeddable cards"
- page should open with embedded card preview and code to copy
- copy the code to test it out (make sure to change "snapcraft.io" in the iframe URL to demo URL: https://snapcraft-io-canonical-websites-pr-1621.run.demo.haus/)
- check if iframe renders as expected

screenshot 2019-02-20 at 13 30 30


Lukewh:
I think the embed HTML needs some work - we should probably have a `max-width` and the height doesn't seem to be enough sometimes. Can we not use JavaScript to resize the frame to the right size from within the frame?

https://codepen.io/Lukewh/pen/mvgELo

hughesjordan:
Looks good so far!

bartaz:
@Lukewh Ah, when the width gets above a threshold bigger spacing jumps in.

I don't think we can change iframe width from the script within the frame, especially not cross-domain.
It would have to be a parent document that observes the frame. But not sure if it would work cross-domain either.

As one of future enhancements (especially when we start adding switches for more content in the card) I think we can add JavaScript on publisher page to adapt height (and HTML snippet) to fit current setup of the card. But to make it dynamically in third party website will probably be not possible.

We could add some more restrictions to default iframe HTML, but I guess in the end it will be up to user to adjust some props to fit their needs.

PR #1634, opened 2019-02-24
Upgrade talisker requests

tbille:
# Summary

- Upgrade to talisker 0.11.1
- Configure talisker metrics for each request.Session; this will allow us to get more metrics

# QA

- create `.env.local` with: `TALISKER_NETWORKS=172.17.0.1`
- `./run`
- Check that metrics for requests are there and working on `/_status/metrics`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ requests==2.21.0
 requests-cache==0.4.13
 responses==0.9.0
 ruamel.yaml==0.15.72
-talisker==0.11.0
+talisker[gunicorn]==0.14.2
 
 # Temporary fix to SSO redirect issue
 Werkzeug==0.14.1
--- a/webapp/api/store.py
+++ b/webapp/api/store.py
@@ -1,4 +1,5 @@
 import os
+import talisker.requests
 from webapp import api
 from webapp.api.exceptions import (
     ApiResponseDecodeError,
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -16,7 +16,8 @@ requests==2.20.0
 requests-cache==0.4.13
 responses==0.9.0
 ruamel.yaml==0.15.72
-talisker==0.11.0
+talisker==0.13.0
```
jpmartinspt: Do we need the two lines?

tbille:
Waiting for this to be merged before merging https://github.com/canonical-ols/talisker/pull/428

webapp/api/store.py
```diff
@@ -75,6 +76,8 @@ def __init__(self, store=None, testing=False, cache=True):
             self.session = api.requests.Session()
         else:
             self.session = api.requests.CachedSession(timeout=(1, 6))
+            talisker.requests.configure(self.session)
```
bloodearnest: Unless there's a particular reason, I would configure the session regardless of testing or caching - it should affect neither of those things?

webapp/app.py
```diff
@@ -38,6 +38,14 @@ def create_app(testing=False):
 
     if not testing:
         talisker.flask.register(app)
+        import webapp.api.blog
+        import webapp.api.dashboard
+        import webapp.api.sso
+
+        talisker.requests.configure(webapp.api.blog.api_session)
```
bloodearnest: Same as above - why only add talisker stuff when not testing? It shouldn't break when testing, and keeps your tests closer to production.

tbille: this would imply a refactor on our test suite since we don't run the tests with talikser. For now I would rather keeping .it like this and maybe work on this on another task

PR #1638, opened 2019-02-26
Cache static assets longer

jpmartinspt:


webapp/config.py
```diff
@@ -24,3 +24,6 @@ class ConfigurationError(Exception):
     raise ConfigurationError("`WEBAPP` is not configured")
 
 WEBAPP_EXTRA_HEADERS = {}
+
+SEND_FILE_MAX_AGE_DEFAULT = 10*365*24*60*60
```
hatched: Can you add a comment as to how long this is for?

jpmartinspt: Done!

barrymcgee:
Still to do...

- Check in all templates that static assets are being properly fingerprinting.

webapp/config.py
```diff
@@ -24,3 +24,6 @@ class ConfigurationError(Exception):
     raise ConfigurationError("`WEBAPP` is not configured")
 
 WEBAPP_EXTRA_HEADERS = {}
+
+# Ten years default cache time on static files
+SEND_FILE_MAX_AGE_DEFAULT = 10 * 365 * 24 * 60 * 60
```
tbille: is this used somewhere?

PR #1640, opened 2019-02-27
FSF MOOS

igorljubuncic:
This is the updated FSF MOOS guide.

The YAML shown on the template page builds correctly, and the snap runs.

The description primarily highlights the benefits of snaps to the robotics audience. I've made the changes after consulting with Kyle.

I've also updated the test section to reflect the correct invocation of the snap, as the snap name and the application name are different.

### DIFF
--- a/templates/home/_fsf_moos.html
+++ b/templates/home/_fsf_moos.html
@@ -24,8 +24,7 @@
 
 <b>parts</b>:
   <b>moos</b>:
-    <b>source</b>: https://github.com/themoos/ \
-	core-moos/archive/10.0.2.a-release.tar.gz
+    <b>source</b>: https://github.com/themoos/core-moos/archive/10.0.2.a-release.tar.gz
     <b>build-packages</b>:
       - gcc
       - g++
--- a/templates/home/_fsf_moos.html
+++ b/templates/home/_fsf_moos.html
@@ -2,16 +2,46 @@
   <div class='col-6'>
     <h4>Why are snaps good for MOOS projects?</h4>
     <ul>
-      <li>Snaps are easy to discover and install. Millions of users can browse and install snaps graphically in the Snap Store or from the command-line.</li>
-      <li>Snaps install and run the same across Linux. They bundle the exact version of MOOS/MOOS-IvP required and your app’s library dependencies.</li>
-      <li>Snaps automatically update to the latest version. Four times a day, users’ systems will check for new versions and upgrade in the background.</li>
-      <li>Upgrades are not disruptive. Because upgrades are not in-place, users can keep your app open as it’s upgraded in the background.</li>
-      <li>Upgrades are safe. If your app fails to upgrade, users automatically roll back to the previous revision.</li>
+      <li>Bundle all the runtime requirements, including the exact version of MOOS/MOOS-IvP and system libraries you need.</li>
+      <li>Expand the distributions supported beyond just Ubuntu.</li>
+      <li>Directly and reliably control the delivery of application updates using existing infrastructure.</li>
+      <li>Extremely simple creation of daemons.</li>
     </ul>
+  </div>
+
+  <div class='col-6'>
+    <h4>How moos-ping defines snapcraft.yaml:</h4>
+    <code class="p-code-yaml"><pre>
+<b>name</b>: moos-ping
+<b>version</b>: '0.1'
+<b>summary</b>: MOOS Ping Example
+<b>description</b>: |
+  The main communication mechanism for all 
+  MOOS apps.
+
+<b>confinement</b>: devmode
+<b>base</b>: core18
+
+<b>parts</b>:
+  <b>moos</b>:
+    <b>source</b>: https://github.com/themoos/ \
+	core-moos/archive/10.0.2.a-release.tar.gz
+    <b>build-packages</b>:
+      - gcc
+      - g++
+      - make
+    <b>plugin</b>: cmake
+    <b>configflags</b>:
+      - -DCMAKE_INSTALL_PREFIX=/
+
+<b>apps</b>:
+  <b>MOOSDB</b>:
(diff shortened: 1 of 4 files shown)


evandandrea:
@kyrofa can you review this?

templates/home/_fsf_moos.html
```diff
@@ -2,16 +2,46 @@
...
     <h4>Why are snaps good for MOOS projects?</h4>
     <ul>
-      <li>Snaps are easy to discover and install. Millions of users can browse and install snaps graphically in the Snap Store or from the command-line.</li>
-      <li>Snaps install and run the same across Linux. They bundle the exact version of MOOS/MOOS-IvP required and your app’s library dependencies.</li>
-      <li>Snaps automatically update to the latest version. Four times a day, users’ systems will check for new versions and upgrade in the background.</li>
-      <li>Upgrades are not disruptive. Because upgrades are not in-place, users can keep your app open as it’s upgraded in the background.</li>
-      <li>Upgrades are safe. If your app fails to upgrade, users automatically roll back to the previous revision.</li>
+      <li>Bundle all the runtime requirements, including the exact version of MOOS/MOOS-IvP and system libraries you need.</li>
+      <li>Expand the distributions supported beyond just Ubuntu.</li>
+      <li>Directly and reliably control the delivery of application updates using existing infrastructure.</li>
```
kyrofa: Can we reword this slightly (or add a new li) that mentions how safe upgrades are (rollbacks etc., like the last li being removed).

templates/home/_fsf_moos.html
```diff
@@ -2,16 +2,46 @@
...
+      <li>Extremely simple creation of daemons.</li>
     </ul>
   </div>
 
+  <div class='col-6'>
+    <h4>How moos-ping defines snapcraft.yaml:</h4>
```
kyrofa: This fits the pattern established by the other languages/frameworks, but I'm not sure it makes sense here given that "moos-ping" is not an actual thing. It's just an example; no one's going to know what "moos-ping" is. Would "Here's an example MOOS snapcraft.yaml" (or something similar) work better?

templates/home/_fsf_moos.html
```diff
@@ -2,16 +2,46 @@
...
+<b>confinement</b>: devmode
+<b>base</b>: core18
+
+<b>parts</b>:
+  <b>moos</b>:
+    <b>source</b>: https://github.com/themoos/ \
+	core-moos/archive/10.0.2.a-release.tar.gz
+    <b>build-packages</b>:
+      - gcc
+      - g++
```
kyrofa: This is missing the part as well as the app that actually makes it do anything interesting (this does no pinging). Is that on purpose, or an oversight? The flow suggests cloning this, which has more: https://github.com/snapcraft-docs/moos-ping/blob/master/snapcraft.yaml

templates/home/_fsf_moos.html
igorljubuncic: Yes, I will re-add it.

templates/home/_fsf_moos.html
igorljubuncic: If I understand correctly, MOOS is a communication mechanism through which apps interface. So this might be semantically "incorrect" but it would make more sense to newcomers to the guide. I will make the change.

templates/home/_fsf_moos.html
igorljubuncic: The reason for this change is deliberate - inline with the other guides, I wanted to provide a simple one-part one-app yaml that is easy to follow. Do you think you can provide a simpler example or a simpler implementation(similar to ros/ros2)?

kyrofa: Sadly, no. MOOS is a source-only piece of software, and it doesn't ship its examples within it. Any MOOS snap will always be at least two parts: MOOS itself, and then one's own components. If you strip out the examples and just use MOOSDB it's a much less interesting example-- it's like snapping up a "dbus example" consisting of only the dbus daemon. I don't have strong feelings on this, my goal is simply to inform.

What is the primary purpose of the FSF? To show folks how to use snapcraft with some simple examples? Or to guide folks toward creating a snap of their stuff in the most straight-forward way possible? If it's the former, then yeah, just using MOOSDB is fine, although it doesn't teach much more than how to use the cmake plugin. If it's the latter, then I'll point out that, after this FSF, the next question that a MOOS dev will have is "Okay, how do I add my own stuff to this snap?"

igorljubuncic: I believe the primary purpose is to show people relatively simple examples of snaps with focus on specific languages (and plugins), which will then give them the appetite to explore more and try additional snaps. @evandandrea what do you think on this one?

templates/home/_fsf_moos.html
```diff
@@ -2,16 +2,46 @@
...
+<b>description</b>: |
+  The main communication mechanism for all 
+  MOOS apps.
+
+<b>confinement</b>: devmode
+<b>base</b>: core18
+
+<b>parts</b>:
+  <b>moos</b>:
+    <b>source</b>: https://github.com/themoos/ \
```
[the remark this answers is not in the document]
igorljubuncic: Agreed.

templates/home/_fsf_moos.html
evandandrea: Still outstanding.

templates/home/_fsf_moos.html
evandandrea: This should be updated to match the style used in the ROS first snap flow:
>Here's an example snapcraft.yaml that uses MOOS:

templates/home/_fsf_moos.html
evandandrea: Simple as possible. If MOOS requires more to "do something interesting", let's include the more.

igorljubuncic:
Currently on hold until we provide a suitable example. Please do not merge.

PR #1653, opened 2019-03-05
Summary option for publicise embeddable card

### LINKED ISSUE #1604: Add optional summary to embeddable card

bartaz:
Fixes #1604

Adds optional summary to embedded card

### QA
- ./run or https://snapcraft-io-canonical-websites-pr-1653.run.demo.haus/
- go to Publicize page of any snap, Embeddable card section
- "Show summary" option should be available to select
- Select "Show summary", summary should appear in the iframe (embedded code should update respectively)
- Check if summary works with and without the button

screenshot 2019-03-05 at 16 55 28

### DIFF
--- a/static/js/publisher/publicise.js
+++ b/static/js/publisher/publicise.js
@@ -64,7 +64,7 @@ const getCardEmbedHTML = (snapName, options) => {
   return `&lt;iframe src="https://snapcraft.io${getCardPath(
     snapName,
     options
-  )}" frameborder="0" width="100%" height="320px" style="border: 1px solid" &gt;&lt;/iframe&gt;`;
+  )}" frameborder="0" width="100%" height="320px" style="border: 1px solid #CCC; border-radius: 2px;"&gt;&lt;/iframe&gt;`;
 };
 
 // get form state from inputs
--- a/static/js/publisher/publicise.js
+++ b/static/js/publisher/publicise.js
@@ -38,7 +38,18 @@ function initSnapButtonsPicker() {
 // EMBEDDABLE CARDS
 
 const getCardPath = (snapName, button) => {
-  return `/${snapName}/embedded?button=${button}`;
+  const path = `/${snapName}/embedded`;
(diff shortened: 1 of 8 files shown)


hughesjordan:
All seems to work fine @bartaz one thing I noticed is that the right border does not always display properly, being cut off by the iframe perhaps?
screenshot 2019-03-05 at 4 14 49 pm

Also, is it possible to resize the width of suru background so the different shades of grey can be seen?

bartaz:
@hughesjordan The border issue is strange - the border is not in the card, but on the frame itself, so something else is making it disappear.
I noticed that in Chrome and made a CSS workaround that fixed that for me, but apparently it still happens.

I assume you are on Chrome as well? What is your screen resolution or window size?

hughesjordan:
Yep using Chrome, it doesn't seem to be specific to window size as when I resize it seems to show and hide as below, occurs on both my Mac and external display:
mar-06-2019 9-19-42 am

bartaz:
@hughesjordan Yes, just noticed the same. If I fix it in CSS for certain widths it starts to disappear in others. Looks like a Chrome bug to me, not sure how to fix that. I'll try to see if it happens only on snapcraft.io or if frame is embedded elsewhere too.

hughesjordan:
@bartaz Not sure if this is a Vanilla thing (and it's very minor) but the radio buttons seems to have additional margin causing them to be mis-aligned with the checkbox beneath.
screenshot 2019-03-06 at 2 02 15 pm

Otherwise looks good.

bartaz:
@hughesjordan It is vanilla bug (that I reported https://github.com/vanilla-framework/vanilla-framework/issues/2186).
If we want I can fix that in snapcraft.io when we add more checkboxes (screenshot), but then we will have to undo this fix when Vanilla is updated with a bugfix.

hughesjordan:
Ah okay cool, not to worry about it for now then. Cheers Bartek!

PR #1655, opened 2019-03-06
List disputed names

Lukewh:
Fixes https://github.com/CanonicalLtd/snap-squad/issues/921

# USE STAGING API'S IF YOU'RE GOING TO CREATE SOME DISPUTES

Implemented as per:
screenshot 2019-03-06 at 11 04 01 am

# USE STAGING API'S IF YOU'RE GOING TO CREATE SOME DISPUTES

## QA
- Pull the branch
# USE STAGING API'S IF YOU'RE GOING TO CREATE SOME DISPUTES
- `./run`
- If you have disputed names, they should be displayed as red with a warning icon
- Other registered snaps (unpublished) should be shown, hovering over the row should show a link to documentation

# USE STAGING API'S IF YOU'RE GOING TO CREATE SOME DISPUTES

## Screenshot
screenshot_2019-03-06 my published snaps linux software in the snap store 1

# USE STAGING API'S IF YOU'RE GOING TO CREATE SOME DISPUTES

### DIFF
--- a/templates/publisher/account-snaps.html
+++ b/templates/publisher/account-snaps.html
@@ -138,7 +138,7 @@ My published snaps — Linux software in the Snap Store
                       {% if dispute_pending %}
                         &nbsp;<span class="p-tooltip p-tooltip--btm-center" aria-describedby="{{ snap }}-dispute-pending">
                           <i class="p-icon--warning"></i>
-                          <span class="p-tooltip__message" role="tooltip" id="{{ snap }}-dispute-pending">Name being disputed</span>
+                          <span class="p-tooltip__message" role="tooltip" id="{{ snap }}-dispute-pending">Name dispute in progress</span>
                         </span>
                       {% endif %}
                     </span>
@@ -195,7 +195,7 @@ My published snaps — Linux software in the Snap Store
                   </td>
                   <td>
                     {% if dispute_pending %}
-                      <span class="p-snapcraft-dispute-list__muted">(Snap name disputed)</span>
+                      <span class="p-snapcraft-dispute-list__muted">(Name dispute in progress)</span>
                     {% else %}
                       <a href="https://docs.snapcraft.io/releasing-your-app" target="_blank" class="p-link--external p-snapcraft-dispute-list__link u-float--right">Publish to this name</a>
                     {% endif %}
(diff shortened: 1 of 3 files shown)


hughesjordan:
Juan mentioned a few small tweaks yesterday.. to drop the red an have 'Snap name disputed' in brackets:
screenshot 2019-03-07 at 9 58 51 am

Also probably a separate issue but worth noting, we will need to display an icon in the 'My published snaps' table too:
screenshot 2019-03-07 at 10 07 46 am

Lukewh:
Updated screenshot:
screenshot_2019-03-07 my published snaps linux software in the snap store

I also added the tooltip to the main table:
screenshot_2019-03-07 my published snaps linux software in the snap store 1

therealjuan:
Just one small comment, can we have the warning on the same position on both tables? After the snap name?

Lukewh:
Screenshot_2019-03-11  My published snaps — Linux software in the Snap Store
Moved @therealjuan

jpmartinspt:
@Lukewh Code looks fine, if anything I would suggest http://jinja.pocoo.org/docs/2.10/templates/#with-statement, just to avoid writing that comparison all the time. Maybe not for now, but might come in handy if we keep adding status checks to that table.

templates/publisher/account-snaps.html
```diff
@@ -169,18 +179,32 @@ <h3 class="p-heading--five">Join the community</h3>
...
+                <tr class="p-snapcraft-dispute-list__item">
+                  <td width="25%">
+                    {{ snap }}
+                    {% if dispute_pending %}
+                      &nbsp;<i class="p-icon--warning p-snapcraft-dispute-list__icon"></i>
+                    {% endif %}
+                  </td>
+                  <td>
+                    {% if dispute_pending %}
+                      <span class="p-snapcraft-dispute-list__muted">(Snap name disputed)</span>
```
jpmartinspt: I think (snap name disputed) makes it sound like the dispute is not pending anymore. I would maybe use (snap name in dispute). Thoughts @therealjuan @Lukewh?

therealjuan: Name dispute in progress?

Lukewh: Updated :+1:

jpmartinspt: :+1:

PR #1665, opened 2019-03-08
Add optional channel map to embeddable card

### LINKED ISSUE #1603: Add channels options to embeddable card settings

bartaz:
Fixes #1603

Add option to show/hide all channels in embeddable card

## QA

- ./run or https://snapcraft-io-canonical-websites-pr-1665.run.demo.haus/
- go to Publicise page of any snap
- 'Show all channels' option should be available
- All options should be checked by default
- Uncheck 'Show all channels', channel map should be removed from preview (replaced with version name)
- Try different configurations of options, make sure they work as expected

screenshot 2019-03-08 at 15 55 49

### DIFF
--- a/static/js/publisher/publicise.js
+++ b/static/js/publisher/publicise.js
@@ -39,30 +39,26 @@ function initSnapButtonsPicker() {
 
 const getCardPath = (snapName, options = {}) => {
   const path = `/${snapName}/embedded`;
-  let params = "";
+  let params = [];
 
   if (options.button) {
-    params = `button=${options.button}`;
+    params.push(`button=${options.button}`);
+  }
+
+  if (options["show-channels"]) {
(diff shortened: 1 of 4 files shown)


hughesjordan:
@bartaz looking good!

static/js/publisher/publicise.js
```diff
@@ -39,30 +39,26 @@ function initSnapButtonsPicker() {
 
 const getCardPath = (snapName, options = {}) => {
   const path = `/${snapName}/embedded`;
-  let params = "";
+  let params = [];
```
Lukewh: you changed it!? Now I've wasted some shower time thinking about your way!

bartaz: Previous worked fine with 1-2 params, but doesn't scale well.

PR #1677, opened 2019-03-11
Update home page banner to new slanted strip design

### LINKED ISSUE #1391: Homepage branding refresh build
Implement the "slanted backgrounds" of https://github.com/canonicalltd/snapcraft-design/issues/545
Just the header for now (the issue says the footer isn't signed off).

We'd like to make this in a way that's reusable and purely aesthetic, so as little extra markup as possible. @anthonydillon and I discussed it a little and if we could use SVG's with masks and / or pseudo elements that would be ideal.

This is the first site to implement this so will help our decisions in how to implement for vanilla :)!

bartaz:
Fixes #1391 

### QA
- ./run or https://snapcraft-io-canonical-websites-pr-1677.run.demo.haus/
- go to store home page
- enjoy brand new slanted strip design
- compare with [design](https://user-images.githubusercontent.com/2741678/48841400-3fb61480-ed89-11e8-81ea-fae097442164.png)
- play with different screen sizes to make sure it works nice
- check in different browsers

Screenshot 2019-03-11 at 16 46 57
Screenshot 2019-03-11 at 16 47 25

### DIFF
--- a/static/sass/_patterns_strip_slanted.scss
+++ b/static/sass/_patterns_strip_slanted.scss
@@ -1,0 +1,55 @@
+// based on https://github.com/vanilla-framework/vanilla-framework/pull/2051
+
+@mixin p-strip-slanted {
+  $horizontal-edge-padding: $spv-inter--regular-scaleable;
+  $slanted-edge-padding: $horizontal-edge-padding * 2; // coloured area takes half the rectangle surface area; to visually compensate, we multiply it by 2
+  $bleed: -3px; // stretch outside to compensate for aliasing; otherwise edge of image underneath becomes visible as you resize
+
+  %vf-strip-slanted {
+    background-position: 50% 50%;
+    margin-top: $bleed;
+    overflow: hidden;
+    position: relative;
+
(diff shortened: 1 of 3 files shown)


anthonydillon:
@bartaz in the future it would be good to link to the design in the QA steps. Code and QA is ok. Guess designer will do the design review but would like to check in the QA step too.

bartaz:
@anthonydillon Designs are linked from issue this fixes, but good point, that linking in QA steps would be useful.

As for the pattern itself - because it relies on 2 images (one being applied on ::after) it's hard to define these images outside of SCSS.
Maybe similar (but simplified) effect can be achieved with CSS gradients where colours can be given as variables - but this still would end up in SCSS (as mixin, or pattern?). Also CSS gradients will require more care to be fully cross browser and may be more error prone than this solution with predefined backgrounds.

But certainly more flexible pattern could be a nice small side project to explore.

lyubomir-popov:
@bartaz jusrt one ting, can you pull the backgrounds 1px up to hide the white line between header and hero due to subpixel position errors:

image

Issue #1671, opened 2019-03-11
Convert screenshots to React component


Lukewh:
And implement new design as described in https://github.com/canonical-websites/snapcraft.io/issues/1609

Related issues:
Negates: https://github.com/canonical-websites/snapcraft.io/issues/411, https://github.com/canonical-websites/snapcraft.io/issues/600, https://github.com/canonical-websites/snapcraft.io/issues/848

PR #1681, opened 2019-03-12
Convert screenshots to React and implement new design

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1671

## QA

- Pull the branch
- `./run`
- You should be able to manipulate screenshots successfully with a mouse
- You should be able to manipulate screenshots successfully with a keyboard (tab, delete, return keys)
- If you have over 5 images uploaded (blender) you should receive a warning with slightly different styled images
- Turn JS off and you should still get a read-only view of your screenshots


static/js/publisher/form/media.js
```diff
@@ -0,0 +1,225 @@
...
+
+  componentDidUpdate(prevProps, prevState) {
+    this.props.updateState(prevState.mediaData);
+    if (
+      prevState.mediaData.filter(item => item.status !== "delete").length >
+      this.props.mediaLimit
+    ) {
+      // Do the query lookup only when we need it - otherwise every state
+      // update will trigger a DOM lookup
+      const warning = document.querySelector(".js-media-limit-warning");
```
bartaz: Why toggling external DOM element? Can't warning be rendered by React only if needed?

static/js/publisher/form/media.js
```diff
@@ -0,0 +1,225 @@
...
+    const input = document.createElement("input");
+    input.type = "file";
+    input.accept = "image/x-png,image/gif,image/jpeg";
+    input.name = "screenshots";
+    input.hidden = "hidden";
+    input.addEventListener("change", () => {
+      this.mediaChange(input);
+    });
+
+    const form = ReactDOM.findDOMNode(this).closest("form"); // eslint-disable-line react/no-find-dom-node
```
bartaz: It's better to access DOM of a component using ["refs"](https://reactjs.org/docs/refs-and-the-dom.html).

static/js/publisher/form/media.js
```diff
@@ -0,0 +1,225 @@
...
+    throw new Error("No media holder El");
+  }
+
+  ReactDOM.render(
+    <Media mediaData={images} updateState={updateState} />,
+    mediaHolderEl
+  );
+}
+
+export { Media as default, MediaItem, initMedia };
```
bartaz: I think it would be better to separate components to their own files and have any functionality that uses them (`initMedia`) separately.
Also, why exporting `MediaItem` here when it seems to have its own file?

static/js/publisher/form/media.test.js
```diff
@@ -0,0 +1,341 @@
+import Media, { initMedia } from "./media";
+import React from "react";
+import { render, fireEvent } from "react-testing-library";
```
bartaz: :clap:

static/js/publisher/form/media.js
Lukewh: It's in the template for no-js rendering. Didn't want to duplicate the element

bartaz:
When there are some screenshots on the list I can click them and they gain visible focus. But it doesn't seem like I can do anything more to them? Focus on "Add image" seems to allow using keyboard to add screenshot, but what about focus on image?

static/js/publisher/form/media.js
Lukewh: can you use a ref for an element outside of the ReactDOM?

static/js/publisher/form/media.js
Lukewh: :+1: will do this (and remove the MediaItem export - legacy of when these all used to be in the same file)

Lukewh:
> When there are some screenshots on the list I can click them and they gain visible focus. But it doesn't seem like I can do anything more to them? Focus on "Add image" seems to allow using keyboard to add screenshot, but what about focus on image?

@bartaz focus on an image and press delete/ backspace to remove the image.

templates/publisher/listing.html
```diff
@@ -303,27 +303,47 @@
...
           <div class="row" data-validation-name="screenshots">
-            <div class="col-12">
+            <div class="col-2">
+              <label>Images:</label>
+            </div>
+            <div class="col-8">
+              <div class="p-notification--caution js-media-limit-warning{% if screenshot_urls|length < 5 %} u-hide{% endif %}">
+                <p class="p-notification__response">You have over 5 images uploaded. Not all image will be visible to users.</p>
+              </div>
+              <div class="p-listing-images p-fluid-grid js-media-holder">
```
bartaz: Seems like we are rendering this both in template and in JS.
Any reason for that? I can get that it will make it render without JS as well, and will probably appear earlier on the page before React re-renders it. But the functionality wouldn't work without JS anyway. None of this form would. So I'm not sure it's worth duplicating in template and React. With any change to class names or markup they can get out of date here.

templates/publisher/listing.html
```diff
@@ -303,27 +303,47 @@
...
+                    <img src="{{ image }}" />
+                  </div>
+                {% endfor %}
+                {% for empty in range(0, remainder) %}
+                  <div class="p-listing-images__image is-empty p-fluid-grid__item--small js-media-item-holder">
+                  </div>
+                {% endfor %}
+              </div>
+            </div>
+            {#<div class="col-12">
```
bartaz: No need to keep this around.

templates/publisher/listing.html
Lukewh: Interesting. I just tried without JS and you're right. I thought we still saved title etc. without JS, obviously not! I guess we rely on the "diff" input being submitted.
It would probably be better to have the diff input list which input's have changed, and use that list to construct the POST to dashboard. If ! diff list we should POST all fields.

bartaz: I think the plan was to move it all to React/JS eventually, so having a fallback support for pure POST was not needed.
As much as I like the idea of progressive enhancement and having things working without JS in this case (and with moving much of rendering to React) it may not be worth implementing and maintaining.

static/js/publisher/form/media.js
bartaz: No. You can store a ref to elements rendered by React. But at least this would allow replacing `findDOMNode` if I read it correctly.

bartaz:
One thing that I'm missing a bit is some kind of status message when marking screenshots to delete.

In current implementation screenshots are marked and a message is shown:
Screenshot 2019-03-12 at 14 32 12

It makes it clear that these changes were not saved yet.

This PR changes that, when screenshot is 'deleted' it is removed from the view, but there is no other sign showing that it needs to be saved. It gives the impression that screenshots are removed immediately, when they are not until 'Save' is clicked.

static/js/publisher/form/media.js
Lukewh: Ah, I see. Fixed!

templates/publisher/listing.html
```diff
@@ -303,27 +303,22 @@
...
-            <div class="col-12">
-              <div class="p-screenshots-toolbar" id="screenshots-toolbar">
-                <label>Screenshots (up to 5):</label>
-                <div class="p-screenshots-toolbar__buttons">
-                  <button class="p-button-neutral js-add-screenshots"><i class="p-icon--plus"></i></button>
-                  <button class="p-button-neutral js-delete-screenshot"><i class="p-icon--delete"></i></button>
-                  <button class="p-button-neutral js-fullscreen-screenshot  u-no-margin--top"><i class="p-icon--fullscreen"></i></button>
-                </div>
+            <div class="col-2">
+              <label>Images:</label>
```
bartaz: Here it says 'Images', but couple lines above it says 'Media' (in error notification) - is this intended?

templates/publisher/listing.html
```diff
@@ -516,12 +511,13 @@
       snapIconInput: "snap_icon",
       form: "market-form",
       formNotification: "market-form-status",
-      screenshotsToolbar: "screenshots-toolbar",
-      screenshotsWrapper: "snap-screenshots",
+      // screenshotsToolbar: "screenshots-toolbar",
+      // screenshotsWrapper: "snap-screenshots",
```
bartaz: Can be removed I guess?

templates/publisher/listing.html
Lukewh: I think it should be "Images" in the notification for now. Will update :+1:

bartaz:
Why is the first image larger than the others?

Screenshot 2019-03-13 at 11 26 31

Lukewh:
@bartaz following the spec: https://docs.google.com/document/d/1KgprCc1WiHbtCtTQ1vSP_EIpPaC44U7UAIwTcMSAkhE/edit#

hughesjordan:
> Why is the first image larger than the others?
> 
> Screenshot 2019-03-13 at 11 26 31

@bartaz Intention is to indicate which out of your 5 images will be the primary image. Any thoughts?

bartaz:
Delete icon in hidden behind images in the 'over 5 screenshots' part. Seems like applying `filter` causes this.

Screenshot 2019-03-13 at 12 05 21

Lukewh:
Good catch @bartaz

static/js/publisher/form/media.js
```diff
@@ -0,0 +1,270 @@
...
+        </div>
+      );
+    }
+    return false;
+  }
+
+  renderBlankMedia(numberOfBlank) {
+    const blankMedia = [];
+    for (let i = 0; i < numberOfBlank; i++) {
+      const classes = mediaClasses.slice(0);
```
bartaz: That `slice` confused me, didn't realise it's just cloning the array.
We have transpilers in place, maybe that would be clearer:
```suggestion
      const classes = [...mediaClasses];
```

bartaz:
@hughesjordan What is 'primary image'? If there are screenshots on snap details page they are just shown as a list, it doesn't seem like first one is special in any way:

Screenshot 2019-03-13 at 12 29 38

So I don't know why it is shown differently in listing page.
And if there is any special meaning of "primary image" it would be good to have it explained somewhere (in a tooltip?).

hughesjordan:
@bartaz 

> it doesn't seem like first one is special in any way
>
It's not special other than the fact it is first e.g. you don't choose to put the worst out of 5 images first.

To me a 'primary image' is the image that comes first, enlarging the first image in the assets interface was an effort make that **even more** apparent.

May be unnecessary as generally we order things left to right anyway & visually it looks nicer to have them all sized the same (designs already produced see below)

Image added A

The question is do we want an additional way to say to the publisher 'this will be the image that visitors see first' or is natural ordering enough? Could do something like this?

Screenshot 2019-03-13 at 12 12 17 pm

bartaz:
@hughesjordan I think if publisher seen a snap page with screenshots or previewed their own it's pretty clear for them how screenshots are displayed and ordered. I don't think there is a need for any other highlighting of first image if it doesn't have any additional meaning.

If I as a user see screenshot marked as "Main image" I start to think: "what does it mean? is it displayed somewhere else, is it featured is some other view?".
For now it's just a screenshot, one of five. It's the first one - and it's in exactly same order as on snap details page, so it's quite self explanatory.

At least for me trying to give first image some additional significance when there isn't really any is only a source of unnecessary confusion.

static/sass/_snapcraft_market_screenshots.scss
```diff
@@ -1,79 +1,77 @@
 @mixin snapcraft-market-screenshots {
-  .p-screenshots-toolbar {
-    align-items: baseline;
-    display: flex;
-    justify-content: space-between;
+  .p-listing-images {
+    align-items: flex-start;
```
bartaz: This doesn't seem to change anything (?).

As far as I remember `align-items` is about aligning items orthogonally to flex direction (so vertically in this case) - which is not needed (until container is bigger then the items).

Lukewh: Ah this is left over from when there was the larger image - it ensured subsequent images were aligned with the top. Will remove

bartaz:
Current 'flexible grid' doesn't seem to work well for screenshots, as lack of margin starts to be visible on smaller screen sizes:
Screenshot 2019-03-13 at 14 28 45

This is not an issue on home page where we use it for logos, as they usually have enough spacing themselves (so even if items are close to each other, they don't look close).
With screenshots taking all space of square items it becomes more visible, and doesn't look nice.

static/js/publisher/form.js
```diff
@@ -136,16 +136,19 @@ function initForm(config, initialState, errors) {
...
+  if (config.mediaHolder) {
+    const screenshots = state.images.filter(
+      image => image.type === "screenshot"
     );
+    initMedia(config.mediaHolder, screenshots, nextState => {
+      const noneScreenshots = state.images.filter(
+        item => item.type !== "screenshot"
+      );
+      const newState = Object.assign({}, state);
+      newState.images = noneScreenshots.concat(nextState);
```
bartaz: Not necessary change, but quite nice shorthand for `Object.assign` to consider:

```js
const newState = {
   ...state,
   images: noneScreenshots.concat(nextState)l
}
```

static/js/publisher/form/media.test.js
```diff
@@ -0,0 +1,314 @@
...
+
+describe("Media", () => {
+  const imageSelector = ".js-media-item-holder:not(.is-empty)";
+  const addImageSelector = ".p-listing-images__add-image";
+
+  it("should render without an image", () => {
+    const { container } = render(
+      <Media mediaData={[]} updateState={() => {}} />
```
bartaz: Is setting these props necessary? They seem to be set in `defaultProps` already to the same values. So I guess tests could only set the ones that are different(?)

static/js/publisher/form/media.js
```diff
@@ -0,0 +1,271 @@
...
+          {mediaList.map((item, i) => (
+            <MediaItem
+              key={item.url}
+              url={item.url}
+              type={item.type}
+              status={item.status}
+              remove={this.markForDeletion}
+              overflow={i > 4}
+              focused={this.focused}
+              blurred={this.blurred}
```
bartaz: I find these short function names (`focused`, `blurred`, `remove`) a bit confusing, especially when mixed with other "value" props like `focus`. It's not clear from their name that they are functions/callbacks and it makes the code bit harder to read I think. Naming like `focusHandler` or `onItemFocus` or similar would make it clearer that it's a function and that it's meant to be handling some event.

This is both about naming class functions (`this.focused`) and props of component.

hughesjordan:
@bartaz Okay lets make them all the same size

hughesjordan:
> One thing that I'm missing a bit is some kind of status message when marking screenshots to delete.
> 
> In current implementation screenshots are marked and a message is shown:
> Screenshot 2019-03-12 at 14 32 12
> 
> It makes it clear that these changes were not saved yet.
> 
> This PR changes that, when screenshot is 'deleted' it is removed from the view, but there is no other sign showing that it needs to be saved. It gives the impression that screenshots are removed immediately, when they are not until 'Save' is clicked.

@bartaz @Lukewh Not sure that we need this? To the user all they need to know is that the images that they see in front of them are the ones that will appear on the listing if and when they click on save. If they try to leave the page with unsaved changes we will prompt them anyway.

Also the way it works is strange... I can add 3 images and without having saved, then delete them. It will say 3 images to be deleted when nothing has been uploaded.

bartaz:
If images that are not saved get deleted and appear as 'to delete' - that seems like a bug.

If we want to show this summary at all - I'll leave that for UX to decide. For me they seem useful - they give additional context on what is yet unsaved. Sure - we don't do it for other fields, but why not if we can for screenshots? Especially that adding screenshots require uploading which is a longer action, so it's good to let users know what to expect.

therealjuan:
Can we get rid of this messaging as it feels like is overcomplicating things? Also as @bartaz comments is not happening in other fields and the main reason why we have the sticky element is to make sure people understand there are unsaved changes. Also, we have a dialog alert once you leave the page.

If we really see this problem arising for users, we can always add it back or look on ways to improve so 🔪 

On other news: Demo service does not seem to work with this PR

Issue #1682, opened 2019-03-12
Unable to remove icon set for snap


anonymouse64:
### Expected behaviour

Through the store listing page I should be able to remove an icon that was previously uploaded for the snap using the store listing. 

I basically don't want to have an icon set for the snap at all so that it uses the default icon in the store listing instead.

### Steps to reproduce the problem

One can only upload a new icon, not remove a current one.

PR #1683, opened 2019-03-13
Allow removing of snap icon

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1682

## Done

- New layout (label to the left)
- Added "Change" button (rather then clicking on the image)
- Made the image slightly bigger - why not?
- Added a "Remove" button

## QA

- Pull the branch
- `./run`
- Add remove an icon, reload the page etc.
- See that it works as expected

### DIFF
--- a/static/js/publisher/form.js
+++ b/static/js/publisher/form.js
@@ -19,7 +19,14 @@ const IS_CHROMIUM =
   typeof window.chrome !== "undefined" &&
   window.navigator.userAgent.indexOf("Edge") === -1; // Edge pretends to have window.chrome
 
-function initSnapIconEdit(changeIcon, removeIcon, iconId, iconInputId, state) {
+function initSnapIconEdit(
+  changeIcon,
+  removeIcon,
+  iconId,
+  iconInputId,
+  state,
+  updateFormState
+) {
   const snapIconEl = document.getElementById(iconId);
   const snapIconInput = document.getElementById(iconInputId);
   const changeIconEl = document.querySelector(changeIcon);
@@ -41,6 +48,7 @@ function initSnapIconEdit(changeIcon, removeIcon, iconId, iconInputId, state) {
     });
 
     updateState(state, { images });
+    snapIconEl.classList.remove("u-hide");
     removeIconEl.classList.remove("u-hide");
   });
 
@@ -56,9 +64,10 @@ function initSnapIconEdit(changeIcon, removeIcon, iconId, iconInputId, state) {
 
     const images = state.images.filter(image => image.type !== "icon");
 
-    updateState(state, { images });
-
     snapIconInput.value = "";
+    updateState(state, { images });
+    updateFormState();
+    snapIconEl.classList.add("u-hide");
     removeIconEl.classList.add("u-hide");
   });
 
@@ -157,7 +166,8 @@ function initForm(config, initialState, errors) {
       config.snapIconRemove,
       config.snapIcon,
       config.snapIconInput,
-      state
+      state,
+      updateFormState
     );
   }
 
--- a/static/js/publisher/form.js
+++ b/static/js/publisher/form.js
@@ -19,9 +19,11 @@ const IS_CHROMIUM =
   typeof window.chrome !== "undefined" &&
   window.navigator.userAgent.indexOf("Edge") === -1; // Edge pretends to have window.chrome
 
-function initSnapIconEdit(iconElId, iconInputId, state) {
+function initSnapIconEdit(changeIcon, removeIcon, iconId, iconInputId, state) {
+  const snapIconEl = document.getElementById(iconId);
   const snapIconInput = document.getElementById(iconInputId);
-  const snapIconEl = document.getElementById(iconElId);
+  const changeIconEl = document.querySelector(changeIcon);
+  const removeIconEl = document.querySelector(removeIcon);
 
   snapIconInput.addEventListener("change", function() {
     const iconFile = this.files[0];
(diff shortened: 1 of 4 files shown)


bartaz:
The icon seems to be unnecessary large:

Screenshot 2019-03-14 at 16 51 56

It's displayed way bigger than in any other place in UI.
Previous version of listing page used 80x80px size, but the biggest size we display the icon in public views currently is 60x60px on snap details page. So I think that similar size would be enough for this view.

bartaz:
I think it would be nicer if visually we could make it more consistent:

Screenshot 2019-03-15 at 10 06 56

It looks a bit strange with 'Edit' on top of icon and 'Delete' separately next to it.
Maybe both edit and delete could be small buttons next to icon?
Or we don't need edit icon at all (similar to screenshots)?

@therealjuan @hughesjordan Thoughts?

therealjuan:
It might be better to have a button next to the image with Edit icon and a link with reset to default

`[       Icon       ] [Edit snap icon] _Reset to default_`

Lukewh:
> I think it would be nicer if visually we could make it more consistent:
> 
> Screenshot 2019-03-15 at 10 06 56
> 
> It looks a bit strange with 'Edit' on top of icon and 'Delete' separately next to it.
> Maybe both edit and delete could be small buttons next to icon?
> Or we don't need edit icon at all (similar to screenshots)?
> 
> @therealjuan @hughesjordan Thoughts?

The different with screenshots is that the only action is to delete them, you don't change them in place.
I didn't want to put the delete over the icon as it's already quite small and the edit icon covers half the image... I'm not sure if there's a way to nicely implement something similar to https://github.com/canonical-websites/snapcraft.io/pull/1696#issuecomment-472973907 with such a small image?

static/js/publisher/form.js
```diff
@@ -39,11 +40,35 @@ function initSnapIconEdit(iconElId, iconInputId, state) {
...
+    e.preventDefault();
+    snapIconEl.src =
+      "https://assets.ubuntu.com/v1/be6eb412-snapcraft-missing-icon.svg";
+
+    const images = state.images.filter(image => image.type !== "icon");
+
+    updateState(state, { images });
+
+    snapIconInput.value = "";
+    snapIconInput.closest("form").dispatchEvent(
```
bartaz: Why do we need to dispatch event manually here? Especially "change" event on a form object?

Lukewh:
> It might be better to have a button next to the image with Edit icon and a link with reset to default
> 
> `[ Icon ] [Edit icon] _Reset to default_`

I something similar originally, but moved to use the remove icon to try and keep it consistent with the other "media" inputs on the page. Feels like a good opportunity to make things consistent.

static/js/publisher/form.js
Lukewh: Good point, I should just `updateState` :+1:

Lukewh: ... It's already `updateState`ing so yeah can remove :+1::+1:

therealjuan:
So then we can just include edit icon next to the image and leave the delete as in here but without the square button, just a simple button. 

image

static/js/publisher/form.js
bartaz: Not sure what you want to trigger here. There is some function (`checkForm`?) to update the Revert/Submit buttons state based on form state. But I'm not sure if `updateState` doesn't to it internally as well...

Lukewh: I'm not sure either now - I think because I was resetting the value I needed to cause the form to update

Lukewh:
> So then we can just include edit icon next to the image and leave the delete as in here but without the square button, just a simple button.
> 
> image

:+1: works for me

therealjuan:
Great good morning btw!

therealjuan:

bartaz:
As much as it works with screenshots I'm not sure if overlaying 'delete' on icon looks good.

Screenshot 2019-03-15 at 11 03 54

Icons are quite small so 'delete' covers a bigger space than in screenshot.
Also, icon is quite often important part of the brand, so covering it with icon may just look bad from that point of view as well. It has already been discussed before when half of the icon was covered with 'Edit' overlay.

templates/publisher/listing.html
```diff
@@ -95,19 +95,31 @@
...
           <div class="col-2">
-            <div class="p-media-object__image--large p-editable-icon">
-              <img id="snap_icon_image"
-                {% if icon_url %}
-                  src="{{ icon_url }}" alt="{% if display_title %}{{ display_title }}{% else %}{{ snap_title }}{% endif %} snap"
-                {% else %}
-                  src="https://assets.ubuntu.com/v1/be6eb412-snapcraft-missing-icon.svg" alt=""
-                {% endif %}
-              />
+            <label class="p-editable-icon__label">Snap icon:</label>
```
bartaz: Could we link this label with icon input so clicking on it would open file dialog?

hughesjordan:
@bartaz @Lukewh @therealjuan how about this?

Empty:
Screenshot 2019-03-15 at 11 19 15 am

With icon:
Screenshot 2019-03-15 at 11 19 19 am

Hover:
Screenshot 2019-03-15 at 11 19 26 am

hughesjordan:
Equally this could work, without the need for the tooltip:
Screenshot 2019-03-15 at 12 22 16 pm

Lukewh:
> Equally this could work, without the need for the tooltip:
> Screenshot 2019-03-15 at 12 22 16 pm

@bartaz what do you think of ^

bartaz:
I think reverting after removing icon doesn't work correctly.

1. Go to a snap with an icon
2. Click on 'delete icon' icon
3. Icon is removed, + appears
4. Click 'Revert' on the form
5. Nothing happens

static/sass/_snapcraft_market.scss
```diff
@@ -14,45 +14,92 @@
...
+        top: 0;
+      }
+
+      img {
+        background: $color-x-light;
+        height: 100%;
+        width: 100%;
+
+        &[src=""] {
+          display: none;
```
bartaz: Why is it possible to have img without `src`?

hughesjordan:
@Lukewh one small thing I've just noticed.. perhaps not necessary to have this overlay as the icons can be square.
Screenshot 2019-03-18 at 9 23 10 am

Lukewh:
> @Lukewh one small thing I've just noticed.. perhaps not necessary to have this overlay as the icons can be square.
> Screenshot 2019-03-18 at 9 23 10 am

Do we use a square version anywhere?

hughesjordan:
@Lukewh these?
Screenshot 2019-03-18 at 9 27 40 am

Lukewh:
> @Lukewh these?

It's pretty bad I didn't notice this when I had a hand in redoing that page.... I'll remove the circle overlay :+1:

Lukewh:
Updated @hughesjordan

bartaz:
I think the circle overlay was there because icon could be displayed as circle or square depending on the context. I think we are showing them as squares (or simply however they are in the image) now everywhere, so overlay is indeed not needed.

hughesjordan:
@Lukewh 👍

Issue #111, opened 2017-11-03
Research snap search API


nottrobin:
Docs here: https://search.apps.ubuntu.com/docs/

Come up with a very basic plan for how to wire up this app to talk to that API.

tbille:
# Report

Here is the Snapcraft enpoints documentation: https://search.apps.ubuntu.com/docs/#snap-specific-endpoints

I am going to add an endpoint to the backend:
```
GET /search?q={searched_string:string}&limit={limit:int}&offset={offset:int}
```

It will have 3 parameters

- **q** *string*: searched string
- **offset** *int*: number of snaps already searched
- **limit** *int*: number of snaps 

**Returns:** The endpoint will return a template with `limit` snaps.

The snapcraft search endpoint doesn't provides the `offset` and the `limit` parameters. I will develop a small algorithm to convert `offset, limit` to `page, size` already provided by the snapcraft search endpoint. This is to get the same type of endpoint as we have on https://www.ubuntu.com/search?q=snap&limit=10&offset=20

bartaz:
@tbille 

Probably @nottrobin can comment on that, but I'm not sure we need to convert `page` and `size` into `offset` and `limit`. Both are rather technical implementation details and don't matter much from front-end or user perspective.

Or course if there is a particular reason for using `offset` and `limit` (other then just to be the same as on ubuntu.com search), I'm fine with that.

But snapcraft.io pages will use snap store APIs, so I suppose they all will use `page` `size` consistently, so maybe it would be better to stick with what API provides.

nottrobin:
@bartaz I asked @tbille to use `offset` and `limit`.

Interfaces are important and shouldn't be chosen lightly. I suggested `limit` and `offset` precisely to fit in with the www.ubuntu.com search and the standard suggested in [this book](https://pages.apigee.com/rs/apigee/images/api-design-ebook-2012-03.pdf), which I find very sensible. This was carefully considered at the time.

I think standardisation in our URL interfaces across our websites is certainly useful - our developers and the users of our sites (who will be far greater in number than the users of the Snap Store API) will appreciate the consistency.

I actually also think there's a small benefit to being different from the Store API to indicate that this is an independent service. People shouldn't assume it provides all the features of the API, which it definitely won't, or that it even returns exactly the same data from the features that it does share, which it probably won't. Also, not being closely tied to that API's standards leaves the Store API to iterate freely. If they change that interface in v2, we won't have to play catch-up.

[...]

PR #1687, opened 2019-03-13
JetBrains publisher page

tbille:
Fixes https://github.com/CanonicalLtd/snap-squad/issues/885
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1688
Publisher pages template.

JetBrains example:
JetBrains - Snap Store

**Note:** Moved fluid-grid content to use grid instead of flex


bartaz:
@therealjuan Changing "flex grid" from flexbox to grid is a significant change to styling that affects multiple changes. We shouldn't do it as part of JetBrains publisher page PR.

I can already see couple of places where styles are broken and I don't know if this is because of grid or some other code.

On home page OS logos become very big on 'medium' screens - they shouldn't do that (old flex implementation keeps them as 4 in a row), same happens on /build page with OS logos:
Screenshot 2019-03-14 at 08 45 34

On Store page (and other places where snaps are listed - publisher page as well) spacing is broken - there is way too much space between snap name and publisher name:

Screenshot 2019-03-14 at 08 47 57

There may be other places affected as well, we would need to check where this pattern is used on the site.

webapp/store/views.py
```diff
@@ -231,6 +234,37 @@ def brand_search_snap():
...
+                os.path.join(flask.current_app.root_path, file), "r"
+            ) as stream:
+                data = yaml.load(stream)
+        except Exception:
+            data = None
+
+        return data
+
+    @store.route("/publisher/jetbrains")
+    def publisher_details_kde():
```
bartaz: That is not KDE anymore
```suggestion
    def publisher_details_jetbrains():
```

webapp/store/views.py
```diff
@@ -231,6 +234,37 @@ def brand_search_snap():
             status_code,
         )
 
+    # @store.route(
+    #     '/publisher/<regex("[a-z0-9-]*[a-z][a-z0-9-]*"):publisher_name>'
+    # )
```
bartaz: This could be removed, for now we only have hardcoded publisher pages.

webapp/store/content/publisher-kde.yaml
```diff
@@ -0,0 +1,865 @@
+publisher: KDE
```
bartaz: I guess we don't need this file until we decide to add KDE page?

evandandrea:
I know it's been raised before, but so I have something written to remind myself of the rationale: how do you feel about not showing the publisher name for the snaps when they all match?

therealjuan:
Sure, @evandandrea and will include the one line summary perhaps which is more useful. 

Additionally, we will remove the `Featured snaps` title and the `15 published snaps` underneath as it is confusing as visually they already look featured.

evandandrea:
This looks much better. Thanks @therealjuan!

webapp/store/views.py
```diff
@@ -231,6 +234,33 @@ def brand_search_snap():
...
+                os.path.join(flask.current_app.root_path, file), "r"
+            ) as stream:
+                data = yaml.load(stream)
+        except Exception:
+            data = None
+
+        return data
+
+    @store.route("/publisher/jetbrains")
+    def publisher_details_jetbrains():
```
bartaz: @tbille @jpmartinspt We would need some backend look on this. It was quickly glued together with bits from other parts of the app (we read YAML for first snap flow), but I'm not sure if that is good to go as it is.

Also this endpoint needs a test not to decrease coverage.

webapp/store/views.py
```diff
@@ -231,6 +235,30 @@ def brand_search_snap():
...
+
+        return data
+
+    @store.route("/publisher/<publisher>")
+    def publisher_details_jetbrains(publisher):
+        """
+        A view to display the publisher details page for specific publisher.
+        """
+
+        context = _get_file("store/content/publishers/" + publisher + ".yaml")
```
bartaz: Is it safe to pass `publisher` directly from URL to file path without any regexp in route or any other validation?

tbille: this is an excellent question. Let me check.

templates/store/publisher-details.html
```diff
@@ -0,0 +1,212 @@
...
+    <div class="row">
+      <div class="p-snap-heading">
+        {% if icon_url %}
+          <img class="p-snap-heading__icon" src="{{ icon_url }}" alt="{{ publisher }} snap" />
+        {% else %}
+          <img class="p-snap-heading__icon" src="https://assets.ubuntu.com/v1/6fbb3483-snapcraft-default-snap-icon.svg" alt="" />
+        {% endif %}
+        <div class="p-snap-heading__title">
+          <h1 class="p-heading--two p-snap-heading__name">{{ publisher }}</h1>
+          <p class="p-snap-heading__publisher">
```
Lukewh:
```suggestion
          <p class="p-snap-heading__publisher u-no-margin--bottom">
```
Removes the extra blank space to balance the top and bottom of the strip a bit, imo.

templates/store/publisher-details.html
```diff
@@ -0,0 +1,212 @@
...
+
+{% if icon_url %}
+  {% block meta_image %}
+    {{ icon_url }}
+  {% endblock %}
+{% endif %}
+
+
+{% block content %}
+  <div class="p-strip--dark is-shallow" style="background: #262626">
```
Lukewh: If we don't use` p-strip--dark` anywhere else can we override the default vanilla `#111` with `#262626`?

templates/store/publisher-details.html
```diff
@@ -0,0 +1,212 @@
...
+        </div>
+
+      </div>
+    </div>
+  </div>
+
+  {% if banner_url %}
+  <section
+    class="p-strip--image is-deep"
+    style="min-height: 400px; background-image:url('{{ banner_url }}'); background-position: center;">
```
Lukewh: can we put these in a pattern too (apart from `background-image` obvs?

templates/store/publisher-details.html
```diff
@@ -0,0 +1,212 @@
...
+            <img src="https://assets.ubuntu.com/v1/be6eb412-snapcraft-missing-icon.svg" class="p-media-object--snap__img" alt="">
+            {% endif %}
+
+            <div class="p-media-object--snap__content">
+              <h4 class="p-media-object--snap__title">
+                {{ snap.title }}
+              </h4>
+              <div class="p-media-object--snap__desc">
+                <p>{{ snap.summary }}</p>
+                </p>
```
Lukewh: Extra `p`

templates/store/publisher-details.html
```diff
@@ -0,0 +1,212 @@
...
+    </div>
+  </div>
+  {% endif %}
+
+  <div class="p-strip is-shallow">
+    <div class="row">
+      <div class="col-12 p-fluid-grid__publisher-snaps">
+      {% for snap in snaps %}
+        <div class="p-fluid-grid__item">
+          <a class="p-media-object--snap{% if snap_no_margin %}u-no-margin--bottom{% endif %}" href="/{{ snap.package_name }}">
```
Lukewh: You can get rid of the `{% if snap_no_margin %}` block, it's not defined anywhere

templates/store/publisher-details.html
```diff
@@ -0,0 +1,212 @@
...
+            {% else %}
+            <img src="https://assets.ubuntu.com/v1/be6eb412-snapcraft-missing-icon.svg" class="p-media-object--snap__img" alt="">
+            {% endif %}
+
+            <div class="p-media-object--snap__content">
+              <h4 class="p-media-object--snap__title">
+                {{ snap.title }}
+              </h4>
+              <div class="p-media-object--snap__desc">
+                <p>{{ snap.summary }}</p>
```
Lukewh: Can we update `templates/store/_media-object-snap-partial.html` to handle this small variation?

```
// templates/store/_media-object-snap-partial.html
{% if snap_show_summary %}
  <p>{{ snap.summary }}</p>
{% else %}
  ...the current <p> in that template
{% endif %}
```

```
// publisher-details.html
{% for snap in snaps %}
  {% set snap_show_summary=true %}
  {% include("store/_media-object-snap-partial.html") %}
{% endfor %}
```

webapp/store/content/publishers/jetbrains.yaml
```diff
@@ -0,0 +1,155 @@
...
+    {
+        "developer_validation": "verified",
+        "icon_url": "https://dashboard.snapcraft.io/site_media/appmedia/2017/11/WebStorm_1282x.png",
+        "origin": "jetbrains",
+        "package_name": "webstorm",
+        "publisher": "JetBrains",
+        "summary": "Powerful IDE for modern JavaScript development",
+        "title": "WebStorm",
+    }
+]
```
Lukewh: this file is a bit painful, can't we just provide snap names and automate getting the information? I know it won't be the most performant, but with this we have to update whenever they make a change? Could we not pre-fetch to fill this file out on startup or something so at least we just need to redeploy to get their latest?

webapp/store/views.py
```diff
@@ -231,6 +235,30 @@ def brand_search_snap():
             status_code,
         )
 
+    def _get_file(file):
```
jpmartinspt: What about _get_context?

webapp/store/views.py
```diff
@@ -231,6 +235,30 @@ def brand_search_snap():
...
+        try:
+            with open(
+                os.path.join(flask.current_app.root_path, file), "r"
```
jpmartinspt: We could maybe define the publishers yaml directory in settings and just append the publisher name to it?

webapp/store/content/publishers/jetbrains.yaml
tbille: This could be done as an improvement instead I agreee. I think for now as a proof of concept we can keep this like this and re-iterate on it later (?)

webapp/store/views.py
therealjuan: 👍

templates/store/publisher-details.html
therealjuan: 👍

templates/store/publisher-details.html
therealjuan: 👍

templates/store/publisher-details.html
therealjuan: 👍

templates/store/publisher-details.html
therealjuan: 👍

templates/store/publisher-details.html
therealjuan: 👍

webapp/store/content/publishers/jetbrains.yaml
therealjuan: https://github.com/canonical-websites/snapcraft.io/issues/1708

templates/store/_media-object-snap-partial.html
```diff
@@ -10,14 +10,18 @@ <h4 class="p-media-object--snap__title">
...
+        <p>
+          <span class="u-off-screen">Publisher: </span>{{ snap.origin }}
+          {% if snap.developer_validation and snap.developer_validation == VERIFIED_PUBLISHER %}
+          <span class="p-verified p-tooltip p-tooltip--top-center" aria-describedby="{{ snap.package_name }}-tooltip">
+                        <img src="https://assets.ubuntu.com/v1/75654c90-rosette.svg" />
+                        <span class="p-tooltip__message u-align--center" role="tooltip" id="{{ snap.package_name }}-tooltip">Verified account</span>
+                      </span>
+          {% endif %}
+      {% endif %}
       </p>
```
Lukewh: `<p>` needs to go up inside the `{% endif %}`'s

therealjuan: 👍

Lukewh:
Some of the spacing isn't ideal on mobile
Screenshot_2019-03-19 JetBrains in the Snap Store
Screenshot_2019-03-19 JetBrains in the Snap Store(1)
Screenshot_2019-03-19 JetBrains in the Snap Store(2)

kwm14:
Design not needed 😢

Issue #1690, opened 2019-03-14
Add restrictions information


Lukewh:

therealjuan:
AFAIK this is already added to the publisher page so closing.

PR #1696, opened 2019-03-14
Upload and edit banners

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1689
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1690

## QA
- Pull the branch
- `./run`
- Follow the guidance displayed below the banner area "Show banner restrictions"
- Any file that doesn't meet the restrictions should fail with an error
- You should be able to set, remove and change the image

### DIFF
--- a/static/js/libs/fileValidation.js
+++ b/static/js/libs/fileValidation.js
@@ -42,6 +42,34 @@ function baseRestrictions(file, restrictions) {
   });
 }
 
+function imageWhitelistHandler(file, image, whitelist) {
+  const errors = whitelist.filter(whitelistItem => {
+    if (whitelistItem.fileName) {
+      let fileName = file.name.split(".");
+      fileName = fileName.slice(0, fileName.length - 1).join(".");
+      if (fileName !== whitelistItem.fileName) {
+        return false;
+      }
+    }
+    if (whitelistItem.accept) {
+      if (!whitelistItem.accept.includes(file.type)) {
+        return false;
+      }
+    }
+    if (whitelistItem.dimensions) {
+      if (
+        image.naturalWidth !== whitelistItem.dimensions[0] ||
+        image.naturalHeight !== whitelistItem.dimensions[1]
+      ) {
+        return false;
+      }
+    }
+    return true;
+  });
+
+  return errors.length > 0;
+}
+
 function imageRestrictions(file, restrictions) {
   return new Promise((resolve, reject) => {
     if (!restrictions.accept || restrictions.accept[0].indexOf("image") < 0) {
@@ -61,14 +89,12 @@ function imageRestrictions(file, restrictions) {
         height: height
       };
 
-      if (restrictions.whitelist && restrictions.whitelist.dimensions) {
-        if (
-          width === restrictions.whitelist.dimensions[0] &&
-          height === restrictions.whitelist.dimensions[1]
-        ) {
-          resolve(file);
-          return;
-        }
+      if (
+        restrictions.whitelist &&
+        imageWhitelistHandler(file, image, restrictions.whitelist)
+      ) {
+        resolve(file);
+        return;
       }
 
       const aspectRatio = width / height;
--- a/static/js/libs/fileValidation.js
+++ b/static/js/libs/fileValidation.js
@@ -56,16 +56,14 @@ function imageRestrictions(file, restrictions) {
       const width = image.naturalWidth;
       const height = image.naturalHeight;
 
-      if (
-        (file.name.indexOf("banner.") === 0 &&
-          width === 1218 &&
-          height === 240) ||
-        (file.name.indexOf("banner-icon.") === 0 &&
-          width === 240 &&
-          height === 240)
-      ) {
-        resolve(file);
-        return;
+      if (restrictions.whitelist && restrictions.whitelist.dimensions) {
+        if (
+          width === restrictions.whitelist.dimensions[0] &&
+          height === restrictions.whitelist.dimensions[1]
+        ) {
+          resolve(file);
+          return;
+        }
       }
 
       const aspectRatio = width / height;
@@ -112,13 +110,25 @@ function imageRestrictions(file, restrictions) {
         }
 
         if (hasAspectError) {
-          reject([
-            `has a width (${width} pixels) that is ${aspectRatio.toFixed(
-              2
-            )}x its height (${height}). Its width needs to be between ${aspectRatioMin[0] /
-              aspectRatioMin[1]}x and ${aspectRatioMax[0] /
-              aspectRatioMax[1]}x the height.`
-          ]);
+          const min =
+            Math.round((aspectRatioMin[0] / aspectRatioMin[1]) * 100) / 100;
+          const max =
+            Math.round((aspectRatioMax[0] / aspectRatioMax[1]) * 100) / 100;
+          const actual = Math.round(aspectRatio * 100) / 100;
+
+          const message = [
+            `has a width (${width} pixels) that is ${actual}x its height (${height} pixels).`
+          ];
+
+          if (min === max) {
+            message.push(`Its width needs to be ${min}x the height.`);
+          } else {
+            message.push(
+              `Its width needs to be between ${min}x and ${max}x the height.`
+            );
+          }
+
+          reject([message.join(" ")]);
           return;
         }
       }
--- a/static/js/libs/fileValidation.js
+++ b/static/js/libs/fileValidation.js
@@ -84,11 +84,6 @@ function imageRestrictions(file, restrictions) {
       const width = image.naturalWidth;
       const height = image.naturalHeight;
 
-      file.meta = {
-        width: width,
-        height: height
-      };
-
       if (
         restrictions.whitelist &&
         imageWhitelistHandler(file, image, restrictions.whitelist)
@@ -98,7 +93,6 @@ function imageRestrictions(file, restrictions) {
       }
 
       const aspectRatio = width / height;
-      file.meta = aspectRatio;
       let hasDimensionError = false;
       if (restrictions.width) {
         if (
--- a/static/js/libs/fileValidation.js
+++ b/static/js/libs/fileValidation.js
@@ -56,6 +56,11 @@ function imageRestrictions(file, restrictions) {
       const width = image.naturalWidth;
       const height = image.naturalHeight;
 
+      file.meta = {
+        width: width,
+        height: height
+      };
+
       if (restrictions.whitelist && restrictions.whitelist.dimensions) {
         if (
           width === restrictions.whitelist.dimensions[0] &&
@@ -67,6 +72,7 @@ function imageRestrictions(file, restrictions) {
       }
 
       const aspectRatio = width / height;
+      file.meta = aspectRatio;
       let hasDimensionError = false;
       if (restrictions.width) {
         if (
--- a/static/js/publisher/form.js
+++ b/static/js/publisher/form.js
@@ -6,6 +6,7 @@ import { categories } from "./market/categories";
 import { storageCommands } from "./market/storageCommands";
 import { initMedia } from "./market/initMedia";
 import { initIcon } from "./market/initIcon";
+import { initBanner } from "./market/initBanner";
 
 // https://gist.github.com/dperini/729294
 // Luke 07-06-2018 made the protocol optional
@@ -120,17 +121,24 @@ function initForm(config, initialState, errors) {
 
   if (config.mediaHolder) {
     const screenshots = state.images.filter(
-      image => image.type === "screenshot"
+      image => image.type === "screenshot" && !image.isBanner
     );
     initMedia(config.mediaHolder, screenshots, newImages => {
       const noneScreenshots = state.images.filter(
-        item => item.type !== "screenshot"
+        item => item.type !== "screenshot" || item.isBanner
       );
       const newState = {
         ...state,
         images: noneScreenshots.concat(newImages)
       };
       updateState(state, newState);
+      updateFormState();
+    });
+  }
+
+  if (config.bannerHolder) {
+    initBanner(config.bannerHolder, state, nextState => {
+      updateState(state, nextState);
       updateFormState();
     });
   }
--- a/static/js/publisher/form.js
+++ b/static/js/publisher/form.js
@@ -121,11 +121,11 @@ function initForm(config, initialState, errors) {
 
   if (config.mediaHolder) {
     const screenshots = state.images.filter(
-      image => image.type === "screenshot" && !image.isBanner
+      image => image.type === "screenshot"
     );
     initMedia(config.mediaHolder, screenshots, newImages => {
       const noneScreenshots = state.images.filter(
-        item => item.type !== "screenshot" || item.isBanner
+        item => item.type !== "screenshot"
       );
       const newState = {
         ...state,
@@ -137,8 +137,19 @@ function initForm(config, initialState, errors) {
   }
 
   if (config.bannerHolder) {
-    initBanner(config.bannerHolder, state, nextState => {
-      updateState(state, nextState);
+    let banners = state.images.filter(image => image.type === "banner");
+    initBanner(config.bannerHolder, banners, image => {
+      let newImages = state.images.filter(image => image.type !== "banner");
+
+      if (image) {
+        newImages = newImages.concat([image]);
+      }
+
+      const newState = {
+        ...state,
+        images: newImages
+      };
+      updateState(state, newState);
       updateFormState();
     });
   }
--- a/static/js/publisher/form/AccordionHelp.js
+++ b/static/js/publisher/form/AccordionHelp.js
@@ -25,17 +25,17 @@ class AccordionHelp extends React.Component {
     const { name, children } = this.props;
 
     const classNames = ["row"];
-    let verb = "Hide";
+    let label = `Hide ${name.toLowerCase()}`;
     if (!open) {
       classNames.push("u-hide");
-      verb = "Show";
+      label = `Show ${name.toLowerCase()}`;
     }
 
     return (
       <Fragment>
         <p className={`p-form-help-text ${open ? "u-no-margin--bottom" : ""}`}>
           <a role="button" onClick={this.toggleHelp}>
-            {verb} {name.toLowerCase()}
+            {label}
           </a>
         </p>
         <div className={classNames.join(" ")}>{children}</div>
--- a/static/js/publisher/form/AccordionHelp.js
+++ b/static/js/publisher/form/AccordionHelp.js
@@ -38,9 +38,7 @@ class AccordionHelp extends React.Component {
             {verb} {name.toLowerCase()}
           </a>
         </p>
-        <div className={classNames.join(" ")}>
-          <div className="col-8">{children}</div>
-        </div>
+        <div className={classNames.join(" ")}>{children}</div>
       </Fragment>
     );
   }
--- a/static/js/publisher/form/AccordionHelp.js
+++ b/static/js/publisher/form/AccordionHelp.js
@@ -1,0 +1,58 @@
+import React, { Fragment } from "react";
+import PropTypes from "prop-types";
+
+class AccordionHelp extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.toggleHelp = this.toggleHelp.bind(this);
+
+    this.state = {
+      open: false
+    };
+  }
+
+  toggleHelp() {
+    const { open } = this.state;
+
+    this.setState({
+      open: !open
+    });
+  }
+
+  render() {
+    const { open } = this.state;
+    const { name, children } = this.props;
+
+    const classNames = ["row"];
+    let verb = "Hide";
+    if (!open) {
+      classNames.push("u-hide");
+      verb = "Show";
+    }
+
+    return (
+      <Fragment>
+        <p className={`p-form-help-text ${open ? "u-no-margin--bottom" : ""}`}>
+          <a role="button" onClick={this.toggleHelp}>
+            {verb} {name.toLowerCase()}
+          </a>
+        </p>
+        <div className={classNames.join(" ")}>
+          <div className="col-8">{children}</div>
+        </div>
+      </Fragment>
+    );
+  }
+}
+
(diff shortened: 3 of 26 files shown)


static/js/publisher/form/banner.js
```diff
@@ -0,0 +1,260 @@
...
+    if (this.state.bannerIcon.url) {
+      images.push(this.state.bannerIcon);
+    }
+
+    this.props.updateImageState(images);
+  }
+
+  bannerImageChangeHandler() {
+    const file = this.bannerImageInput.files[0];
+    if (file.name !== "banner.jpg" && file.name !== "banner.png") {
```
bartaz: The file name restriction is purely internal, because screenshots are used to store them.
It's a pity that having a dedicated UI we still expose this internal limitation to the user by forcing them to name their images properly.

Could we rename the file on our front- or backend to the names required by the store?

Lukewh: We're in discussions on how we're going to improve banners (to be used across both the website and desktop store) this is an interim implementation.

I've added [this issue](https://github.com/canonical-websites/snapcraft.io/issues/1698) so we can discuss it further outside this PR :+1:

hughesjordan:
@Lukewh niiiice, one thing though, a personal preference but not too keen on the edit icon when we have the delete top right. What about something like this?

Screenshot 2019-03-14 at 5 19 38 pm

Lukewh:
> @Lukewh niiiice, one thing though, a personal preference but not too keen on the edit icon when we have the delete top right. What about something like this?
> 
> Screenshot 2019-03-14 at 5 19 38 pm

@hughesjordan Would you want the same for the icon?

webapp/publisher/snaps/logic.py
```diff
@@ -310,3 +328,25 @@ def convert_date(date_to_convert):
     """
     date_parsed = parser.parse(date_to_convert).replace(tzinfo=None)
     return date_parsed.strftime("%B %Y")
+
+
+def categorise_media(media):
```
tbille: nice one, but could this be refactored to make one iteration on the array media instead?

webapp/publisher/snaps/views.py
```diff
@@ -278,12 +278,9 @@ def get_listing_snap(snap_name):
     is_on_stable = logic.is_snap_on_stable(snap_details["channel_maps_list"])
 
     # Filter icon & screenshot urls from the media set.
-    icon_urls = [
-        m["url"] for m in snap_details["media"] if m["type"] == "icon"
-    ]
-    screenshot_urls = [
-        m["url"] for m in snap_details["media"] if m["type"] == "screenshot"
-    ]
+    (icon_urls, screenshot_urls, banner_urls) = logic.categorise_media(
```
tbille: can't you remove the parenthesis?

```
icon_urls, screenshot_urls, banner_urls = logic.categorise_media(...)
```

hughesjordan:
@Lukewh To make these consistent with the suggestion for the add/remove/edit icon we could do the following:

Screenshot 2019-03-15 at 12 20 13 pm

hughesjordan:
@Lukewh if I add and then remove either the banner or icon, I then can't upload the same banner or icon afterwards

Lukewh:
> @Lukewh if I add and then remove either the banner or icon, I then can't upload the same banner or icon afterwards

Should be sorted now @hughesjordan :+1:. Was working fine in firefox, but derped in chrome

hughesjordan:
@Lukewh a few teeny things

I feel like the remove icon could do with being darker so that it stands out more & could increase the opacity of the white overlay it sits on. Currently in images below I think it gets lost a bit. On mouse over the colour could also turn to red. (Should be the same for all asset upload areas)

Side note - these icons should only appear on hover for the images area too.

Screenshot 2019-03-18 at 9 55 52 am
Screenshot 2019-03-18 at 9 56 00 am

Text on the mouse over state of the background should probably read 'Edit background' to match the resting state.

hughesjordan:
@Lukewh don't hate me for this but the icon looks like it is off pixel +1 otherwise

Screenshot 2019-03-18 at 2 15 00 pm

webapp/publisher/snaps/logic.py
```diff
@@ -310,3 +328,27 @@ def convert_date(date_to_convert):
...
```
bartaz: Do we have a test for it?

Lukewh: I cover it in the listing page post and get endpoint testing

Lukewh: https://codecov.io/gh/canonical-websites/snapcraft.io/pull/1696/diff?src=pr&el=tree#D2-329

templates/publisher/listing.html
```diff
@@ -543,6 +561,9 @@
         {% for screenshot_url in screenshot_urls %}
           { url: {{ screenshot_url|tojson }}, type: "screenshot", status: "uploaded" },
         {% endfor %}
+        {% for banner_url in banner_urls %}
+          { url: {{ banner_url|tojson }}, type: "screenshot", status: "uploaded", is_banner: true },
```
bartaz: Nitpicking a bit, but as `is_banner` is a JS property maybe making it camelCase `isBanner` would make sense?

Lukewh: :+1: I guess because I was doing it in the template I automatically went into pymode

webapp/publisher/snaps/logic.py
```diff
@@ -2,6 +2,7 @@
 import datetime
 from dateutil import parser
 from json import dumps
+from re import search as re_search
```
tbille: could you please use `import re` and and use `re.search` thanks :love_hotel:

static/js/publisher/form/banner.js
```diff
@@ -0,0 +1,262 @@
...
+        <input
+          name="banner-icon"
+          type="file"
+          hidden="hidden"
+          onChange={this.bannerIconChangeHandler}
+          ref={item => (this.bannerIconInput = item)}
+        />
+        <p className="p-form-help-text">
+          Adding a featured banner will increase your chances of being featured
+          on snapcraft.io and in GNOME software but does not immediately make
```
bartaz: Missing comma?
```suggestion
          on snapcraft.io and in GNOME software, but does not immediately make
```

webapp/publisher/snaps/logic.py
```diff
@@ -2,6 +2,7 @@
 import datetime
 from dateutil import parser
 from json import dumps
+import re
```
tbille:
```
import datetime
import hashlib
import re
from json import dumps

from dateutil import parser
```

static/js/publisher/form.js
```diff
@@ -175,11 +176,11 @@ function initForm(config, initialState, errors) {
 
   if (config.mediaHolder) {
     const screenshots = state.images.filter(
-      image => image.type === "screenshot"
+      image => image.type === "screenshot" && !image.isBanner
```
bartaz: `if (!image.isBanner) { image.isHulk = true; }` #lamenerdjokes

bartaz:
LGTM, but I guess should be Don't merge until API update is live?

Lukewh:
> LGTM, but I guess should be `Don't merge` until API update is live?

It went live an hour ago :+1:

bartaz:
@Lukewh Seems that I can't add icon or banner if I have 5 screenshots:
Screenshot 2019-03-19 at 15 27 43

bartaz:
What about snaps that have legacy banners already in screenshots?
Shouldn't they be 'migrated' to current banner?
Does API/clients use legacy banner or only new banner?
If legacy banners are not supported should we prompt users to delete them and use new banners?

Screenshot 2019-04-09 at 07 47 14

bartaz:
Banner restrictions say max size is 3840 x 1440 pixels, but such image is not accepted because it has wrong aspect ratio:

Screenshot 2019-04-09 at 08 00 39

bartaz:
It seems that aspect ratio restrictions have height and width confused. It is not possible to upload any banner (apart from whitelisted legacy one):

Screenshot 2019-04-09 at 08 06 30

static/js/publisher/form/banner.js
```diff
@@ -0,0 +1,219 @@
...
+            <small>
+              Accepted image formats include: <b>JPEG & PNG files</b>
+              <br />
+              Min resolution: <b>720 x 240 pixels</b>
+              <br />
+              Max resolution: <b>3840 x 1440 pixels</b>
+              <br />
+              Recommended (legacy) size: <b>1218 x 240 pixels</b>
+              <br />
+              Aspect ratio: <b>1:3</b>
```
bartaz: I guess this should be 3:1 (width:height)

static/js/publisher/market/restrictions.js
```diff
@@ -49,4 +49,27 @@ const ICON_RESTRICTIONS = {
...
+    min: 720,
+    max: 3840
+  },
+  height: {
+    min: 240,
+    max: 1440
+  },
+  aspectRatio: {
+    min: [1, 3],
+    max: [1, 3]
```
bartaz: 3:1 here as well.

bartaz:
The banner 'box' has 5:1 aspect ratio (legacy banner), while the expected image ratio is 3:1 which means the box will be bigger when proper banner is used.
If the expected banner is 3:1 the area should also have 3:1 aspect ratio so that there is no change in size when correct banner is used.
Choosing a legacy banner could then make the box smaller to contain it.

Screenshot 2019-04-09 at 08 12 23

static/js/publisher/form/AccordionHelp.js
```diff
@@ -0,0 +1,58 @@
...
+
+    return (
+      <Fragment>
+        <p className={`p-form-help-text ${open ? "u-no-margin--bottom" : ""}`}>
+          <a role="button" onClick={this.toggleHelp}>
+            {verb} {name.toLowerCase()}
+          </a>
+        </p>
+        <div className={classNames.join(" ")}>
+          <div className="col-8">{children}</div>
```
bartaz: If it's expected to have just one column, why using `row` and `col-` classes at all?
But, if we want to reuse it to build something like 2 column markdown help, maybe it can just render anything that it gets in `{children}`? So it will be parent component that will decide how to split content into columns?

```suggestion
          {children}
```

static/js/publisher/form/banner.js
```diff
@@ -0,0 +1,219 @@
...
+        url: url,
+        file,
+        name: file.name,
+        status: "new",
+        type: "banner"
+      });
+    }
+  }
+
+  handleRemoveImageClick(e) {
```
bartaz: Not sure if there is a simple way to abstract it, but seems like we are duplicating a lot of same logic in icon/screenshots/banner components.
While they all reuse file input component, the 'wrapper' still reimplements a lot of similar stuff.

static/sass/_snapcraft_market.scss
```diff
@@ -116,4 +118,82 @@
...
+      }
+
+      .p-market-banner__change {
+        border: 1px solid $color-link;
+        border-radius: $border-radius;
+        opacity: 1;
+      }
+    }
+
+    .p-market-banner__remove {
```
bartaz: I guess we have very similar styles in screenshots as well.
Would be good to extract them into some common pattern, so when we want to change them we can do it consistently from one place.

static/sass/_snapcraft_market.scss
```diff
@@ -116,4 +118,82 @@
...
+      position: absolute;
+      right: 0;
+      top: 0;
+    }
+  }
+
+  .p-icon--delete {
+    @include snapcraft-icon-delete($color-mid-dark);
+
+    &:hover {
```
bartaz: Icon only changes colour when hovering over icon itself, not when hovering over whole clickable "remove" area.

Screenshot 2019-04-09 at 08 31 59

templates/publisher/listing.html
```diff
@@ -101,7 +101,7 @@ <h2 class="p-heading--four">
 
           <div class="row" data-validation-name="icon">
             <div class="col-2">
-              <label class="p-editable-icon__label" for="snap_icon">Snap icon:</label>
+              <label class="p-editable-icon__label">Snap icon:</label>
```
bartaz: Why removing the label?

Lukewh:
> What about snaps that have legacy banners already in screenshots?
> Shouldn't they be 'migrated' to current banner?
> Does API/clients use legacy banner or only new banner?
> If legacy banners are not supported should we prompt users to delete them and use new banners?
> 
> Screenshot 2019-04-09 at 07 47 14

I thought about adding a tooltip to the screenshots area, but we could add a bigger info box if there's banner and banner-icon:
"We're migrating to a new banner system, while you can still upload a banner background, the banner-icon is being deprecated. [Read more](link to a blog post)"
Blog post doesn't exist yet, but seems like a good opportunity to write a more complete update on what's changing?

wdyt?

static/js/publisher/form/banner.js
Lukewh: I blindly followed the spec, reeeeaaaaalllllyyyy good catch, thanks @bartaz

static/js/publisher/form/banner.js
Lukewh: Yeah I was thinking of doing a follow up to do abstract to another component (`MediaEdit` or something) - along with the scss that's related.

bartaz:
> "We're migrating to a new banner system, while you can still upload a banner background, the banner-icon is being deprecated. [Read more](link to a blog post)"
Blog post doesn't exist yet, but seems like a good opportunity to write a more complete update on what's changing?

Sounds good to me. If there is any migration or deprecation plan it should be clearly explained to users: how long old banners will work? will the banner icon stop being displayed in clients, etc?

templates/publisher/listing.html
```diff
@@ -330,7 +340,7 @@ <h2 class="p-heading--four">Contact information</h2>
         <div class="row">
           <div class="col-2">
             <label for="developer-website">Developer website:</label>
-          </div>
+        </div>
```
bartaz: This seems like unnecessary indent move.

bartaz:
Now that legacy banner is deprecated and cannot be uploaded I think that 'Add background' box should have the aspect ratio of 3:1, so when a valid banner is selected it doesn't resize:

Screenshot 2019-04-09 at 08 12 23

static/js/publisher/form/banner.js
```diff
@@ -0,0 +1,217 @@
...
+          <div className="col-8">
+            <p>
+              <small>
+                Accepted image formats include: <b>JPEG & PNG files</b>
+                <br />
+                Min resolution: <b>720 x 240 pixels</b>
+                <br />
+                Max resolution: <b>4320 x 1440 pixels</b>
+                <br />
+                Recommended (legacy) size: <b>1218 x 240 pixels</b>
```
bartaz: It should be removed if legacy banners are not accepted anymore.

static/js/libs/fileValidation.js
```diff
@@ -56,19 +84,21 @@ function imageRestrictions(file, restrictions) {
...
-          height === 240)
+        restrictions.whitelist &&
+        imageWhitelistHandler(file, image, restrictions.whitelist)
       ) {
         resolve(file);
         return;
       }
 
       const aspectRatio = width / height;
+      file.meta = aspectRatio;
```
bartaz: Is this `meta` used anywhere?
It's first set to `{ width, height }`, here overridden by `aspectRatio`, but it doesn't appear anywhere else in the diff?

static/js/publisher/form/AccordionHelp.js
```diff
@@ -0,0 +1,56 @@
...
+    if (!open) {
+      classNames.push("u-hide");
+      verb = "Show";
+    }
+
+    return (
+      <Fragment>
+        <p className={`p-form-help-text ${open ? "u-no-margin--bottom" : ""}`}>
+          <a role="button" onClick={this.toggleHelp}>
+            {verb} {name.toLowerCase()}
```
bartaz: That's very specific "verb+name" approach. I guess having just 2 props for 'show label' and 'hide label' would be  a bit more generic and reusable. But it can be changed later if needed elsewhere.

static/js/publisher/form/banner.js
```diff
@@ -0,0 +1,217 @@
...
+        >
+          <div
+            className={backgroundClasses.join(" ")}
+            tabIndex="0"
+            onKeyDown={this.keyboardEvent}
+            onFocus={this.toggleFocus}
+            onBlur={this.toggleFocus}
+          >
+            {background && <img src={background} />}
+            {!background && (
```
bartaz: Why not using `background ?  : ( ... )` here?
It's not a nicest syntax, but at least it's closer to `if ... else` that it seems we want to achieve here.

static/js/publisher/form/banner.test.js
```diff
@@ -0,0 +1,167 @@
...
+
+const backgroundSelector = ".p-market-banner__image img";
+
+function renderBanner(bannerImage, updateState, restrictions) {
+  if (bannerImage && updateState) {
+    return render(
+      <Banner bannerImage={bannerImage} updateImageState={updateState} />
+    );
+  }
+  if (bannerImage && updateState) {
```
bartaz: Seems to be a duplicate of the code above?

static/js/publisher/form/banner.test.js
```diff
@@ -0,0 +1,167 @@
...
+    return render(
+      <Banner bannerImage={bannerImage} updateImageState={updateState} />
+    );
+  }
+  if (updateState) {
+    return render(<Banner updateImageState={updateState} />);
```
bartaz: Isn't it equivalent of `render(<Banner bannerImage={undefined} updateImageState={updateState} />)`?
So that `if`s wouldn't be needed at all and params would be directly passed to the component (if they are defined or not)?

static/js/publisher/form/banner.test.js
```diff
@@ -0,0 +1,167 @@
...
+        ).toEqual(0);
+        expect(container.querySelector(backgroundSelector).src).toEqual(
+          "banner.png"
+        );
+        done();
+      }, 500);
+    });
+
+    it("should show an error if a restriction isn't met", done => {
+      const { container } = renderBanner(null, null, true);
```
bartaz: Extracting `renderBanner` resulted in tests being harder to read.
I have no idea what `null, null, true` means. Just rendering `<Banner restrictions={{ accept: ["image/jpeg"] }} />` here inline would make it much more clear. I don't see such specific banner configurations to be reused much to justify having extracted renderBanner function (and component props obscured in function params).

static/js/publisher/form/fileInput.js
```diff
@@ -186,7 +187,7 @@ class FileInput extends React.Component {
         className={localClassName.join(" ")}
         onClick={this.fileClickHandler}
         onKeyDown={this.keyboardEventHandler}
-        tabIndex={0}
+        tabIndex={noFocus ? null : 0}
```
bartaz: I find bool props that are negations quite confusing and harder to comprehend.
Could it be 'isFocusable' that is `true` by default and can be turned off if needed?

templates/publisher/listing.html
```diff
@@ -511,6 +518,9 @@ <h2 class="p-heading--four">Additional information</h2>
         {% for screenshot_url in screenshot_urls %}
           { url: {{ screenshot_url|tojson }}, type: "screenshot", status: "uploaded" },
         {% endfor %}
+        {% for banner_url in banner_urls %}
+          { url: {{ banner_url|tojson }}, type: "banner", status: "uploaded" },
```
bartaz: Since we assume there is only one banner (and only read one in JS anyway), could it be just filtered as one image here, and not passed as unnecessary array that we read just first item of?

static/js/libs/fileValidation.js
Lukewh: Ah I think I might have been using it for something else. Don't think the meta is used for anything so removed it :+1: good catch.

static/js/publisher/form/banner.test.js
Lukewh: Yeah, the tests were more complex before, I've reduced them down a lot because much of the functionality is handled by the fileInput component now!

bartaz:
Couple visual issues:

1. In the banner place + and "Add background" seem quite far away

Screenshot 2019-04-26 at 07 26 34

2. When dragging label on banner is different colour then on images

Screenshot 2019-04-26 at 07 22 36


3. When dropping before image shows up, it seems like whole banner shrinks to height 0 and appears again causing rest of the page content to jump

banner-drop

4. Summary and description are now far away down the form. These are 2 most basic and easy to fill fields, possibly the first ones that developer can update on the page. It would be nice if we could make them more accessible and visible by keeping it above media.

Lukewh:
> Couple visual issues:
> 
>     1. In the banner place + and "Add background" seem quite far away
> 

:+1: damn chrome

> 
>     2. When dragging label on banner is different colour then on images
> 
> 

:+1: fixed

> 
>     3. When dropping before image shows up, it seems like whole banner shrinks to height 0 and appears again causing rest of the page content to jump
> 
> 

:+1: should be fixed when I push

> 
>     4. Summary and description are now far away down the form. These are 2 most basic and easy to fill fields, possibly the first ones that developer can update on the page. It would be nice if we could make them more accessible and visible by keeping it above media.

:+1: @hughesjordan thoughts? I was thinking maybe move them up to be under "category"?

>     5. There is very little space between the fields. This combined with help text and 'Show restrictions links' make it a bit confusing what applies to what:
> 

:+1: @hughesjordan did mention this before and I ignored him it seems

hughesjordan:
> 1. Summary and description are now far away down the form. These are 2 most basic and easy to fill fields, possibly the first ones that developer can update on the page. It would be nice if we could make them more accessible and visible by keeping it above media.

@bartaz @Lukewh When re-organising the layout of this page, ordering like this was deliberate to keep the elements the same here as they would appear on the details page.

> 2. There is very little space between the fields. This combined with help text and 'Show restrictions links' make it a bit confusing what applies to what:

Here is the design on Zeplin to provide better spacing: https://zpl.io/2ZrDpJq

bartaz:
Not sure if it's related to this PR, but became visible now: drop targets for images and banner have different widths:

Screenshot 2019-04-26 at 13 12 28

Issue #1674, opened 2019-03-11
Add screenshots requirements overlay


Lukewh:
Related issues:
https://github.com/canonical-websites/snapcraft.io/issues/413

Issue #413, opened 2018-03-19
Screenshots - No help text for screenshot requirements


Greggless:
No requirements explained:
- on initial view
- on initial upload
- on applying changes

It would be more helpful if the screenshot requirements were explained up front, or the files were validated on initial upload to 'staging' and more helpful error text given _E.G. The image you selected must be in PNG or JPEG format_
# 
Triaged from #365

Greggless:
- Closely related to: https://github.com/CanonicalLtd/snapcraft-design/issues/180
- Relevant comment: [Hiya!](https://github.com/CanonicalLtd/snapcraft-design/issues/180#issuecomment-351763568)
- How this is executed will depend on how @matthewpaulthomas solves the [related issue^^](https://github.com/CanonicalLtd/snapcraft-design/issues/180)
- Partially solves: #412 

## On initial view
### What it looks like now:

![image](https://user-images.githubusercontent.com/23459065/37958167-a81f1240-31a7-11e8-8a58-e2672630b152.png)

### What it could look like:
![image](https://user-images.githubusercontent.com/23459065/37958976-ce80ef24-31a9-11e8-9e33-5386721bebda.png)

## On initial upload
(No 'initial' upload for first implementation of screenshots)
- But if we can, can we omit .svg and other incompatible file types from being selected in the file picker? @bartaz 

## On applying changes
![image](https://user-images.githubusercontent.com/23459065/37959308-c04e247a-31aa-11e8-8387-524514188f5e.png)

matthewpaulthomas:
> But if we can, can we omit .svg and other incompatible file types from being selected in the file picker? 

We can `<input type="file" accept="image/jpeg,image/png"`…`>`. However:
- I haven’t found a straight answer to whether/which browsers currently use this to prevent selection of unwanted types.
- Even if all browsers do, you always need server validation as well, because clients can lie.

If most/all browsers currently do prevent unwanted types, though, that could mean we don’t need extra text “Format: .PNGs and .JPEG” ahead of time. We can have just error feedback for that requirement.

> ## On applying changes
>…
> ### What it could look like:

[...]

PR #1701, opened 2019-03-14
Image restrictions

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1674
Fixes https://github.com/canonical-websites/snapcraft.io/issues/413

## Done
- Only trigger a form state update when the state changes in a meaningful way
- ~Added a new simplified "p-overlay" pattern (to be rolled out to other pages (snap details))~
- Added the restrictions text

## QA
- Pull the branch
- `./run`
- See the media restrictions overlay when you click the button
- Ensure media uploading/ removing and saving still works


bartaz:
@hughesjordan @Lukewh 

I feel like we are a bit inconsistent with how we show additional help on this page.
Supported markdown syntax is displayed inline under the input (just expands below).
Image restrictions show a small modal even when it doesn't really seem to contain more text then markdown help.

Screenshot 2019-03-18 at 11 01 39

Ideally I think both should be the same: either inline expanding, or in modal.

static/sass/_snapcraft_patterns_overlay.scss
```diff
@@ -0,0 +1,25 @@
+@mixin snapcraft-p-overlay {
+  .p-overlay {
```
bartaz: There is already a Vanilla modal pattern (that we use on snap details page): https://docs.vanillaframework.io/en/patterns/modal

Is there any reason to create this as separate pattern?

hughesjordan:
@bartaz @Lukewh good spot, I had originally included a modal incase we wanted to show best practice examples for assets (the 'guidelines' bit), which may clutter the UI.

But considering thats not a thing atm and that markdown help is especially useful to see inline when editing your description, I'd say we should go that route. Can the markdown help be repurposed for the others?

hughesjordan:
@bartaz @Lukewh Just to give an idea of what I was thinking:

Screenshot 2019-03-18 at 11 44 23 am

therealjuan:
Can we get together with Advocacy to set the guidelines and create the relevant new issue and leave as is for now?

static/sass/_snapcraft_patterns_overlay.scss
Lukewh: "Added a new simplified "p-overlay" pattern (to be rolled out to other pages (snap details))" from description. But this is removed now so not required.

static/js/publisher/form/media.js
```diff
@@ -13,15 +13,14 @@ class Media extends React.Component {
...
+
     this.state = {
-      mediaData: props.mediaData
+      mediaData: props.mediaData,
+      restrictionsVisible: false
     };
   }
 
-  componentDidUpdate(prevProps, prevState) {
-    this.props.updateState(prevState.mediaData);
```
bartaz: Why can't it be called on each component update?
I understand that state contains more than just `mediaData`, so there will be some unnecessary calls (when opening/closing image restrictions), but is this such a heavy call that we need to avoid it?
Also, if mediaData didn't change, maybe it should be up to `updateState` function to decide whether to do something or not?

Lukewh: As it's passed through to the form to update the state and update the form, which includes diffing and updating localStorage etc. I figured we're better off only updating and causing that chain reaction once, when we know we need to. And so we don't need to add some extra logic to do what the `updateState` callback does for free.

bartaz: OK, still I feel like it should be `updateState` to check if anything changed and only update if necessary.
But if we want to keep that in the component maybe it would make sense to keep it in one place (`componentDidUpdate`) and do something like:
```
if (prevState.mediaData !== this.state.mediaData) {
    this.props.updateState(this.state.mediaData);
}
```

Of course that assumes that mediaData is easily comparable, so needs to be updated in immutable way.

Lukewh: I'll do a little test, my worry is that we're passing `File`'s through the state. I'm not sure how comparing them would work

bartaz: @Lukewh It's just comparing 2 arrays, so items aren't even looked at. Only if array (reference) changed they will be considered different. If we modify array by pushing or changing items in place this comparison would not work.

Way around it may be serialising with `JSON.stringify` and comparing that, but that's some overhead and not fully bulletproof, so if it get's that complicated let's leave it as it is now.

static/js/publisher/form/media.js
```diff
@@ -36,7 +41,7 @@ class Media extends React.Component {
   }
 
   mediaChange(input) {
-    const newMediaData = this.state.mediaData;
+    const newMediaData = [...this.state.mediaData];
```
bartaz: 👍

PR #1712, opened 2019-03-19
File validation

Lukewh:
## Done
Checks a file selected by a file input against a preset list of requirements.

## QA
- Pull the branch
- `./run`
- Try and upload different files that do and do not conform to the [spec](https://docs.google.com/document/d/1kR1MIwU1__q-0aPckhnOWOvSWDZMOt9I-KFwQBeAVCI/edit#heading=h.17b7d6xrozfw)

## Follow ups
- Convert icon to react component
- implement drag to upload

### DIFF
--- a/static/js/libs/fileValidation.js
+++ b/static/js/libs/fileValidation.js
@@ -3,16 +3,20 @@ function baseRestrictions(file, restrictions) {
     const errors = [];
     if (restrictions.accept) {
       if (restrictions.accept.indexOf(file.type) === -1) {
-        errors.push("File type is incorrect");
+        errors.push("file type is incorrect");
       }
     }
 
     if (restrictions.size) {
       if (restrictions.size.max && file.size > restrictions.size.max) {
-        errors.push(`File size is over ${restrictions.size.max}MB`);
+        errors.push(
+          `file size is over ${(restrictions.size.max / 1000000).toFixed(2)}MB`
+        );
       }
       if (restrictions.size.min && file.size < restrictions.size.min) {
-        errors.push(`File size is below ${restrictions.size.min}MB`);
+        errors.push(
+          `file size is below ${(restrictions.size.min / 1000000).toFixed(2)}MB`
+        );
       }
     }
 
@@ -31,7 +35,6 @@ function imageRestrictions(file, restrictions) {
       return;
     }
 
-    const errors = [];
     const url = URL.createObjectURL(file);
     const image = new Image();
 
@@ -39,52 +42,60 @@ function imageRestrictions(file, restrictions) {
       const width = image.naturalWidth;
       const height = image.naturalHeight;
       const aspectRatio = width / height;
+      let hasDimensionError = false;
       if (restrictions.width) {
-        if (restrictions.width.max && width > restrictions.width.max) {
-          errors.push(`Image width above ${restrictions.width.max} pixels`);
-        }
-        if (restrictions.width.min && width < restrictions.width.min) {
-          errors.push(`Image width below ${restrictions.width.min} pixels`);
+        if (
+          (restrictions.width.max && width > restrictions.width.max) ||
+          (restrictions.width.min && width < restrictions.width.min)
+        ) {
+          hasDimensionError = true;
         }
       }
       if (restrictions.height) {
-        if (restrictions.height.max && height > restrictions.height.max) {
-          errors.push(`Image height above ${restrictions.height.max} pixels`);
-        }
-        if (restrictions.height.min && height < restrictions.height.min) {
-          errors.push(`Image height below ${restrictions.height.min} pixels`);
+        if (
+          (restrictions.height.max && height > restrictions.height.max) ||
+          (restrictions.height.min && height < restrictions.height.min)
+        ) {
+          hasDimensionError = true;
         }
       }
+
+      if (hasDimensionError) {
+        reject([
+          `has dimensions ${width} x ${height} pixels. It needs to be at least ${
+            restrictions.width.min
+          } x ${restrictions.height.min} and at most ${
+            restrictions.width.max
+          } x ${restrictions.height.max} pixels.`
+        ]);
+        return;
+      }
+
+      let hasAspectError = false;
       if (restrictions.aspectRatio) {
         const aspectRatioMax = restrictions.aspectRatio.max;
         const aspectRatioMin = restrictions.aspectRatio.min;
         if (
-          aspectRatioMax &&
-          aspectRatio > aspectRatioMax[0] / aspectRatioMax[1]
+          (aspectRatioMax &&
+            aspectRatio > aspectRatioMax[0] / aspectRatioMax[1]) ||
+          (aspectRatioMin &&
+            aspectRatio < aspectRatioMin[0] / aspectRatioMin[1])
         ) {
-          errors.push(
-            `Image aspect ratio is above ${aspectRatioMax[0]}:${
-              aspectRatioMax[1]
-            }`
-          );
+          hasAspectError = true;
         }
-        if (
-          aspectRatioMin &&
-          aspectRatio < aspectRatioMin[0] / aspectRatioMin[1]
-        ) {
-          errors.push(
-            `Image aspect ratio is below ${aspectRatioMin[0]}:${
-              aspectRatioMin[1]
-            }`
-          );
-        }
-      }
 
-      if (errors.length > 0) {
-        reject(errors);
-      } else {
-        resolve(file);
+        if (hasAspectError) {
+          reject([
+            `has a width (${width} pixels) that is ${aspectRatio.toFixed(
+              2
+            )}x its height (${height}). Its width needs to be between ${aspectRatioMin[0] /
+              aspectRatioMin[1]}x and ${aspectRatioMax[0] /
+              aspectRatioMax[1]}x the height.`
+          ]);
+          return;
+        }
       }
+      resolve(file);
     });
 
     image.src = url;
--- a/static/js/libs/fileValidation.js
+++ b/static/js/libs/fileValidation.js
@@ -1,0 +1,106 @@
+function baseRestrictions(file, restrictions) {
+  return new Promise((resolve, reject) => {
+    const errors = [];
+    if (restrictions.accept) {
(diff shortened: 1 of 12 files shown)


static/js/libs/fileValidation.test.js
```diff
@@ -0,0 +1,321 @@
...
+        set src(newSrc) {
+          this._src = newSrc;
+          this.element.dispatchEvent(new Event("load"));
+        }
+
+        get src() {
+          return this._src;
+        }
+      }
+      window.Image = Image;
```
bartaz: Not sure how the tests are run in jest, but if we really need to override global objects, we should revert it after test, so it doesn't influence others.

static/js/libs/fileValidation.js
```diff
@@ -0,0 +1,106 @@
...
+      reject(errors);
+    } else {
+      resolve(file);
+    }
+  });
+}
+
+function imageRestrictions(file, restrictions) {
+  if (!restrictions.accept || restrictions.accept[0].indexOf("image") < 0) {
+    return file;
```
bartaz: Why returning `file`? It's quite a different return value from `Promise`, is it expected that return value is checked every time it's used?

static/js/libs/fileValidation.test.js
Lukewh: I looked this up - because I couldn't find another way to do this... So I found (after digging) in the Jest docs this:
https://jestjs.io/docs/en/configuration#setupfiles-array
> Since every test runs in its own environment

https://jestjs.io/docs/en/configuration#globals-object
> Note that, if you specify a global reference value (like an object or array) here, and some code mutates that value in the midst of running a test, that mutation will not be persisted across test runs for other test files.

And finally
https://jestjs.io/docs/en/configuration#testenvironment-string
> Note: TestEnvironment is sandboxed. Each test suite will trigger setup/teardown in their own TestEnvironment.

From those I took it to mean that each test suite (file) has it's own environment - so as long as that functionality is required in that test file we should be A.OK :+1:

__BUT__ in larger test suites (form.test.js for example) we should definitely revert.

static/js/libs/fileValidation.js
Lukewh: I was under the impression (I can't remember where from) that you can return a normal value within a promise chain. I could be wrong, but this does seem to work... Or I haven't tested it thoroughly enough

bartaz: Maybe when Promise is expected a value can be treated as resolved promise... but `imageRestrictions().then()` would not work if it returns file I guess?

static/js/libs/fileValidation.test.js
bartaz: OK. It still could be a good practice to just to it with cleanup in before/after callbacks, because any suite can become "larger" at some point and devs adding tests may not be aware that other test case is modifying globals.

static/js/libs/fileValidation.js
Lukewh: Yeah I think that's how it works. Probably better to resolve the promise for compatibility - will update

static/js/libs/fileValidation.js
```diff
@@ -0,0 +1,106 @@
...
+  return new Promise((resolve, reject) => {
+    const errors = [];
+    if (restrictions.accept) {
+      if (restrictions.accept.indexOf(file.type) === -1) {
+        errors.push("File type is incorrect");
+      }
+    }
+
+    if (restrictions.size) {
+      const fileSize = file.size / 1000000;
```
bartaz: Wouldn't it be easier to make the restriction value in bytes to avoid floating point comparisons?

static/js/publisher/form/fileUpload.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  keyboardEventHandler(e) {
+    if (e.key === "Enter") {
+      this.addFileHandler();
+    }
+  }
+
+  render() {
+    const { classes, children } = this.props;
```
bartaz: I think that for consistency with native `className` it would be nicer to have `className` prop on component instead of custom array `classes`. So regardless of what it is just using `className` would work.

static/js/publisher/form/fileUpload.js
```diff
@@ -0,0 +1,82 @@
...
+      >
+        {children}
+      </div>
+    );
+  }
+}
+
+FileUpload.propTypes = {
+  classes: PropTypes.arrayOf(PropTypes.string),
+  children: PropTypes.shape({}),
```
bartaz: Why not `PropTypes.node`?

static/js/publisher/form/fileUpload.js
```diff
@@ -0,0 +1,82 @@
...
+  inputName: PropTypes.string,
+  parent: PropTypes.shape({}),
```
bartaz: Is this parent in DOM? Why not `PropTypes.node`?

static/js/publisher/form/media.js
```diff
@@ -215,13 +232,15 @@ class Media extends React.Component {
...
+  updateState: () => {},
+  restrictions: {}
 };
 
 Media.propTypes = {
   mediaLimit: PropTypes.number,
   mediaData: PropTypes.array,
-  updateState: PropTypes.func
+  updateState: PropTypes.func,
+  restrictions: PropTypes.shape({})
```
bartaz: If there is no expected shape why not just `.object`?

static/js/libs/fileValidation.js
Lukewh: yeah, it's just not very friendly - but I guess they won't change that often

static/js/publisher/form/fileUpload.js
```diff
@@ -0,0 +1,82 @@
...
+      </div>
+    );
+  }
+}
+
+FileUpload.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.node,
+  inputName: PropTypes.string,
+  parent: PropTypes.node,
```
bartaz: Having a second look at that it seems that we are passing a component here, not a node (but I'm not sure how they correlate to each other exactly in React).
Looking at the code it seems that we are interested in `parent.holder` so maybe a shape expecting `holder` node would be better?

bartaz: BTW, that's quite a strange pattern for React anyway, to pass a component and do something with it.
Is it necessary? Can't this component just render inputs in itself? Is there a case when input really has to be put elsewhere in the DOM?

Lukewh: Yeah, the issue is that the `parent.holder` node isn't defined when `FileUpload` get's rendered (as it's a child of the parent). While typing that I did think maybe if `this.holder` was defined in the constructor of the pare, when the element has been rendered the reference will be updated.... Leave it with me, I want to try something :)

tests/publisher/endpoint_testing.py
```diff
@@ -12,7 +12,7 @@
 
 class BaseTestCases:
     """
-    This class has a set of test classes that should be inherited by endpoint
+    This class has a set of test className that should be inherited by endpoint
```
bartaz: LOL. Seems like one rename too far 😄

Lukewh: Ooops!

static/js/publisher/form/fileUpload.js
```diff
@@ -0,0 +1,102 @@
+import React from "react";
+
+import PropTypes from "prop-types";
+
+import { validateRestrictions } from "../../libs/fileValidation";
+
+class FileUpload extends React.Component {
```
bartaz: `FileUpload` name is a bit misleading, it doesn't seem to have anything to do with 'upload'. Maybe something like `FileValidator` or `FileValidationInput` would be more accurate?

Lukewh: maybe just "File"? As it's essentially just a wrapper around a file input?

bartaz: `File` sound too much like native `File` class, so maybe `FileInput`?

bartaz:
I guess the error message could be worded a bit better instead of just showing file name:
Screenshot 2019-03-25 at 10 34 35

Maybe something like: "Cannot use [filename]:" + list of errors?

matthewpaulthomas:
> Maybe something like: "Cannot use [filename]:" + list of errors?

The linked spec has a single message for any dimension error: “{filename} has dimensions {width}×{height}. It needs to be at least {min_width}×{min_height} and at most {max_width}×{max_height}.” I think that would be more helpful than the four separate current messages, since reminding you of the whole range would discourage overcompensation errors. As a bonus, it would make the code shorter, since the four error cases would have the same code path (`if (`too narrow` || `too wide` || `too short` || `too tall` ) { errors.push(`…).

The spec’s error message for aspect ratios seems to be a bit out of date, assuming a set of precise ratios. Instead, I’d suggest the same approach as for dimensions: “{filename}’s width ({width}) is {aspect} × its height ({height}). To work in the directory, it must be between {minimum}× and {maximum}×.” For example: “foo.jpg’s width (3800) is 7.6× its height (500). To work in the directory, it must be between 0.5× and 2×.”)

hughesjordan:
@Lukewh looking good, maybe could add units to each of the messages? Also, there seems to be an extra space after the file name.

Screenshot 2019-03-26 at 12 16 43 pm

For the aspect ratio one perhaps we could make it read: "**It's width** needs to be between xx and xx **the height**"
Screenshot 2019-03-26 at 12 18 11 pm

static/js/libs/fileValidation.js
```diff
@@ -0,0 +1,117 @@
...
+            aspectRatio < aspectRatioMin[0] / aspectRatioMin[1])
+        ) {
+          hasAspectError = true;
+        }
+
+        if (hasAspectError) {
+          reject([
+            `has a width (${width}) that is ${aspectRatio.toFixed(
+              2
+            )}x it's height (${height}). It needs to be between ${aspectRatioMin[0] /
```
matthewpaulthomas:
```suggestion
            )}x its height (${height}). It needs to be between ${aspectRatioMin[0] /
```

Lukewh: thanks @matthewpaulthomas :+1:

PR #1724, opened 2019-03-20
Make whole content of a embedded card clickable as a link to the store

bartaz:
Fixes https://github.com/CanonicalLtd/snap-squad/issues/941

Makes whole embedded card clickable, so it can link to snap details page even if there is no 'Store Button' in it.

### QA

- ./run or https://snapcraft-io-canonical-websites-pr-1724.run.demo.haus/
- Sign in, go to Publicise page of any snap, choose Embeddable cards
- Whole card preview should be clickable and link to snap details page
- Turn off store button option, card should be still clickable


hughesjordan:
All good, nice one @bartaz

PR #1734, opened 2019-03-22
Move the FSF up on the page

evandandrea:
Not convinced by how it looks at the top, but this PR will at least give us something to debate around.

Screenshot 2019-03-25 at 15 00 41

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -43,6 +43,20 @@
   </div>
 </div>
 
+<div class="p-strip">
+    <div class="row">
+      <div class="col-12">
+        <h3 class="p-heading--two">Learn how to snap an app <br class="u-hide--large" /> in 30 minutes</h3>
+        <p>What language or framework does your app use?</p>
+      </div>
+    </div>
+    {% include "partials/fsf_language.html" %}
+  </div>
+
+  <div class="row">
+      <hr />
+    </div>
+
 <section class="p-strip">
   <div class="row">
     <h3 class="p-muted-heading">Official snaps from major publishers</h3>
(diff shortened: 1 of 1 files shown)


templates/index.html
```diff
@@ -43,6 +43,20 @@ <h2>The app store for Linux</h2>
   </div>
 </div>
 
+<div class="p-strip">
+    <div class="row">
```
bartaz: Seems like indentation got a bit shifted after copy/pasting.

therealjuan:
I wonder if we could just leave the publishers still at the very top and replace the FSF module with something more along the lines of:

image

matthewpaulthomas:
That’s roughly what I had in mind for when the number of languages/frameworks became too wide for a single row of icons. (I was surprised when the icons were spread out so much that only 12 wrapped to two rows.)

I think a greater distinction between the tabs and their contents would help people understand where they start and stop and how they interact. (This would be even more important if the section is moved away from the bottom of the page.) A couple of extra hairlines would be enough:

front-page-languages-frameworks-vertical

evandandrea:
>I wonder if we could just leave the publishers still at the very top and replace the FSF module with something more along the lines of…

Where would the FSF module sit on the page?

evandandrea:
I'm not convinced by having the example yaml on a hidden-by-default tab. There are differences in benefits to users of different languages –bundling dependencies in C/C++ is a lifesaver, but solved by the language in Java and Go– however, there is enough commonality between the [listed benefits](https://snapcraft.io/first-snap) that I worry each language tab will feel too similar.

Does someone who's not written as many snaps have an opinion? What would motivate you to continue?

therealjuan:
I was referring to something like this @evandandrea:

FSF Updated

We should implement this and improve the layout afterwards.

templates/index.html
evandandrea: Sorry about that. Fixed as 81c16b1

evandandrea:
>We should implement this and improve the layout afterwards.

Sure, 81c16b1 puts it beneath the major publishers.

PR #1735, opened 2019-03-25
Updated layout for listing page

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1705

## QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/<snap>/listing
or
- Use the demo
- See the new layout - ensure fields and combinations still work as expected
- :+1:

### DIFF
--- a/templates/publisher/listing.html
+++ b/templates/publisher/listing.html
@@ -81,6 +81,12 @@
             </div>
         {% endif %}
 
+        <div class="row">
+          <h2 class="p-heading--five">
+            Listing details
+          </h2>
+        </div>
+
         {% if field_errors and field_errors['icon'] %}
           <div class="row">
             <div class="col-12">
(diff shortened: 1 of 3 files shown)


templates/publisher/listing.html
```diff
@@ -162,18 +144,18 @@
             <select
               name="primary_category"
               {% if snap_categories.categories[0] %}
-                value="{{ snap_categories.categories[0].name }}"
+              value="{{ snap_categories.categories[0].name }}"
```
bartaz: From code readability point of view I think indent in the template code (between `{% if %}`, `{% for %}`, etc is more important than indent in rendered HTML.

templates/publisher/listing.html
```diff
@@ -439,66 +449,35 @@
                 name="public_metrics_territories"
                 id="public-metrics-territories"
                 {% if not contains(public_metrics_blacklist, 'installed_base_by_country_percent') %}
-                checked="checked"
-                {% endif %}
-                {% if not public_metrics_enabled %}
-                disabled="disabled"
-                {% endif %}
+              checked="checked"
```
bartaz: Something strange happened to indent there (and couple similar sections below)

bartaz:
Headings with `p-heading--five` seem quite small.
I feel like `p-heading--four` would work a bit better:

Screenshot 2019-03-25 at 11 43 33

templates/publisher/listing.html
Lukewh: Ah thanks, I think my editor did some tomfoolery when moving things around

hughesjordan:
@Lukewh Can we make it so that the remove icons are offset by the same distance from the inputs?
Screenshot 2019-03-26 at 10 00 46 am

Also, when on mobile there are a few issues. The remove 'snap icon' icon is now to the left and the remove 'category' icon underneath as it's in another column. Maybe something to save and tackle all mobile issues at once.
Screenshot 2019-03-26 at 10 31 27 am

Last thing would be spacing, if you click on the image below you can see a comparison between how it is currently and a suggestion.

The suggestion (image on the right) has a little more spacing between inputs to give them a little more breathing room and then more generous spacing between the `icon/title/category`, `video/images` and `summary/description` areas. This is to group the relevant areas here as they would be presented on the snap listing page.
Comparison

PR #1755, opened 2019-03-27
Removes “by” from publisher name

matthewpaulthomas:
A big advantage of snaps is that the author and publisher are usually the same person. In the few cases where they aren’t, it’s important that we don’t imply that the publisher is the author: it can upset the real author, and sour them on the platform.

Dropping the “by” from the store page:
- avoids this implication
- marginally simplifies the page
- is consistent with how the publisher is presented on the [App Store Preview](https://itunes.apple.com/us/app/skype-for-iphone/id304878510), [Microsoft Store](https://www.microsoft.com/en-us/p/skype/9wzdncrfj364?activetab=pivot:overviewtab), and [Google Play](https://play.google.com/store/apps/details?id=com.skype.raider&hl=en_US).

### DIFF
--- a/templates/store/_snap-header-information.html
+++ b/templates/store/_snap-header-information.html
@@ -1,4 +1,4 @@
-by {{ display_name(publisher, username) }}
+{{ display_name(publisher, username) }}
 {% if developer_validation and developer_validation == VERIFIED_PUBLISHER %}
 <span class="p-verified p-tooltip p-tooltip--top-center" aria-describedby="{{ package_name }}-tooltip">
   <img src="https://assets.ubuntu.com/v1/75654c90-rosette.svg" />


matthewpaulthomas:
&lt;hughsie&gt; popey, uncool: https://snapcraft.io/snap-store -- "by Canonical" apparently....
&lt;mclasen&gt; the bigger problem here is clearly the screenshot
…
&lt;hughsie&gt; the bigger problem for me is that it says "by Canonical"
&lt;kenvandine&gt; hughsie: that's the publisher of snap in the store
&lt;hughsie&gt; that's not what it says dude
&lt;hughsie&gt; Snap Store
&lt;hughsie&gt; by Canonical
…
&lt;kenvandine&gt; hughsie: yeah, i know what it sounds like :)
&lt;kenvandine&gt; that "by" is the publisher of the snap
&lt;mclasen&gt; still looks like gnome software to me :/
&lt;kenvandine&gt; i don't think GNOME wants to be publishing the snap only fork :)
&lt;kenvandine&gt; mclasen: it mostly is
&lt;kenvandine&gt; we change less than endless does :)
&lt;kenvandine&gt; we really aren't trying to claim it's something completely new by canonical
`*` mclasen likes the endless changes
&lt;kenvandine&gt; that "by" just means who publishes the snap in the store
…
&lt;kenvandine&gt; we specifically didn't call it gnome-software because we don't want user to see it and think it handles everything gnome-software does
&lt;kenvandine&gt; we didn't want any confusion
&lt;hughsie&gt; well, in reducing confusion you've managed to anger the upstream maintainer...

Issue #1703, opened 2019-03-18
Create React component for dragging and dropping


Lukewh:
A generic component that can wrap other components

PR #1757, opened 2019-03-28
Drag to upload files

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1703
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1672
Fixes https://github.com/canonicalltd/snap-squad/issues/915

## QA

- Pull the branch
- `./run`
- Drag a file onto the browser window
- Drop it somewhere other than the dropzone
- Drag another file onto the browser window
- Drop it on the dropzone, should upload
- Try again with different sized images, different number of images etc.

### DIFF
--- a/static/js/publisher/form/fileInput.js
+++ b/static/js/publisher/form/fileInput.js
@@ -11,13 +11,25 @@ class FileInput extends React.Component {
     this.fileClickHandler = this.fileClickHandler.bind(this);
     this.fileChangedHandler = this.fileChangedHandler.bind(this);
     this.keyboardEventHandler = this.keyboardEventHandler.bind(this);
+    this.dragOverHandler = this.dragOverHandler.bind(this);
+    this.dropHandler = this.dropHandler.bind(this);
+    this.dragExitHandler = this.dragExitHandler.bind(this);
+    this.dragLeaveHandler = this.dragLeaveHandler.bind(this);
+
+    this.cancelDragging = this.cancelDragging.bind(this);
+
+    this.state = {
+      isDragging: false,
+      canDrop: false,
+      overLimit: false
+    };
   }
 
   componentDidMount() {
     const { restrictions, inputName } = this.props;
 
-    // Handle input creation outside of the react lifecycle so as to avoid
-    // rerendering and side-effects. We need the input to maintain a consistent
+    // Handle input creation outside of the react lifecycle to avoid
+    // re-rendering and side-effects. We need the input to maintain a consistent
     // DOM state throughout the component lifecycle.
     this.input = document.createElement("input");
     this.input.type = "file";
@@ -29,6 +41,29 @@ class FileInput extends React.Component {
     this.input.addEventListener("change", this.fileChangedHandler);
 
     this.holder.appendChild(this.input);
+
+    document.addEventListener("dragover", this.dragOverHandler);
+    document.addEventListener("drop", this.dropHandler);
+
+    if ("ondragexit" in document) {
+      // Firefox fires dragexit and we can use the event.relatedTarget to see what the
+      // new element taking focus is. When leaving the window, this is set to null
+      document.addEventListener("dragexit", this.dragExitHandler);
+    } else {
+      // Chrome does not implement dragexit, but it does fire an event as soon as the
+      // mouse leaves the window, setting the mouse position to 0, 0
+      document.addEventListener("dragleave", this.dragLeaveHandler);
+    }
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener("dragover", this.dragOverHandler);
+    document.removeEventListener("drop", this.dropHandler);
+    if ("ondragexit" in document) {
+      document.removeEventListener("dragexit", this.dragExitHandler);
+    } else {
+      document.removeEventListener("dragleave", this.dragLeaveHandler);
+    }
   }
 
   fileClickHandler() {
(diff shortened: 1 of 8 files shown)


bartaz:
When I hover with a file over drop target is sometimes start to blink (appears and dissapears).
Drop target sometimes is rendered above all 5 screenshots, sometimes only on last "Add image" box.

Also, I drag and drop one image, but it's added twice.

drop

static/js/publisher/form/fileInput.js
```diff
@@ -11,6 +11,35 @@ class FileInput extends React.Component {
...
+    this.dropHandler = this.dropHandler.bind(this);
+
+    this.state = {
+      isDragging: false,
+      canDrop: false
+    };
+  }
+
+  componentDidMount() {
+    document.addEventListener("drag", this.preventDefaultDragHandler);
```
bartaz: Does it need to be in `FileInput` component?
We have it rendered multiple times in a page (5 instances for screenshots). Which means these events are added 5 times to document.

It seems that all we need it one drop target for screenshots, so maybe having these events set by `Media` component would be better? Or does it need access to internals of `FileInput`?

Lukewh: The issue is that to enable the "drop here" when you drag onto the window we need to add a listener to the document.

We could have those set by the parent component, but it means `FileInput` isn't self contained. We could remove all previous document event listeners on mount, so we only have 1 set of document eventlisteners (for the global preventDefault stuff) but won't help with the state setting of each component.

hughesjordan:
If I choose not to release a file on the drop zone and return it to the desktop, the assets UI remains in drop to upload mode. Only way to change the view back is to drop a file in the browser.

Mar-29-2019 9-52-58 am

bartaz:
And dropped images are still doubled in Chrome.

static/js/publisher/form/fileInput.js
```diff
@@ -26,9 +37,32 @@ class FileInput extends React.Component {
       this.input.accept = restrictions.accept;
     }
     this.input.hidden = true;
+    this.input.multiple = true;
```
bartaz: This makes it possible to choose multiple files when clicking "Add" as well.

While being a nice UX improvement it has some drawbacks:
- it makes it quite easy to choose many images at once and therefor reach the limit of 5 causing warnings
- if single input contains multiple files they cannot be removed one by one (we cannot remove one file from input and keep others) - this can lead to some bugs or unexpected behaviour when input containing multiple files changes value
- and because it's not possible to remove files from input partially if many files are chosen and added but only few kept (rest removed) all of them will be uploaded anyway (from single input)

static/js/publisher/form/media.js
```diff
@@ -22,8 +22,13 @@ class Media extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    if (prevState.mediaData !== this.state.mediaData) {
-      this.props.updateState(this.state.mediaData);
+    const { mediaData } = this.state;
+    const { mediaLimit } = this.props;
+    const mediaLength = mediaData.filter(item => item.status !== "delete")
```
bartaz: No need to filter anymore (once you rebase).

static/js/publisher/form/fileInput.js
Lukewh: > This makes it possible to choose multiple files when clicking "Add" as well.
> 
> While being a nice UX improvement it has some drawbacks:
> 
>     * it makes it quite easy to choose many images at once and therefor reach the limit of 5 causing warnings
> 
The way I implemented it you do get the "you have over 5 images" warning and the save button stays disabled, until you're at 5 images. I think this is ok, but would be good to have some UX feedback on it.

>     * if single input contains multiple files they cannot be removed one by one (we cannot remove one file from input and keep others) - this can lead to some bugs or unexpected behaviour when input containing multiple files changes value
> 
- Because the view is rendered from the state, you can remove 1-by-1 visually and....

>     * and because it's not possible to remove files from input partially if many files are chosen and added but only few kept (rest removed) all of them will be uploaded anyway (from single input)
> 
- ...only the images that get sent in the images array get uploaded to the store by the backend, so even if there are multiple images only those that are desired will be uploaded to the store (though yes, they'll be uploaded to our backend, then disregarded).

static/js/publisher/form/media.js
Lukewh: :ok_hand:

static/js/publisher/form/fileInput.js
bartaz: It does cause issues.

Not sure about specific test case, but if you drag multiple files in, remove some (to make space for more), drag some more, remove some, drag, etc. Not all of screenshots are saved.

bartaz: I'm afraid we may need to have to limit screenshots interactions to one file until we switch from using file inputs and keep file objects via File API in state (and POST form via ajax instead of HTML).

hughesjordan:
Not sure if you have fixed this, but in some cases the save button is still enabled when you have over 5 images:

Screenshot 2019-04-02 at 4 14 00 pm

With regards to the multi-upload, I think it's a nice to have that would be great if we could get it working nicely. Maybe we don't need to add additional images and then make the user remove. Could we just fill un-used slots instead? If there are no slots available, then there is no means of uploading?

static/js/publisher/form/fileInput.js
```diff
@@ -55,12 +88,85 @@ class FileInput extends React.Component {
...
+
+    this.setState({
+      isDragging: true,
+      canDrop: false
+    });
+
+    if (e.target === this.holder) {
+      this.setState({
+        canDrop: true,
+        cantDrop: false
```
bartaz: Is this needed? Or do you want just to be double sure that user can drop? ;)

Lukewh: Ah, bad rebase conflict artifact, thanks

bartaz:
Could we be a bit less "errory" about multiple files drop? It's not an error and users fault that they want to drop multiple files. I'd rather say "Sorry, you can drop one image at a time".
Screenshot 2019-04-03 at 11 10 16

And could we detect it earlier? Before user drops? To show drop space as 'disabled' with different text then "Aaaaand... drop"?

Lukewh:
> Could we be a bit less "errory" about multiple files drop? It's not an error and users fault that they want to drop multiple files. I'd rather say "Sorry, you can drop one image at a time".
> Screenshot 2019-04-03 at 11 10 16
Makes sense :+1:
> And could we detect it earlier? Before user drops? To show drop space as 'disabled' with different text then "Aaaaand... drop"?
Unfortunately not, you only get the list of files on the "drop" even

bartaz:
Drop space is enabled when anything is dragged to a page, not only files.
It can be link, image from other page, etc. Anything that can be dragged with a mouse.

It works fine (nothing gets dropped), but is a bit confusing as drop area looks like it will accept the input.
It seems it can be detected, because gmail only shows drop area when dragging files, not anything else.

Issue #1767, opened 2019-04-02
Screenshots not saved if one is deleted


bartaz:
### Steps to reproduce the problem

There is some weird behaviour when uploading screenshots.

1. Go to listing page (where you don't have many screenshots)
2. Add 2 new screenshots (without saving)
3. Delete first of new screenshots (before saving) - second new screenshots is still on the list
4. Save (in attempt to save second added screenshot)
5. Page reloads but no new screenshots appear

If you add more screenshots (3-4) and delete one, some are saved, but no all...

### Expected behaviour

All screenshots that are new and not deleted should be saved.

PR #1769, opened 2019-04-02
Manage fileInput inputs outside of react lifecycle. Don't clear on ch…

Lukewh:
…ange

Fixes https://github.com/canonical-websites/snapcraft.io/issues/1767

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004
- Follow this steps in https://github.com/canonical-websites/snapcraft.io/issues/1767

### DIFF
--- a/static/js/publisher/form/fileInput.js
+++ b/static/js/publisher/form/fileInput.js
@@ -16,6 +16,9 @@ class FileInput extends React.Component {
   componentDidMount() {
     const { restrictions, inputName } = this.props;
 
+    // Handle input creation outside of the react lifecycle so as to avoid
+    // rerendering and side-effects. We need the input to maintain a consistent
+    // DOM state throughout the component lifecycle.
     this.input = document.createElement("input");
     this.input.type = "file";
     this.input.name = inputName;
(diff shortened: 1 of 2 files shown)


static/js/publisher/form/fileInput.js
```diff
@@ -13,11 +13,19 @@ class FileInput extends React.Component {
...
   }
 
-  componentDidUpdate() {
-    const { active, clear } = this.props;
-    if (active && clear && this.input) {
-      this.input.value = "";
+  componentDidMount() {
+    const { restrictions, inputName } = this.props;
+
+    this.input = document.createElement("input");
```
bartaz: Is this needed or would removing `clear` be enough of a fix?

Sure, it kind of prevents React from accidentally rendering new input fields on render, but I guess that React is smart enough to not rerender DOM when not needed (in this case if restrictions and inputName don't change)?

bartaz: Maybe it is needed... I did a quick test with just old `componentDidUpdate` removed to prevent from clearing the input and it worked better, but I still had cases of missing uploads when adding/removing many screenshots at once in crazy order.

But please add  a comment explaining why we add input manually here instead of properly rendering it with React.

PR #1770, opened 2019-04-02
Remove deleted screenshots from state

bartaz:
Removes deleted screenshot from form state (instead of just changing their status).
This cleans up different parts of code and will help with reordering screenshots.

### QA

- ./run or demo
- go to listing page of any snap
- add screenshots, delete screenshots, just test screenshots from different angles
- everything should work as before

### DIFF
--- a/static/js/publisher/form.js
+++ b/static/js/publisher/form.js
@@ -146,13 +146,6 @@ function initForm(config, initialState, errors) {
 
   let state = JSON.parse(JSON.stringify(initialState));
 
-  const stateInput = document.createElement("input");
-  stateInput.type = "hidden";
-  stateInput.name = "state";
-  stateInput.value = "";
-
-  formEl.appendChild(stateInput);
-
   const diffInput = document.createElement("input");
   diffInput.type = "hidden";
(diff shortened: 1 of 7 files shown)


static/js/publisher/form.js
```diff
@@ -146,13 +146,6 @@ function initForm(config, initialState, errors) {
 
   let state = JSON.parse(JSON.stringify(initialState));
 
-  const stateInput = document.createElement("input");
```
Lukewh: so we no longer need the state as an input, for any values?

webapp/publisher/snaps/views.py
```diff
@@ -380,13 +380,19 @@ def post_listing_snap(snap_name):
             except ApiError as api_error:
                 return _handle_errors(api_error)
 
+            print(changes["images"])
```
Lukewh: pls

static/js/publisher/form.js
bartaz: We don't need state input for saving. Preview uses it's own state input, so this one can be removed.

PR #1772, opened 2019-04-03
Drag and drop screenshots to reorder

### LINKED ISSUE #1673: Enable dragging and dropping to sort screenshots

bartaz:
Fixes #1673 

Enables drag and drop to reorder screenshots on publisher listing page.

### QA

- ./run or https://snapcraft-io-canonical-websites-pr-1772.run.demo.haus/
- go to listing page of any snap with screenshots
- drag a screenshot to reorder, if you drop it in different place change should be detected (Revert and Save should be enabled)
- if you drag back to same place (old order) Revert would be disabled again
- make sure adding screenshots works (and new screenshots can be reordered)
- make sure removing screenshots work (and dnd works after removing)
- make screen smaller (so screenshots don't fit into one line), make sure reordering works

dnd

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "react": "^16.4.1",
     "react-dom": "^16.4.1",
     "react-redux": "^5.1.1",
+    "react-sortable-hoc": "^1.8.3",
     "react-testing-library": "^6.0.0",
     "redux": "^4.0.1",
     "redux-mock-store": "^1.5.3",
--- a/static/js/publisher/form/media.js
+++ b/static/js/publisher/form/media.js
@@ -2,8 +2,7 @@ import React, { Fragment } from "react";
 
 import { PropTypes } from "prop-types";
 
-import MediaItem, { mediaClasses } from "./mediaItem";
-import FileInput from "./fileInput";
+import SortableMediaList from "./mediaList";
 
 class Media extends React.Component {
   constructor(props) {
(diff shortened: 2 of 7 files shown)


Lukewh:
When you have no images, upload 5, reorder and save - when the page reloads they're in the upload order. Maybe the inputs need reordering?

hughesjordan:
Hey @bartaz a few issues:

- This part doesn't seem to work for me, revert button is still enabled:
> * if you drag back to same place (old order) Revert would be disabled again


- When I add an image, re-order that image and click on save, it then re-appears in last slot.

- After saving any changes, when I re-order any image the 'Changes applied' notification does not disappear.

bartaz:
@Lukewh it seems that our backend doesn't take order to account when it comes to new screenshots. It just adds new screenshots to the end of the list (in order of upload).
While reordering inputs may be a hacky workaround, it will only help if all screenshots are new, not if new ones are mixed between old ones. So I guess it would be better to fix it in our backend logic.

bartaz:
@hughesjordan @Lukewh Updated so that order of newly added screenshots is correctly preserved.

For other issues:

> if you drag back to same place (old order) Revert would be disabled again

I see it working for myself. @hughesjordan Is there any specific case when revert button is enabled if order is same as saved?

> After saving any changes, when I re-order any image the 'Changes applied' notification does not disappear.

That is the case also for removing screenshots, etc. So I don't feel like it's an issue with this specific PR.

hughesjordan:
@bartaz does not seem to be any specific case for me:

Apr-04-2019 9-37-13 am

bartaz:
@hughesjordan This doesn't seem to be related to screenshots, so I reported it separately (#1772).

Does screenshot reordering work good apart from that?

hughesjordan:
@bartaz apart from that its good! 👍🏻

Issue #1791, opened 2019-04-05
VSCode snap has blog post about games related


Lukewh:
https://snapcraft.io/code has a related blog post "Retro Style Games on Linux" because of "Codename-LT".

`sc:snap:codename-lt` matches to `sc:snap:code`

PR #1792, opened 2019-04-05
Better blog post filtering

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1791

## QA
- Pull the branch or use the demo
- `./run` if you've pulled
- Visit http://0.0.0.0:8004/code
- See that the blog posts are related to vscode

Also check other snaps like http://0.0.0.0:8004/gitkraken to ensure their posts still come through

### DIFF
--- a/webapp/blog/views.py
+++ b/webapp/blog/views.py
@@ -195,22 +195,24 @@ def snap_posts(snap):
     articles = []
 
     if blog_tags:
-        blog_tags_ids = logic.get_tag_id_list(blog_tags)
-        try:
-            blog_articles, total_pages = api.get_articles(blog_tags_ids, 3)
-        except ApiError:
-            blog_articles = []
+        blog_tags_ids = logic.get_tag_id_list(blog_tags, snap)
 
-        for article in blog_articles:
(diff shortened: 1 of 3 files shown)


webapp/blog/views.py
```diff
@@ -195,22 +195,24 @@ def snap_posts(snap):
...
-            )
-            articles.append(
-                {
-                    "slug": transformed_article["slug"],
-                    "title": transformed_article["title"]["rendered"],
-                }
-            )
+        blog_tags_ids = logic.get_tag_id_list(blog_tags, snap)
+
+        if blog_tags_ids:
```
jpmartinspt: I don't think this will ever be empty if blog_tags is not empty because if it can't get the id it will just raise a `KeyError`.

Lukewh: `get_tag_id_list` returns an empty list if tags are found, but don't match the required name.

So the next part of the code would run (without any tag filtering) and return the latest 3 posts from the blog either way

jpmartinspt: You are absolutely right, sorry didn't read the `get_tag_id_list` function properly!

Issue #1763, opened 2019-04-02
Avoid Asciinema automatic replay


Lin-Buo-Ren:
### Current behaviour
Asciinema cast in the listing page replays automatically after playback

### Expected behaviour
Asciinema cast in the listing page stops after playing

### Steps to reproduce the problem
Check out <https://snapcraft.io/tree-strict>

### Specs

- _URL:_ https://snapcraft.io/tree-strict
- _Operating system:_ Ubuntu 18.04
- _Browser:_ 66.0.2 from Ubuntu 18.04

PR #1797, opened 2019-04-05
Don't loop asciinema on details pages

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1763#event-2252195260

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/<snap_with_asciinema>
- Watch the asciinema, see that it doesn't loop

### DIFF
--- a/templates/partials/_video.html
+++ b/templates/partials/_video.html
@@ -6,5 +6,5 @@
           webkitallowfullscreen mozallowfullscreen allowfullscreen
           src="{{ video.url }}?loop=1&title=0&byline=0&portrait=0&transparent=0"></iframe>
 {% elif video.type == "asciinema" %}
-  <script src="{{ video.url }}" id="asciicast-{{ video.id }}" async data-autoplay="1" data-loop="1" data-preload="0"></script>
+  <script src="{{ video.url }}" id="asciicast-{{ video.id }}" async data-autoplay="1" data-preload="0"></script>
 {% endif %}


bartaz:
If we disable looping for asciinema, we should disable it for other videos as well for consistency.

Lukewh:
> If we disable looping for asciinema, we should disable it for other videos as well for consistency.

It's a different beast from advocacy "as long as the videos still loop +1"

bartaz:
> "as long as the videos still loop +1"

🤷‍♂️

evandandrea:
>"as long as the videos still loop +1"

I can see the rationale for not looping asciinema (as demonstrated in #1763). You want to have the video end with some static text that you can look over. Normal videos, on the other hand, more typically demonstrate some graphical functionality and seeing it in a tight loop can be helpful.

therealjuan:
UX+1 but created a new issue as youtube and asciinema should behave similarly as it can be confusing to have 2 video implementations acting differently. It is more useful to have graphical context looping but should be up to the visitor to decide if it wants to keep looping the video.

One potential benefit of stopping playback on the video is that users might want to scroll down and dig more on the given snap they are visiting. With auto-looping, this might means that users will skip looking at the rest of the page and just leave once they have seen the page a couple of times.

New issue: https://github.com/canonical-websites/snapcraft.io/issues/1801

Issue #1765, opened 2019-04-02
'Latest release' is empty if last revisions is not yet released


bartaz:
If you have a snap that has latest revision that was uploaded, but not yet released to any channel this snap will show empty value in 'Latest release', even if it was released before.

<img width="1030" alt="Screenshot 2019-04-02 at 09 51 03" src="https://user-images.githubusercontent.com/83575/55385185-da3f1500-552c-11e9-905d-24cf14be54ec.png">

### Expected 
Latest release should show channel name of latest released revision. If latest revision is not released yet it should look at the revisions before.

### Cause

Noticed when fixing #1764 - we are currently only checking status of latest revision (`snaps[snap].latest_revisions[0].status == "Published"`) - we should probably look through whole `latest_revisions` list to check which channel was released to.

PR #1804, opened 2019-04-09
Show latest release in snap list

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1765

## QA
- Pull the branch
- `./run`
- All snaps that have a release should show a channel and version

Note: I expect the tests to fail - my local setup has borked so I couldn't get any meaningful test failure messages.

### DIFF
--- a/templates/publisher/account-snaps.html
+++ b/templates/publisher/account-snaps.html
@@ -82,9 +82,9 @@ My published snaps — Linux software in the Snap Store
           <section class="p-strip--light is-shallow p-snapcraft-first-snap__notification">
             <div class="row">
               <div class="col-9">
-                {% if snaps[snap].latest_revisions[0].channels[0] %}
+                {% if snaps[snap].latest_release %}
                   <p>
-                    You've released {{ snap }} to the "{{ snaps[snap].latest_revisions[0].channels[0] }}" channel!
+                    You've released {{ snap }} to the "{{ snaps[snap].latest_release.channels[0] }}" channel!
                   </p>
                 {% else %}
                   <p>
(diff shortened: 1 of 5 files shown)


webapp/publisher/snaps/logic.py
```diff
@@ -25,6 +25,13 @@ def get_snaps_account_info(account_info):
 
     now = datetime.datetime.utcnow()
 
+    for snap in user_snaps:
+        snap_info = user_snaps[snap]
+        for revision in snap_info["latest_revisions"]:
+            if len(revision["channels"]) > 0:
+                snap_info["latest_revision"] = revision
```
bartaz: The name may be a bit confusing.
Technically "latest_revision" is `latest_revisions[0]`, the one we are interested in is latest released revision. So I'd be for calling it exactly what it is, or at least `latest_release` for short.

templates/publisher/account-snaps.html
```diff
@@ -159,15 +163,16 @@ <h3 class="p-heading--five">Join the community</h3>
                   {% endif %}
                 </td>
                 {% if published %}
-                  <td>{{ snaps[snap].latest_revisions[0].channels[0] }}</td>
+                  <td>{{ snaps[snap].latest_revision.channels[0]|title }}</td>
```
bartaz: I'm not sure if we want to capitalise channel names. They are kind of part of snapcraft brand and are usually lower case ("edge" instead of "Edge").

Thoughts @therealjuan ?

webapp/publisher/snaps/logic.py
Lukewh: Much better name :+1:

PR #1809, opened 2019-04-11
Newsletter signup on blog

Lukewh:
Fixes https://github.com/canonical-websites/snapcraft.io/issues/1807

## QA
- Pull branch
- `./run`
- Visit http://0.0.0.0:8004/blog or https://snapcraft-io-canonical-websites-pr-1809.run.demo.haus/blog
- Sign up to the newsletter - you should see a "subscibed" message next to the button

### DIFF
--- a/static/js/public/snap-details/reportSnap.js
+++ b/static/js/public/snap-details/reportSnap.js
@@ -1,4 +1,5 @@
 import "whatwg-fetch";
+import { buttonEnabled, buttonLoading } from "../../libs/formHelpers";
 
 const showEl = el => el.classList.remove("u-hide");
 const hideEl = el => el.classList.add("u-hide");
@@ -17,7 +18,7 @@ function toggleModal(modal, show) {
 }
 
 function initForm(modal) {
-  buttonEnabled(modal.querySelector("button[type=submit]"));
+  buttonEnabled(modal.querySelector("button[type=submit]"), "Submit report");
 
   showEl(modal.querySelector(".js-report-snap-form"));
   hideEl(modal.querySelector(".js-report-snap-success"));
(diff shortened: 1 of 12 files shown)


bartaz:
Something for later maybe, but it would be good if any kind of email notification (or ideally verification email) would be sent.
Right now apart from 'subscribed' text showing up there is nothing in email confirming that it was successful. Also it feels like anyone can put anyones email there and subscribe them to blog without confirmation.

static/js/public/snap-details/reportSnap.js
```diff
@@ -36,17 +37,6 @@ function showError(modal) {
   showEl(modal.querySelector(".js-report-snap-error"));
 }
 
-function buttonLoading(button) {
```
bartaz: Cool that we can reuse that 👍

Lukewh:
> Something for later maybe, but it would be good if any kind of email notification (or ideally verification email) would be sent.
> Right now apart from 'subscribed' text showing up there is nothing in email confirming that it was successful. Also it feels like anyone can put anyones email there and subscribe them to blog without confirmation.

Yeah, I think once we have this signup in place, we should definitely make the experience tighter and nicer. I'm not sure what the limitations are with Marketo but I'll have a chat with Lucia and see what she says :+1:

webapp/blog/views.py
```diff
@@ -44,8 +45,15 @@ def homepage():
         filter = None
 
     try:
+        get_page = 1
```
bartaz: A little comment here would be nice to explain that we offset because of newsletter card

Lukewh: Ah, thought I'd added one :+1:

PR #1829, opened 2019-04-16
Add JSON endpoint to register snap name

bartaz:
Adds a JSON endpoint to register snap name (to be used in first snap flow).

### QA
- make sure tests pass

### DIFF
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -806,6 +806,33 @@ def post_register_name():
     return flask.redirect(flask.url_for("account.get_account"))
 
 
+@publisher_snaps.route("/register-snap/json", methods=["POST"])
+@login_required
+def post_register_name_json():
+    snap_name = flask.request.form.get("snap-name")
+
+    if not snap_name:
+        return flask.jsonify({"success": False})
+
+    try:
+        response = api.post_register_name(
+            session=flask.session, snap_name=snap_name
+        )
+    except ApiResponseErrorList as api_response_error_list:
+        return (
+            flask.jsonify(
+                {"success": False, "errors": api_response_error_list.errors}
+            ),
+            api_response_error_list.status_code,
+        )
+    except ApiError as api_error:
+        return _handle_errors(api_error)
+
+    response["success"] = True
+
+    return flask.jsonify(response)
+
+
 @publisher_snaps.route("/register-name-dispute")
 @login_required
 def get_register_name_dispute():
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -812,23 +812,30 @@ def post_register_name_json():
     snap_name = flask.request.form.get("snap-name")
 
     if not snap_name:
-        return flask.jsonify({"success": False})
+        return (
+            flask.jsonify({"errors": [{"message": "Snap name is required"}]}),
+            400,
+        )
 
     try:
         response = api.post_register_name(
             session=flask.session, snap_name=snap_name
         )
     except ApiResponseErrorList as api_response_error_list:
+        for error in api_response_error_list.errors:
+            # if snap name is already owned treat it as success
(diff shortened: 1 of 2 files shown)


webapp/publisher/snaps/views.py
```diff
@@ -806,6 +806,33 @@ def post_register_name():
     return flask.redirect(flask.url_for("account.get_account"))
 
 
+@publisher_snaps.route("/register-snap-json", methods=["POST"])
```
bartaz: @jpmartinspt @tbille I'm not sure about the URL. If we want a JSON version endpoint of a route that we already have (/register-name) what would be good url? `register-name-json`, `register-name/json`?

Or maybe it could be on the same route/URL and just respond with JSON based on headers?

tbille: I prefer the second name `register-name/json`

webapp/publisher/snaps/views.py
```diff
@@ -806,6 +806,33 @@ def post_register_name():
...
+        return (
+            flask.jsonify(
+                {"success": False, "errors": api_response_error_list.errors}
+            ),
+            api_response_error_list.status_code,
+        )
+    except ApiError as api_error:
+        return _handle_errors(api_error)
+
+    response["success"] = True
```
tbille: why not using status codes?

webapp/publisher/snaps/views.py
```diff
@@ -806,6 +806,33 @@ def post_register_name():
...
+        )
+    except ApiResponseErrorList as api_response_error_list:
+        return (
+            flask.jsonify(
+                {"success": False, "errors": api_response_error_list.errors}
+            ),
+            api_response_error_list.status_code,
+        )
+    except ApiError as api_error:
+        return _handle_errors(api_error)
```
jpmartinspt: I would just `return flask.jsonify({"errors": [error]}), 500`, for consistency of always returning JSON.

webapp/publisher/snaps/views.py
```diff
@@ -806,6 +806,33 @@ def post_register_name():
     return flask.redirect(flask.url_for("account.get_account"))
 
 
+@publisher_snaps.route("/register-snap/json", methods=["POST"])
+@login_required
+def post_register_name_json():
+    snap_name = flask.request.form.get("snap-name")
+
+    if not snap_name:
+        return flask.jsonify({"success": False})
```
jpmartinspt: Is the success key to help on the client side? Or could we just return an error message and a 400 status code?

webapp/publisher/snaps/views.py
bartaz: @jpmartinspt Doesn't `_handle_errors` do something more (detecting kind of error etc) that we would want to keep?

webapp/publisher/snaps/views.py
bartaz: Old habit from build.snapcraft I suppose.
But sending list of errors in JSON with a proper status code probably would be enough to tell if there was an error or success in front-end.

webapp/publisher/snaps/views.py
bartaz: @jpmartinspt I tried to remove `_handle_errors` and just return JSON, but the tests fails when I do that. I think our `BaseTestCases.EndpointLoggedIn` assumes `_handle_errors` is used and has some tests that assure it's in place (to check expired macaroons, etc).

I agree that for JSON endpoints it would be good to have JSON error handling, so maybe that's something we could improve later: having a "JSON equivalent" of `_handle_errors` and/or `BaseTestCases.EndpointLoggedIn`.

jpmartinspt: Yeah I had a similar issue with the metrics endpoint ended up not using the `BaseTestCases.EndpointLoggedIn`.
I think you are right, we might need an `_handle_errors_json`, I'm happy if you leave handle_errors for now, but maybe we could raise a ticket to look into this.

PR #1838, opened 2019-04-17
Featured snap background colours

Lukewh:
Fixes https://github.com/canonicalltd/snap-squad/issues/969

## Done

- Updated `_media-object-snap-partial` include to be the new snap layout
- Updated IOT page with new include
- Updated search results with new include
- Updated publisher pages with new include
- Use client to work out background for featured snaps
- Fallback to #eee if the colour-figurer-outer fails

## QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/iot
- Visit http://0.0.0.0:8004/publisher/jetbrains
- Ensure the snap components(?) are consistent throughout
- Ensure the store page has featured banners that have the #eee background
- ...
- **DON'T MERGE THIS PR** waiting on https://bugs.launchpad.net/software-center-agent/+bug/1826006


webapp/app.py
```diff
@@ -20,6 +20,7 @@
 from webapp.publisher.views import account
 from webapp.snapcraft.views import snapcraft_blueprint
 from webapp.store.views import store_blueprint
+from webapp.proxies.views import proxies_blueprint
```
tbille: hello you, who are you?

tbille: oups

therealjuan:
Just noticed the featured banner on publisher pages have tremendous icons:



Also the initial background colour for featured snap banner is still `#999` so might be better to have our old friend `rgba(238,238,238)`

Everything else is great!

therealjuan:
LGTM but would LEB (look even better) once CORS is there!

kwm14:
## Design review
Looking really good @Lukewh 👍 just a couple of comments below 😉 

### IoT
- [ ] Margins between rows of snaps look big or inconsistent per category if we can match the space between each row similarly to the http://0.0.0.0:8004/store page 

Screenshot 2019-04-29 at 10 34 14

### JetBrains
- [ ] Featured snap icons still looking rather large

Screenshot 2019-04-29 at 10 35 20

- [ ] Can we set a limit of 2 lines on snap descriptions as per http://0.0.0.0:8004/store styling 😉 

Screenshot 2019-04-29 at 10 35 37

### KDE
- [ ] Same as JetBRains above #2linelimit

Screenshot 2019-04-29 at 10 43 08

kwm14:
LGTM 💯

static/js/public/store-categories.js
```diff
@@ -1,8 +1,29 @@
-import { Swiper, Navigation } from "swiper/dist/js/swiper.esm";
 import "whatwg-fetch";
-import { CATEGORY_CONFIG } from "../config/swiper.config";
+import * as Fac from "fast-average-color";
```
bartaz: Why is it `* as Fac` instead of just `Fac`?

bartaz:
There is a blink of darker background colour on load.

If there is a default grey colour applied, it could be applied by CSS immediately.

featured

therealjuan:
I'll have a play with what we can do @bartaz

static/sass/_patterns_media-object--snap.scss
```diff
@@ -1,109 +1,107 @@
...
+
+    .p-media-object__image {
+      max-height: 3rem;
+      max-width: 3rem;
     }
 
-    &.is-placeholder {
-      .p-media-object--snap__img {
-        background: $color-light;
+    &.is--placeholder {
```
bartaz: Why double dashes? I think Vanilla uses one dash in their `is-something` class names?

static/sass/styles.scss
```diff
@@ -240,6 +240,9 @@ $color-navigation-background: #252525;
 @import 'snapcraft-p_grid';
 @include snapcraft-p-grid;
 
+@import 'snapcraft_p-featured-snap';
+@include snapcraft-p-featured-snaps;
```
bartaz: Nit picking, but why the inconsistency between file name and mixin (snap vs snaps)?

static/js/public/store-categories.js
```diff
@@ -1,8 +1,29 @@
...
-Swiper.use([Navigation]);
+function getColour(holder) {
+  function extractAndSet(image, parent) {
+    const fac = new Fac();
+    const colour = fac.getColor(image, { defaultColor: [238, 238, 238] });
+    parent.style.backgroundColor = colour.rgb;
+    parent.classList.add(colour.isDark ? "is--dark" : "is--light");
+  }
+
+  const srcs = holder.querySelectorAll(".p-generated-featured-snap__icon img");
```
bartaz: Why `srcs` and not `images`? I expected to see URLs as values

static/js/public/store-categories.js
```diff
@@ -15,18 +36,8 @@ function getCategory(holder) {
...
-          navigation: {
-            nextEl: `.swiper-button__next--${category}`,
-            prevEl: `.swiper-button__prev--${category}`
-          }
-        },
-        CATEGORY_CONFIG
-      )
-    );
+
+    getColour(holder);
```
bartaz: By looking at the coded I get am impression that colour calculations are only going to happen for categories on store page, but not on other pages (like publisher details, iot, etc).
Is `getColour` used elsewhere?

static/sass/_patterns_media-object--snap.scss
```diff
@@ -1,109 +1,107 @@
...
-      width: auto;
-    }
-
-    &__content {
-      display: flex;
-      flex: 1 1 auto;
-      flex-direction: column;
-      padding-right: $sph-intra;
+  %media-object {
+    .p-media-object__details {
```
bartaz: Does it have anything to do with Vanilla `p-media-object` (https://docs.vanillaframework.io/en/patterns/media-object)?
It seems to use and override same class names, not sure if that's intentional or not...

static/sass/_snapcraft_p-featured-snap.scss
```diff
@@ -0,0 +1,88 @@
+@mixin snapcraft-p-featured-snaps {
+  $light-transparent: rgba(255, 255, 255, 0);
+  $light-semi-opaque: rgba(255, 255, 255, .65);
+  $light-opaque: rgba(255, 255, 255, .8);
+  $dark-transparent: rgba(0, 0, 0, 0);
+  $dark-semi-opaque: rgba(0, 0, 0, .65);
+  $dark-opaque: rgba(0, 0, 0, .8);
+
+  .p-generated-featured-snap {
```
bartaz: The naming is a bit confusing... We have `p-media-object--is-featured`, `p-featured-snap`, `p-media-object--is-featured` `snapcraft-p-grid--featured`... And I don't really know which of these are part of same pattern, which are different, are they meant to be used together or separately?

static/js/public/store-categories.js
Lukewh: not yet. my next task is to do the same for the category pages so I'll probably extract it somewhere then

static/sass/_patterns_media-object--snap.scss
Lukewh: yeah, it expects `.p-media-object` as a base and modifies a few things if the classname has `p-media-object p-media-object--snap`

I changed this from how it was before because it repeated a lot of the `p-media-object` class and used different names within.

static/sass/_snapcraft_p-featured-snap.scss
Lukewh: :+1: I've reduced this down and used the wrapping `snapcraft-p-grid--featured` to alter sizes

static/sass/_snapcraft_p-featured-snap.scss
```diff
@@ -0,0 +1,90 @@
...
+      .p-heading--three {
+        padding-top: $sp-large;
+      }
+    }
+
+    &:hover {
+      text-decoration: none;
+    }
+
+    &.is--dark {
```
bartaz: Same double dash as in before...

static/sass/_snapcraft_p-featured-snap.scss
```diff
@@ -0,0 +1,90 @@
...
+      .p-generated-featured-snap__title {
+        color: $color-x-light;
+      }
+
+      .p-muted-heading {
+        color: $light-opaque;
+      }
+    }
+
+    &.is--light {
```
bartaz: Same double dash as in before...

static/sass/_snapcraft_p-featured-snap.scss
bartaz: Thanks. It's better now.
But I still don't like the `p-generated-featured-snap` name. Why "generated"? I would guess it's because we "generate" a colour background for it, but I don't think this implementation detail should define the name of the pattern.

I know we have 'featured' snap elsewhere, so this would need to be something different... Maybe it could have something to do with banner? p-snap-banner? or p-featured-banner? As I suppose in future we could or will show snap banner images as background if provided?

webapp/store/views.py
```diff
@@ -73,17 +73,24 @@ def store_view():
...
-                snap_searched="", category="featured", size=24, page=1
+                snap_searched="", category="featured", size=10, page=1
             )
         except ApiError:
             featured_snaps_results = []
 
         featured_snaps = logic.get_searched_snaps(featured_snaps_results)
 
+        if len(featured_snaps) > 0 and featured_snaps[0]["icon_url"] == "":
+            featured_snaps = featured_snaps[:-1]
```
bartaz: Not sure exactly what happens here... If first snap doesn't have an icon we remove it?
What if second one doesn't have icon as well?

webapp/store/views.py
```diff
@@ -270,15 +277,22 @@ def store_category(category):
...
-                snap_searched="", category=category, size=24, page=1
+                snap_searched="", category=category, size=10, page=1
             )
         except ApiError as api_error:
             status_code, error_info = _handle_errors(api_error)
 
         snaps_results = logic.get_searched_snaps(category_results)
 
+        if len(snaps_results) > 0 and snaps_results[0]["icon_url"] == "":
+            snaps_results = snaps_results[:-1]
```
bartaz: Same comment as above.

webapp/store/views.py
```diff
@@ -270,15 +277,22 @@ def store_category(category):
...
+
+        for snap in snaps_results:
+            snap["show_summary"] = True
```
bartaz: If all snaps are getting "show_summary" why do we need it at all? Can't it be treated as True by default?

webapp/store/views.py
Lukewh: The first snap will have the "featured" banner, unless it doesn't have an icon. By default we get 10 snaps (1 main, then 3 rows of 3). 
If we don't show the first snap as a banner - we don't want to have a hanging 10th snap in the grid, so this removes it.

webapp/store/views.py
Lukewh: Same reply as above :D

PR #1856, opened 2019-04-24
First snap flow - choose snap name

### LINKED ISSUE #1786: FSF improvements - choose snap name at the beginning of the flow
Screenshot 2019-04-05 at 08 40 00

bartaz:
Fixes #1786

Adds a form to choose snap name at the beginning of the package step of first snap flow

### QA

- ./run or demo
- go to first snap flow
- in the beginning of the Package snap there should be a 'Choose a snap name' form
- choose a snap name, try using invalid name to see if validation works
- 'Choose' button should be disabled until valid name is used
- Click 'Choose snap name' button, page should reload, first step should be marked done, second step should be open
- Make sure that name chosen by user is used in next steps (in some warning notifications and commands)
- Chosen snap name should be used in the register name form on the last step of first snap flow
- Go back and choose different language for first snap flow
- Snap name chosen in different language should not be used

#### Check different language/OS configurations to see all is fine

Screenshot 2019-04-24 at 14 51 22

### DIFF
--- a/static/js/public/accordion.test.js
+++ b/static/js/public/accordion.test.js
@@ -1,18 +1,5 @@
 import initAccordion, { initAccordionButtons } from "./accordion";
 
-// jsdom (in version that comes with jest) doesn't support .closest
-// so we need to polyfill that
-// shouldn't be needed when jsdom used by jest is updated to v11.12.0
-window.Element.prototype.closest = function(selector) {
-  var el = this;
-  while (el) {
-    if (el.matches(selector)) {
-      return el;
-    }
-    el = el.parentElement;
-  }
-};
-
 describe("accordion", () => {
   let button1;
   let button2;
--- a/static/js/public/accordion.test.js
+++ b/static/js/public/accordion.test.js
@@ -1,0 +1,142 @@
+import initAccordion, { initAccordionButtons } from "./accordion";
+
+// jsdom (in version that comes with jest) doesn't support .closest
+// so we need to polyfill that
+// shouldn't be needed when jsdom used by jest is updated to v11.12.0
+window.Element.prototype.closest = function(selector) {
+  var el = this;
+  while (el) {
+    if (el.matches(selector)) {
+      return el;
+    }
+    el = el.parentElement;
+  }
+};
+
+describe("accordion", () => {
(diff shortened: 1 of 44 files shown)


Lukewh:
On the "register your snap name" section on push:
After signing in the "Register snap name" button was disabled, until I updated the input (removed and re-added character)

bartaz:
@Lukewh Ah, good point, need to update it, so that when name is already set and valid button should be enabled.

hughesjordan:
@bartaz so far as I can see it all works nicely!

Can we increase the top and bottom padding on the `p-accordion__panel` element to give the content within the accordions more breathing room?

Screenshot 2019-04-25 at 9 59 46 am

evandandrea:
Shouldn't this check to make sure the name isn't already registered?

hughesjordan:
> Shouldn't this check to make sure the name isn't already registered?

@evandandrea It should do eventually but I think Bartek's plan is to be work on it as a separate feature as it involves a lot more work. @bartaz feel free to correct me if I'm wrong!

bartaz:
@evandandrea @hughesjordan At this point of the flow it's not possible with current API. Name registering API needs authorisation, so even to check if name is registered user needs to be signed in.

In current shape of the first snap flow it will be possible only in the last step after user signs in.

hughesjordan:
@bartaz sorry just spotted that step 2 should have a 'Continue' button which when clicked collapses that accordion and marks the step as complete with the green check icon

hughesjordan:
@bartaz nice! 👍

static/js/public/accordion.test.js
```diff
@@ -0,0 +1,142 @@
+import initAccordion, { initAccordionButtons } from "./accordion";
+
+// jsdom (in version that comes with jest) doesn't support .closest
+// so we need to polyfill that
+// shouldn't be needed when jsdom used by jest is updated to v11.12.0
+window.Element.prototype.closest = function(selector) {
```
Lukewh: https://jestjs.io/docs/en/configuration#setupfiles-array can we do this instead? We could just include out polyfill file and everything should be auto-filled?

PR #1861, opened 2019-04-25
MOOS FSF: Remove "ping" references

kyrofa:
Also update to use the latest MOOS, and stop referring to the "official" moos snap.


templates/home/_fsf_moos.html
```diff
@@ -3,40 +3,34 @@
     <h4>Why are snaps good for MOOS projects?</h4>
     <ul>
       <li>Bundle all the runtime requirements, including the exact version of MOOS/MOOS-IvP and system libraries you need.</li>
-      <li>Expand the distributions supported beyond just Ubuntu.</li>
```
kyrofa: Not a concern for MOOS.

PR #1877, opened 2019-04-30
Add snap installs graph to publisher page

jpmartinspt:
Adds snap installs graph to the publisher homepage.

TODO:

- [x] Limit snaps to 10? (Which snaps?)
- [x] Scale better for small screens

### DIFF
--- a/static/js/publisher/metrics/graphs/snapInstalls.js
+++ b/static/js/publisher/metrics/graphs/snapInstalls.js
@@ -1,0 +1,227 @@
+import Chart from "chart.js";
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { schemeCategory10 } from "d3-scale-chromatic";
+import { scaleOrdinal } from "d3-scale";
+
+class SnapInstalls extends Component {
+  constructor(props) {
+    super(props);
+
+    const metrics = [];
+    const error = false;
+    const loading = true;
+
+    this.state = {
+      metrics,
+      error,
+      loading
+    };
+  }
+
+  transformMetrics(json) {
+    let data = {
+      series: [],
+      max: 0,
+      buckets: json.buckets
+    };
+
+    const mapValues = (value, index) => {
+      if (data.max < value) {
+        data.max = value;
+      }
+
+      return {
+        x: json.buckets[index],
+        y: value
+      };
+    };
+
+    json.snaps.forEach(snap => {
+      // Grab the continued installs serie
+      const continuedInstalls = snap.series.filter(
+        serie => serie.name === "continued"
+      )[0];
+
+      const item = {
+        name: snap.name,
+        values: continuedInstalls.values.map(mapValues)
+      };
+
+      data.series.push(item);
+    });
+    return data;
+  }
+
+  getMetrics(snaps, token) {
+    // Limit snaps to 10 for performance reasons
+    const filteredSnaps = {};
+
+    Object.keys(snaps)
+      .slice(0, 9)
+      .forEach(key => {
+        filteredSnaps[key] = snaps[key];
+      });
+
+    fetch("/snaps/metrics/json", {
+      method: "POST",
+      body: JSON.stringify(filteredSnaps),
+      headers: {
+        "Content-Type": "application/json",
+        "X-CSRFToken": token
+      }
+    })
+      .then(response => {
+        if (!response.ok) {
+          throw "Could not fetch data.";
+        }
+        return response.json();
+      })
+      .then(json => {
+        const loading = false;
+        const metrics = this.transformMetrics(json);
+        this.setState({ loading, metrics }, this.draw);
+      })
+      .catch(error => {
+        this.setState({ error });
+      });
+  }
+
+  draw() {
+    if (this.canvas) {
+      const metrics = this.state.metrics;
+      const ctx = this.canvas.getContext("2d");
+
+      const color = scaleOrdinal(schemeCategory10);
+      const fontFamily =
+        '"Ubuntu", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif';
+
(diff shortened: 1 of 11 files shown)


templates/publisher/account-snaps.html
```diff
@@ -214,3 +224,17 @@ <h4 class="p-heading--five">Registered names ({{ registered_snaps|count }})</h4>
     </section>
   {% endif %}
 {% endblock %}
+
+{% block scripts %}
+  <script src="{{ static_url('js/dist/publisher.js') }}"></script>
+  <script>
+    Raven.context(function () {
+      var options = {
+        snaps: [{% for name, snap in snaps.items() %}"{{snap['snap-id']}}"{%endfor%}],
```
evandandrea: You're missing a comma between entries.

bartaz:
The graph here seems to have a different "style" to the ones on "Metrics" page for a snap. It's especially visible with a tooltip that has completely different look and feel.
It would be nice if we could keep them consistent in design, so they feel like parts of the same thing.

Screenshot 2019-05-06 at 08 11 12

bartaz:
Graph takes a bit to load and displays empty space while loading. And when it loads it makes the whole page jump quite significantly:

installs

1. Graph (when rendered) seems pretty large. It takes around half of my screen (that is not in small resolution). Does it need to be that big?

2. Some loading/spinner could be shown while loading, to give user a clue something is going to show up in the space.

3. It would be nice if the space 'reserved' for the graph when it's loading would be the same as rendered graph to prevent jumping on the page. Especially that the area that gets pushed down is probably the most interesting on the page and mostly interacted with (snaps list). Obviously we don't want to keep half of the screen empty for loading, so (see point 1.) would be nice to consider making the graph smaller in height.

jpmartinspt:
@bartaz Style is still the same but addressed the other comments, let me know if it's looking better.

popey:
I don't get a nice graph when I use the demo site :(

image

static/js/publisher/metrics/metrics.js
```diff
@@ -34,4 +38,16 @@ function renderMetrics(metrics) {
   territoriesMetrics(metrics.territories.selector, metrics.territories.metrics);
 }
 
-export default renderMetrics;
+/**
+ * Render publisher page metrics
+ * @param {Object} options An object with rendering options.
+ */
+function renderPublisherMetrics(options) {
```
bartaz: Name of the `renderPublisherMetrics` function seems to be more specific than `renderMetrics`, but they both practically render publisher metrics... just different ones.

As this are metrics for 'My snaps' page, maybe it could be called `renderMySnapsMetrics` or (because it renders snap installs) just `renderSnapInstalls`?

static/js/publisher/metrics/graphs/snapInstalls.js
```diff
@@ -0,0 +1,219 @@
+import Chart from "chart.js";
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { schemeCategory10 } from "d3-scale-chromatic";
+import { scaleOrdinal } from "d3-scale";
+
+class SnapInstalls extends Component {
```
bartaz: This component does a lot: loads data, renders error/loading state, draws graph...

It would be nice if we could at least split it into two: one "SnapInstallsChart" that would be just a stateless component that would draw a chart based on data passed in props, other one "SnapInstallsController" would deal with data handling and it would render "SnapInstallsChart" once data is fetched.

Does that make sense, or does chart.js handle it better when done in place?

static/js/publisher/metrics/graphs/snapInstalls.js
```diff
@@ -0,0 +1,219 @@
...
+      };
+
+      data.series.push(item);
+    });
+    return data;
+  }
+
+  getMetrics(snaps, token) {
+    // Limit snaps to 10 for performance reasons
+    const filteredSnaps = {};
```
bartaz: Not sure if it needs to be done in this PR, but would it make sense to load more snaps data later? So get snaps metrics in batches of 10 for example, and render them adding them to graph when they load?

static/js/publisher/metrics/graphs/snapInstalls.js
```diff
@@ -0,0 +1,219 @@
...
+      });
+  }
+
+  draw() {
+    if (this.canvas) {
+      const metrics = this.state.metrics;
+      const ctx = this.canvas.getContext("2d");
+      const color = scaleOrdinal(schemeCategory10);
+
+      const datasets = metrics.series.map((serie, i) => {
```
bartaz: I think `series` is both singular and plural.

bartaz:
Another visual issue (not sure if it is possible with chart.js - if not maybe another reason to reconsider using it...) - it doesn't use ubuntu font. So any text rendered on chart (axis, data, tooltips) are rendered in different font then rest of the page.

Would it be possible at least to do tooltip in CSS rather then print on canvas? Does chart.js have any hooks/callbacks or something that would allow custom tooltips or more control over styling?

Screenshot 2019-05-10 at 14 47 34

jpmartinspt:
@bartaz Swapped font-family. It's possible to do custom HTML/CSS tooltips too.

static/js/publisher/metrics/graphs/snapInstalls.js
jpmartinspt: I agree we would need better handling of what/how many snaps to display, but possibly as a follow-up PR.

static/js/publisher/metrics/graphs/snapInstalls.js
jpmartinspt: It can be split, not sure we want to do it before deciding if we are good to go with chart.js or not?

bartaz: @jpmartinspt Yes, I guess it's not worth putting more work if we are not sure.

Would be good to get UX and design feedback first to see if chart.js is feasible of doing what we need design wise.

jpmartinspt:
Merged into feature branch.

PR #1878, opened 2019-04-30
Update kde snaps (as icons have updated etc.)

Lukewh:
## QA

- Pull the branch
- `./run`
- Make sure all the snaps click through and look ok


bartaz:
Oh we should really start getting these from API...

Even if we want a handcrafted and ordered list, it would be good to only keep snap names in the list, but get details (in batch) from API. But I don't think it's possible to get batch of snap details based on list of snap names now from API.

bartaz:
Not really related to this PR, so maybe good to fix it separately, but seems that icons of snap media item push the text to the right when they load.

It's especially seen on slow connection, but also on fast one it's noticeable sometimes that text moves when loading:
icons

All snaps are meant to have icons (default or not), so would be good to reserve the space for them even before image loads.

Lukewh:
> Oh we should really start getting these from API...
> 
> Even if we want a handcrafted and ordered list, it would be good to only keep snap names in the list, but get details (in batch) from API. But I don't think it's possible to get batch of snap details based on list of snap names now from API.

Just had this conversation with @evandandrea we've got a potential item on the store roadmap for next cycle for publisher API support.

Batching isn't possible and looping through and getting each snap on every page load would be slow.

Lukewh:
> Not really related to this PR, so maybe good to fix it separately, but seems that icons of snap media item push the text to the right when they load.
> 
> It's especially seen on slow connection, but also on fast one it's noticeable sometimes that text moves when loading:
> icons
> 
> All snaps are meant to have icons (default or not), so would be good to reserve the space for them even before image loads.

Yeah, this seems to be an issue with Vanilla, even if you set a width on an image this jump happens. I tested it out with a codepen (that I can't find now).. Will redo the pen and add an issue to vanilla!

PR #1880, opened 2019-04-30
Simple notification component and show error for releases

Lukewh:
## QA

- Pull the branch
- `./run`
- Visit https://snapcraft.io/charm/releases
- See an error message


static/js/publisher/release.js
```diff
@@ -13,17 +14,31 @@ const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
...
+          channelMapsList={channelMapsList}
+          releasesData={releasesData}
+          options={options}
+        />
+      </Provider>,
+      document.querySelector(id)
+    );
+  } else {
+    ReactDOM.render(
+      <div className="row">
```
bartaz: If there is an error in releases data maybe we can just render notification in HTML on server side, not try to run releases JS at all? We know that there is an error even in the python view, no need to pass it all the way to React.

Lukewh: ... :mindblownemoji:

PR #1881, opened 2019-05-01
Force image sizes to prevent load-jumping

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/pull/1878#issuecomment-487961744

To get around https://github.com/canonical-web-and-design/vanilla-framework/issues/2175 this set's specific image sizes for media-object images.

## QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/store and http://0.0.0.0:8004/publisher/kde
- Throttle your connection/ browser [chrome](https://css-tricks.com/throttling-the-network/) [firefox](https://blog.nightly.mozilla.org/2016/11/07/simulate-slow-connections-with-the-network-throttling-tool/)
- See that the area icons should be aren't collapsed before the page loads

### DIFF
--- a/static/sass/_patterns_media-object--snap.scss
+++ b/static/sass/_patterns_media-object--snap.scss
@@ -44,8 +44,6 @@
 
     .p-media-object__image {
       height: 3rem;
-      max-height: 3rem;
-      max-width: 3rem;
       width: 3rem;
     }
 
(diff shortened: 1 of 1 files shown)


static/sass/_patterns_media-object--snap.scss
```diff
@@ -43,17 +43,17 @@
     }
 
     .p-media-object__image {
+      height: 3rem;
       max-height: 3rem;
```
bartaz: There is probably no need for `max-` width and height if `width` and `height` are both specified with the same values.

PR #1890, opened 2019-05-07
New category page

Lukewh:
Fixes https://github.com/canonicalltd/snap-squad/issues/970

## Does not:
Implement infinite scroll - do we want to do this? Really? Do we?

## To consider:
There are 20 featured snaps per category, which leaves a hanging snap - what should we do about that?

## QA
- Pull the branch
- `./run`
- Click through the "See more" buttons

## Note
- Featured category should not be limited to 20 featured snaps (all the snaps are featured)


kwm14:
## Design review
### Featured snaps
- [ ] Padding looking tight at the bottom the list of snaps

Screenshot 2019-05-07 at 11 40 14

### Development snaps
- [ ] Agree that widow snap looks odd. The previous discussion with @evandandrea when designing featured snaps he mentioned [here](https://github.com/canonical-web-and-design/snapcraft.io/issues/1719#issuecomment-477248859) that "To avoid widow snaps/odd numbers in a row we can change Top 20 snaps to a different number to cater"

Screenshot 2019-05-07 at 11 43 51

- [ ] Add more padding below the title or above first line of snaps as it's looking tight. '1rem' would do the trick 😉 

Screenshot 2019-05-07 at 11 59 31

kwm14:
# Design review
Thanks for updating @Lukewh 
- Design LGTM 👍 
- Bonus points for updating snap icon size and headings 😃

static/js/public/store-categories.js
```diff
@@ -72,4 +75,4 @@ function storeCategories() {
   }
 }
 
-export { storeCategories };
+export { storeCategories, getColour };
```
bartaz: It feels a bit confusing to have `getColour` exported from `store-categories.js`. 
It is used by categories, and apparently it's used elsewhere - maybe it would be good to have it as separate small module and import wherever needed?

webapp/store/views.py
```diff
@@ -132,26 +133,14 @@ def search_snap():
...
-        offset = flask.request.args.get("offset", default=0, type=int)
+        size = 48
 
-        try:
-            page = floor(offset / size) + 1
-        except ZeroDivisionError:
-            size = 10
-            page = floor(offset / size) + 1
+        if page == 1:
+            size = 47
```
bartaz: Why is size 48 or 47? Could we have these magic numbers explained in comments?

hughesjordan:
Displaying the page here feels a little unusual, I didn't even see it at first. I'd expect it to be with the pagination, what do you think @kwm14?

Also, spotted another of my test snaps that needs to be eradicated.
Screenshot 2019-05-08 at 9 56 54 am

templates/store/search.html
```diff
@@ -18,21 +18,6 @@
   <section id="main-content" class="p-strip--image is-shallow snapcraft-banner-background">
   <div class="row">
     <form action="/search" class="p-form p-form--inline p-form--search">
-      <div class="p-form__group p-form__group--no-grow">
```
bartaz: This this dropdown removed, is it now possible to search within a category? Or do we don't want this functionality anymore?

tests/store/tests_search.py
```diff
@@ -206,23 +175,16 @@ def test_search_q_with_results_but_no_total(self):
...
         self.assert_context("category_display", None)
-        self.assert_context("categories", [])
+        self.assert_context("featured_snaps", [])
         self.assert_context(
-            "snaps",
+            "searched_snaps",
             [
                 {"package_name": "toto"},
                 {"package_name": "tata"},
                 {"package_name": "tutu"},
```
bartaz: 🤣

kwm14:
> Displaying the page here feels a little unusual, I didn't even see it at first. I'd expect it to be with the pagination, what do you think @kwm14?
> 
> Also, spotted another of my test snaps that needs to be eradicated.
> Screenshot 2019-05-08 at 9 56 54 am

Good spot @hughesjordan, agree it does look very odd positioned there. Either remove it or keep within the pagination as you suggest for a first pass.

Another Vanilla plug for @Lukewh 😉 but in 2.0 we're releasing a newly designed page pagination which will solve our issue here.

pagination

webapp/store/logic.py
```diff
@@ -378,3 +393,13 @@ def get_videos(media):
     return [
         get_video_embed_code(m["url"]) for m in media if m["type"] == "video"
     ]
+
+
+def promote_snap_with_icon(snaps):
+    snap_with_icon = next(snap for snap in snaps if snap["icon_url"] != "")
+
+    snap_with_icon_index = snaps.index(snap_with_icon)
```
jpmartinspt: If there is no snap with icon this will raise an exception, is that intended?

webapp/store/logic.py
```diff
@@ -99,6 +99,21 @@ def convert_navigation_url(url, link):
     return url
 
 
+def build_pagination_link(snap_searched, snap_category, page):
+    params = []
+
+    if snap_searched != "":
```
jpmartinspt: I think `if snaps_searched:` is enough since "" evaluates to `False`

webapp/store/logic.py
```diff
@@ -99,6 +99,21 @@ def convert_navigation_url(url, link):
...
+        params.append("q=" + snap_searched)
+
+    if snap_category != "":
```
jpmartinspt: Same as above.

templates/store/search.html
Lukewh: It's still possible by URL hacking, but we decided to remove it as it's overly complicated and dead-end promoting on the UI

webapp/store/logic.py
Lukewh: Good catch thanks

hughesjordan:
@Lukewh 👍

webapp/store/views.py
```diff
@@ -132,26 +133,17 @@ def search_snap():
...
-        offset = flask.request.args.get("offset", default=0, type=int)
+        # The default size should be 48 (rows of 4)
+        size = 48
 
-        try:
-            page = floor(offset / size) + 1
-        except ZeroDivisionError:
-            size = 10
-            page = floor(offset / size) + 1
+        # Page 1 has a snap at the top, and a few rows of 3, followed by rows
```
bartaz: It seems that that is the case only when showing first category page.
First page of standard search (like: https://snapcraft-io-canonical-web-and-design-pr-1890.run.demo.haus/search?q=snap) should not be limited to 47 snaps.

If the page size varies between pages - are we sure we are not skipping some results when calculating what to get from API?

Lukewh: Aaah good catch with normal search results! Forgot about those :D!

Total results is 141.
Page 1 has 47, so the last page is calculated as 3 exactly.
Page 2 has 48, so the last page is calculated as 2.9375, which is rounded up to 3
Winning

Total results is 144.
Page 1 has 47, so the last page is calculated as 3.063...., which is rounded up to 4
Page 2 has 48, so the last page is calculated as 3 exactly.
Uh-oh

I'm of the opinion that the number of people that skip to the last page is minimal and the chance of this happening is low. If someone does jump beyond the "last" page, the snaps from the last page are shown - so nobody will see a blank screen (https://snapcraft-io-canonical-web-and-design-pr-1890.run.demo.haus/search?q=snap&page=5010202010)

webapp/store/views.py
```diff
@@ -162,29 +156,71 @@ def search_snap():
...
+            if total_pages:
+                links["last"] = logic.build_pagination_link(
+                    snap_searched=snap_searched,
+                    snap_category=snap_category,
+                    page=total_pages,
+                )
+
+        featured_snaps = []
+
+        number_of_featured_snaps = 19
```
bartaz: One more magic number worth explaining in comments (especially that it depends on layout, so when design is updated we need to take that into account).

PR #1891, opened 2019-05-07
Annotated snapcraft yaml

### LINKED ISSUE #1896: Build annotated snapcraft.yaml and download for python tutorial

bartaz:
Fixes #1896

Adds annotated snapcraft.yaml viewer to Python tutorial in first snap flow (Package step).
Allows to download snapcraft.yaml file generated with a snap name chosen by user.

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-1891.run.demo.haus/
- go to Package step of Python first snap flow
- choose name, go to third step (Download your snapcraft.yaml)
- annotated yaml should be shown (with a name chosen before)
- tooltips should be shown when hovering over different sections of the file
- button below it should download the snapcraft.yaml file with a name chosen by user

**Please make sure to read the copy text of "Downloa your snapcraft.yaml" section. My "D" key is misbehaving recently an I may have mae some typos ;)**

Screenshot 2019-05-09 at 15 56 28

### DIFF
--- a/webapp/first_snap/views.py
+++ b/webapp/first_snap/views.py
@@ -85,6 +85,36 @@ def get_language(language):
     )
 
 
+@first_snap.route("/<language>/snapcraft.yaml")
+def get_language_snapcraft_yaml(language):
+    filename = f"first_snap/content/{language}/package.yaml"
+    snapcraft_yaml_filename = f"first_snap/content/{language}/snapcraft.yaml"
+    snap_name_cookie = f"fsf_snap_name_{language}"
+    steps = get_file(filename)
+
+    if not steps:
+        return flask.abort(404)
+
+    snap_name = steps["name"]
+
+    if snap_name_cookie in flask.request.cookies:
+        snap_name = flask.request.cookies.get(snap_name_cookie)
+
+    snapcraft_yaml = get_file(snapcraft_yaml_filename, {"${name}": snap_name})
+
+    if not snapcraft_yaml:
+        return flask.abort(404)
+
+    content = StringIO()
+    yaml.dump(snapcraft_yaml, content)
+
+    resp = flask.Response(content.getvalue(), mimetype="text/yaml")
+    resp.headers.extend(
+        {"Content-Disposition": "attachment;filename=snaprcaft.yaml"}
+    )
+    return resp
+
+
 @first_snap.route("/<language>/<operating_system>/package")
 def get_package(language, operating_system):
     filename = f"first_snap/content/{language}/package.yaml"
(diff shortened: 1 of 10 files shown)


webapp/first_snap/views.py
```diff
@@ -47,11 +85,45 @@ def get_language(language):
...
+
+    if not steps:
+        return flask.abort(404)
+
+    snap_name = steps["name"]
+
+    if snap_name_cookie in flask.request.cookies:
+        snap_name = flask.request.cookies.get(snap_name_cookie)
+
+    snapcraft_yaml = get_file(snapcraft_yaml_filename, {"${name}": snap_name})
```
bartaz: @jpmartinspt On one hand it's nice we could reuse `get_file` here and let it do the replacement. On the other hand we parse the YAML just to dump it back into text form again and download as a file.

I suppose we could do file reading and replacing without involving YAML library, but that would mean reimplementing `get_file` (just skipping YAML parsing).

Thoughts?

Lukewh:
with the extra padding, it's annoying that "local" drops down a line.
We should maybe update the descriptions... AGAIN... to have the break at the right place?

Screenshot_2019-05-09 Snapcraft - Snaps are universal Linux packages

bartaz:
@Lukewh This file is copied from offlineimap github repo, I would rather not edit it for style reasons. There will always be some screen width where it will wrap.

To make it better on a default big screen maybe I can just bump with width from col-8 to col-9 (this should make files within 80 char lines display without wrapping)

Lukewh:
Also I'm pretty sure the `command: bin/test-offlineimap-lukewh` should still be `command: bin/test-offlineimap` the command that run's hasn't changed... Though I could be wrong?

evandandrea:
Yes, it should be:
```
apps:
  test-offlineimap-nnnn:
    command: bin/offlineimap
```

The app name and the snap name have to match, but the command path can be anything.

bartaz:
@evandandrea Ah, thanks. So the command is not affected by snap name change, the command itself comes from what is in code not snapcraft.yaml?

evandandrea:
Precisely that.

Lukewh:
> @evandandrea Ah, thanks. So the command is not affected by snap name change, the command itself comes from what is in code not snapcraft.yaml?

@bartaz yeah if you check the repo, there's the `bin/` folder with `offlineimap` executable in it, it's just the command a user will run on the cli `test-offlineimap-name` that you want to change to avoid conflicts and to match the snap name

hughesjordan:
@bartaz can we make sure that spacing is consistent within each accordion, above and below the content there should be more space.

Also, we're missing the last sentence as seen below and can we update the spacing there too please?

Screenshot 2019-05-09 at 1 23 09 pm

Last thing would be to add some text to instruct the user what to do with the downloaded yaml file:

Screenshot 2019-05-09 at 1 31 03 pm

PR #1895, opened 2019-05-08
Hiri case study

Lukewh:
Fixes https://github.com/CanonicalLtd/snap-squad/issues/984

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/hiri
- See the link to the case study in place of the first blog post

### DIFF
--- a/static/js/public/snap-details/blog-posts.js
+++ b/static/js/public/snap-details/blog-posts.js
@@ -99,7 +99,7 @@ function snapDetailsPosts(
     if (posts.length > 0 && showOnSuccessSelector) {
       const showOnSuccess = document.querySelector(showOnSuccessSelector);
       if (showOnSuccess) {
-        showOnSuccess.style.display = "block";
+        showOnSuccess.classList.remove("u-hide");
       }
     }
   });
--- a/static/js/public/snap-details/blog-posts.js
+++ b/static/js/public/snap-details/blog-posts.js
@@ -17,6 +17,8 @@ class BlogPosts {
     this.holder = document.querySelector(holderSelector);
     this.template = document.querySelector(templateSelector);
 
+    this.limit = 3;
+
     if (!this.holder) {
       throw new Error(`${holderSelector} does not exist`);
     }
@@ -46,11 +48,17 @@ class BlogPosts {
           });
         }
 
-        posts.forEach(post => {
+        const cols = 12 / this.limit;
+
+        posts.forEach((post, index) => {
+          if (index >= this.limit) {
+            return;
+          }
           let postHTML = this.template.innerHTML;
           Object.keys(post).forEach(key => {
             postHTML = postHTML.split("${" + key + "}").join(post[key]);
           });
+          postHTML = postHTML.split("${size}").join(`col-${cols}`);
           postsHTML.push(postHTML);
         });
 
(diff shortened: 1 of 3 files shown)


templates/store/snap-details.html
```diff
@@ -192,22 +192,48 @@ <h4>Share this snap</h4>
...
+          </div>
+          <div class="col-8">
             <h3 class="p-heading--four">Related blog posts</h3>
           </div>
         </div>
-        <div class="row u-equal-height" data-js="blog-article-list" data-snap="{{ package_name }}">
+        <div class="row u-equal-height" data-js="blog-articles">
+          <div class="col-4">
+            <h3 class="p-heading--four">
+              <a href="https://www.ubuntu.com/engage/hiri-case-study">Hiri successfully monetises their snap to dispel myths around proprietary Linux apps </a>
```
anthonydillon: Should this be an external link?

templates/store/snap-details.html
```diff
@@ -192,22 +192,48 @@ <h4>Share this snap</h4>
...
+            </h3>
+          </div>
+          <div class="col-8" data-js="blog-article-list" data-snap="{{ package_name }}" data-limit="2"></div>
         </div>
       </div>
-    </div>
+    {% else %}
+      <div data-js="blog-articles" style="display: none;">
```
anthonydillon: You should use `u-hide` class instead of inline styles were possible

templates/store/snap-details.html
```diff
@@ -192,22 +192,48 @@ <h4>Share this snap</h4>
   </div>
 
   {% if not IS_BRAND_STORE %}
-    <div data-js="blog-articles" style="display: none;">
+    {% if package_name == "hiri" %}
```
bartaz: Not sure how long it is supposed to live here, but maybe we could move it to some partial, so it doesn't clutter main template too much?

templates/store/snap-details.html
```diff
@@ -192,22 +192,48 @@ <h4>Share this snap</h4>
...
+            <h3 class="p-heading--four">
+              <a href="https://www.ubuntu.com/engage/hiri-case-study" class="p-link--external">Hiri successfully monetises their snap to dispel myths around proprietary Linux apps </a>
+            </h3>
+          </div>
+          <div class="col-8" data-js="blog-article-list" data-snap="{{ package_name }}" data-limit="2"></div>
         </div>
       </div>
-    </div>
+    {% else %}
+      <div data-js="blog-articles"class="u-hide">
```
bartaz: Parsers would handle this, but a space would be nice:
```suggestion
      <div data-js="blog-articles" class="u-hide">
```

static/js/public/snap-details/blog-posts.js
```diff
@@ -83,13 +91,15 @@ function snapDetailsPosts(
     throw new Error("Snap not defined");
   }
 
+  this.limit = blogPosts.holder.dataset.limit;
```
bartaz: Hmm... I wonder why don't we do more of that instead passing lots of options via JSON to function. But I guess this works only for a small settings like a number or short string.

Anyway, shouldn't there be some default value if it's not set in DOM?

Lukewh: I used `this.limit` rather than `blogPosts.limit`. I thought the function was in the class. The class does have a default of 3 though :)

PR #1898, opened 2019-05-09
Email preferences

Lukewh:
Show's whether a publisher is signed up to the blog newsletter and allows them to toggle their subscription.

## QA

- You'll need `MARKETO_URL`, `MARKETO_CLIENT_ID` and `MARKET_CLIENT_SECRET` in your `.env.local` (if you're familiar with Marketo, go ahead and get them - otherwise ask me :))
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/account/details
- You should see an "Email preferences" section at the bottom of the page with a toggle for subscribing
- Toggle the checkbox and save
- Ensure toggling/ untoggling and saving works as you'd expect

### DIFF
--- a/static/js/publisher/submitEnabler.js
+++ b/static/js/publisher/submitEnabler.js
@@ -1,0 +1,35 @@
+function submitEnabler(formSelector, buttonSelectors) {
+  if (!formSelector) {
+    throw new TypeError("`formSelector` argument is required");
+  }
+
+  if (!buttonSelectors || buttonSelectors.length === 0) {
+    throw new TypeError("At least one `buttonSelectors` must be defined");
+  }
+
+  const formEl = document.querySelector(formSelector);
+
+  if (!formEl) {
+    throw new Error(`${formSelector} is not a valid element`);
+  }
+
+  const buttonEls = buttonSelectors.map(selector =>
+    document.querySelector(selector)
+  );
+
+  buttonEls.forEach(button => {
+    if (button) {
+      button.setAttribute("disabled", "disabled");
+      button.classList.add("is--disabled");
+    }
+  });
+
+  formEl.addEventListener("change", () => {
+    buttonEls.forEach(button => {
+      button.removeAttribute("disabled");
+      button.classList.remove("is--disabled");
+    });
+  });
+}
+
+export { submitEnabler as default };
--- a/static/js/publisher/submitEnabler.js
+++ b/static/js/publisher/submitEnabler.js
@@ -1,3 +1,5 @@
+import shallowDiff from "../libs/shallowDiff";
+
 function submitEnabler(formSelector, buttonSelectors) {
   if (!formSelector) {
     throw new TypeError("`formSelector` argument is required");
@@ -17,6 +19,14 @@ function submitEnabler(formSelector, buttonSelectors) {
     document.querySelector(selector)
   );
 
+  const initialState = new FormData(formEl);
+
+  const initialStateJson = {};
+
+  for (const [key, value] of initialState.entries()) {
+    initialStateJson[key] = value;
+  }
+
   buttonEls.forEach(button => {
     if (button) {
       button.setAttribute("disabled", "disabled");
@@ -25,9 +35,22 @@ function submitEnabler(formSelector, buttonSelectors) {
   });
 
   formEl.addEventListener("change", () => {
+    const newState = new FormData(formEl);
+    const newStateJson = {};
+
+    for (const [key, value] of newState.entries()) {
+      newStateJson[key] = value;
+    }
+
+    const diff = shallowDiff(initialStateJson, newStateJson);
     buttonEls.forEach(button => {
-      button.removeAttribute("disabled");
-      button.classList.remove("is--disabled");
+      if (diff) {
+        button.removeAttribute("disabled");
+        button.classList.remove("is--disabled");
+      } else {
+        button.setAttribute("disabled", "disabled");
+        button.classList.add("is--disabled");
+      }
     });
   });
 }
--- a/webapp/api/marketo.py
+++ b/webapp/api/marketo.py
@@ -49,13 +49,16 @@ LEADS = "".join(
 class MarketoApi:
     # Marketo isn't fast, so give it plenty of time to make a connection,
     # and respond
-    api_session = api.requests.Session(timeout=(2, 12))
-    token = ""
+    def __init__(self, api_session=api.requests.Session(timeout=(2, 12))):
+        self.api_session = api_session
+        self.token = None
 
-    def __init__(self):
-        self.authenticate()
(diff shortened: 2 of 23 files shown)


webapp/publisher/views.py
```diff
@@ -78,16 +78,53 @@ def get_account_details():
...
         "username": flask_user["nickname"],
         "displayname": flask_user["fullname"],
         "email": flask_user["email"],
+        "subscriptions": subscriptions,
     }
 
     return flask.render_template("publisher/account-details.html", **context)
 
 
+@account.route("/details", methods=["POST"])
```
Lukewh: @tbille @jpmartinspt I'd appreciate some help testing this route. I've tried many things and wasted a lot of time, in the end I've just removed the tests.

hughesjordan:
@Lukewh I'd expect the button to be disabled until a change has been made and I'd also expect to see a notif to assure the user that preferences have been saved.

webapp/publisher/views.py
```diff
@@ -78,16 +80,57 @@ def get_account_details():
...
+        subscribed_to_newsletter = False
+        if (
+            "snapcraftnewsletter" in marketo_subscribed
+            and marketo_subscribed["snapcraftnewsletter"]
+        ):
+            subscribed_to_newsletter = True
+    except Exception:
+        pass
+
+    if subscribed_to_newsletter is not None:
```
jpmartinspt: Think you can remove the `is not None`

webapp/api/marketo.py
```diff
@@ -0,0 +1,126 @@
...
+        try:
+            body = response.json()
+        except ValueError as decode_error:
+            api_error_exception = ApiResponseDecodeError(
+                "JSON decoding failed: {}".format(decode_error)
+            )
+            raise api_error_exception
+
+        if not response.ok:
+            if "success" in body and body["success"] is False:
```
jpmartinspt: Could be `if body.get("success") is False:`

webapp/publisher/views.py
```diff
@@ -12,12 +12,14 @@
...
     MissingUsername,
 )
-
+from webapp.decorators import login_required
 
 account = flask.Blueprint(
     "account", __name__, template_folder="/templates", static_folder="/static"
 )
 
+marketo = marketo_api.MarketoApi()
```
jpmartinspt: You seem to be redefining `marketo` in both views, could we just make use of this instance instead?

webapp/publisher/views.py
jpmartinspt: I do see a test for it, anything in particular you wanted help with?

webapp/publisher/views.py
Lukewh: Ah, the two in the methods are from a bad merge, thanks :)!

webapp/publisher/views.py
Lukewh: it has 3 states though, None (nothing came back from the server), False (you're not subscribed) and True (you're subscribed). Is there a better way of doing this?

hughesjordan:
@Lukewh when interacting with the checkbox and reverting it to its previous state without saving preferences, the button should also revert to the previous state.

Lukewh:
> @Lukewh when interacting with the checkbox and reverting it to its previous state without saving preferences, the button should also revert to the previous state.

Sorted :+1:

webapp/publisher/views.py
jpmartinspt: You could update `subscriptions` inside the try instead of setting `subscribed_to_newsletter` and just get rid of this `if`?

webapp/publisher/views.py
```diff
@@ -78,16 +80,55 @@ def get_account_details():
...
+    # if anything fails, just continue and don't show
+    # this section
+    try:
+        marketo_user = marketo.get_user(flask_user["email"])
+        marketo_subscribed = marketo.get_newsletter_subscription(
+            marketo_user["id"]
+        )
+        subscribed_to_newsletter = False
+        if (
+            "snapcraftnewsletter" in marketo_subscribed
```
jpmartinspt: I think this could also be just `if marketo_subscribed.get("snapcraftnewsletter")`.

webapp/api/marketo.py
```diff
@@ -0,0 +1,126 @@
...
+                    map(lambda error: error["message"], body["errors"]),
+                )
+
+                raise api_error_exception
+            else:
+                raise ApiResponseError("Unknown error from api", 500)
+
+        return body
+
+    def authenticate(self):
```
tbille: tests are failing because you are making an API call when initializing the application.

Can't we get the token without making this request? Otherwise we will need to think of a way to mock this API call on the setup of every test.

Lukewh: I think maybe won't work. I think we might need to initialize and do the handshake on every request that requires it. Maybe you can have a look at the Marketo API, but to me it seems like the token lifetime is really short.

tbille: @jpmartinspt 	I'll let you investigate

static/js/publisher/submitEnabler.js
```diff
@@ -0,0 +1,58 @@
...
+
+  formEl.addEventListener("change", () => {
+    const newState = new FormData(formEl);
+    const newStateJson = {};
+
+    for (const [key, value] of newState.entries()) {
+      newStateJson[key] = value;
+    }
+
+    const diff = shallowDiff(initialStateJson, newStateJson);
```
bartaz: Seems like most of the `shallowDiff` logic of marking fields as added/changed/removed, giving previous and current value is not really needed if all we want is a bool check?
Or is it used elsewhere?

Even on listing page where we have a similar check (a possible reuse case) we only list changed/added fields and their values - I'm not sure if we need all of the details `shallowDiff` provides?

Lukewh: How do you feel about adding a flag to get either the complete output (or something closer to the listing page diff check), or a boolean?

I could add an issue to update the listing page state stuff to use this.

bartaz: I don't mind - I just felt it does much more that we need it to. Adding even more conditional logic to it seems even more unnecessary ;)

I'm fine with leaving it as it is not to add more work. I just feel that it means we are having a code to maintain that does something we don't need.

therealjuan:
image

Can we make the check work okay? :)

Apart from that works beatifully!

Lukewh:
@therealjuan this is site wide. I've created a separate issue for it https://github.com/canonical-web-and-design/snapcraft.io/issues/2020

anthonydillon:
@therealjuan I believe this will be resolved with the Vanilla v2 upgrade: https://docs.vanillaframework.io/base/forms/#checkbox

therealjuan:
So let's mergey merge then

PR #1912, opened 2019-05-14
Moved devmode icon to right hand side

### LINKED ISSUE #1911: Release UI – Move devmode icon to right hand side
Right now on the Release UI when a revision is in devmode, there is an icon on the left-hand side. This shifts all content a few pixels inside the cell:

image

Moving the devmode icon to the right hand side will avoid this happening:
image

therealjuan:
### QA Steps

Fixes #1911 Moving devmode icon to right hand side

1. Pull the branch or open the demo service
2. Go to `htop/releases` to see revisions with devmode where the icon is on the right-hand side
3. #winning

### DIFF
--- a/static/js/publisher/release/components/releasesTableCell.js
+++ b/static/js/publisher/release/components/releasesTableCell.js
@@ -31,13 +31,13 @@ class ReleasesTableCell extends Component {
   renderRevision(revision, isPending) {
     return (
       <Fragment>
-        {isInDevmode(revision) && (
-          <span className="p-release-data__icon">
-            <DevmodeIcon revision={revision} showTooltip={false} />
-          </span>
-        )}
         <span className="p-release-data__info">
           <span className="p-release-data__title">{revision.revision}</span>
+          {isInDevmode(revision) && (
+            <span className="p-release-data__icon u-float-right">
+              <DevmodeIcon revision={revision} showTooltip={false} />
+            </span>
+          )}
           <span className="p-release-data__meta">{revision.version}</span>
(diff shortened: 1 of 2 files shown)


bartaz:
@therealjuan Are we fine from design point of view with icon being quite far away from revision when there are not so many architectures?

Screenshot 2019-05-14 at 12 05 13

bartaz:
Icon will 'drop' down for long revision names:

Screenshot 2019-05-14 at 12 14 09

But, it doesn't really look broken and 7 digit revision numbers are probably not something we will see very soon.

static/js/publisher/release/components/releasesTableCell.js
```diff
@@ -31,13 +31,13 @@ class ReleasesTableCell extends Component {
...
       <Fragment>
-        {isInDevmode(revision) && (
-          <span className="p-release-data__icon">
-            <DevmodeIcon revision={revision} showTooltip={false} />
-          </span>
-        )}
         <span className="p-release-data__info">
           <span className="p-release-data__title">{revision.revision}</span>
+          {isInDevmode(revision) && (
+            <span className="p-release-data__icon u-float-right">
```
bartaz: The `p-release-data__icon` class doesn't seem to be needed anymore (removed from styles).

therealjuan:
From a design standpoint is fine, and we probably will be updating the devmode icon soon as it is not quite prescriptive. 

Revision numbers will go away at some point, once sets are in place so no issue from that point of view.

PR #1913, opened 2019-05-14
Improve speed on blog pages

jpmartinspt:
## Done
Improve speed of blog pages and reduce the number of HTTP calls to the wordpress API made to render these pages.

## QA

- ./run
- Make sure /blog and /blog/<slug> endpoints work as in production.

### DIFF
--- a/webapp/api/blog.py
+++ b/webapp/api/blog.py
@@ -9,7 +9,7 @@ API_URL = os.getenv(
 TAGS = [2996]  # 'snapcraft.io'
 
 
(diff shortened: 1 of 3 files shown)


webapp/api/blog.py
```diff
@@ -9,7 +9,7 @@
 TAGS = [2996]  # 'snapcraft.io'
 
 
-api_session = api.requests.CachedSession(expire_after=3600)
+api_session = api.requests.Session(timeout=(1, 5))
```
tbille: keep CacheSession AND LET'S MERGE THAT!!!

PR #1919, opened 2019-05-15
Check if user is snap collaborator for snap details editing

### LINKED ISSUE #1813: "Update its listing information here." link should be availabe to the snap collaborators
### Current behavior
screenshot

### Expected behaviour
screenshot

1. Set another developer account as collaborator in the snap's dashboard
1. Login to the collaborator account
1. Browse the snap's listing page

### Specs

- _URL:_ <https://snapcraft.io/_snap_identifier_> (as collaborator)
- _Operating system:_ Ubuntu 16.04 AMD64
- _Browser:_ Mozilla Firefox 66.0.2 from Ubuntu archive

bartaz:
Fixes #1813

Show notification on snap details page for snap owners and collaborators with a link to editing details of listing page.

### QA

- ./run or https://snapcraft-io-canonical-web-and-design-pr-1919.run.demo.haus/
- go to snap details page of a snap shared with you (that you didn't create yourself)
- snap details page should show notification about editing the snap listing
- go to snap details page of snap you don't own (and are not a collaborator of) - no notification should be on the page

Screenshot 2019-05-17 at 09 37 06

### DIFF
--- a/webapp/login/views.py
+++ b/webapp/login/views.py
@@ -8,6 +8,7 @@ from webapp.api import dashboard
 from webapp.api.exceptions import ApiCircuitBreaker, ApiError, ApiResponseError
 from webapp.extensions import csrf
 from webapp.login.macaroon import MacaroonRequest, MacaroonResponse
+from webapp.publisher.snaps import logic
 
 login = flask.Blueprint(
     "login", __name__, template_folder="/templates", static_folder="/static"
@@ -70,6 +71,9 @@ def after_login(resp):
             "image": resp.image,
             "email": account["email"],
         }
+        snaps, names = logic.get_snaps_account_info(account)
+        flask.session["user_snaps"] = snaps
+
     except ApiCircuitBreaker:
         flask.abort(503)
     except Exception:
--- a/webapp/login/views.py
+++ b/webapp/login/views.py
@@ -72,7 +72,7 @@ def after_login(resp):
             "email": account["email"],
         }
         snaps, names = logic.get_snaps_account_info(account)
-        flask.session["user_snaps"] = snaps
+        flask.session["user_snaps"] = list(snaps.keys())
 
     except ApiCircuitBreaker:
         flask.abort(503)
--- a/webapp/login/views.py
+++ b/webapp/login/views.py
@@ -71,8 +71,8 @@ def after_login(resp):
             "image": resp.image,
             "email": account["email"],
         }
-        snaps, names = logic.get_snaps_account_info(account)
-        flask.session["user_snaps"] = list(snaps.keys())
+        owned, shared = logic.get_snap_names_by_ownership(account)
+        flask.session["user_shared_snaps"] = shared
 
     except ApiCircuitBreaker:
         flask.abort(503)
--- a/webapp/publisher/snaps/logic.py
+++ b/webapp/publisher/snaps/logic.py
@@ -7,10 +7,10 @@ from json import dumps
 def get_snaps_account_info(account_info):
     """Get snaps from the account information of a user
 
-    :param account_info The account informations
+    :param account_info: The account informations
 
-    :return A list of snaps
-    :return A list of registred snaps
+    :return: A list of snaps
+    :return: A list of registred snaps
     """
     user_snaps = {}
     registered_snaps = {}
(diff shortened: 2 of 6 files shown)


webapp/store/snap_details_views.py
```diff
@@ -237,6 +240,20 @@ def snap_details(snap_name):
             os_metrics = None
             country_devices = None
 
+        if authentication.is_authenticated(flask.session):
+            try:
+                account_info = publisher_api.get_account(flask.session)
+            except ApiError as api_error:
+                # TODO: handle errors including macaroon refresh
```
bartaz: @tbille @jpmartinspt 

Is it OK for us to be calling publisher API and logic from snap details view? We want to get list of users snaps (if they are signed in) to know if they can edit current snap.
Also, calling publisher API would require different error handling then snap details views, at least for macaroon refresh redirect to work...

Can it be done here? Or should we do it differently?

tbille: mmm I am not sure this is the best way to do this. Let's chat @jpmartinspt if we could do something better.

bartaz: @tbille Other idea we discussed with Luke was to store list of user snaps in session when the log in. But not sure what other implications this may have.

tbille: this was this idea that I had. Since this is not something sensitive.

jpmartinspt: I think I like the session idea too.

bartaz: @jpmartinspt I guess as long as we use it just for notification it will not be as bad, probably may be hardly noticeable.
We may consider refreshing the data in session when visiting "my snaps" page, so at least no logout will be required.

Lukewh:
Hmmm @jpmartinspt @bartaz 

```
/usr/local/lib/python3.6/dist-packages/werkzeug/wrappers.py:1116: UserWarning: The "b'session'" cookie is too large: the value was 5347 bytes but the header required 26 extra bytes. The final size was 5373 bytes but the limit is 4093 bytes. Browsers may silently ignore cookies larger than this.
  samesite=samesite
```

jpmartinspt:
@Lukewh @bartaz too much stuff in session.. we need to try and store name/id only for each snap.

bartaz:
@jpmartinspt @Lukewh Updated, should only keep list of snaps in session now.

bartaz:
@jpmartinspt @Lukewh I updated it once again. Now only the list of snap names that are shared with user is kept in the session. User own snaps are not (as we can check that using their username). This should be the minimal amount of information we can store in session to keep the functionality.

webapp/login/views.py
```diff
@@ -70,6 +71,9 @@ def after_login(resp):
             "image": resp.image,
             "email": account["email"],
         }
+        owned, shared = logic.get_snap_names_by_ownership(account)
+        flask.session["user_shared_snaps"] = list(shared)
```
jpmartinspt: Do we need to call `list` on shared? Seems like the `logic.get_snap_names_by_ownership` function already returns a list.

bartaz: Ah, good catch, it's a leftover from old code then keys needed converting to list. Will fix that.

webapp/publisher/snaps/logic.py
```diff
@@ -50,6 +50,29 @@ def get_snaps_account_info(account_info):
     return user_snaps, registered_snaps
 
 
+def get_snap_names_by_ownership(account_info):
+    """Get list of snaps names user is collaborator of
+
+    :param account_info The account informations
+
+    :return A list of owned snaps names
```
tbille: syntax:

```
:param account_info: blabla

:returns: bla bla
```

webapp/publisher/snaps/logic.py
```diff
@@ -50,6 +50,29 @@ def get_snaps_account_info(account_info):
     return user_snaps, registered_snaps
 
 
+def get_snap_names_by_ownership(account_info):
+    """Get list of snaps names user is collaborator of
```
tbille: of what!!! the mystery  isn't resolved

Lukewh: of snap names... no mystery

webapp/publisher/snaps/logic.py
bartaz: I think I copied it from somewhere... but will fix.

webapp/publisher/snaps/logic.py
bartaz: @Lukewh So it is correct English? I wasn't sure when writing it as well ;)

Lukewh: The only thing I'd change is "snaps" > "snap" as the "names" pluralises...

bartaz: @Lukewh but it's a list of "names" not snaps... so.. "list of snaps' names ..." maybe?

PR #1924, opened 2019-05-17
WIP: Remove hardcoded snap names from first snap flow content

bartaz:
In order to simplify first snap flow views and content .yaml files this PR removes hardcoded snap names from each language content (like `test-offlineimap-{name}`) and replaces them with generated names based on language name (like `test-python-[yourname]`).

### QA
- ./run or demo
- go to first snap flow, any language
- if you haven't chosen any snap name before default snap name based on language should be used

Screenshot 2019-05-17 at 11 57 06

### DIFF
--- a/tests/first_snap/tests_views.py
+++ b/tests/first_snap/tests_views.py
@@ -142,7 +142,7 @@ class FirstSnap(TestCase):
         self.assert_context("language", "python")
         self.assert_context("os", "linux")
         self.assert_context("user", None)
-        self.assert_context("snap_name", "test-offlineimap-{name}")
+        self.assert_context("snap_name", "test-python-[yourname]")
 
     def test_get_push_snap_name(self):
         response = self.client.get(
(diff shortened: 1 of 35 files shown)


bartaz:
@evandandrea Do we need hardcoded project specific suggested snap names for first snap flow such as `test-offlineimap-{name}` or would it be fine to base such name on name of the language, like `test-python-[yourname]`? This would simplify both the code and .yaml files with content.

evandandrea:
I'm not sure about dropping the app name for the language it's built in. It's not a generic python example, it's offlineimap. However, I do like using yourname when they've already signed into SSO.

@igorljubuncic as you own the FSF from the Advocacy side, what are your thoughts?

bartaz:
Sure, the names can be project specific, but from tutorial point of view users choose a language, not a project. And in current flow we ask them to choose the snap name even before they download the code, so before they even know what they are naming.

igorljubuncic:
I also think we should not use generic names because it can be confusing. We might end up with 100 python snaps that do completely different things - test-python-igor and test-python-evan don't tell us anything about functionality.

We also need to use FSF as a best practice for new devs - so they adopt the same approach when they build their applications.

Appending username automatically for logged in users is a good idea.

bartaz:
@igorljubuncic Ok thanks. I'll close this then.
For having username added automatically I opened a separate issue: https://github.com/canonical-web-and-design/snapcraft.io/issues/1937

PR #1926, opened 2019-05-20
Skip featured colouring if parent can't be found

Lukewh:
Fixes various sentry errors relating to `t is null` `Cannot read property 'style' of null` etc.

## QA

- Pull the branch
- Make sure the featured background colours still load
- Do the same on some category pages

### DIFF
--- a/static/js/libs/colours.js
+++ b/static/js/libs/colours.js
@@ -17,13 +17,15 @@ function getColour(holder, imageSelector, imageParentSelector) {
   if (images.length > 0) {
     for (let i = 0, ii = images.length; i < ii; i += 1) {
       const parent = images[i].closest(imageParentSelector);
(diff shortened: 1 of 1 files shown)


static/js/libs/colours.js
```diff
@@ -17,6 +17,9 @@ function getColour(holder, imageSelector, imageParentSelector) {
   if (images.length > 0) {
     for (let i = 0, ii = images.length; i < ii; i += 1) {
       const parent = images[i].closest(imageParentSelector);
+      if (!parent) {
+        continue;
```
bartaz: Would be good to avoid arbitrary breaking out of the loop.
This can be easily avoided here by putting the image logic below inside of `if (parent) { ...` instead.

PR #1927, opened 2019-05-20
Reduce number of requests for getSnapCount in fsf

Lukewh:
To reduce the number of idle requests to check if a snap has been uploaded.


bartaz:
Now that we have accordion there we could probably also only send the requests when user has relevant part of accordion open.

PR #1930, opened 2019-05-20
Fix twitter card meta tags

anthonydillon:
## Done
Updated the twitter card meta tags across layouts.

## QA
- Enter the demo URL of some pages into https://cards-dev.twitter.com/validator
- Check a details page (for example, /code)


anthonydillon:
Infact, not sure this is making it any better but wanted to test it on the demo.

PR #1947, opened 2019-05-31
Show better error message on incorrect aspect ratios

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/1936
and supersedes https://github.com/canonical-web-and-design/snapcraft.io/pull/1941

### QA Steps

1. Pull the branch or open the demo service
2. Open edit listing page for a snap you own
3. Try to upload this image 1444x482.png as a Featured Banner:
1444x482
4. You should get an error message: `1941-banner.png (1444 x 482 pixels) does not have the correct aspect ratio: it needs to be 3:1 (e.g., 1446 x 482 pixels)`
5. The tests should also pass and have 100% coverage to cover:
 - A suggested image size that would exceed and dimension restrictions (it instead suggests the alternative dimension change)
 - A range of allowed aspect ratios for example when trying to use the same image for a media item `1941-banner.png (1444 x 482 pixels) does not have the correct aspect ratio: it needs to be between 1:2 and 2:1`
6. #winning

### DIFF
--- a/static/js/libs/fileValidation.test.js
+++ b/static/js/libs/fileValidation.test.js
@@ -354,7 +354,7 @@ describe("validateRestrictions", () => {
         });
 
         expect(validation.errors).toEqual([
-          "has a width (30 pixels) that is 3x its height (10 pixels). Its width needs to be between 1x and 2x the height."
+          "(30 x 10 pixels) does not have the correct aspect ratio: it needs to be between 1:1 and 2:1"
         ]);
       });
 
@@ -365,7 +365,7 @@ describe("validateRestrictions", () => {
 
         generateImage({
           naturalWidth: 10,
-          naturalHeight: 30
+          naturalHeight: 15
         });
 
         const aspectRatio = {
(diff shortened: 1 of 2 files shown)


bartaz:
It suggests size that is bigger than chosen image, and as far as I understand we expect it to suggest smaller dimensions so image can be cropped?
Screenshot 2019-06-03 at 12 44 54

bartaz:
Still couple of edge cases.

When image is very close to 3:1 ratio error message shows suggested size same as image size:

Screenshot 2019-06-03 at 14 41 04


When width is too large (compared to height) suggested size adapts height - again suggesting size that is bigger then current image:
Screenshot 2019-06-03 at 14 42 28

static/js/libs/fileValidation.test.js
```diff
@@ -379,7 +379,105 @@ describe("validateRestrictions", () => {
         });
 
         expect(validation.errors).toEqual([
-          "has a width (10 pixels) that is 0.33x its height (30 pixels). Its width needs to be between 0.5x and 0.33x the height."
+          "(10 x 30 pixels) does not have the correct aspect ratio: it needs to be between 1:2 and 1:3"
```
bartaz: 10x30 seems like it is 1:3 (so fits in range)... why do we expect it to error?

Lukewh: Yeah, I've made the logic better for this. Rather than assuming `min` and `max` aspect ratios (in the restrictions config) are always the correct way round it now checks that the ratio is between or equal to the two ratios.

therealjuan:
+1

PR #1948, opened 2019-05-31
Move all the YAML logic into helpers

jpmartinspt:
# Done
Moved all YAML helper functions to helpers and re-use them across the codebase.

# QA
- ./run
- Make sure all features that rely on YAML still work:
  - Livestream
  - Publisher pages
  - FSF
  - More snaps feature should work `http://0.0.0.0:8004/pycharm-professional`

Fixes #1948

### DIFF
--- a/webapp/first_snap/views.py
+++ b/webapp/first_snap/views.py
@@ -2,13 +2,12 @@ import os
 import flask
 import re
 from io import StringIO
-from ruamel.yaml import YAML
 
+from webapp import helpers
 from webapp.first_snap import logic
 
 YAML_KEY_REGEXP = re.compile(r"([^\s:]*)(:.*)")
 
-yaml = YAML()
 
 first_snap = flask.Blueprint(
     "fist_snap_flow",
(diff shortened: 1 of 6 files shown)


bartaz:
Something changed in a way YAML is parsed?

The function that reads and renders parts of snapcraft.yaml for FSF doesn't work same as it did before. It renders objects in `{ ... }` for some reason:
Screenshot 2019-05-31 at 15 50 09

webapp/first_snap/views.py
```diff
@@ -2,13 +2,12 @@
...
 import re
 from io import StringIO
-from ruamel.yaml import YAML
 
+from webapp import helpers
 from webapp.first_snap import logic
 
 YAML_KEY_REGEXP = re.compile(r"([^\s:]*)(:.*)")
 
-yaml = YAML()
```
bartaz: @jpmartinspt In FSF we use `YAML()` instead of `YAML(typ="safe")`. The default `YAML()` config allows to preserve original yaml file formatting that is used to render back snapcraft.yaml with correct formatting.

I guess we should use `YAML()` for all our yaml reading or allow passing a `typ` argument to `read_yaml` to override it. As I understand using `YAML(typ='rt')` is same as `YAML()`: https://yaml.readthedocs.io/en/latest/basicuse.html

jpmartinspt:
@bartaz Good catch, fixed now to use a different type for FSF.

webapp/helpers.py
```diff
@@ -1,9 +1,19 @@
...
 from ruamel.yaml import YAML
 from werkzeug.routing import BaseConverter
 
-yaml = YAML(typ="safe")
+
+_yaml = YAML(typ="rt")
+_yaml_safe = YAML(typ="safe")
+
+
+def get_yaml_loader(typ="safe"):
```
Lukewh: is there a reason this is `typ` and not `type`? I even added the e to `typ` in this comment before I proof-read

bartaz: It's same as in YAML internally. Weird, but probably best to keep it consistent with what it's used for.

jpmartinspt: `type` is a reserved keyword in python, probably the same reason the yaml lib uses `typ`.

PR #1949, opened 2019-05-31
Distro page template suru background

### LINKED ISSUE #1950: Implement distro page suru background

bartaz:
Fixes #1950 

Adds suru background with distro specific gradient to distro page.

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-1949.run.demo.haus
- go to snap distro install page for any snap (/debian and /arch are available)
- check suru background on different page widths

https://snapcraft-io-canonical-web-and-design-pr-1949.run.demo.haus/install/code/arch
https://snapcraft-io-canonical-web-and-design-pr-1949.run.demo.haus/install/slack/debian

### DIFF
--- a/templates/store/snap-distro-install.html
+++ b/templates/store/snap-distro-install.html
@@ -24,7 +24,7 @@
       "screenshot": "{{ screenshots[0] }}",
       {% endif %}
       "image": "{{ icon_url }}",
-      "operatingSystem": "linux",
+      "operatingSystem": "Linux, {{ distro_name }} ",
       "offers": {
         "price": {% if prices and prices['USD'] %}{{ prices['USD'] }}{% else %}0.00{% endif %},
         "priceCurrency": "USD"
--- a/templates/store/snap-distro-install.html
+++ b/templates/store/snap-distro-install.html
@@ -44,159 +44,150 @@
 {% endblock %}
 
 {% block content %}
-  <div class="p-strip--light is-shallow distro-background">
-    <style>
-      .distro-background {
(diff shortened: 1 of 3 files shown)


Lukewh:
The Arch version feels a bit strangely cut off and top aligned. The design in the linked to issue looks like it's got a negative top margin, which I think would benefit the Arch logo (but not necessarily the Debian logo) @hughesjordan @therealjuan thoughts?

templates/store/snap-distro-install.html
```diff
@@ -44,159 +44,150 @@
...
-        <div class="p-snap-heading__title">
-          <h1 class="p-heading--two p-snap-heading__name" data-live="title">{{ snap_title }}</h1>
-          <div class="u-hide u-show--small">
-            {% include 'store/_snap-header-information.html' %}
+      <div class="p-card">
+        <div class="row">
+          <div class="details-block">
+            <div class="p-snap-heading">
+              {% if icon_url %}
+                <img class="p-snap-heading__icon" src="{{ icon_url }}" alt="{{ snap_title }} snap" data-live="icon" />
```
Lukewh: should probably remove the `data-live` attributes

bartaz: Ah, forgot about those.

bartaz: @Lukewh on a second thought... They don't really hurt here and I plan to actually move these common parts to some partials to reuse them in both snap details and here to keep them consistent later. So the `data-live` would need to be shared in both places as well (unless we add them only based on some parameter).

PR #1956, opened 2019-06-04
Add builder blueprint

jpmartinspt:
## Done
Adds the blueprint where we will move build.snapcraft.io
functionality to, and moves the build hompage to it.

## QA
- `./run`
- Make sure `/build` is still accessible


webapp/builder/views.py
```diff
@@ -0,0 +1,16 @@
+import flask
+
+
+def builder_blueprint():
```
tbille: unless you think you think you'll pass parameters i don't think a function is really needed here.

tbille:
wanna add dumb tests to prepare the code testing in here?

PR #1970, opened 2019-06-05
If logged in, append fsf snap name with username

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/1937

## QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/first-snap
- Not logged in, the flow should work as before
- Log in
- Go through the steps as before - the snap name should at the "Package snap" step, should be suffixed with "-USERNAME" instead of "-{name}", but unlike if the name has been previously picked - you still need to "Choose snap name"


bartaz:
Could we make sure it's replaced in other steps as well?

Screenshot 2019-06-05 at 12 59 19

PR #1982, opened 2019-06-10
Add animated scroll to distro install instructions

bartaz:
Adds animated scroll to distro install instructions.

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-1982.run.demo.haus/
- go to distro page of any snap https://snapcraft-io-canonical-web-and-design-pr-1982.run.demo.haus/install/slack/arch
- click on "Install snap" button
- page should scroll smoothly to the install instructions
- there should be a small offset margin above card with instructions

### DIFF
--- a/templates/store/snap-distro-install.html
+++ b/templates/store/snap-distro-install.html
@@ -86,7 +86,7 @@
               {% include 'store/snap-details/_snap_heading_data.html' %}
 
               <div class="p-snap-install-buttons">
-                <a class="p-button--neutral p-snap-install-buttons__install" href="#install">
+                <a class="p-button--neutral p-snap-install-buttons__install js-install" href="#install">
                     Install snap
                 </a>
               </div>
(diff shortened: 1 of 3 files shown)


templates/store/snap-distro-install.html
```diff
@@ -221,6 +221,16 @@ <h4 class="p-card__title">
   <script src="{{ static_url('js/dist/public.js') }}"></script>
   <script>
     Raven.context(function () {
+      try {
+        document.querySelector(".js-install")
+          .addEventListener('click', function (event) => {
+            event.preventDefault();
+            snapcraft.public.scrollTo("#install", 300, -20);
```
anthonydillon: Do we need to scroll to this instead of jump? If this to add the offset for a sticky header?

bartaz: @anthonydillon It was a UX request to have animated scroll. Offset is so that the card that we scroll to is not sticking to the top of the window, but has nice margin.

hughesjordan:
works nicely thanks @bartaz

templates/store/snap-distro-install.html
Lukewh: Would be nice to add `#install` to the url and auto-jump to the anchor on page load.
Also could we make the "scrollTo" click event more generic?
```
try {
  snapcraft.public.scrollTo(link, {options}); //options with nice defaults, including duration and offset
} catch...
```

`scrollTo` would then:
- Figure out if link exists and fail fast
- Attach the click handler
- Figure out the element to jump to based on the `href` (and check that exists)
- Currently (if I'm reading right) `scrollTo` won't check if the `#target` exists and won't fallback to anchor tags?

kwm14:
Niiice 👍

PR #1989, opened 2019-06-12
activeDevices module refactor, updated dashboard overview graph

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/1868

## QA
- Pull the branch
- `./run`
- If you have no snaps or all your snaps have no users, the graph should show
- If you have snaps, the graph should show
- If you have a lot of snaps you should get a little loader and the snaps should be added 1 by 1
- Resizing should work
- http://0.0.0.0:8004/<snap_name>/metrics should work as before


static/sass/_snapcraft_metrics.scss
```diff
@@ -1,6 +1,7 @@
 @mixin snapcraft-metrics {
   $opaque-white: rgba(255, 255, 255, 1);
   $transparent-white: rgba(255, 255, 255, 0);
+  $semi-transparent-white: rgba(255, 255, 255, .8);
```
bartaz: Are the other 2 used at all?

static/js/libs/arrays.js
```diff
@@ -33,4 +33,39 @@ function arraysEqual(oldArray, newArray) {
...
 
-export { arraysEqual };
+/**
+ * Splits an array into chunks
+ * @param {array} arr
+ * @param {number} chunkSize
+ * @param {function} fn A function that modifies each chunk
+ * @returns {array}
+ */
+function arrayChunk(arr, chunkSize, fn) {
```
bartaz: Does `fn` serve any purpose that simple `.map` on chunked array wouldn't do?

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -3,418 +3,730 @@ import { format } from "d3-format";
...
+      .on("mouseout", this.cancelTooltip.bind(this));
+
+    this.showTooltips = true;
+
+    return this;
+  }
 }
 
-export { activeDevices as default };
+export { Graph };
```
bartaz: Does `Graph` have any "active devices specific" functionality? It seems to be used elsewhere, so maybe the file could be renamed?

bartaz: Or maybe the graph should be called ActiveDevicesGraph if that's specific implementation that shows this particular type of data.

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -3,418 +3,730 @@ import { format } from "d3-format";
...
 import { isMobile } from "../../../libs/mobile";
 import debounce from "../../../libs/debounce";
 import { sortChannels } from "../../../libs/channels";
+import { arraysMerge } from "../../../libs/arrays";
 
-function showGraph(el) {
-  el.style.opacity = 1;
-}
+class Graph {
+  constructor(holderSelector, rawData, options) {
```
bartaz: Would be nice to have a short comment with jsdoc for those.
It's hard to conclude from the code what shape data should have, what options are available...

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -3,418 +3,730 @@ import { format } from "d3-format";
...
-        item.skip = true;
-      }
-    });
+  yScaleGenerator() {
+    return scaleLinear()
+      .rangeRound([
+        this.height - this.padding.top - this.padding.bottom,
+        this.padding.top
+      ])
+      .nice()
```
bartaz: Why `xScaleGenerator` is not `.nice`? - me having no idea what `nice` does...

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -3,418 +3,730 @@ import { format } from "d3-format";
...
+      .domain([0, this.maxYValue + Math.ceil(this.maxYValue * 0.1)]);
   }
 
-  // If we've added anything to the 'other' series,
-  // add it to the array
-  if (other.value > 0) {
-    dataArr.push(other);
+  prepareSVG() {
+    // this.svg.attr("width", this.holder.clientWidth);
```
bartaz: Remove if unnecessary.

static/js/publisher/metrics/graphs/activeDevices.js
```diff
@@ -3,418 +3,730 @@ import { format } from "d3-format";
...
-    .attr("transform", `translate(${margin.left},${margin.top})`);
-
-  // Prepare the data
-  const _data = [];
-  const _keys = [];
-  let max = 0;
-  activeDevices.buckets.forEach((bucket, i) => {
-    let obj = {
-      date: utcParse("%Y-%m-%d")(bucket)
+  prepareStackedData() {
```
bartaz: Do all those methods need to be public? It's hard to discover what is meant to be a public API for `Graph` and what is just internal methods.

static/js/publisher/metrics/metrics.js
```diff
@@ -26,13 +23,37 @@ function renderMetrics(metrics) {
...
-    metrics.activeDevices.annotations
-  );
+  const graph = new Graph(metrics.activeDevices.selector, activeDevices, {
+    stacked: true,
+    area: true,
+    graphType: metrics.activeDevices.type,
+    defaultTrack: metrics.defaultTrack,
+    annotations: metrics.activeDevices.annotations
+  })
+    .prepareData()
```
bartaz: Is `prepareData` optional? Can it not be called, or called in different order?
If that's a necessary step of rendering graph (that doesn't take any additional params) why not just do it automatically in constructor or internally before `render`? Does dev need to call it intentionally?

static/sass/_snapcraft_metrics.scss
Lukewh: pff

static/js/publisher/metrics/graphs/activeDevices.js
Lukewh: "Extends the domain so that it starts and ends on nice round values." :D!
I can't find `notNice` in the docs :(

hughesjordan:
@Lukewh is there a way to only display whole numbers on the install axis?

Screenshot 2019-06-14 at 10 57 06 am

static/js/publisher/metrics/metrics.js
bartaz: Is this needed still? Seems like graph constructor calls it if there is any data(?)

Lukewh: Ah thanks, missed that

static/js/publisher/metrics/graphs/activeDevicesGraph.js
```diff
@@ -0,0 +1,736 @@
...
+      `<span class="snapcraft-graph-tooltip__title">${this.tooltipTimeFormat(
+        dateData.date
+      )}</span>`,
+      tooltipRows(dateData, currentHoverKey),
+      `</span>`,
+      `</div>`
+    ].join("");
+  }
+
+  mouseMove(d, index, nodes) {
```
bartaz: It's clear that some (if not most) of the methods are meant for private use only (`mouseMove`, `tooltipTemplate`, etc). It would be nice to have those more clearly separated from public class API.

It's really hard to tell right now which methods are meant for public API use. `enableToopltip` seems to be public, but is `cancelTooltip` public or private?

This class contains such a lot of very specific code (that depens a lot on knowing d3 and our metrics data structure) that it's really hard to understand what it does internally, but also how to use it.
Any sort of documentation (even if just in comments above public API methods) would be helpful. And maybe prefixing private methods with `_` to make them distinct and less likely to be called from outside?

therealjuan:
A few comments:

* Would it be possible to include a black line whenever you mouse over a date similarly to this example from digitalocean?

image

* The style of the graphs are too different, the ones in the dashboard are less "curvy" and also the area is empty so we should have a unified style:

image

image

Considering that might be even better to completely switch the tooltip to something more easier in the eye...

Artboard

Happy to jump in a call to discuss or if you want to take care of it in a separate PR.

Lukewh:
Ah I can't see the low churn - didn't realise the curvyness was different, will update that.

The two graphs represent different things. The metrics page is a stacked area chart to show the total amount of various versions. Whereas the dashboard is showing individual snaps (not stacked).

With 51 snaps, I find this:
Screenshot_2019-06-17  My published snaps — Linux software in the Snap Store (1)

harder to understand than this:
Screenshot_2019-06-17  My published snaps — Linux software in the Snap Store (2)

The former suggests stacked data, and that the green band is a 0.5% of a total - when in actual fact it's 50% less than the top snap.

I like the new tooltip style though. I think as we're on a feature branch let's get this dashboard graph correct, then I'll do another PR to the feature branch to update the tooltip?

PR #2002, opened 2019-06-13
Included install instructions specific for the distro here

### LINKED ISSUE #1996: Snap detail page: Include link to distro pages (to discuss)
Right now we are linking to documentation on different options to enable snap support on Linux distributions but would be good to link to the relevant instructions for specific Linux distributions somehow.

### Option A

This is an example of how it could be done:
image

## Sequence of events
1. User is interested to install skype on its system but does not have Snaps enabled
2. User chooses install Skype. One click yay!
3. User chooses its Linux distro
4. User is redirected to the relevant distro page instructions to enable snaps

### Option B
Leave as is

### Option C
Include relevant messaging as part of the snap detail page to allow users to get snaps on their system. Think of a banner within the detail pages:

image

Up for us to discuss during triaging

therealjuan:
Fixes #1996 

## QA Steps

1. Pull the branch or visit the demo URL
2. Visit any snap you want
3. Scroll down the bottom to see links to the relevant distributions you want to get snaps support
4. Visit the link
5. You should see instructions on how to install snap support on the distribution
6. #winning


static/sass/_patterns_media-object--small.scss
```diff
@@ -0,0 +1,15 @@
+@mixin p-media-object--small {
+  .p-media-object--small {
```
bartaz: I feel like we don't really need to introduce another pattern.
If all it's doing is making image smaller, we can just set image size in HTML.

Also forcing `__details` to have top margin makes it usable only with "centered" heading.

I believe same (or at least similar results) can be achieved with `.p-media-object`:
```html
<a class="p-media-object u-vertically-center" href="/install/{{package_name}}/arch">
    <img class="p-media-object__image" width="2.5rem" height="2.5rem" src="https://assets.ubuntu.com/v1/feca0fc0-Distro_Logo_ArchLinux.svg" alt="">
    <span class="p-media-object__details">
         <h4 class="p-media-object__title p-heading--five u-no-margin--bottom">Arch Linux</h4>
    </span>
</a>
```

templates/store/snap-details/_distro-instructions-for-snap-support.html
```diff
@@ -0,0 +1,109 @@
...
+        <p>Choose your Linux distribution to get detailed installation instructions. If yours is not shown, get more details on the <a href="https://docs.snapcraft.io/installing-snapd" target="_blank">installing snapd documentation</a>.</p>
+        </div>
+    </div>
+</div>
+<div class="p-strip is-shallow u-no-padding--top">
+    <div class="row" id="distros">
+        <div class="col-12 snapcraft-p-grid">
+            <a class="p-media-object--small" href="/install/{{package_name}}/arch">
+            <span class="p-media-object__image">
+                <img class="p-media-object__image" src="https://assets.ubuntu.com/v1/feca0fc0-Distro_Logo_ArchLinux.svg" alt="">
```
bartaz: `p-media-object__image` class is duplicated on span and img. I guess span is not necessary at all?

bartaz:
It's a bit weird that we put these at the bottom of the page, when install instructions are on top (under the install button).

therealjuan:
Just added your comments from the review :)

The idea of having it at the bottom of the page is that only people with no snap support enabled will benefit and look for further information down the page. We assume if you arrive to snapcraft.io you would be somewhat familiar with snaps. This also lands home snaps not being ubuntu-centric has they work in many other distributions.

The main point for doing this is to make sure they link back and forth so the missing link is to refer back to the details page from the distro instructions but that can be done in a future PR.

static/sass/styles.scss
```diff
@@ -65,6 +65,9 @@ $color-navigation-background: #252525;
 @import 'patterns_media-object--large';
 @include p-media-object--large;
 
+@import 'patterns_media-object--small';
```
bartaz: This file doesn't exist anymore, so this should fail to compile.

bartaz:
@therealjuan There is one leftover from `p-media-object--small`, but looks good otherwise!

static/sass/styles.scss
bartaz: I guess that's why the demo doesn't start...

therealjuan:
Just updated the PR and re-launched the demo service. Ready ready!

hughesjordan:
Nice 👍

bartaz:
@therealjuan Aren't the image dimensions needed: `width="2.5rem" height="2.5rem"`? They seem bigger than intended originally without it. Especially on mobile screens.

bcardiff:
Hi @therealjuan would it make sense to have an anchor to that section? I understand the rationale [mentioned here](https://github.com/canonical-web-and-design/snapcraft.io/pull/2002#issuecomment-501829508) but with that anchor, that section can be used as a menu linked from each project website I guess.

Lukewh:
I've extended on and referred back to you comment [here](https://github.com/canonical-web-and-design/snapcraft.io/issues/2010) @bcardiff

bcardiff:
Thanks @Lukewh. My concern is that `https://snapcraft.io/{{snap}}#distros` which uses an `id` of the table can work, but it does not show the title of the section, causing a loss of context. Probably it was not thought for that use initially.

Lukewh:
@bcardiff :+1:

PR #2009, opened 2019-06-14
Fix capitalisation of openSUSE

popey:
It's "openSUSE", not "OpenSUSE"


bartaz:
Monday morning and I thought I have deja vu. But no, it's different from #2008 :)

PR #2013, opened 2019-06-17
Dispute reserved name

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/1885

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/register-snap
- Try and register "bitcoin-test"
- Error should show saying the name is reserved
- Click the link in the notification
- Add some details
- SUCCESS
- High five the nearest person


bartaz:
When a reserved name is registered notification shows up telling dispute can be filled or new name registered, but the form below still has previous reserved name in it - which we know can't be registered.
Screenshot 2019-06-17 at 11 30 42

I think input should be cleared in this case to make sure another name is chosen.

Also, the only link to reserved name dispute page is in notification, which is not very visible (especially if it was visited before). Would be good to have this more prominent (with a button?)

Lukewh:
@bartaz I've just followed the same pattern as the already registered flow. Perhaps both of these could be revisited in a separate issue/ PR?
Screenshot_2019-06-17 Snapcraft - Snaps are universal Linux packages

tests/publisher/tests_reserved_name_dispute.py
```diff
@@ -0,0 +1,136 @@
+import responses
```
jpmartinspt: I think some of this tests already existed in `tests/publisher/test_register_name_dispute.py`?

therealjuan:
This works as expected. Tested with a combination of reserved and taken names and work as expected.

PR #2017, opened 2019-06-17
Add build github oauth flow

jpmartinspt:


webapp/builder/views.py
```diff
@@ -8,4 +10,40 @@
...
+    if not logic.valid_auth_secret(flask.session, secret):
+        # Secret does not match our copy warn about potential CSRF attack
+        raise GithubAuthException(
+            "Returned shared secret does not match generated shared secret"
+        )
+
+    code = flask.request.args.get("code")
+    response = logic.exchange_code_for_token(flask.request, code)
+
+    data = response.json()
```
tbille: it already happened that we got an invalid json. You should surround it with try catch ValueError (if I remember well)

jpmartinspt: JsonDecodeError?

webapp/builder/logic.py
```diff
@@ -0,0 +1,66 @@
+import os
+import secrets
+import urllib.parse
+
+import requests
```
tbille: `from webapp import api` here is our custom request. Where you can do: `api.requests`

and could you register the Session that you use in talisker: `talisker.requests.configure(YOUR_SESSION)` this will allow us to get more info iin the logs about the api calls that we are making

webapp/builder/views.py
```diff
@@ -8,4 +10,40 @@
...
+    flask.session["github_auth"] = True
+    flask.session["github_token"] = data["access_token"]
+
+    flask.flash("Authenticated with GitHub", "positive")
+    return flask.redirect(flask.url_for("builder.index"))
+
+
+@builder.errorhandler(GithubAuthException)
+def github_auth_exception_handler(e):
+    flask.flash(f"{e}", "negative")
```
tbille: should we send an error to sentry?

jpmartinspt:
@tbille Added some tests and refactored a bit.

PR #2018, opened 2019-06-18
Releases common version

### LINKED ISSUE #1959: Release UI: Detect shared version in channel across archs
- [ ] If all archs from a channel include the same version name the version will be removed from the revision cell and added below the channel / set name

- [ ] Show multiple versions once there are more than 1 different version within the set

bartaz:
Fixes #1959 

Shows version under channel name in releases table.

### QA

- ./run or https://snapcraft-io-canonical-web-and-design-pr-2018.run.demo.haus/
- go to publisher releases page of any snap
- under channel name in each row version should be displayed:
  - version number if all revisions in this channel have same version
  - "Multiple versions" if revisions have different versions
  - nothing if channel has no revisions
- if all revisions have same version in channel versions are not shown in individual cells
- hovering over channel name/version shows tooltip with version breakdown per arch
- this all applies only to 4 channel rows, not to 'available revisions' row at the bottom

Screenshot 2019-06-18 at 13 01 11

### DIFF
--- a/static/js/publisher/release/components/releasesTableCell.js
+++ b/static/js/publisher/release/components/releasesTableCell.js
@@ -25,7 +25,7 @@ class ReleasesTableCell extends Component {
     this.props.undoRelease(revision, `${track}/${risk}`);
   }
 
-  renderRevision(revision, isPending) {
+  renderRevision(revision, isPending, showVersion) {
     return (
       <Fragment>
         <span className="p-release-data__info">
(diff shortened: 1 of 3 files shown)


therealjuan:
Comments:

* Add tooltip to the version below the channel name if they share the same version:
image
* Ideally we would compile all versions and architectures when there are multiple versions:
image

* There will be a follow-up PR moving the selector to a different area

therealjuan:

PR #2019, opened 2019-06-18
Drag'n'drop to promote channel

### LINKED ISSUE #1961: Release UI: Drag'n'drop basic implementation (library, POC...)
- [x] install react-dnd and any necessary dependencies
- [x] make a test implementation of some draggable component (available revisions?)

### LINKED ISSUE #1962: Release UI: Promote from available revisions to a specific channel

### LINKED ISSUE #1963: Release UI: Promote from channel to channel

bartaz:
Fixes #1961 
Fixes #1962 
Fixes #1963 

Adds drag'n'drop to releases page.
Channels can now be promoted using drag'n'drop.

### QA

This is new quite complex feature. Please make sure to QA on different snaps (different number of architectures, devmode revisions, closed channels, etc) and in different browsers (it uses native drag'n'drop, so browser implementations may differ).

- ./run or https://snapcraft-io-canonical-web-and-design-pr-2019.run.demo.haus/
- go to releases page or any snap (many snaps ideally)
- try to promote some stuff using drag'n'drop
- make sure that things that can be normally promoted can also be promoted with drag'n'drop
- make sure that things that can't be normally promoted (empty channel, devmode) can't be promoted with drag'n'drop
- click on things, try drag'n'drop in different UI states

Screenshot 2019-06-25 at 16 35 31

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,8 @@
     "prop-types": "^15.6.2",
     "raven-js": "^3.24.1",
     "react": "^16.4.1",
+    "react-dnd": "^8.0.0",
+    "react-dnd-html5-backend": "^8.0.0",
     "react-dom": "^16.4.1",
     "react-redux": "^5.1.1",
     "react-sortable-hoc": "^1.8.3",
--- a/static/js/publisher/release.js
+++ b/static/js/publisher/release.js
@@ -3,8 +3,10 @@ import ReactDOM from "react-dom";
 import { createStore, applyMiddleware, compose } from "redux";
 import thunk from "redux-thunk";
 import { Provider } from "react-redux";
-import ReleasesController from "./release/releasesController";
+import { DndProvider } from "react-dnd";
+import HTML5Backend from "react-dnd-html5-backend";
 
+import ReleasesController from "./release/releasesController";
 import releases from "./release/reducers";
 
 // setup redux store with thunk middleware and devtools extension:
@@ -22,12 +24,14 @@ const initReleases = (id, snapName, releasesData, channelMapsList, options) => {
 
   ReactDOM.render(
     <Provider store={store}>
-      <ReleasesController
-        snapName={snapName}
-        channelMapsList={channelMapsList}
-        releasesData={releasesData}
-        options={options}
-      />
+      <DndProvider backend={HTML5Backend}>
+        <ReleasesController
+          snapName={snapName}
+          channelMapsList={channelMapsList}
+          releasesData={releasesData}
+          options={options}
+        />
+      </DndProvider>
     </Provider>,
     document.querySelector(id)
   );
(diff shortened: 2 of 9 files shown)


therealjuan:
A few comments:

* Can we use a pointer icon instead of the cross? 
image

* Would it be possible to make the whole cell draggable?
* There should be a separate issue for moving the promote actions as part of the new options (cog) area instead of just the bin?
* I will probably have a review on the visuals with Mr. @kwm14

bartaz:
Thanks for comments @therealjuan 

> * Can we use a pointer icon instead of the cross?

We can, but it's most common for click interaction, not drag. 
But there are some dnd related [cursors in CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#Values) we may try (not sure about their browser compatibility, but we can try and have fallback to "move" cross):

Screenshot 2019-06-19 at 20 50 08

> * Would it be possible to make the whole cell draggable?

It would I guess, but it's get tricker, because we have interactive elements in it (things that have tooltips, things that are clickable), and I'm not sure how they will work together.
Tooltips are most annoying, because if you start drag with tooltip open it will likely be visible during drag, etc. We will need to find some workarounds for that.
Also when we lated make revision cells draggable: they are clickable as a whole, they have tooltip as a whole, they may have a 'cancel' icon in it that is clickable... and we will make it draggable as a whole...
And all those elements are actually both draggable and drop targets at the same time (which is a bit tricky by itself).

So... long story short - I guess it may be possible, but complicates the implementation and may lead to even more weird cases to work around.

> * There should be a separate issue for moving the promote actions as part of the new options (cog) area instead of just the bin?

Ideally yes - we can do it before or later the drag and drop is finished.

therealjuan:
Some comments to discuss tomorrow:

* Once you drag a channel into another one, might be good to highlight as well the channel that you end up promoting to
image
* Could we use grab instead of the cursor to drag? Similarly to what happens with a message on gmail
* Can we update the highlighted colour to use `rgb(250, 230, 190)`?

bartaz:
> * Once you drag a channel into another one, might be good to highlight as well the channel that you end up promoting to

Done. If all revisions in given channel are promoted, channel will be highlighted as well.

> * Could we use grab instead of the cursor to drag? Similarly to what happens with a message on gmail

Turns out to be hard. We can change the cursor of the handle before it's dragged, but once dragging starts it's browser that controls cursor and we have little control over it. To fix that we would need to create [custom implementation of drag'n'drop](https://github.com/react-dnd/react-dnd/issues/325) functionality which is probably not worth the cost just to change the cursor.
> * Can we update the highlighted colour to use `rgb(250, 230, 190)`?
Done.

Lukewh:
Not sure if there's anything we can do about it, but letting go makes the dragged element reset it's drag centre to the top left of the element.
Peek 2019-06-26 09-12

Lukewh:
If you click to select a different revision for an architecture, then drag, should we de-obscure the main table?
Peek 2019-06-26 09-16

bartaz:
> Not sure if there's anything we can do about it, but letting go makes the dragged element reset it's drag centre to the top left of the element.

I don't see it happening (both in Chrome or Firefox). So this may be some native issue. Not sure if we have any control on that.

therealjuan:
Desk review done. Looking juanderful!

PR #2023, opened 2019-06-18
Move tracks dropdown to left side of heading

### LINKED ISSUE #2021: Release UI: Move track selector to end of heading
To make sure the selectors work similarly, we can move the track selector to the end of the heading:

`Revisions available to install in [TRACK_NAME]`

image

The only caveat is when there is only one track we should only see:
`Revisions available to install`
image

bartaz:
Fixes #2021

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-2023.run.demo.haus
- go to releases page of any snap with multiple tracks (https://snapcraft-io-canonical-web-and-design-pr-2023.run.demo.haus/skype/releases)
- track dropdown should be part of the heading on left side
- track dropdown should work as before
- check snap without tracks to see if they are not affected

Screenshot 2019-06-18 at 13 43 21

### DIFF
--- a/static/js/publisher/release/components/releasesHeading.js
+++ b/static/js/publisher/release/components/releasesHeading.js
@@ -1,4 +1,4 @@
-import React, { Component } from "react";
+import React, { Component, Fragment } from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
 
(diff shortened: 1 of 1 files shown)


therealjuan:
Can we add a little more margin on the left? :)

image

bartaz:
@therealjuan Talk to Vanilla guys ;) That's how vanilla inline forms are displayed, I'd prefer not to add custom styles unless necessary.

therealjuan:
How about just adding a simple space at the end of the heading? 

P.S. Vanilla guys are sat next to me.I'm scared

bartaz:
@therealjuan There is a space there already (without it they are just glued together), but we may try to force another one...

kwm14:
Vanilla guy here 👋  what version of Vanilla are we using?

In 2.0 we have a base value of `1rem` between components which will give you the space desired.
Screenshot 2019-06-18 at 15 51 53

bartaz:
@kwm14 We are on 1.7.1 here still ;) So I guess when we migrate to 2.0 we will want it closer than Vanilla does 🤣

PR #2014, opened 2019-06-17
Removed logo in fedora according to guidelines

therealjuan:
Fixes #2007 

### QA Steps

1. Pull the branch or open the demo service
2. Open any distro pages with Fedora instructions like `/install/spotify/fedora`
3. Check the logo is not visible in the background according to [Fedora logo guidelines](https://fedoraproject.org/wiki/Logo/UsageGuidelines)
4. Visit another distro pages like openSuse `/install/spotify/opensuse`
5. Notice the logo should be there
6. #winning

PR #2024, opened 2019-06-18
White logos in distros

therealjuan:
Attempt to fix the problem with #2014 by making use of white version of logos

### QA Steps

2. Visit any distro page (i.e. `/install/skype/mint` or `/install/skype/arch`)
3. Check the logo on the top right is white
4. Notice the white logo adapts between mobile and desktop
5. #winning

### DIFF
--- a/static/sass/_patterns_distro_banner.scss
+++ b/static/sass/_patterns_distro_banner.scss
@@ -56,8 +56,8 @@
   .distro-banner__logo {
     filter: brightness(40);
     float: right;
-    margin-top: 3rem;
-    width: 9rem;
+    margin-top: 1.5rem;
+    width: 5rem;
   }
 
   @media only screen and (min-width: $breakpoint-medium) {
@@ -67,10 +67,15 @@
     }
   }
 
+  @media only screen and (min-width: $breakpoint-large) {
+    .distro-banner__logo {
+      width: 9rem;
+    }
+  }
+
   @media only screen and (max-width: $breakpoint-medium) {
     .distro-banner__logo {
-      margin-top: 1.5rem;
-      width: 6rem;
+      display: none;
     }
   }
 }
--- a/static/sass/_patterns_distro_banner.scss
+++ b/static/sass/_patterns_distro_banner.scss
@@ -60,8 +60,4 @@
     }
   }
 
-  .distro-banner__logo {
-    filter: grayscale(100%) opacity(20%);
-    mix-blend-mode: multiply;
-  }
 }
--- a/static/sass/_patterns_distro_banner.scss
+++ b/static/sass/_patterns_distro_banner.scss
@@ -53,6 +53,13 @@
     right: 2rem;
   }
 
+  .distro-banner__logo {
+    filter: brightness(40);
+    float: right;
+    margin-top: 3rem;
+    width: 9rem;
+  }
+
   @media only screen and (min-width: $breakpoint-medium) {
     .distro-code-snippet {
       // to align code snippet with paragraph of text in column next to it
@@ -60,4 +67,10 @@
     }
   }
 
+  @media only screen and (max-width: $breakpoint-medium) {
+    .distro-banner__logo {
+      margin-top: 1.5rem;
+      width: 6rem;
+    }
+  }
 }
--- a/static/sass/_patterns_distro_banner.scss
+++ b/static/sass/_patterns_distro_banner.scss
@@ -55,7 +55,6 @@
 
   .distro-banner__logo {
     float: right;
-    margin-top: 1.5rem;
     width: 7rem;
   }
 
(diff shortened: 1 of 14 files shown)


bartaz:
Not sure if we want to show logos on small screens at all, they may overlap with text:

Screenshot 2019-06-18 at 13 58 18

kwm14:
## Design review
Looking good @therealjuan just a couple of comments:

- [ ] update `margin-top: 1.5rem` on **/debian**, currently, the logo sits on the baseline compared to all the other distro logos which sit vertically center

Screenshot 2019-06-18 at 13 55 08

- [ ] Remove logo on mobile a suggested by @bartaz above, as it overlaps when viewing a long snap name 😉 

Screenshot 2019-06-18 at 13 58 25

therealjuan:
Just changing the sizes to be different depending on the breakpoint so they are consistent @kwm14 @bartaz 

Re-review all the tings!

static/sass/_patterns_distro_banner.scss
```diff
@@ -53,15 +53,29 @@
     right: 2rem;
   }
 
+  .distro-banner__logo {
+    filter: brightness(40);
```
bartaz: Why do we need brightness filter for logos that are already white?

kwm14:
@therealjuan your fix for **/debian** has caused other distro logos to reach for the top 😮 #notwinning #seaofgreen

Screenshot 2019-06-18 at 14 59 26
Screenshot 2019-06-18 at 15 00 25
Screenshot 2019-06-18 at 15 00 56

static/sass/_patterns_distro_banner.scss
therealjuan: Leftover! Will remove :)

therealjuan:
Thanks for bringing this into our attention @kwm14 

I have updated all SVG logos so they have the same dimensions as this will help with positioning elements in space.

templates/store/snap-distro-install.html
```diff
@@ -68,7 +68,7 @@
       <div class="row">
         <div class="mobile-col-2 col-8"></div>
         <div class="mobile-col-2 col-4">
-          <img class="distro-banner__logo" src="{{ distro_logo }}" />
+          <img class="distro-banner__logo" src="{{ distro_logo_mono }}" />
```
bartaz: Now that this logo is not in 'background' anymore we could possibly move it out of here completely and have it next to heading in HTML. This may help with position it in more standard way in grid.

So I would move this whole `<div class="mobile-col-2 col-4">` with logo next to `<h1>` element. `<h1>` would need to be wrapped in `col-8` as well. So something like:

```html
<div class="row">
  <div class="col-8">
     <h1 class="p-heading--two u-no-margin is-light">How to install {{ snap_title }}<br/> on <b>{{ distro_name }}</b></h1>
  </div>
  <div class="col-4 mobile-col-2">
    <img class="distro-banner__logo" src="{{ distro_logo_mono }}" />
 </div>
</div>
```

bartaz: And we could use `u-vertically-center` then to make sure heading and logo are centrally aligned `<div class="row u-vertically-center">`

kwm14:
@therealjuan as we need to use Fedora's official logo. Looking at their color palette they have a range of colours, the majority which won't work as a background color.

But if we use a value from their [Other official colors](https://fedoraproject.org/wiki/Logo/UsageGuidelines#Colors) it could work if they approve...

Screenshot 2019-06-18 at 16 25 08

templates/store/snap-distro-install.html
```diff
@@ -65,16 +65,17 @@
...
-      </div>
     </div>
     <div class="p-strip">
       <div class="row">
-        <h1 class="p-heading--two u-no-margin is-light">How to install {{ snap_title }}<br/> on <b>{{ distro_name }}</b></h1>
+        <div class="col-8">
+           <h1 class="p-heading--two u-no-margin is-light">How to install {{ snap_title }}<br/> on <b>{{ distro_name }}</b></h1>
+        </div>
+        {% if distro_logo_mono %}
+        <div class="col-4 mobile-col-2">
```
bartaz: Great! Thanks for moving it here.

2 small things we can inprove:
- `mobie-col-2` is not needed (we hide it on mobile)
- to hide it on mobile we can use `u-hide--small` class and avoid custom styles

static/sass/_patterns_distro_banner.scss
```diff
@@ -53,15 +53,21 @@
...
       padding-top: .5rem;
     }
   }
 
-  .distro-banner__logo {
-    filter: grayscale(100%) opacity(20%);
-    mix-blend-mode: multiply;
+  @media only screen and (max-width: $breakpoint-medium) {
+    .distro-banner__logo {
+      display: none;
```
bartaz: To hide it on mobile we can use `u-hide--small` class and avoid custom styles.

kwm14:
Design updates LGTM 👍

PR #2030, opened 2019-06-20
FSF - pagination as buttons

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/1969

## QA

- Pull branch
- `./run`
- Visit http://0.0.0.0:8004/first-snap
- Go through the flow with both linux and macos
- Ensure everything works as expected
- Yay!

### DIFF
--- a/templates/first-snap/_layout_fsf.html
+++ b/templates/first-snap/_layout_fsf.html
@@ -10,11 +10,9 @@
   </div>
 
   {% if self.fsf_pagination() %}
-  <div class="p-strip u-no-padding snapcraft-banner-background">
+  <div class="p-strip u-no-padding">
     <div class="row">
-      <div class="p-pagination">
         {% block fsf_pagination %}{% endblock %}
-      </div>
     </div>
   </div>
   {% endif %}
--- a/templates/first-snap/_layout_fsf.html
+++ b/templates/first-snap/_layout_fsf.html
@@ -10,10 +10,8 @@
   </div>
 
   {% if self.fsf_pagination() %}
-  <div class="p-strip u-no-padding">
-    <div class="row">
-        {% block fsf_pagination %}{% endblock %}
-    </div>
+  <div class="row">
+    {% block fsf_pagination %}{% endblock %}
   </div>
   {% endif %}
 {% endblock %}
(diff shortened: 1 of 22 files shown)


kwm14:
## Design review
Looking good @Lukewh only have one comment which might be for @therealjuan to answer 😃 

- [ ] Visually looks odd to have both '**Chevron**' and '**Previous/Next**' inside the button, feel like we only need to have one

### Current
Screenshot 2019-06-20 at 11 30 25

### Proposed
**Previous/Next**
Screenshot 2019-06-20 at 11 24 49

**Chevron**
Screenshot 2019-06-20 at 11 25 30

My preference is the chevron. Looks cleaner 😉 thoughts?

Lukewh:
@kwm14 I quickly added the chevrons before pushing! My worry with just chevrons is it's not as accessible.

therealjuan:
I believe that having the buttons already positioned left and right is already suggesting previous and next and having the cheveron as well, so including the text inside the button is like a third layer of the same meaning. Probably we can do with just only 2 of them: Chevron + Button position.

kwm14:
Thanks, @Lukewh updates LGTM 👍

therealjuan:

bartaz:
Possibly something for separate task, but maybe the 'Continue' button on very first step could get the chevron (and maybe change in position) as well?

Screenshot 2019-06-21 at 14 12 57

templates/first-snap/_layout_fsf.html
```diff
@@ -10,11 +10,9 @@
   </div>
 
   {% if self.fsf_pagination() %}
-  <div class="p-strip u-no-padding snapcraft-banner-background">
+  <div class="p-strip u-no-padding">
```
bartaz: If pagination now doesn't need snapcraft background I guess it doesn't need to be in own strip anymore?
Maybe moving it together with content into one strip would allow to get rid of some `u-no-padding` workarounds?

hughesjordan:
@Lukewh not sure if it's just because this is a demo but the final 'Go to listing' button within the 'Publish to store' step still appears to be clickable when in a disabled state giving a 404 error.

hughesjordan:
@Lukewh looks good 👍🏻

PR #2043, opened 2019-06-26
Disable ```fences``` code blocks in preview

### LINKED ISSUE #2033: Using multi line code blocks in a listing description is inconsistent between the preview and actual page
### Expected behaviour

I expected the preview to show the same as the actual page (not to format multi line code blocks)

1) Create a listing description which has a multi line code block (multiple lines wrapped by triple backticks)
Screenshot from 2019-06-20 23-26-52

2) Click preview and notice how it formats the block as a code block
Screenshot from 2019-06-20 23-27-29

3) Now click save and view the actual store page, notice how this does not format the code block
Screenshot from 2019-06-20 23-28-04

I understand that the store page uses a subset of markdown (which does not support multi line code blocks), so it is the preview which is wrong here and should format in the same way as the actual page ?

### Specs

 - Browser: Firefox 67

bartaz:
Fixes #2033

### QA
- run or https://snapcraft-io-canonical-web-and-design-pr-2043.run.demo.haus/
- go to listing page of any snap
- edit description to contain code block using fences ```
- preview and make sure code block is not rendered as code block
- use code block using 4 spaces indent
- preview to make sure spaces indented code block is rendered as code block
- make sure inline code in ticks ` works

### DIFF
--- a/static/js/publisher/preview.js
+++ b/static/js/publisher/preview.js
@@ -23,7 +23,8 @@ const md = new MarkdownIt({
   "escape",
   "strikethrough",
   "image",
-  "html_inline"
+  "html_inline",
+  "fence"
 ]);
 
 // For the different elements we might need to change different properties


bartaz:
This disables ``` code blocks in preview, but there still is a little difference in how code blocks are rendered by serverside vs clientside markdown:

Screenshot 2019-06-26 at 09 06 08

With ``` code blocks disabled client side markdown starts to intepret those as inline `code` while server side ignores them completely.

We could try workaround that by disabling `backticks` rule in client side markdown, but this disables inline code as well, so we would need to reimplement it ourselves - which is not trivial (trying to match single backticks while ignoring tripple backticks is not a simple regexp to write).

I think we are getting to a limit of how similar we can render markdown on client and server using 2 completely different markdown implementations. So I guess we will have to live with some inconsistencies.
The only way to have really same markdown in preview would be to use server side markdown rendering (either by rendering whole page on serverside and reload preview on each change or by having a small markdown preview endpoint API that would render description on request).

PR #2045, opened 2019-06-27
Release single revision using drag'n'drop

### LINKED ISSUE #1964: Release UI: Promote within a channel to another channel (same arch)

bartaz:
Fixes #1964

### QA

- ./run or https://snapcraft-io-canonical-web-and-design-pr-2045.run.demo.haus/
- individual revision cells in releases table should have a drag handle
- you should be able to drag the revision to other channel in same architecture
- you should not be able to drag revision to different architecture
- you should not be able to drag empty cells (with no revisions)
- you should not be able to drag devmode revisions to stable/candidate
- make sure dragging whole channel still works as before
- try with different snaps in different browsers

Screenshot 2019-06-27 at 13 50 48

### DIFF
--- a/static/js/publisher/release/components/releasesTableCell.js
+++ b/static/js/publisher/release/components/releasesTableCell.js
@@ -234,6 +234,9 @@ const ReleasesTableCell = props => {
         ref={drag}
         className="p-release-data p-tooltip p-tooltip--btm-center"
       >
+        <span className="p-release-data__handle">
+          <i className="p-icon--drag" />
+        </span>
         {isChannelPendingClose ? (
           <CloseChannelInfo />
         ) : currentRevision ? (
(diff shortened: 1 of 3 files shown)


Lukewh:
Could we prevent tooltips on dragging - I seemed to hit this weird bug a couple of times, this first time without trying to replicate it... Replicating it seemed harder than it appearing randomly.

Peek 2019-06-27 13-05

therealjuan:
Would it make sense to just include a tooltip on a cell if the version is different from the rest or if the revision is in devmode?

bartaz:
@Lukewh @therealjuan 
The issue mentioned by Luke seems to be some Firefox weirdness. I made it better, but still may happen in some edge cases. It would require some additional work around drag'n'drop to fix it. If we see it happening a lot we can try to do it separately.

If we want to reconsider when to show tooltips let's discuss it separately as well. We changed a bit what we show it the table, so it may be worth revisiting the tooltips as well.

therealjuan:
Sounds good to me! +1d

bartaz:
If only I didn't break styles for devmode revisions...

Screenshot 2019-06-28 at 09 57 43

Lukewh:
Screenshot_2019-06-28  Releases and revision history for lukewh-test-snap
shame vanilla icons don't align... But it works again :+1:

PR #2051, opened 2019-06-28
Set and remove default tracks

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/1966
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/1967

**NOTE: You will need a snap with tracks to QA this. Your best option is to set your local env to use staging and I'll add you as a collaborator on a snap I have with tracks.**

## QA
- Pull the branch
- `./run`
- Change track from latest
- Click the "Set as default track" button
- A modal should appear with some text explaining what's going to happen
- Click "Set {track} as default track"
- The default track should be updated (you should see a notification)
- Reload the page to ensure the change has taken effect
- Click "Clear default track"
- Follow similar steps as before
- Do some more clicking and setting/ unsetting


static/js/publisher/release/releasesController.js
```diff
@@ -60,6 +63,39 @@ class ReleasesController extends Component {
     this.props.updateReleases(releasesData.releases);
   }
 
+  fetchSetDefaultTrack(track) {
```
bartaz: Ideally we should avoid adding any new functionality to the controller as this was mainly used in pre-redux implementation.

Fetches can be done inside actions. This way they can be triggered by any component without needing to pass reference to this handler via props through components tree.

Lukewh: :+1: I did it separate to start with, but wanted to follow the pattern already there. Will move it back out :D

bartaz: Thanks!

static/js/publisher/release/components/defaultTrackModifier.js
```diff
@@ -0,0 +1,220 @@
...
+
+    openModal({
+      title: "Clear default track",
+      content: `By clearing the default track, any device that installed the snap with ${
+        defaultTrack.track
+      } track specified will get updates from the latest track. Would you like to proceed?`,
+      actions: [
+        {
+          className: "p-button--positive u-no-margin--bottom u-float--right",
+          onClick: this.confirmClearDefaultTrackHandler,
```
bartaz: This feels a bit overkill to pass classNames etc through redux state. At least the styling bits should be handled by modal component hopefully.

But most certainly we shouldn't pass functions through redux and keep them as part of the store. Store is meant to be serialisable and immutable at any point, functions are not.

In case of `closeModal` callback, it seems that all we do is calling `closeModal` action - I guess modal can do it itself, it's not this component that needs to handle this?
It may be more tricky in case of more specific handlers like `confirmClearDefaultTrackHandler`. But maybe in their case we could also pass a redux action object (`{ type: "ACTION", payload: { ... }}` rather than function?

static/js/publisher/release/components/defaultTrackModifier.js
```diff
@@ -0,0 +1,220 @@
...
+  handleCloseModal() {
+    const { closeModal } = this.props;
+
+    closeModal();
+  }
+
+  confirmDefaultTrackHandler() {
+    const { currentTrack, setDefaultTrack, csrfToken, snapName } = this.props;
+
+    setDefaultTrack(snapName, csrfToken, currentTrack);
```
bartaz: It seems that all the params for `setDefaultTrack` are already in the store, so the action creator function could access them via `getState`. No need to pass them through props and to action.

So I guess we could have `setDefaultTrack()` - that sets current track as default and `resetDefaultTrack` - that sets it to null.

static/js/publisher/release/actions/modal.js
```diff
@@ -0,0 +1,15 @@
+export const OPEN_MODAL = "OPEN_MODAL";
+export const CLOSE_MODAL = "CLOSE_MODAL";
+
+export function openModal(payload) {
+  return {
+    type: OPEN_MODAL,
+    payload: { payload }
```
bartaz: This causes nested `payload` object, so action looks like: `{ type: "OPEN", payload: { payload: { ... } }` and as a result reducer also contains `{ isOpen: true, payload: { ... }}` when it should probably contain modal props directly.

static/js/publisher/release/releasesController.js
```diff
@@ -262,23 +269,21 @@ class ReleasesController extends Component {
...
       <Fragment>
         <div className="row">
-          {this.state.error && (
-            <Notification status="error" appearance="negative">
-              {this.state.error}
-            </Notification>
-          )}
-          <ReleasesHeading />
+          {visible && <Notification />}
+          <ReleasesHeading defaultTrack={this.state.defaultTrack} />
```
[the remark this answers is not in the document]
bartaz: More so, it doesn't seem like ReleasesHeading needs this prop at all.

static/js/publisher/release/components/notification.js
```diff
@@ -43,13 +45,25 @@ export default class Notification extends Component {
 }
 
 Notification.propTypes = {
-  children: PropTypes.node,
+  content: PropTypes.node,
```
bartaz: These are breaking changes.
This component is used in other places (like RevisionsList component) when it uses `children`.
Currently it doesn't display at all and throws errors in console.

To see it you need to look at snap with some devmode revisions in revisions history. Notification should appear warning about devmode revisions.

bartaz: I think we need 2 components:
1st that is previous implementation of purely visual notification - so any component can render it anywhere.
2nd is "global" notification connected to redux store that takes content from the store and is shown/hidden by actions on the store.

static/js/publisher/release/components/defaultTrackModifier.js
```diff
@@ -0,0 +1,196 @@
...
+  componentDidUpdate(prevProps) {
+    const {
+      defaultTrack,
+      currentTrack,
+      closeModal,
+      showNotification,
+      snapName
+    } = this.props;
+    if (defaultTrack.track !== prevProps.defaultTrack.track) {
+      closeModal();
```
bartaz: Why does it happen here?

We have `setDefaultTrack` action. It could dispatch any actions on success - including closing modal and even more accurately success messages (that should be direct results of fetch, not this component).

static/js/publisher/release/components/modal.js
```diff
@@ -0,0 +1,147 @@
...
+};
+
+const mapActionButtonDispatchToProps = dispatch => {
+  const dispatchObj = {
+    dispatch
+  };
+
+  Object.keys(actions).forEach(actionKey => {
+    if (typeof actions[actionKey] === "function") {
+      dispatchObj[actionKey] = () => dispatch(actions[actionKey]());
```
bartaz: While it's nice that in theory any action can be dispatched this way it quickly fails because most of actions creator functions take arguments which are not passed here in any way.
So this will only work for actions that don't take any arguments.
Because of that it may be better to explicitly list only actions that are supported.

static/js/publisher/release/reducers/options.js
```diff
@@ -0,0 +1,3 @@
+export default function(state = {}) {
```
bartaz: Would be good to name this function for consistency. And some comment would be helpful to mention that currently these are only set in initial state.

static/js/publisher/release/actions/defaultTrack.test.js
```diff
@@ -0,0 +1,69 @@
...
+        const store = mockStore({
+          options: {
+            snapName: "test",
+            csrfToken: "test"
+          },
+          currentTrack: "test",
+          defaultTrack: "nope"
+        });
+
+        return store.dispatch(setDefaultTrack()).then(() => {
```
bartaz: Awesome, thank you! 👍

static/js/publisher/release/actions/globalNotification.test.js
```diff
@@ -0,0 +1,28 @@
...
+      });
+    });
+  });
+
+  describe("closeNotification", () => {
+    it("should create an action to hide the notification", () => {
+      const hideNotificationAction = hideNotification();
+
+      expect(hideNotificationAction.type).toBe(HIDE_NOTIFICATION);
+      expect(hideNotificationAction.paylaod).toBeUndefined();
```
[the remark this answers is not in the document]
Lukewh: Good catch :D

static/js/publisher/release/components/notification.js
```diff
@@ -51,5 +56,8 @@ Notification.propTypes = {
     "information"
   ]),
   status: PropTypes.string,
-  onRemoveClick: PropTypes.func
+  canDismiss: PropTypes.bool,
+  hideNotification: PropTypes.func
```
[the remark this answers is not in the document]
Lukewh: Removing the boolean means that on any globalNotification action the `hideNotification` function will always be provided - meaning you wouldn't be able to create a persistent globalNotification via redux... Which is probably ok.

webapp/publisher/snaps/views.py
```diff
@@ -699,6 +700,41 @@ def post_close_channel(snap_name):
     return flask.jsonify(response)
 
 
+@publisher_snaps.route("/<snap_name>/releases/default-track", methods=["POST"])
+@login_required
+def post_default_track(snap_name):
+    data = flask.request.json
+
+    if not data:
+        return flask.jsonify({})
```
jpmartinspt: We could return a `400`.

static/js/publisher/release/components/notification.js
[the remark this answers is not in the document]
bartaz: Hmmm... There can be `canDismiss` in redux state if we need it. And then `GlobalNotification` component may pass `hideNotification` prop to Notification component only when it's set to true?

It's not a big deal, if that makes things complicated lets leave it as it is.

static/js/publisher/release/components/modal.js
Lukewh: we could pass an `actionArguments` array? But we run the risk of those arguments having functions etc. passed through, then we need to check those etc. etc... I think a whitelist is a good idea that can be added to as needed

static/js/publisher/release/components/modal.js
```diff
@@ -0,0 +1,147 @@
...
+        // If an action is passed, perform the specific action
+        this.props[onClickAction.reduxAction]();
+      }
+    } else {
+      // Otherwise dispatch the action object
+      dispatch(onClickAction);
+    }
+
+    this.setState({
+      loading: true
```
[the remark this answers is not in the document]
Lukewh: The loading is specifically for the button (not the entire modal) so we'd probably want to add another action & reducer.
We'd have to call it in a similar way, unless we added a dispatch to every action that *could* trigger the state change. But then we might get some odd behaviour like the "Submit" button turning into a "Loading..." button because some background fetch/ state change triggered the modal button state change.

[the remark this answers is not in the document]
bartaz: OK. Right. Button would need to know what is loading.
Let's not complicate it and leave as it is.

static/js/publisher/release/components/modal.js
bartaz: We don't need a full blown solution. It's safe to assume that for now we can only trigger actions without arguments from modal. At least as functions. You can always pass an action object, which should cover most cases anyway.
So I would just explicitly list only `setDefaultTrack` and `clearDefaultTrack`.

hughesjordan:
@therealjuan @Lukewh  Looking at this in action I'm wondering whether or not we need to have both behaviours of setting and clearing default track? Is the concept just as easily understood with the 'Set as default track' button on all tracks other than the default? (Maybe there's a reason I'm missing, less clicks?)

In any case, if it's beneficial to have the 'clear', should we not also have the 'Set as default track' button available when viewing the latest track and another is set as the default?

Lukewh:
@hughesjordan from the spec:
"Setting the default track to latest is forbidden, as it doesn't add any particular value and it may be confused with turning off the feature. The subtle but important difference between setting the default to track to latest and unsetting it is that the former makes the client to stick to latest (if it requests somerisk will receive latest/somerisk and will get stick to it, so further refreshes will ask for latest/somerisk), while the later be just the normal behaviour (if it requests somerisk will receive latest/somerisk, but further refreshes will ask for somerisk, no track specified, which would allow future changes in default tracks or aliases or similar features). Note that unsetting a default track for a snap is done by setting it to null through the normal setting endpoint."

hughesjordan:
@Lukewh thanks makes sense 👍

PR #2052, opened 2019-06-28
Revision history dnd

### LINKED ISSUE #1965: Release UI: Promote from history within channels (same arch)

bartaz:
Fixes #1965
- Adds possibility for dragging revision from history panel to promote it
- Refactors d'n'd by extracting shared functionality and styles

### QA
- ./run or  https://snapcraft-io-canonical-web-and-design-pr-2052.run.demo.haus/
- click on revision cell to open history panel
- drag any revisions in history to promote it to some channel
- make sure dragging revision cells and channel rows still work

Screenshot 2019-07-01 at 15 33 29

### DIFF
--- a/static/js/publisher/release/components/revisionsList.js
+++ b/static/js/publisher/release/components/revisionsList.js
@@ -252,6 +252,7 @@ class RevisionsList extends Component {
         <table className="p-revisions-list">
           <thead>
             <tr>
+              <th width="30px" />
               <th
                 className={!isReleaseHistory ? "col-checkbox-spacer" : ""}
                 width="150px"
@@ -285,7 +286,7 @@ class RevisionsList extends Component {
               )
             ) : (
               <tr>
-                <td colSpan="4">
+                <td colSpan="5">
                   <em>No releases</em>
                 </td>
               </tr>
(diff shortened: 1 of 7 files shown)


therealjuan:
If you open up the contextual menu on a channel and drag another channel, perhaps the menu should close automatically?

bartaz:
@therealjuan Good catch, thanks. There are some issues with menus/tooltips not hiding on rows/cells that are not active in dnd (because or devmode revisions or other reasons). And because they are not using drag'n'drop they are unaware of it themselves (so can't hide).
To fix that we would need some global state, so that whole UI knows drag is in progress and can react to that.

Let's have it fixed separately, as it's not related to history at all.

therealjuan:
Another one to discuss about revision history:

image
Should the current revision `14239` have white background whereas the other revisions should have a background color like #f7f7f7 and on mouse over become white? Basically working like the channels in other parts of the interface.

Something like this:

image

static/js/publisher/release/components/revisionsList.js
```diff
@@ -252,6 +252,7 @@ class RevisionsList extends Component {
         <table className="p-revisions-list">
           <thead>
             <tr>
+              <th width="30px" />
```
Lukewh: _looks the other way_

bartaz: If only there was a better way to do it without unnecessary CSS overhead ¯\_(ツ)_/¯

bartaz:
Added separate issue #2053 to cover highlighting active revision in history mentioned by @therealjuan

PR #2056, opened 2019-07-03
Fix search when category not existing

tbille:
Fixes https://sentry.is.canonical.com/canonical/ubuntu-com/issues/2325/?query=is%3Aunresolved

When category not existing and no query paramenter.

# qa 

- `./run`
- http://127.0.0.1:5000/search?category=toto
- Should get a page telling you couldn't find snaps on this category.

### DIFF
--- a/templates/store/search.html
+++ b/templates/store/search.html
@@ -72,8 +72,12 @@
         {% endif %}
     {% else %}
       <div class="row">
-        <h2>We couldn't find anything for your search <div class="no-search-query">{{query}}.</div></h2>
-        <h4>Try using some different, or more general keywords.</h4>
+        {% if query %}
+          <h2>We couldn't find anything for your search <div class="no-search-query">{{query}}.</div></h2>
+          <h4>Try using some different, or more general keywords.</h4>
+        {% else %}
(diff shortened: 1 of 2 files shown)


templates/store/search.html
```diff
@@ -72,8 +72,12 @@ <h3>All {{ category_display|lower }} snaps</h3>
         {% endif %}
     {% else %}
       <div class="row">
-        <h2>We couldn't find anything for your search <div class="no-search-query">{{query}}.</div></h2>
-        <h4>Try using some different, or more general keywords.</h4>
+        {% if query %}
+          <h2>We couldn't find anything for your search <div class="no-search-query">{{query}}.</div></h2>
+          <h4>Try using some different, or more general keywords.</h4>
+        {% else %}
+          <h2>We couldn't find anything in the category <div class="no-search-query">{{category_display}}.</div></h2>
```
bartaz: Shouldn't we say that category doesn't exist? Or we don't know that, and we only know it doesn't return any results?

tbille: it's the same for me. What would you prefer?

Lukewh: "We couldn't find the category 'name'." ?

PR #2063, opened 2019-07-03
Disable setting and clearing default tracks

Lukewh:
As the default track conversation continues, we don't want people to get stuck in a hole because of the messaging we're providing.

### DIFF
--- a/static/js/publisher/release/components/releasesHeading.js
+++ b/static/js/publisher/release/components/releasesHeading.js
@@ -4,7 +4,6 @@ import { connect } from "react-redux";
 
 import { setCurrentTrack } from "../actions/currentTrack";
 import { getTracks } from "../selectors";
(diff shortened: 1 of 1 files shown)


static/js/publisher/release/components/releasesHeading.js
```diff
@@ -51,9 +50,11 @@ class ReleasesHeading extends Component {
             </Wrap>
           </h4>
         </div>
-        <div className="col-6">
-          {tracks.length > 1 && <DefaultTrackModifier />}
-        </div>
+        {/* disabled until default tracks are +1'd
```
tbille: Jinja comments
```
{#

#}
```

Lukewh: in javascript? no thanks

tbille: ooooh I do the understaning now

PR #2069, opened 2019-07-04
GitHub accounts connect

### LINKED ISSUE #1031: [Backend] Give template list of licences

jpmartinspt:
Adds the possibility of connecting a github account to your snapcraft.io account.

Fixes https://github.com/canonical-web-and-design/snap-squad/issues/1031

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -14,6 +14,7 @@ import webapp.api
 import webapp.helpers as helpers
 from webapp.blog.views import blog
 from webapp.builder.views import builder
+from webapp.database import init_db
 from webapp.extensions import csrf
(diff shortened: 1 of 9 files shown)


webapp/app.py
```diff
@@ -57,6 +58,8 @@ def create_app(testing=False):
     else:
         init_brandstore(app)
 
+    init_db()
```
tbille: here I think this shouldn't be enabled in the brandstore.
It's a POC but just so we don't forget (you can add a comment or whatever)

jpmartinspt: Moved in the `if`.

PR #2070, opened 2019-07-04
Dashboard graphs

Lukewh:
Dashboard graphs... I would link an issue but apparently they've all been closed?

## QA
- Pull the branch
- `./run`
- Graph


webapp/store/content/publishers/kde.yaml
```diff
@@ -564,8 +564,8 @@ snaps: [
         "origin": "kde",
         "package_name": "rocs",
         "publisher": "KDE",
-        "summary": "Graph Theory Tool for Professors and Students.",
```
bartaz: Find/Replace all ? ;)

Lukewh: Weird! Must have been! Will put that back in :D

static/js/publisher/metrics/axis.js
```diff
@@ -55,7 +55,7 @@ function cullYAxis(ticks) {
 
 /**
  * Update graph x and y axis formatting.
- * @param {HTMLElement} el Graph wrapping element.
```
bartaz: Another probably unnecessary replace.

PR #2095, opened 2019-07-18
Use default track for channel map if set

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/1968

## QA
- Set your env to point to staging (as described in the README)
- Pull this branch
- `./run`
- Visit http://0.0.0.0:8004/lukewh-test
- The channel map dropdown at the top should have the updated text
- Clicking the dropdown and clicking "install >" on a row should display the appropriate install instructions:
  - As track 2.0 is the default track the install instructions should simple be `snap install lukewh-test`
  - The beta version should be `snap install lukewh-test --edge`
  - The 1.0/beta should be `snap install lukewh-test --channel=1.0/beta`
- Also inspect the "View in Desktop store" buttons to ensure they have the correct format of `snapname?channel` if required.

- You should also check that any other snaps that doesn't have a default track set still show "latest" properly. You can do this pointing the codebase to production/ use the demo
- Also check against production that `/node` is still hardcoded to 10 (temporary)

### DIFF
--- a/webapp/store/snap_details_views.py
+++ b/webapp/store/snap_details_views.py
@@ -104,11 +104,7 @@ def snap_details_views(store, api, handle_errors):
         # to use 10, rather then latest
         default_track = helpers.get_default_track(details["name"])
         if not default_track:
-            default_track = (
-                details["default-track"]
-                if details["default-track"]
-                else "latest"
-            )
+            default_track = details.get("default-track", "latest")
 
         lowest_risk_available = logic.get_lowest_available_risk(
             channel_maps_list, default_track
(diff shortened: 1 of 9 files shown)


therealjuan:
I can't see the default track to be 2.0 on your snap:

image

:(

therealjuan:
Step 1: it works
Step 2: As specified
Step 3: #winning

webapp/store/snap_details_views.py
```diff
@@ -103,6 +103,12 @@ def _get_context_snap_details(snap_name):
         # until default tracks are supported by the API we special case node
         # to use 10, rather then latest
         default_track = helpers.get_default_track(details["name"])
+        if not default_track:
+            default_track = (
+                details["default-track"]
```
jpmartinspt: Can be `details.get("default-track", "latest")`.

PR #2098, opened 2019-07-22
Updated default track messaging

Lukewh:
## QA
- Set environment to staging as described in the README
- Pull this branch
- `./run`
- Visit http://0.0.0.0:8004/lukewh-test/releases
- Change the track dropdown
- Set the default track and check all messaging makes sense
- Repeat for clearing the default track


therealjuan:
Works. Thanks.

PR #2119, opened 2019-07-29
Use blog module

tbille:
# Summary

- Use blog module

# QA

- make sure that all features provided bu the old blog are still there:
- Article page
- Index
- Marketo intergration
- Series
- snap related articles (nextcloud)


webapp/app.py
```diff
@@ -80,8 +82,9 @@ def init_snapcraft(app, testing=False):
     app.register_blueprint(store_blueprint(testing=testing))
     app.register_blueprint(account, url_prefix="/account")
     app.register_blueprint(publisher_snaps)
-    app.register_blueprint(blog, url_prefix="/blog")
     init_docs(app, "/docs")
+    init_blog(app, "/blog")
+    # app.register_blueprint(blog, url_prefix="/blog")
```
jpmartinspt: Can be removed.

webapp/app.py
```diff
@@ -48,7 +50,7 @@ def create_app(testing=False):
         init_extensions(app)
 
         talisker.flask.register(app)
-        talisker.requests.configure(webapp.api.blog.api_session)
+        # talisker.requests.configure(webapp.api.blog.api_session)
```
jpmartinspt: Can be removed.

webapp/blog/views.py
```diff
@@ -1,225 +1,32 @@
...
-        blog_tags_ids = logic.get_tag_id_list(blog_tags, snap)
-
-        if blog_tags_ids:
+        if blog_tags:
             try:
-                blog_articles, total_pages = api.get_articles(blog_tags_ids, 3)
+                blog_articles, total_pages = wordpress_api.get_articles(
+                    blog_tags["id"], 3
+                )
             except ApiError:
```
jpmartinspt: I think ApiError won't work here, since `wordpress_api` does not raise this exception.

webapp/blog/views.py
```diff
@@ -234,28 +41,35 @@ def snap_posts(snap):
...
-            article, featured_image=None, author=None
-        )
-        articles.append(
-            {
-                "slug": transformed_article["slug"],
-                "title": transformed_article["title"]["rendered"],
-            }
+        try:
+            blog_articles, total_pages = wordpress_api.get_articles(series)
+        except ApiError:
```
jpmartinspt: I think `ApiError` won't work here, since `wordpress_api` does not raise this exception.

webapp/blog/views.py
```diff
@@ -1,225 +1,32 @@
...
+    blog.init_app(app, "Snapcraft Blog", [2996], "", url_prefix)
 
+    @app.route(url_prefix + "/api/snap-posts/<snap>")
+    def snap_posts(snap):
         try:
-            categories_list = api.get_categories()
+            blog_tags = wordpress_api.get_tag_by_name(
+                "".join(["sc:snap:", snap])
+            )
         except ApiError:
```
jpmartinspt: I think `ApiError` won't work here, since `wordpress_api` does not raise this exception.

webapp/app.py
tbille: done

webapp/blog/views.py
tbille: Same

webapp/blog/views.py
```diff
@@ -1,226 +1,32 @@
...
-        "categories": categories,
-        "used_categories": category_cache,
-        "filter": filter,
-        "newsletter_subscribed": newsletter_subscribed,
-    }
-
-    return flask.render_template("blog/index.html", **context)
-
+            blog_tags = wordpress_api.get_tag_by_name(
+                "".join(["sc:snap:", snap])
```
jpmartinspt: If you feel like it you could swap this to use f-string.

PR #2126, opened 2019-07-31
Build list snap builds

jpmartinspt:
Add a tab in snap details page where the user can see a list of builds for the snap.

### DIFF
--- a/webapp/api/launchpad.py
+++ b/webapp/api/launchpad.py
@@ -1,0 +1,6 @@
+from launchpadlib.launchpad import Launchpad
+
+
+launchpad = Launchpad.login_anonymously(
+    "storefront", "production", version="devel"
+)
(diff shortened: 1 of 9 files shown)


webapp/api/launchpad.py
```diff
@@ -0,0 +1,29 @@
+from webapp import api
```
tbille: as mentionned in chat. Can we use the launchpadlib to avoid depending of self written API files.

webapp/api/launchpad.py
```diff
@@ -0,0 +1,6 @@
+from launchpadlib.launchpad import Launchpad
+
+
+launchpad = Launchpad.login_anonymously(
+    "snappy", "production", version="devel"
```
tbille: is this an account: `snappy`?

jpmartinspt: From the launchpadlib docs: "The first argument to Launchpad.login_anonymously() is a string that identifies the web service client".

tbille: then use `storefront` instead

PR #2142, opened 2019-08-06
Vanilla snap details page

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2130

## QA
- Pull the branch
- `./run`
- Visit multiple snaps http://0.0.0.0:8004/toto http://0.0.0.0:8004/spotify http://0.0.0.0:8004/slack http://0.0.0.0:8004/vlc etc.
- Ensure everything works and looks as expected specifically:
  - channel map
  - video/ image layout
  - distro install instruction
  - all the above at various window sizes (mobile, tablet etc.)
- Visit multiple snap distro install pages http://0.0.0.0:8004/install/toto/arch http://0.0.0.0:8004/install/spotify/debian etc.
- Ensure everything looks and works as expected specifically:
  - "Install" button jump
  - "Show more" link on description


bartaz:
@Lukewh Oh, seems like we both looked at distro pages.

There is an issue with spacing in the install instruction card:
Screenshot 2019-08-06 at 11 44 16
https://snapcraft-io-canonical-web-and-design-pr-2142.run.demo.haus/install/microk8s/manjaro

I reported it to Vanilla: https://github.com/canonical-web-and-design/vanilla-framework/issues/2447
Not sure if there is a simple workaround to use before it's fixed.

Also another issue on wide screens:
Screenshot 2019-08-06 at 11 46 51

Which can be fixed by adding `u-clearfix` to div with `u-fixed-width` above the snaps. I have the fix in my own branch as well...

templates/store/snap-details/_channel_map.html
```diff
@@ -1,7 +1,7 @@
 <div id="js-channel-map" class="p-channel-map is-closed">
   <div id="channel-map-install" class="p-channel-map__tab">
     <div class="p-strip--light is-shallow">
-      <div class="row">
+      <div class="u-fixed-width">
```
bartaz: We use `u-fixed-width` a lot in this migration, and I'm not sure if it's good approach in long term.
Ideally this should be done with a `row` and `col-12`. I think Vanilla `u-fixed-width` is meant to be used when you don't want to or can't use the grid.
We can and should use the grid, but just use `u-fixed-width`, because class name replace is quicker fix. I already had issues with blindly use of `u-fixed-width` with floats in it.

Sure, we want to make this migration as quick as possible, and we could say 'we will update this later'. But practically speaking - will this 'later' be any time soon, or will we stay with `u-fixed-width` for a long time? 

@Lukewh Thoughts? As much as I hate adding these divs with `col-12` just to make things work I start to feel we should do it now properly.

bartaz: Unless we really create a post-migration task to get rid of `u-fixed-width` that we do soon after.

bartaz:
Embeddable card modal layout is broken on snap details:

Screenshot 2019-08-06 at 12 54 37

templates/store/snap-details/_channel_map.html
Lukewh: I changed to `u-fixed-width` as per https://ubuntu.com/blog/vanilla-framework-2-0-upgrade-guide 2.4:
"Previously, a .row with a single col-12 or a col-12 on its own may have been used to display content in a fixed-width container. The nested solution adds unnecessary markup and a lone col-12 will no longer work.

A simple way to make an element fixed width, use the utility .u-fixed-width, which does not need columns."

Lukewh:
> Embeddable card modal layout is broken on snap details:
Yeah - I saw this, but thought we could cover it when sorting the publicise page? It's hard to know when to fix what.

Lukewh:
> @Lukewh Oh, seems like we both looked at distro pages.

If we've done the same thing, should be ok :D

therealjuan:
## Channel map

image
 Drop down arrow is a little tight on the channel map

## Report a snap

image

Report a snap dialog has too much padding on the bottom?

## Distro pages

image

Distro pages have very tight padding on the top and the sides? Right now is like this but maybe we can use this opportunity to add more margin on the top of the title

image

Paragraphs on distro pages are missing a margin bottom

image

There are some strange margin things happening here between snaps and the bottom elements and within the image on the card :)

Lukewh:
@therealjuan 
> ## Channel map
> 
> Drop down arrow is a little tight on the channel map

I've removed the `min-width: 0` rule, so it uses the vanilla default now

> ## Report a snap
> 
> Report a snap dialog has too much padding on the bottom?

Done

> ## Distro pages
>
> Distro pages have very tight padding on the top and the sides? Right now is like this but maybe we can use this opportunity to add more margin on the top of the title

How much padding you want mr mr?

> Paragraphs on distro pages are missing a margin bottom

@bartaz reported to vanilla [here](https://github.com/canonical-web-and-design/vanilla-framework/issues/2447)

> There are some strange margin things happening here between snaps and the bottom elements and within the image on the card :)

I think you made these images originally -- can you upload wider versions or even better SVG versions?

therealjuan:
On a snap detail page of a publisher with publisher page, the featured snaps are floating:

image

bartaz:
@Lukewh I may be wrong, but I don't think that layout of snap details embedded modal has anything to do with publisher pages (apart from using same JS logic). If I recall correctly the layout is separate in some partial.

Lukewh:
> @Lukewh I may be wrong, but I don't think that layout of snap details embedded modal has anything to do with publisher pages (apart from using same JS logic). If I recall correctly the layout is separate in some partial.

I'll take a look :+1:

templates/store/snap-details/_channel_map.html
bartaz: OK. If they recommend it, than it certainly should clear floats (like row did).

Lukewh: Apparently using `u-clearfix` is the correct way to do it :man_shrugging: https://github.com/canonical-web-and-design/vanilla-framework/issues/2451

kwm14:
## Design review
Looking good @Lukewh just a couple of small updates 😺 
**Snap heading**
- [ ] Float/position buttons to the left on the mobile viewport

Screenshot 2019-08-06 at 16 57 31

**Share embeddable card**
- [ ] Code snippet component has a weird white top and bottom border...?!?

Screenshot 2019-08-06 at 17 01 07

**Distro pages**
- [ ] Enable snaps on the section headings looking too close for comfort

Screenshot 2019-08-06 at 17 04 03

- [ ] Random margin-right on 'More things to do...' cards

Screenshot 2019-08-06 at 17 05 58

therealjuan:
The problem with the more things to do is due to the images being used, we need to re-export as SVGs so they adapt to the width instead fo being fixed.

Lukewh:
> ## Design review
> 
> Looking good @Lukewh just a couple of small updates smiley_cat
> **Snap heading**
> 
>     * [ ]  Float/position buttons to the left on the mobile viewport
> 

https://snapcraft.io/minetest-luk3yx
Long names need lots of space. It works the same as snapcraft.io currently

> 
> **Share embeddable card**
> 
>     * [ ]  Code snippet component has a weird white top and bottom border...?!?
> 

This should be fixed as of 5 minutes ago :D

> 
> **Distro pages**
> 
>     * [ ]  Enable snaps on the section headings looking too close for comfort

Will be fixed when https://github.com/canonical-web-and-design/vanilla-framework/issues/2447 gets released
 
>     * [ ]  Random margin-right on 'More things to do...' cards

Waiting on updated assets as @therealjuan commented :D

kwm14:
@Lukewh wow, that was fast 👍 refreshed and have seen your updates. And spoke to @therealjuan about the images 😄

Lukewh:
[BLOCKED] Waiting on @hughesjordan to provide updated images for the "More things to do..." section

therealjuan:
https://assets.ubuntu.com/v1/9d4affec-distro_img_01.svg
https://assets.ubuntu.com/v1/0f7350d8-distro_img_02.svg

@Lukewh  @hughesjordan  ^^^ images as SVGs

PR #2145, opened 2019-08-06
Fix annotated code styles in first snap flow

bartaz:
Fixes styling of annotated code in first snap flow.

### QA

- go to first snap flow "package snap" step
- click on generate snapcraft.yaml
- annotated code should look fine, hovering should show tooltips
- compare with production

### DIFF
--- a/static/sass/_patterns_annotated_code.scss
+++ b/static/sass/_patterns_annotated_code.scss
@@ -10,6 +10,12 @@
     text-align: left;
     text-shadow: none;
 
+    code {
+      background: none;
+      box-shadow: none;
+      white-space: pre;
+    }
+
     code b {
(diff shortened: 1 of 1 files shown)


Lukewh:
Might be good to propose this to vanilla @bartaz wdyt?

bartaz:
@Lukewh you mean annotated code as a pattern? We may give it a try. Possibly it should be quite isolated, although it may contain some snapcraft/yaml specific stuff for formatting.

Lukewh:
@bartaz yeah true, I guess in it's current form it doesn't cater to different markup... Maybe we should just make sure other squads are aware we have it for now and formalise in Vanilla if they want to use it?

PR #2163, opened 2019-08-08
Added title to docs so they appear on search results

### LINKED ISSUE #2161: Docs: Search results have the same title
When searching documentation in snapcraft.io is quite difficult to distinguish between results as they all share the same title (coming from the main template) and you need to rely on the URL and the strip of content appearing from the HTML:

image

A quick fix for this will be to update the Page title to the name of the post.

therealjuan:
Fixes #2161 

### QA Steps
1. Pull the branch or run locally
2. Visit `/docs`
3. Search for something in the docs
4. Results should have different title on results as opposed os current behaviour:

image

### DIFF
--- a/templates/docs/document.html
+++ b/templates/docs/document.html
@@ -1,7 +1,7 @@
 
 {% extends "docs/base.html" %}
 
-{% block meta_title %}Snapcraft documentation | {{ document.title }} {% endblock %}
+{% block meta_title %} {{ document.title }} | Snapcraft documentation{% endblock %}
 
 {% block content_docs %}
 <div class="p-strip is-shallow">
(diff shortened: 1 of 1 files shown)


bartaz:
Can't search on demo because of missing key (I guess it's demo configuration):
Screenshot 2019-08-08 at 13 19 24

templates/docs/document.html
```diff
@@ -1,6 +1,8 @@
 
 {% extends "docs/base.html" %}
 
+{% block meta_title %}Snapcraft documentation | {{ document.title }} {% endblock %}
```
bartaz: I think I'd do it opposite: "Document title | Snapcraft documentation" ... as title seems more relevant to what people want
```suggestion
{% block meta_title %}{{ document.title }} | Snapcraft documentation{% endblock %}
```

PR #2167, opened 2019-08-09
Publisher settings sticky bar consistent with listings

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2139

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/snap_name/settings
- Ensure the page works as before
- The sticky bar should look the same as the listing page
- The sticky bar should be tighter
- Both the listing page and settings page sticky bar buttons should go full-width without margins at mobile size

### DIFF
--- a/static/sass/_snapcraft_market.scss
+++ b/static/sass/_snapcraft_market.scss
@@ -125,6 +125,14 @@
     .p-inline-list__item {
       line-height: 2.3rem;
     }
+
+    [class*="p-button"]:not([class*="p-button__"]) {
+      margin-right: $sph-inner;
+
+      &:last-child {
+        margin-right: 0;
+      }
+    }
   }
 
   .p-market-banner {
--- a/templates/publisher/listing.html
+++ b/templates/publisher/listing.html
@@ -14,7 +14,7 @@ Listing details for {% if display_title %}{{ display_title }}{% else %}{{ snap_t
     <input type="hidden" name="snap_id" value="{{ snap_id }}" />
 
     <div class="snapcraft-p-sticky" id="store-listing-notification">
-      <div class="row">
+      <div class="row u-no-margin--bottom">
         <div class="col-7">
           <p class="snapcraft-p-market-message u-no-margin--bottom">
             {% if private %}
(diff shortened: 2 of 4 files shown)


static/sass/_snapcraft_market.scss
```diff
@@ -125,6 +125,14 @@
     .p-inline-list__item {
       line-height: 2.3rem;
     }
+
+    [class*="p-button"]:not([class*="p-button__"]) {
```
bartaz: Thats very specific, where does it come from?

templates/publisher/listing.html
```diff
@@ -14,7 +14,7 @@
     <input type="hidden" name="snap_id" value="{{ snap_id }}" />
 
     <div class="snapcraft-p-sticky" id="store-listing-notification">
-      <div class="row">
+      <div class="row u-no-margin--bottom">
```
bartaz: Rows don't have margin in Vanilla.
It seems that we need to reset it because we set margin with custom `.market-form .row`, which is bit weird.

Not sure if we need to fix it now, but seems like layout of this page could use updating to remove such oddities.

static/sass/_snapcraft_market.scss
Lukewh: Vanilla adds right margin to some elements, and not to others (whether a `button` or `p`, because we use both we get random spacing. This target's both `p-button--neutral` and `p-button--positive`, but excludes `p-button__text` added for the save spinner.

templates/publisher/listing.html
Lukewh: Yeah I think we should do a pass on the listings page after these vanilla updates to try and remove some of these silly overrides. At the end of the day it's a form, so should just work like the rest of the vanilla forms

static/sass/_snapcraft_market.scss
Lukewh: Maybe we should make this sticky bar a proper component too?

bartaz: Added a not on that to #2168 as well.

kwm14:
Design LGTM 👍

therealjuan:

PR #2186, opened 2019-08-13
Featured banners

Lukewh:
Supersedes https://github.com/canonical-web-and-design/snapcraft.io/pull/2116

## QA
- Pull the branch
- `./run --env FEATURED_BANNERS_ENABLED=True`
- See featured banner images and updated generated banners
- Ensure that clicking "Show banner restrictions" under the banner image lists the recommended size of 1920x640

### DIFF
--- a/static/js/publisher/form/banner.js
+++ b/static/js/publisher/form/banner.js
@@ -161,46 +161,38 @@ class Banner extends React.Component {
           you eligible to be featured.
         </p>
         <AccordionHelp name="banner restrictions">
-          <div className="col-8">
-            <ul className="u-no-margin--bottom">
-              <li>
-                <small>
-                  Accepted image formats include: <b>JPEG & PNG files</b>
-                </small>
-              </li>
-              <li>
-                <small>
-                  Min resolution: <b>720 x 240 pixels</b>
-                  <br />
-                  <i>
-                    For best results on <a href="/store">the store</a> we
-                    {/*
-                       wtf prettier? having 'or' on a newline removes the space
-                       having it on the same line says "add a newline" using {' '}
-                       complains even more!
-                     */}
-                    recommend a resolution of <b>1920 x 640 pixels</b>
-                    &nbsp;or greater
-                  </i>
-                </small>
-              </li>
-              <li>
-                <small>
-                  Max resolution: <b>4320 x 1440 pixels</b>
-                </small>
-              </li>
-              <li>
-                <small>
-                  Aspect ratio: <b>3:1</b>
-                </small>
-              </li>
-              <li>
-                <small>
-                  File size limit: <b>2MB</b>
-                </small>
-              </li>
-            </ul>
-          </div>
+          <ul>
+            <li>
+              <small>
+                Accepted image formats include: <b>JPEG & PNG files</b>
+              </small>
+            </li>
+            <li>
+              <small>
+                Min resolution: <b>720 x 240 pixels</b>
+                <br />
+                <i>
+                  For best results on <a href="/store">the store</a> we
+                  recommend a resolution of <b>1920 x 640 pixels</b> or greater
+                </i>
+              </small>
+            </li>
+            <li>
+              <small>
+                Max resolution: <b>4320 x 1440 pixels</b>
+              </small>
+            </li>
+            <li>
+              <small>
+                Aspect ratio: <b>3:1</b>
+              </small>
+            </li>
+            <li>
+              <small>
+                File size limit: <b>2MB</b>
+              </small>
+            </li>
+          </ul>
         </AccordionHelp>
       </Fragment>
     );
--- a/static/js/publisher/form/banner.js
+++ b/static/js/publisher/form/banner.js
@@ -162,24 +162,44 @@ class Banner extends React.Component {
         </p>
         <AccordionHelp name="banner restrictions">
           <div className="col-8">
-            <p>
-              <small>
-                Accepted image formats include: <b>JPEG & PNG files</b>
-                <br />
-                Min resolution: <b>720 x 240 pixels</b>
-                <br />
-                Max resolution: <b>4320 x 1440 pixels</b>
-                <br />
-                <i>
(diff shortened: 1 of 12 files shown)


kwm14:
## Design review

- [ ] Do we have guidelines for the size of your snap icon? At the newly proposed size, this snap icon is pixelated to the max 😆 

Screenshot 2019-08-13 at 13 49 06

- [ ] The new banner size has increased the snap icon and it's rather on the large size...

Screenshot 2019-08-13 at 14 07 49

- [ ] Proposal to update `$breakpoint-large` and `$breakpoint-medium` to `192px/12rem` and `$breakpoint-small` to `160px/10rem`

- [ ] Should we remove min resolution to stop users uploading a small banner or replace the values with 1920 x 640 pixels?

  Accepted image formats include: **JPEG & PNG files**
  Min resolution: **1920 x 640 pixels**
  Max resolution: **4320 x 1440 pixels**
  Aspect ratio: **3:1**
  File size limit: **2MB**

Lukewh:
> ## Design review
> 
>     * [ ]  Do we have guidelines for the size of your snap icon? At the newly proposed size, this snap icon is pixelated to the max

That icon is 256x256, they've scaled and uploaded that pixelated image :shrug:

>     * [ ]  Should we remove min resolution to stop users uploading a small banner or replace the values with 1920 x 640 pixels?

I'm following what we agreed in the meeting on https://github.com/canonical-web-and-design/snapcraft.io/pull/2116#issuecomment-519024900 if that's changed, I can update it. I worry that would be pulling the rug from under those that have already uploaded a banner? It also means confirming with GS and the store to update the restrictions (which is fine, we just need to confirm that's what we want to do) otherwise people can just upload directly to the API and skip our size check and carry on uploading smaller images.

bartaz:
The generated ones are bit oddly spaced on smaller screen sizes:
Screenshot 2019-08-14 at 09 39 40

static/js/publisher/form/banner.js
```diff
@@ -170,6 +170,8 @@ class Banner extends React.Component {
                 <br />
                 Max resolution: <b>4320 x 1440 pixels</b>
                 <br />
+                Recommended resolution: <b>1920 x 640 pixels</b>
```
bartaz: Maybe it would be worth adding a bit of explanation why it's recommended? That smaller sizes would be pixelated when displayed in full width on store page?

templates/store/_category-partial.html
```diff
@@ -1,10 +1,16 @@
 <div id="js-snap-{{ category }}">
   <div class="u-fixed-width">
     {% if has_featured %}
+      {% if snaps[0].banner_url %}
+        <a class="p-featured-snap p-featured-snap--banner" href="/{{ snaps[0].package_name }}">
```
bartaz: Does it need `p-featured-snap` class? It seems that `p-featured-snap--banner` styles override most of `p-featured-snap` styles anyway.

bartaz:
The text in generated banner (especially snap name) feels too small when banner is full size:

Screenshot 2019-08-14 at 11 49 28

therealjuan:
Currently icons look quite big:
image


Just wondering if we use `col-2` instead of `col-3` :
image


@kwm14 ? @Lukewh ?

kwm14:
> Just wondering if we use `col-2` instead of `col-3` :
> image
> 
> @kwm14 ? @Lukewh ?

Just tried your suggestion. LGTM 👍

Lukewh:
> > Just wondering if we use `col-2` instead of `col-3` :
> > @kwm14 ? @Lukewh ?
> 
> Just tried your suggestion. LGTM +1

So is it official `col-2`? :+1: from everyone?

templates/store/_category-partial.html
```diff
@@ -13,7 +19,7 @@
           {% if loop.index != 1 or snap["icon_url"] == "" %}
               {% if has_featured %}
                   {% set icon_size = "large" %}
-                  <div class="col-4 col-medium-3">
+                  <div class="col-4 col-medium-3 col-small-2">
```
bartaz: `col-small-2` breaks how snaps are displayed on mobile:

Screenshot 2019-08-16 at 09 05 38

bartaz: And funny how 'featured banner' on mobile seems to be smaller and less informative than snaps in regular listing.

Lukewh: Ah I meant to back that change out - I was trying something for a follow-up PR, will remove.
The old way these icons worked is they became smaller at the medium breakpoint so I was testing that they looked ok next to the smaller banner size on mobile, but didn't want to do the PR as part of this one!

kwm14:
## Design review
@Lukewh banner size updates LGTM 👍 

Only one comment on banner restrictions info recommended size '_For best results on the store we recommend a resolution of 1920 x 640 pixels or greater_' feel detached from the max resolution info.

Screenshot 2019-08-16 at 16 32 06

Could we put banner restrictions info into an [unordered list](https://docs.vanillaframework.io/base/typography/#unordered-list) but better than my hack below 😆 

Screenshot 2019-08-16 at 16 31 09

kwm14:
Bullet update LGTM 👍

webapp/configs/snapcraft.py
```diff
@@ -2,4 +2,5 @@
 
 WEBAPP_CONFIG = {"LAYOUT": "_layout.html", "STORE_NAME": "Snap store"}
 
+FEATURED_BANNERS_ENABLED = os.getenv("FEATURED_BANNERS_ENABLED", "false")
```
jpmartinspt: How do you feel about doing something like:
```
_featured_banners_enabled = os.environ.get("FEATURED_BANNERS_ENABLED", default="false")
FEATURED_BANNERS_ENABLED = _featured_banners_enabled.lower() in {"1", "t", "true"}
```

bartaz: It's perfect, thanks! I forgot that this config file is just a python script, so I can parse the value in place instead of the views themselves.

PR #2187, opened 2019-08-13
Small fixes for Vanilla 2 footer

bartaz:
### QA
- ./run or demo
- look at footer
- it should look well
- there should be no dead hiring link
- social icon color should match background


therealjuan:
image

PR #2192, opened 2019-08-14
Remove p-fluid-grid

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2174

## Done
- Added custom col class with modifier
- Updated homepage p-fluid-grid use with the grid
- Updated first snap flow with the grid
- Updated /build with the grid
- Updated listing page screenshots with a custom grid

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004
- Resize the page up and down, make sure that the following flow to new lines properly:
  - `Official snaps from major publishers` logos
  - Language selection grid
  - Linux distro grid
- Visit http://0.0.0.0:8004/build
- Resize the page up and down
- Make sure distro icons flow to new lines properly
- Visit http://0.0.0.0:8004/first-snap/c
- Resize the page up and down
- Make sure the OS boxes flow to new lines properly
- Resize the page up and down
- Make sure the media areas flow properly - add some screenshots, remove some, drag images on etc.


static/js/public/fsf-language-select.js
```diff
@@ -10,7 +10,11 @@ function initFSFLanguageSelect(rootEl) {
   const closeDetails = () => {
     flowDetails.forEach(e => (e.style.display = "none"));
     flowLinks.forEach(l => l.classList.remove("is-open"));
-    window.location.hash = "";
+    if (window.history.pushState) {
```
Lukewh: This stops the page jumping to the top when you resize

static/js/public/fsf-language-select.js
```diff
@@ -50,7 +56,12 @@ function initFSFLanguageSelect(rootEl) {
         window.location.hash = "";
       }
 
-      window.scrollTo(0, link.getBoundingClientRect().top);
+      const viewportOffset = link.getBoundingClientRect().top;
```
Lukewh: I'm not sure how this worked before? `getBoundingClientRect().top` is from the viewport top, no the document top? I also added `bahaviour: "smooth"` which is apparently a thing.

static/js/publisher/form/mediaList.js
```diff
@@ -26,7 +26,14 @@ class MediaList extends React.Component {
         classes.push("u-hide");
       }
 
-      inputs.push(
+      // We always want the "add" item to be before any hidden inputs
```
Lukewh: This is because of the CSS magic I do later with nth-child... The "Add" element was always added to the end so wasn't being caught by the selector.

static/sass/_snapcraft_market_screenshots.scss
```diff
@@ -16,15 +19,49 @@
...
+    grid-column-start: auto;
     height: $screenshot-thumb-size;
     justify-content: center;
     margin-bottom: $spv-inner--small;
-    margin-right: $spv-inner--small;
     min-width: auto;
     position: relative;
     width: $screenshot-thumb-size;
 
+    @media screen and (min-width: 936px) and (max-width: 1152px) {
```
Lukewh: previously mentioned nth-child magic

static/sass/_snapcraft_market_screenshots.scss
```diff
@@ -1,7 +1,10 @@
 @mixin snapcraft-market-screenshots {
   .p-listing-images {
-    justify-content: flex-start;
-    margin-right: -$spv-inner--small;
+    display: grid;
```
bartaz: I think we shouldn't do that.

Removing custom implementation to replace it with Vanilla grid classes (a little bit customized) how we do it in other places is fine, but here we are implementing just another custom styling that seems much more complicated than it was before. I don't think it's worth it.
Especially that it behaves bit weirdly during reordering with drag and drop (spacing between items is off when dragging, especially when they wrap in 2 lines on smaller screens).

static/js/public/fsf-language-select.js
bartaz: Hmm... yeah, I don't know, it seems it worked in "most" cases, but I've seen it scrolling to weird places, but couldn't really reproduce it reliably.
Anyway, thanks for figuring it out and fixing.

static/sass/_snapcraft_market_screenshots.scss
bartaz: If it turns out this is the last place we need out "fluid grid" pattern, let's move any needed fluid grid styles here and remove the `p-fluid-grid`, but I wouldn't rewrite the whole thing in grid just for the sake of it.

Unless we can simply work this out with Vanilla col classes, which apparently is not that straightforward in this case.

templates/index.html
```diff
@@ -49,25 +49,24 @@ <h2>The app store for Linux</h2>
     <h3 class="p-muted-heading">Official snaps from major publishers</h3>
   </div>
   <div class="row">
-    <div class="p-fluid-grid--centered">
-      <span class="p-logo-image p-fluid-grid__item--small u-align--center">
-          <a href="/firefox"><img src="https://assets.ubuntu.com/v1/aae93857-logo-mozilla--snapcraft-homepage.svg" alt="Mozilla" /></a>
-      </span>
-      <span class="p-logo-image p-fluid-grid__item--small u-align--center">
+    <div class="col-small-2 col-medium-3 col-2 col-no-scale u-align--center">
```
bartaz: I think it was suggested we could try using inline images pattern here.
I gave it a try (to try avoid adding customized grid classes) and it kinda works (but not without custom styling):
https://jsfiddle.net/bartaz/oyx83meu/

So, it's one custom modification of a pattern against the other. Not sure which one I prefer.

What I don't like about the col-no-scale col-no-scale--pushed is that it ties a lot into how many logos there are. If we would add a new logo to the list, we need to figure out sizing and when images should drop to new line again.
On the other hand... if we add more images we probably would need some design/style changes anyway...

static/sass/_snapcraft_custom-cols.scss
```diff
@@ -0,0 +1,17 @@
+@mixin snapcraft-custom-cols {
+  .col-no-scale {
+    @media (min-width: 772px) and (max-width: 1083px) {
```
bartaz: Where do these numbers come from? Are they somehow related to breakpoints from vanilla? At least the 772px which seems to be repeated? Would be nice to have those taken from some variables if possible.

I assume the max-widths come from trial and error (just looking at when we want items to drop, or what minimal size we want them to have). Because of that (and their quite arbitrary nature) I'm not really sure how to name those classes. `col-no-scale` doesn't say anything. And it's not like it disabling scaling. It's just it changes the breakpoints of when grid changes. I kind of prefer the `col-land-select` one, as it specifically says for what this col class was made, so maybe we could stick to that?
I guess `col-no-scale` is used in different places and that's why the name is so vague? If it's mostly for logos, maybe call it `col-logo`?

templates/partials/fsf_language.html
```diff
@@ -1,89 +1,110 @@
-<div class="u-fixed-width js-fsf-language-select" id="language">
-  <div class="p-fluid-grid--centered">
-    <a class="p-logo-link p-fluid-grid__item--small p-card p-flow-link" href="/first-snap/python" data-flow-link="python">
+<div class="row js-fsf-language-select" id="language">
+  <div class="col-small-2 col-medium-3 col-2 col-lang-select u-align--center">
```
bartaz: Because the `u-align--center` affects things inside the logo card, maybe we can just add centering to existing classes (.p-logo-link or .p-flow-link) instead of adding utility class?

Lukewh:
Agreed changes:
- [ ] `col-no-scale` -> `col-logos`
- [ ] Use [inline images pattern](https://docs.vanillaframework.io/patterns/inline-images/) for listing screenshots
- [ ] Discuss with @kwm14 if the publisher logos and distro logos should be aligned in style - they have the same functionality, but different treatment (distro's have hover border)

templates/partials/fsf_language.html
Lukewh: Doesn't need to be aligned center anyway... I seemed to make that up.

static/js/publisher/form/mediaList.js
bartaz: Is this still relevant if we don't use grid for media images?

Lukewh: it's not needed, but I think it's nice to have - to make visible items consistently together

bartaz: OK. I don't fully understand what it does, so let's stick to it ;)

Lukewh: Ah - just so you know, the "Add screenshot" was always getting appended to the list of inputs. So you've have - "visible, visible, visible, hidden, hidden, visible" order of elements. This just makes sure it's "visible, visisble, visible, visible, hidden, hidden"... I'm sure that explains it completely :|

kwm14:
Design LGTM 👍

PR #2193, opened 2019-08-16
Vanillarise forms

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2177

## Done
- Updated both listing and settings forms to use `p-form p-form--stacked`

## QA
- Pull the branch
- `./run`
- Make sure things look and work as expected (in relation to field spacing &  help text / error text spacing)
- Try and raise some errors on fields
- Visit http://0.0.0.0:8004/snap_name/settings
- Make sure things look and work as expected (as for the listing page)

### DIFF
--- a/static/js/publisher/form/AccordionHelp.js
+++ b/static/js/publisher/form/AccordionHelp.js
@@ -33,7 +33,7 @@ class AccordionHelp extends React.Component {
 
     return (
       <Fragment>
-        <p className="p-form-help-text u-no-margin">
+        <p className={`p-form-help-text${open ? " u-no-margin--bottom" : ""}`}>
           <a role="button" onClick={this.toggleHelp}>
             {label}
           </a>
--- a/static/sass/_snapcraft_market.scss
+++ b/static/sass/_snapcraft_market.scss
@@ -6,6 +6,11 @@
   $black-overlay-light-color: rgba(0, 0, 0, .2);
   $white-overlay-strong-color: rgba(255, 255, 255, .9);
 
+  // Override baseline alignment
+  .p-form--stacked .p-form__group {
+    align-items: flex-start;
+  }
+
   // hack to make anchor link behave like a button
   .js-market-revert.is-disabled {
 
@@ -20,6 +25,7 @@
   .p-editable-icon {
     cursor: pointer;
     display: inline-block;
+    margin-bottom: .7rem;
     margin-right: 3.2877%;
     position: relative;
 
--- a/static/sass/_snapcraft_market.scss
+++ b/static/sass/_snapcraft_market.scss
@@ -6,21 +6,6 @@
   $black-overlay-light-color: rgba(0, 0, 0, .2);
   $white-overlay-strong-color: rgba(255, 255, 255, .9);
 
-  // Add extra space between groups of related items
-  .snapcraft-p-market__section {
-    margin-bottom: $sp-xx-large;
-  }
-
-  .market-form {
-    .row {
-      margin-bottom: $sp-medium;
-
-      & .row {
-        margin-bottom: 0;
-      }
-    }
-  }
-
   // hack to make anchor link behave like a button
   .js-market-revert.is-disabled {
 
--- a/static/sass/_snapcraft_market.scss
+++ b/static/sass/_snapcraft_market.scss
@@ -7,7 +7,7 @@
   $white-overlay-strong-color: rgba(255, 255, 255, .9);
 
   // Override baseline alignment
-  .p-form--stacked .p-form__group {
+  .p-form--stacked .p-form__group--top {
     align-items: flex-start;
   }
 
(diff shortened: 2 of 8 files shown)


kwm14:
## Design review
Design LGTM and errors work in fields 👍

bartaz:
I find spacing of labels on listing page a bit confusing on 'tall' fields.
Labels for snap icon and images just looks misaligned, but label for featured banner looks aligned to help text below the field it should refer to:

Screenshot 2019-08-16 at 13 35 49

For comparison on settings it seems that large fields (like list of checkboxes) have label nicely aligned to top:

Screenshot 2019-08-16 at 13 36 10

Thoughts @kwm14 ?

Lukewh:
> I find spacing of labels on listing page a bit confusing on 'tall' fields.
> Labels for snap icon and images just looks misaligned, but label for featured banner looks aligned to help text below the field it should refer to:
> 
> 
> For comparison on settings it seems that large fields (like list of checkboxes) have label nicely aligned to top:
> 
> 
> Thoughts @kwm14 ?

I knew I'd forgotten to do something, thanks @bartaz

static/js/publisher/form/AccordionHelp.js
```diff
@@ -33,7 +33,7 @@ class AccordionHelp extends React.Component {
 
     return (
       <Fragment>
-        <p className="p-form-help-text u-no-margin">
+        <p className={`p-form-help-text${open ? " u-no-margin--bottom" : ""}`}>
```
bartaz: I hate writing these and hate reading these :) Some day we need to finally get some nice classname utility function.

On that note, I think its unnecessary optimisation that makes it harder to read: 
```suggestion
        <p className={`p-form-help-text ${open ? "u-no-margin--bottom" : ""}`}>
```

kwm14:
> I find spacing of labels on listing page a bit confusing on 'tall' fields.
> Labels for snap icon and images just looks misaligned, but label for featured banner looks aligned to help text below the field it should refer to:
> 

Ahhhhh yes. I didn't think we could control that, as elements align to the baseline grid. So just assumed that was the norm 💭 

snap

If we can update @Lukewh , then great! 💯 🥇 👍

templates/publisher/settings.html
```diff
@@ -122,102 +120,100 @@
...
+          </div>
         </div>
       </div>
 
-      <div class="p-strip is-shallow">
-        <div class="row">
-          <hr class="u-no-margin--bottom" />
-        </div>
+      <div class="row p-form__group">
+        <hr class="u-no-margin--bottom" />
```
bartaz: I wish there was a nicer way to do these instead of wrapping hr into rows or strips... But let's not overthink it for now.

kwm14: @bartaz @Lukewh @therealjuan do we even need the dividing lines between the listing sections?

Lukewh: I removed them, but it looked a bit odd as they're not really related

kwm14: @Lukewh Ah ok makes sense, just thought I'd ask in case there wasn't a reason for it 😉

static/sass/_snapcraft_market.scss
```diff
@@ -6,19 +6,9 @@
...
-    .row {
-      margin-bottom: $sp-medium;
-
-      & .row {
-        margin-bottom: 0;
-      }
-    }
+  // Override baseline alignment
+  .p-form--stacked .p-form__group {
+    align-items: flex-start;
```
bartaz: It should probably be a modifier (`.p-form__group--top`?) because it breaks alignment in some places:

Screenshot 2019-08-16 at 15 18 07

templates/publisher/listing.html
```diff
@@ -143,179 +144,173 @@ <h2 class="p-heading--four">
...
         </div>
-
-        <div class="js-categories-category1-help-text row{% if snap_categories.categories[0] %} u-hide{% endif %}">
-          <div class="col-8 col-start-large-3">
-            <p class="p-form-help-text u-no-margin--bottom">
-              Include a category and your snap will be better ranked within the Snap Store.
-              It will also allow users to discover your snap via browsing the available categories.
-            </p>
+        <div class="col-12">
+          <div class="js-categories-category1-help-text row{% if snap_categories.categories[0] %} u-hide{% endif %}">
```
bartaz: Maybe it's just me, but I find this harder to read (and possibly more error prone when maintaining, because you need to remember to put space in the conditional. The browser wouldn't care if there is additional space at the end...

```suggestion
          <div class="js-categories-category1-help-text row {% if snap_categories.categories[0] %}u-hide{% endif %}">
```

This happens in several places.

static/js/publisher/form/AccordionHelp.js
Lukewh: Once we're happy with the other stuff, I'll accept this change before merging!

templates/publisher/listing.html
Lukewh: Updated them all - they were inconsistent throughout and I'm happy for it to be this way.

bartaz: Thx

kwm14:
Design updates LGTM 👍

PR #2208, opened 2019-08-23
Tour steps functionality

### LINKED ISSUE #2198: Build tour functionality
- [ ] combine overlay and tooltip component
- [ ] prepare tour JSON data format
- [ ] make it possible to jump between tour states, show and hide it

bartaz:
Fixes #2198 

- Moves demo tour to publisher listing page
- Adds support for multiple steps with content provided via JSON
- Adds support for different positioning of tooltips to adjust to different elements on page (2 possible positions for the sake of demo)

### QA
This is not a complete feature. It's a minimal functional demo of the tour, merged to a feature branch where the work on it will continue.

- ./run or demo
-  sign in and go to listing page of any snap you own
- find a tour question mark (?) button in right bottom corner
- click it to trigger the tour
- navigate between the steps
- skip or finish tour to close
- tour should handle scrolling and resizing of the screen

Screenshot 2019-08-23 at 13 26 05

### DIFF
--- a/static/js/publisher/tour.js
+++ b/static/js/publisher/tour.js
@@ -3,10 +3,10 @@ import ReactDOM from "react-dom";
 
 import Tour from "./tour/tour";
 
-export default function initTour() {
-  const tour = document.createElement("div");
-  document.body.appendChild(tour);
-
-  const target = document.querySelector("h2");
-  ReactDOM.render(<Tour target={target} />, tour);
+export default function initTour({ container, target }) {
+  if (document.contains(container) && document.contains(target)) {
+    ReactDOM.render(<Tour target={target} />, container);
+  } else {
+    throw Error("initTour container or target elements not found in document");
+  }
 }
--- a/static/js/publisher/tour.js
+++ b/static/js/publisher/tour.js
@@ -3,10 +3,13 @@ import ReactDOM from "react-dom";
 
 import Tour from "./tour/tour";
 
-export default function initTour({ container, target }) {
-  if (document.contains(container) && document.contains(target)) {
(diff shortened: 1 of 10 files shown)


therealjuan:
* When resizing there is a moment that the window gets funny so not sure if we want to hide completely the component while resizing or maybe it is due to the animation?

image

* Maybe we need a little bit more margin on the top and bottom of the mask overlay?
image

bartaz:
> When resizing there is a moment that the window gets funny so not sure if we want to hide completely the component while resizing or maybe it is due to the animation?

There is some delay before we recalculate position during resize (because we don't want to do that thousands of times when user is resizing). But the weird behaviour is probably more visible for these right positioned steps, because the tooltip in this case is positioned relative to window right, while the mask to the window left, so when you resize, tooltip stays close to window border while mask stays behind.
As suggested, the easiest workaround would be to hide either the mask 'window' or both the mask and tooltip step while resizing and show them in new position once resize is finished.

> Maybe we need a little bit more margin on the top and bottom of the mask overlay?

The mask shape relies a lot on spacing around targeted element (with a 5px offset). We can easily make the offset bigger. There will be cases where mask will look misaligned because top and bottom spacing have different values - it's a common case in Vanilla, and we can't do much about it.
But making the offset bigger will at least give elements more space in the highlight.

bartaz:
@therealjuan I updated the mask offset to make it a bit bigger, but want to keep it within small spacing of Vanilla to avoid mask showing parts of other elements around.

Also added some functionality to hide the mask when resizing, so it looks less weird.

PR #2215, opened 2019-08-27
Tour with interactive page

bartaz:
Makes the highlighted elements interactive during the tour.

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-2215.run.demo.haus/
- sign in
- go to snap listing page of any snap you own
- click on the tour icon (?) in bottom right
- click through the tour
- while clicking through the tour steps interact with elements that are featured, try to edit them during the tour

### DIFF
--- a/static/sass/_snapcraft_tour.scss
+++ b/static/sass/_snapcraft_tour.scss
@@ -29,6 +29,7 @@
   .p-tour-overlay {
     bottom: 0;
     left: 0;
+    pointer-events: none;
     position: absolute;
     right: 0;
     top: 0;
@@ -39,6 +40,7 @@
     background: $color-overlay;
     bottom: 0;
     left: 0;
+    pointer-events: all;
     position: absolute;
     right: 0;
     top: 0;
@@ -183,6 +185,7 @@
 
     max-width: 27rem;
     overflow: visible; // to make tooltip arrow visible
+    pointer-events: all;
     position: absolute;
     transition: 200ms;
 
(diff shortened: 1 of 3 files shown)


templates/publisher/listing.html
```diff
@@ -206,9 +205,8 @@ <h2 class="p-heading--four">
           <label for="video" class="p-form__label">Video:</label>
         </div>
         <div class="col-8">
-          <div class="p-form__control">
+          <div class="p-form__control" data-tour="listing-video">
```
bartaz: Moving some tour elements around do make sure whole interactive area is highlighted and available during the tour.

static/sass/_snapcraft_tour.scss
```diff
@@ -29,6 +29,7 @@
   .p-tour-overlay {
     bottom: 0;
     left: 0;
+    pointer-events: none;
```
bartaz: Make it possible to click through the transparent 'window' in the overlay.

bartaz:
@anthonydillon Thanks for comments:
> * You can click tabs like Releases as there is an offset of the mask on the element
Probably won't fix that, we want the offset so if any other element is there it will be clickable. Unless we can make steps interactive or not individually, so we can disable clicking on this one completely.

> * When you get to the social one it masks the second help text. Could you move the selected element to a parent?
Good catch, will do!

> * The Get featured, get installed! step doesn't scroll up enough to be in the screen
This should be fixed by #2216

PR #2223, opened 2019-08-30
Adds basic tests for Tour component.

bartaz:
Adds tests to Tour components.

### QA
- ./run test
- tests should pass


bartaz:
The more tests I added the more coverage decrease codecov calculated (because of low coverage of some files touched by tests), so I guess we will just ignore and merge as it is, as it's still more tests than we used to have.

PR #2231, opened 2019-09-05
Revision attributes icons

### LINKED ISSUE #2226: Show an icon next to revisions built on Launchpad

bartaz:
Fixes #2226 
Add Lauchpad icon to revisions built on LP
Adds Lauchpad revisions to 'Available revisions' dropdown

### QA
- ./run or demo
- go to releases page of any snap built on lauchpad
- check if revisions built on LP have icon in releases table
- check if revisions built on LP have icon in releases table tooltip
- check if revisions built on LP have icon in revision history build column
- if all revisions in the channel are from the same build request icon should be shown under channel name as well
- click on available revisions dropdown, Launchpad revisions should be listed there

Screenshot 2019-09-05 at 09 53 04


Lukewh:
I'm wondering whether the icon should be next to the revision number? Next to the version reduces the amount of space for the version:
Screenshot_2019-09-05  Releases and revision history for olaris

Lukewh:
Or perhaps just aligned to the left so that the revision number doesn't clip the history icon?
Screenshot_2019-09-05  Releases and revision history for olaris (1)

bartaz:
@Lukewh I wanted to make it quick to see how we want to improve that.
These elements are quite crowded and specifically styled to make things fit and text to have ellipsis, etc. We can move the icons, but this may require bigger changes in layout of the elements.

kwm14:
@bartaz LGTM. We can look at a visual solution in another issue 😃as it will need a bit more exploration my side 👍  

The only comments would be to remove the launchpad logo from the tooltip as I don't think it's needed and move the icon to the right of the release (so we have a clean keyline)

Screenshot 2019-09-06 at 14 26 41

kwm14:
## Quick idea
Show table row actions on hover...just why it's in my head 🤣 

ideaz

bartaz:
@karlos I moved the icons to the right of version and removed it from tooltips.


> Show table row actions on hover...just why it's in my head 🤣
> ideaz

That's something to explore. May be a bit confusing that  drag handle is not something to click, when others are. Also X icon only shows up when there is a pending release (not yet saved) so most revisions will only have history icon (and drag handle).

kwm14:
> @karlos I moved the icons to the right of version and removed it from tooltips.
> 
Updates looking good @bartaz 
> 
> That's something to explore. Maybe a bit confusing that drag handle is not something to click when others are. Also, X icon only shows up when there is a pending release (not yet saved) so most revisions will only have history icon (and drag handle).
>
Definitely. Ah yes, you make a good point about the dragging interaction 👍 now I know how the interactions work will be interesting to look at. Maybe something in 🇫🇷

PR #2232, opened 2019-09-05
Fix selecting same version in all archs in release UI

### LINKED ISSUE #2230: Release UI: Revisions get unselected when selecting in all architectures
- go to snap with many architectures and same version in all of them
- choose 'All' available revisions
- unselect one of the revisions from the same version
- show history of other arch
- message will show that same version is available in other architectures
- click on 'Select in all architectures'

### Current
Currently selected revisions in same version get **unselected**, new one is selected.
'Select in all architectures' button is still visible and clicking it again: unselects the selected ones and selects new ones.

### Expected
Clicking 'Select in all architectures' should select same version in ALL architectures, so that the button disappears (because all architectures have the same version selected). Currently selected revisions should not be unselected by clicking it.

select

bartaz:
Fixes #2230 

Fixes the issue where revisions from the same architecture were unselected.

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-2232.run.demo.haus/
- go to releases page of a snap with multiple architectures (and multiple versions)
- show "All revisions" in available revisions
- select same version in all architectures
- unselect one architecture
- open history of another architecture, there should be a button 'Select in all architectures'
- clicking on it should select the missing architecture, but not unselect any
- try different combinations of selected revisions: all selected from different versions, all unselected, etc, make sure that behaviour of selecting same version in all architectures works as expected in different cases

Screenshot 2019-09-05 at 17 18 14

### DIFF
--- a/static/js/publisher/release/components/revisionsList.js
+++ b/static/js/publisher/release/components/revisionsList.js
@@ -13,7 +13,7 @@ import { getChannelName, isInDevmode } from "../helpers";
 
 import RevisionsListRow from "./revisionsListRow";
 import { closeHistory } from "../actions/history";
-import { toggleRevision, clearSelectedRevisions } from "../actions/channelMap";
+import { toggleRevision } from "../actions/channelMap";
 import {
   getFilteredReleaseHistory,
   getPendingChannelMap,
@@ -34,7 +34,6 @@ class RevisionsList extends Component {
   }
 
   selectVersionClick(revisions) {
-    this.props.clearSelectedRevisions();
     revisions.forEach(revision => this.props.toggleRevision(revision));
   }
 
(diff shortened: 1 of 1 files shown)


Lukewh:
Peek 2019-09-05 15-36
Is this expected?

bartaz:
@Lukewh Updated as discussed (not clearing already selected revisions), and updated button label to say 'Select in available architectures'

therealjuan:
Perhaps the call to action should be `Select in available architectures` instead of all which is maybe misleading?

bartaz:
> Perhaps the call to action should be Select in available architectures instead of all which is maybe misleading?

@therealjuan Already done!

PR #2249, opened 2019-09-10
Revision attributes sets

### LINKED ISSUE #2225: Group sets of revisions into build sets based on revision attributes

bartaz:
Fixes #2225 

Allows promoting all revisions in same build set at once.
Removes Launchpad option from "Available revisions" menu when there are no lauchpad builds.

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-2249.run.demo.haus/
- go to releases page of any snap with Launchpad builds (like [surl](https://snapcraft-io-canonical-web-and-design-pr-2249.run.demo.haus/surl/releases))
- find any release that is part of a build set (has LP icon)
- try to drag it to promote to different channel
- while dragging architectures of all revisions in the build set should highlight
- you should be able to drop the set on ANY architecture from the set
- when you drop build set all revisions from the set should be promoted to given channel

Screenshot 2019-09-10 at 16 12 37

### DIFF
--- a/static/js/publisher/release/components/availableRevisionsMenu.js
+++ b/static/js/publisher/release/components/availableRevisionsMenu.js
@@ -89,7 +89,19 @@ export class AvailableRevisionsMenu extends Component {
   }
 
   renderItems() {
-    const items = Object.keys(menuLabels);
+    let items = Object.keys(menuLabels);
+
+    items = items.filter(item => {
+      // remove Launchpad menu option if there are no LP builds
+      if (
+        item === AVAILABLE_REVISIONS_SELECT_LAUNCHPAD &&
+        this.props.getFilteredCount(AVAILABLE_REVISIONS_SELECT_LAUNCHPAD) === 0
+      ) {
+        return false;
+      }
+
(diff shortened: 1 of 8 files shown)


Lukewh:
Peek 2019-09-10 16-42
I find this odd behaviour.. It's kind of unrelated to this PR, but we definitely need to show (under "launchpad") rows of the same tag, not rows of a version.
Something like:
Screenshot_2019-09-10  Releases and revision history for surl
This can be a separate issue and PR though.

bartaz:
@Lukewh 

> I find this behaviour odd.

Yes, it's an odd edge case here.
It's caused by the fact that currently when user chooses filter from dropdown we just select latest revision for each architecture separately. So if latest build set doesn't cover all architectures they will get mixed.
We can possibly solve it and only select revisions from one build set there.

Anyway, such situation still would be possible if user promotes these build sets to a channel.

One thing that we may want to explore (but requires some custom handling) is to change the element that is dragged to make it more clear, that it's more than one revision that is dragged.

PR #2262, opened 2019-09-13
Drag'n'drop updates for revision attributes support

bartaz:
Updates release UI drag'n'drop functionality for better handling of build sets.

- removes different drag types implementations - anything being dragged is treated as set of revisions (revision, build set or channel)
- removes single cells as drop targets, making it possible to drop everything on channel level (it will end up in correct architecture anyway)
- during dragging highlights all architectures that will be affected by the drop in given channel

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-2262.run.demo.haus/
- go to releases page of any snap (ideally with some revisions build in Launchpad)
- try promoting a single revision to some channel (from releases table)
- try promoting single revision to some channel (from revisions history of any channel)
- try promoting a revision being part of a build set from launchpad (all revisions should be promoted)
- try promoting whole channel (all revisions from channel should be promoted)
- make sure that other Release functionality still works (history, closing channel, branches, etc)

Screenshot 2019-09-19 at 08 20 44

### DIFF
--- a/static/js/publisher/release/components/releasesTableCell.js
+++ b/static/js/publisher/release/components/releasesTableCell.js
@@ -13,12 +13,7 @@ import {
   getBuildId,
   getRevisionsArchitectures
 } from "../helpers";
-import {
-  useDragging,
-  DND_ITEM_REVISION,
-  DND_ITEM_BUILDSET,
-  Handle
-} from "./dnd";
+import { useDragging, DND_ITEM_REVISIONS, Handle } from "./dnd";
 
 import { toggleHistory } from "../actions/history";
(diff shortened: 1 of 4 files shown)


static/js/publisher/release/components/releasesTableCell.js
```diff
@@ -285,6 +215,7 @@ const ReleasesTableCell = props => {
     props.undoRelease(revision, channel);
   }
 
+  // TODO: still needed?
```
bartaz: Yes. Highlighting same build set on hover is separate feature from drag'n'drop.

PR #2273, opened 2019-09-24
🚧 Use flask-base; remove create_app

### LINKED ISSUE #1850: Remove testing logic from the application
Currently, the snapcraft.io app is written to behave differently under "test mode", in the following ways:

- Don't use `flask_wtf.csrf.CSRFProtect`
- Don't register the app as a Talisker app
- Don't add Talisker settings
- Bypass caching in the `StoreAPI`

Code is here:

https://github.com/canonical-web-and-design/snapcraft.io/blob/43146934ead35f0187afc86cdc9dfb441e3fecd6/webapp/app.py#L40-L46

And here:

https://github.com/canonical-web-and-design/snapcraft.io/blob/4279134b159f2eec556e65e7f015420ca279b146/webapp/api/store.py#L75-L78

I'd suggest that this is undesirable, in all these cases. Our tests risk missing bugs in the CSRF logic, in Talisker logic (which we've had multiple issues with already), and in the fundamental behaviour of the `StoreAPI` object - whose cache may be causing issues we're unaware of.

I experimented with running the tests with each of these features included, and only two caused issues:

- Including `talisker.requests.configure(webapp.api.dashboard.api_session)` leads to multiple `requests.exceptions.ConnectionError` and `requests.exceptions.Timeout` failures, but *tests still pass*
- Using the cache in `StoreAPI` led to multiple test failures in `tests_requests.RequestsCacheTest` and `store.tests_search.GetSearchViewTest`

I'd suggest that both of these should be looked into. The `ConnectionError` and `Timeout` errors might well indicate something wrong in the dashboard API, and the `StoreAPI` failures suggest that we may not be effectively testing the cache behaviour in our tests, or that our tests aren't effectively taking account of the caching done inside the `StoreAPI` object, which they should.

Even if it is entirely legitimate for these errors to show up in tests, but not in production, I think these could easily be, and should be, handled at the testing layer. The `ConnectionError` and `Timeout` errors could be explicitly suppressed, if it's legitimate to do so, and we should be able to find a way to call `StoreAPI` with `cache=False` for the purposes if testing, again if it's legitimate to do so.

More generally, I think it's dangerous to introduce the precedent that it's okay to run our app fundamentally differently for testing than we do for production use. In every case, we should make every possible effort to test the app as it is used in production, and a neat way to achieve this is to keep the app fundamentally unaware of whether it is in "test mode" or not (as also discussed in https://github.com/canonical-web-and-design/docs.snapcraft.io/pull/132#pullrequestreview-222064624).

So I would suggest that we remove the `create_app(testing=False)` function entirely, move `app.testing = True` into the test initialisation step, remove the `if not testing:` conditionals, and make appropriate fixes so the `webapp.api.dashboard.api_session` and the `StoreAPI` caching work without issue.

I'd be happy to submit a PR to this effect myself, I just might need a little help from someone with regard to understanding the `StoreAPI` tests.

nottrobin:
Fixes critical problem with SENTRY_DSN.

Also fixes #1850

QA
--

Check tests pass.

Run the demo, check it works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ webapp/licenses.json
 coverage/
 .webcache/
 .webcache_blog/
+.coverage
--- a/entrypoint
+++ b/entrypoint
@@ -2,7 +2,7 @@
 
 set -e
 
-RUN_COMMAND="talisker.gunicorn webapp.app:create_app() --bind $1 --worker-class sync --workers 5 --name talisker-`hostname`"
+RUN_COMMAND="talisker.gunicorn webapp.app:app --bind $1 --worker-class sync --workers 5 --name talisker-`hostname`"
 
 if [ "${FLASK_DEBUG}" = true ] || [ "${FLASK_DEBUG}" = 1 ]; then
     RUN_COMMAND="${RUN_COMMAND} --reload --log-level debug --timeout 9999"
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,28 +4,21 @@ canonicalwebteam.discourse-docs==0.6.4
 canonicalwebteam.blog==2.0.14
 canonicalwebteam.yaml-responses==1.1.1
 canonicalwebteam.search==0.1.0
-Flask==1.0.2
+canonicalwebteam.flask-base==0.3.3
 Flask-OpenID-Stateless==1.2.6
 Flask-WTF==0.14.2
 humanize==0.5.1
 Markdown==3.0.1
-prometheus_client==0.3.1
 mistune==0.8.4
 pybadges==1.1.0
 pybreaker==0.4.4
 pycountry==17.9.23
 pymacaroons==0.12.0
 python-dateutil==2.6.1
-raven[flask]==6.5.0
 requests==2.21.0
 requests-cache==0.4.13
-responses==0.9.0
+responses==0.10.6
 ruamel.yaml==0.15.72
-talisker[gunicorn]==0.14.3
-
-# Temporary fix to SSO redirect issue
-Werkzeug==0.14.1
 
 # Development dependencies
-black==18.6b4
 Flask-Testing==0.6.2
--- a/tests/first_snap/tests_views.py
+++ b/tests/first_snap/tests_views.py
@@ -4,14 +4,14 @@ from unittest.mock import mock_open, patch
 from flask_testing import TestCase
 
 from webapp import helpers
-from webapp.app import create_app
+from webapp.app import app
 
 
 class FirstSnap(TestCase):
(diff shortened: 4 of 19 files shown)


Lukewh:
Can this be split into multiple PR's? Is this fixing "a critical problem with SENTRY_DSN", updating "Use flask-base; remove create_app" or implementing a suggestion (https://github.com/canonical-web-and-design/snapcraft.io/issues/1850)? I'd rather have the "critical problem" solved without shoehorning in suggestions that haven't been discussed.

nottrobin:
@Lukewh I can try.

The reason I did it this way is because what needs to happen to fix this is to update to latest Talisker. But when you do that, you need the latest Flask and Prometheus or you get clashes. And Raven.

The way we install all those these days in modern sites is using Flask-base. So the ideal way to import those dependencies so we can keep them up-to-date in-step with our other sites is to switch to Flask-base.

But Flask-base implements a lot of things that snapcraft.io implements itself individually.

So, updating to latest dependencies the right way involves replacing a lot of local snapcraft.io logic with Flask-base, which is what I'm doing here.

I'm not sure what middle-ground makes sense.

Lukewh:
@nottrobin that's fine, it would have been good to explain that in the description of the PR - context is key.

Also the tests fail.

nottrobin:
@Lukewh yep fair enough. I know the tests fail :( I'm trying to see if I can do a smaller set of changes that don't break tests. Haven't found it yet.

jpmartinspt:
Can't we just update the requirement to `talisker[flask,gunicorn,raven,prometheus]==0.15.0 `?

nottrobin:
@jpmartinspt oddly, that breaks tests too.

jpmartinspt:
@nottrobin not for me, locally at least.

nottrobin:
@jpmartinspt did you try `./run clean`?

Anyway, https://github.com/canonical-web-and-design/snapcraft.io/pull/2274 works. I'll put this PR on hold, and if someone could merge that that'd be great.

nottrobin:
@jpmartinspt oh sorry you're right it does work. Not sure what mess I got my dependencies into before then. I'll update https://github.com/canonical-web-and-design/snapcraft.io/pull/2274 accordingly.

Lukewh:
This seems to be out of date now. Feel free to reopen with updates, or create new PR's

nottrobin:
Thanks @Lukewh. Yeah this does need doing and we will do it as part of https://github.com/canonical-web-and-design/base-squad/issues/285#issuecomment-539079067, but I think it makes sense to open a new PR when we get to it.

PR #2274, opened 2019-09-25
Update raven and talisker

nottrobin:
This makes snapcraft.io work with newer-format SENTRY_DSN keys.

QA
--

Check tests pass.

Then try running the site with:

``` bash
./run --env SENTRY_DSN=https://1e82fd54e08142c9978f623cb746b965@sentry.is.canonical.com//3
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,23 +9,20 @@ Flask-OpenID-Stateless==1.2.6
 Flask-WTF==0.14.2
 humanize==0.5.1
 Markdown==3.0.1
-prometheus_client==0.3.1
 mistune==0.8.4
 pybadges==1.1.0
 pybreaker==0.4.4
 pycountry==17.9.23
 pymacaroons==0.12.0
 python-dateutil==2.6.1
-raven[flask]==6.5.0
 requests==2.21.0
 requests-cache==0.4.13
 responses==0.9.0
 ruamel.yaml==0.15.72
-talisker[gunicorn]==0.14.3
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -16,12 +16,12 @@ pybreaker==0.4.4
 pycountry==17.9.23
 pymacaroons==0.12.0
 python-dateutil==2.6.1
-raven[flask]==6.5.0
```
tbille: remove line

requirements.txt
```diff
@@ -16,12 +16,12 @@ pybreaker==0.4.4
...
+raven[flask]==6.10.0
 requests==2.21.0
 requests-cache==0.4.13
 responses==0.9.0
 ruamel.yaml==0.15.72
-talisker[gunicorn]==0.14.3
```
tbille: `talisker[flask,gunicorn,raven,prometheus]==0.15.0`

tbille: and you can remove `prometheus_client` dep

tbille:
A way to test sentry is hitting this endpoint: http://0.0.0.0:8004/_status/test/sentry

requirements.txt
nottrobin: And in theory I could have removed Werkzeug too, but I think if I remove all those then it really does break things. I'm experimenting again, but could we merge this in a working state first? I'm going to fix up https://github.com/canonical-web-and-design/snapcraft.io/pull/2273 anyway to move to the modern version of things.

PR #2277, opened 2019-09-26
Refactor releases code into the release redux actions

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2275

## QA
- Pull the branch
- `./run`
- Ensure releasing revisions works as expected


static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,230 @@
...
+    const { csrfToken, snapName } = options;
+    const releases = Object.keys(pendingReleases).map(id => {
+      return {
+        id,
+        revision: pendingReleases[id].revision,
+        channels: pendingReleases[id].channels
+      };
+    });
+
+    hideNotification();
```
bartaz: This should probably be `dispatch(hideNotification())` as `hideNotification` is an action

static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,230 @@
...
+import { cancelPendingReleases } from "./pendingReleases";
+import { releaseRevisionSuccess, closeChannelSuccess } from "./channelMap";
+import { updateRevisions } from "./revisions";
+
+import { getRevisionsMap, initReleasesData } from "../releasesState";
+
+const ERROR_MESSAGE =
+  "There was an error while processing your request, please try again later.";
+
+function updateReleasesData(dispatch, releasesData) {
```
bartaz: This feels like it would fit into being action creator function as it technically just dispatches actions to update the data:
```
function updateReleasesData(releasesData) {
  return dispatch => {
    // init channel data in revisions list
    const revisionsMap = getRevisionsMap(releasesData.revisions);
    initReleasesData(revisionsMap, releasesData.releases);

    dispatch(updateRevisions(revisionsMap));
    dispatch(updateReleases(releasesData.releases));
  }
}
```

And of course when it's used it would be `dispatch(updateReleasesData)`. I guess in theory this action could be useful in other places as well.

static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,230 @@
...
+    redirect: "follow",
+    referrer: "no-referrer"
+  })
+    .then(response => response.json())
+    .catch(() => {
+      throw new Error(ERROR_MESSAGE);
+    });
+}
+
+function fetchUpdatedReleasesHistory(dispatch, csrfToken, snapName) {
```
bartaz: I think this one is not needed as separate function, especially now when `dispatch` needs to be passed in.
All it does is just single `then`. We can do it inline where it's called. Especially that it seems we do it in one place anyway.

static/js/publisher/release/actions/releases.js
Lukewh: I've refactored this out - if there isn't a use case for it now, I figured we didn't need to make it too generic? I'll happily change it to what you suggested if you think it would be better this way though.

static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,177 @@
...
+  if (json.success) {
+    if (json.closed_channels && json.closed_channels.length > 0) {
+      json.closed_channels.forEach(channel => {
+        // make sure channels without track name get prefixed with 'latest'
+        if (RISKS.indexOf(channel.split("/")[0]) !== -1) {
+          // TODO: This should be the default track, not latest
+          channel = `latest/${channel}`;
+        }
+
+        closeChannelSuccess(channel);
```
bartaz:
```suggestion
        dispatch(closeChannelSuccess(channel));
```

static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,177 @@
...
+              dispatch(releaseRevisionSuccess(revision, channel));
+            }
+          });
+        });
+      });
+    });
+  }
+}
+
+function fetchReleases(dispatch, csrfToken, snapName, releases, revisions) {
```
bartaz: I would say this should also be in api module, but not sure what to do with passing of `dispatch`.
Maybe instead of passing `dispatch` we could pass `successCallback` that would just be a function (`if (successCallback) successCallback(revision, channel)`). This way api module can be free of dependencies on dispatch, while still being able to call arbitrary callbacks on success/error?

Lukewh: I guess we could return a list/object of the results from the queue and iterate over them within the action, rather than calling `handleReleaseResponse` within the `fetch` itself - give it one job?

bartaz: That's a good point. Maybe we could use `Promise.all` for that? 
Anyway, one benefit of doing it in here is that each response is handled when it returns and doesn't need to wait for others. Passing it all to array would mean we only handle all once we get all results back.

static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,157 @@
...
+
+    if (errors.length) {
+      message = `${message} ${errors
+        .map(e => e.message)
+        .filter(m => m)
+        .join(" ")}`;
+    }
+  }
+
+  showNotification({
```
bartaz: Should be `dispatch`ed

static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,157 @@
...
+  };
+}
+
+function handleCloseResponse(dispatch, json, channels) {
+  if (json.success) {
+    if (json.closed_channels && json.closed_channels.length > 0) {
+      json.closed_channels.forEach(channel => {
+        // make sure channels without track name get prefixed with 'latest'
+        if (RISKS.indexOf(channel.split("/")[0]) !== -1) {
+          // TODO: This should be the default track, not latest
```
bartaz: I'm not sure if this comment is true (?) If channel is only a risk (like "beta") is it really "default-track/beta"? Or is it "latest/beta"? ... Default tracks are confusing...

Lukewh: it will be `default-track/beta`, though for most people default-track will = latest

static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,164 @@
...
+  }
+
+  return showNotification({
+    status: "error",
+    appearance: "negative",
+    content: message
+  });
+}
+
+// TODO: move inside of this function out
```
bartaz: This comment is not relevant anymore, fixed by this refactoring I think.

static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,164 @@
...
+    };
+
+    dispatch(hideNotification());
+    return fetchReleases(_handleReleaseResponse, releases, csrfToken, snapName)
+      .then(() =>
+        fetchCloses(_handleCloseResponse, csrfToken, snapName, pendingCloses)
+      )
+      .then(() => fetchReleasesHistory(csrfToken, snapName))
+      .then(json => dispatch(updateReleasesData(json)))
+      .catch(error => dispatch(handleReleaseError(error)))
```
bartaz: Now that `handleReleaseError` returns an action the name is a bit confusing (other `handle*` functions are just handlers/callbacks that are not dispatched.

I would at least rename `handleReleaseError` to ... `showErrorNotification` maybe? Or change it so it only returns error message `getErrorMessage(error)` and here do `dispatch(showNotification({ ... message: getErrorMessage(error) })`.

static/js/publisher/release/components/releasesConfirm.js
```diff
@@ -3,18 +3,35 @@ import PropTypes from "prop-types";
...
 
 import { cancelPendingReleases } from "../actions/pendingReleases";
+import { releaseRevisions } from "../actions/releases";
 
 class ReleasesConfirm extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isLoading: false
```
bartaz: Possibly would be worth moving it into redux. But that requires all the actions/reducer boilerplate... So I'm fine with treating it out of scope for this PR.

static/js/publisher/release/actions/releases.test.js
```diff
@@ -1,4 +1,21 @@
-import { UPDATE_RELEASES, updateReleases } from "./releases";
+/* global global, jest */
```
bartaz: Isn't `jest` in globals by default in test env? I don't recall needing to add it.

Lukewh: Might just be my editor, will remove and see if the linter complains!

Lukewh: eslint failed without it

static/js/publisher/release/actions/releases.js
```diff
@@ -1,5 +1,167 @@
...
+          // TODO: This should be the default track, not latest
+          channel = `latest/${channel}`;
+        }
+
+        dispatch(closeChannelSuccess(channel));
+      });
+    }
+  } else {
+    let error = new Error(
+      `Error while closing channels: ${channels.join(", ")}.`
```
bartaz: Hmm.. the only reason to pass `channels` to this function is this error message?
Not a fault of this PR, but it is a bit weird.

static/js/publisher/release/actions/releases.test.js
```diff
@@ -12,4 +29,234 @@ describe("releases actions", () => {
       expect(updateReleases(releases).payload.releases).toEqual(releases);
     });
   });
+
+  describe("handleCloseResponse", () => {
+    it("should dispatch CLOSE_CHANNEL_SUCCESS when successfull", () => {
+      const store = mockStore({});
+
+      handleCloseResponse(
```
bartaz: Now that I don't remember what these functions are doing is a nice time to review this tests for readability.

A bit of nit picking, but it would be nice to pass named variables instead of incline values to the function call.
At least for the API response object. It took me a while to understand what we are passing, why "edge" is mentioned twice etc.

Something like `handleCloseResponse(store.dispatch, apiResponse, ["latest/edge"])` would be a bit more understandable.

static/js/publisher/release/actions/releases.test.js
```diff
@@ -12,4 +29,234 @@ describe("releases actions", () => {
...
+    it("should throw an error if the API responds with an error", () => {
+      expect(() => {
+        handleCloseResponse(() => {}, { error: true, json: "nope" }, [
+          "latest/edge"
+        ]);
+      }).toThrow();
+    });
+  });
+
+  describe("getErrorMessage", () => {
```
bartaz: Not that this test is here it looks like this function would better fit into some helpers module.

static/js/publisher/release/actions/releases.test.js
```diff
@@ -12,4 +29,234 @@ describe("releases actions", () => {
...
+        getErrorMessage({
+          message: "error!",
+          json: { errors: [{ message: "error1" }, { message: "error2" }] }
+        })
+      ).toEqual("error! error1 error2");
+    });
+  });
+
+  describe("handleReleaseResponse", () => {
+    it("should dispatch all the actions", () => {
```
bartaz: "should dispatch all the actions", but we expect only one action?

static/js/publisher/release/actions/releases.test.js
```diff
@@ -12,4 +29,234 @@ describe("releases actions", () => {
...
+          }
+        },
+        pendingCloses: ["latest/edge"],
+        revisions: {
+          "3": revision
+        }
+      });
+
+      global.fetch = jest
+        .fn()
```
bartaz: Prettier doesn't like it as `global.fetch = jest.fn()` ?

static/js/publisher/release/actions/releases.test.js
```diff
@@ -12,4 +29,234 @@ describe("releases actions", () => {
...
+          revision: revision
+        },
+        type: "RELEASE_REVISION_SUCCESS"
+      });
+    });
+  });
+
+  describe("releaseRevisions", () => {
+    beforeEach(() => {
+      global.fetch = jest.fn().mockResolvedValue();
```
bartaz: Why doing this if we mock again in test?

static/js/publisher/release/actions/releases.test.js
```diff
@@ -12,4 +29,234 @@ describe("releases actions", () => {
...
+        },
+        pendingCloses: ["latest/edge"],
+        revisions: {
+          "3": revision
+        }
+      });
+
+      global.fetch = jest
+        .fn()
+        .mockResolvedValueOnce({
```
bartaz: Would be nice to have a comment before each `.mockResolvedOnce` explaining what API call is being mocked

static/js/publisher/release/actions/releases.test.js
```diff
@@ -12,4 +29,234 @@ describe("releases actions", () => {
...
+          json: () => ({
+            releases: [release],
+            revisions: [revision]
+          })
+        });
+
+      return store.dispatch(releaseRevisions()).then(() => {
+        const actions = store.getActions();
+        expect(actions[0]).toEqual({
```
bartaz: I think in case when we compare the whole array `expect(actions).toEqual([ all the actions... ])` would work as well?

static/js/publisher/release/actions/releases.test.js
Lukewh: No it keeps wrapping it

PR #2278, opened 2019-09-27
Release UI: release table row component refactor

bartaz:
Refactors release table components to make it possible to add custom rows for build sets.

### QA

- ./run or demo
- everything should work as before (try drag and drop, promote some things, etc)
- go to releases page of any snap with lauchpad builds
- list of all builds from lauchpad should be rendered as rows below the table
- whole rows and individual revisions should be draggable
- contextual menu of build rows should work as expected
- there should be no history icon on the build revisions

Screenshot 2019-09-27 at 11 57 44

### DIFF
--- a/static/js/publisher/release/components/releasesTableChannelCell.js
+++ b/static/js/publisher/release/components/releasesTableChannelCell.js
@@ -1,0 +1,391 @@
+import React, { Fragment } from "react";
+import PropTypes from "prop-types";
+import { connect } from "react-redux";
+import { parse, distanceInWordsToNow, addDays } from "date-fns";
+
+import { sortChannels } from "../../../libs/channels";
+
+import {
+  getArchitectures,
+  getPendingChannelMap,
+  hasPendingRelease,
+  getBranches
+} from "../selectors";
+import { Handle } from "./dnd";
+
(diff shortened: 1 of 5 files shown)


static/js/publisher/release/components/releasesTableChannelCell.js
```diff
@@ -0,0 +1,407 @@
+import React, { Fragment } from "react";
```
bartaz: This file is mostly extracted functionality that was previously part of ReleasesTableRow component.

static/js/publisher/release/constants.js
```diff
@@ -3,6 +3,7 @@ const CANDIDATE = "candidate";
 const BETA = "beta";
 const EDGE = "edge";
 const AVAILABLE = "AVAILABLE"; // fake 'risk' with all available revisions
+const BUILD = "BUILD"; // fake 'risk' for build sets
```
bartaz: Would be nice to get rid of those special case 'fake' risks. But this may require deeper refactor of release UI data structure and components...

Lukewh: I agree, but it would probably mean adding another data type to each revision/ release

PR #2297, opened 2019-10-04
Image template

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2291

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004
Things should load quicker


templates/_layout.html
```diff
@@ -21,6 +21,7 @@
 
     <link rel="stylesheet" href="{{ static_url('css/styles.css') }}" />
 
+    <script src="https://assets.ubuntu.com/v1/842d27d3-lazysizes.min.js" async></script>
```
bartaz: It probably doesn't need to be the very first one script? Could we move it later, so browser can pick more important ones first? Probably could go below {% block scripts_includes %}

static/sass/styles.scss
```diff
@@ -227,3 +227,11 @@ $color-social-icon-foreground: $color-light;
 
 @import 'snapcraft_tour';
 @include snapcraft-tour;
+
+.lazyload {
```
bartaz: What is this one for?

Lukewh: see https://pypi.org/project/canonicalwebteam.image-template/ 
"Add lazy sizes" header

bartaz: Put this link in a comment of the file please.

bartaz: After doing some QA I think this is actually wrong and it makes it look bad.
What it does it tells the browser to remove any sizing from the images before they load. So even if the img tag is properly sized in HTML (what we do in many places to reduce layout changes) it will be forced to have size of 0 before loading. It makes the page jump when images are loaded which is bad from performance (unnecessary rerenders in browser) and is very badly visible (because images are loaded when user scrolls to them). So perceived performance is actually worse.

Current snapcraft.io (on slow connection) - nothing moves around, images just load:

current-slow

Lazy loading as currently implemented:
lazy-slow

It seems that removing this .lazyload style helps (or we could just reduce it to only do visibility:hidden):
lazy-slow-width

Lukewh: The problem with only doing visibility is that when JS is disabled, you get a default image that has a width and height that pushed the non-js icon... I'm happy on snapcraft to do this though - most people have JS right!? And if that's the case, we should just remove this CSS altogether

PR #2305, opened 2019-10-10
View progressive releases

Lukewh:
Supersedes(partially) https://github.com/canonical-web-and-design/snapcraft.io/pull/2259
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2240

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/snap/releases for a snap with a progressive release
- Click around and make sure things look ok.

This is just the viewing of a progressive release.

Cell, history closed
Screenshot_2019-10-15  Releases and revision history for lukewh-test

History open
Screenshot_2019-10-15  Releases and revision history for lukewh-test (1)

### DIFF
--- a/static/js/libs/channels.js
+++ b/static/js/libs/channels.js
@@ -265,7 +265,7 @@ function sortChannels(channels, options) {
  * @returns string
  */
 function getChannelString(channelObj) {
-  return `${channelObj.track}/${channelObj.risk}${
+  return `${channelObj.track ? `${channelObj.track}/` : ""}${channelObj.risk}${
     channelObj.branch ? `/${channelObj.branch}` : ""
   }`;
 }
--- a/static/js/libs/channels.js
+++ b/static/js/libs/channels.js
@@ -258,4 +258,22 @@ function sortChannels(channels, options) {
   };
 }
 
-export { sortChannels, parseChannel, createChannelTree, sortAlphaNum };
+/**
+ * Get a channel string based on an object containing track, risk and branch
+ *
+ * @param {track: string, risk: string, branch: string} channelObj
+ * @returns string
+ */
+function getChannelString(channelObj) {
+  return `${channelObj.track}/${channelObj.risk}${
+    channelObj.branch ? `/${channelObj.branch}` : ""
+  }`;
+}
+
+export {
+  sortChannels,
+  parseChannel,
+  createChannelTree,
+  sortAlphaNum,
+  getChannelString
+};
--- a/static/js/publisher/release/components/devmodeRevision.js
+++ b/static/js/publisher/release/components/devmodeRevision.js
@@ -3,14 +3,24 @@ import PropTypes from "prop-types";
 
 import { isInDevmode } from "../helpers";
 
-export default function DevmodeRevision({ revision, showTooltip }) {
+export default function DevmodeRevision({
+  revision,
+  showTooltip,
+  isProgressive
+}) {
+  let revisionLabel = revision.revision;
+
+  if (isProgressive) {
+    revisionLabel = `→ ${revisionLabel}`;
+  }
+
   if (isInDevmode(revision)) {
     return (
       <span
         className="p-tooltip p-tooltip--btm-center"
         aria-describedby={`revision-devmode-${revision.revision}`}
       >
-        {revision.revision}*
+        {revisionLabel}*
         {showTooltip && (
           <span
             className="p-tooltip__message u-align--center"
@@ -30,7 +40,7 @@ export default function DevmodeRevision({ revision, showTooltip }) {
     );
   }
 
-  return revision.revision;
+  return revisionLabel;
 }
 
 DevmodeRevision.propTypes = {
--- a/static/js/publisher/release/components/devmodeRevision.js
+++ b/static/js/publisher/release/components/devmodeRevision.js
@@ -45,5 +45,6 @@ export default function DevmodeRevision({
 
 DevmodeRevision.propTypes = {
   revision: PropTypes.object.isRequired,
-  showTooltip: PropTypes.bool
+  showTooltip: PropTypes.bool,
+  isProgressive: PropTypes.bool
 };
--- a/static/js/publisher/release/components/progressiveBar.js
+++ b/static/js/publisher/release/components/progressiveBar.js
@@ -1,7 +1,40 @@
 import React from "react";
 import PropTypes from "prop-types";
 
-class ProgressiveBar extends React.Component {
+const ProgressiveBar = ({ percentage, targetPercentage, readonly }) => (
+  <div className="progressive-bar p-tooltip--btm-center">
+    {!readonly && (
+      <div
+        className="progressive-bar__target"
+        style={{ width: `${targetPercentage}%` }}
+      >
+        <div className="progressive-bar__target-adjust" />
+      </div>
+    )}
+    <div
+      className="progressive-bar__target-value"
+      style={{
+        left: `${targetPercentage ? targetPercentage : percentage}%`
+      }}
+    >
+      <span className="p-tooltip__message" role="tooltip">
+        {targetPercentage ? targetPercentage : percentage}%
+      </span>
+    </div>
+    <div
+      className="progressive-bar__current"
+      style={{ width: `${percentage}%` }}
+    />
+  </div>
+);
+
+ProgressiveBar.propTypes = {
+  percentage: PropTypes.number,
+  targetPercentage: PropTypes.number,
+  readonly: PropTypes.bool
+};
+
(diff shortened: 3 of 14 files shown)


static/js/libs/channels.js
```diff
@@ -258,4 +258,22 @@ function sortChannels(channels, options) {
...
 }
 
-export { sortChannels, parseChannel, createChannelTree, sortAlphaNum };
+/**
+ * Get a channel string based on an object containing track, risk and branch
+ *
+ * @param {track: string, risk: string, branch: string} channelObj
+ * @returns string
+ */
+function getChannelString(channelObj) {
```
bartaz: Seems similar to `release/helpers.js@getChannelName`.

bartaz:
The % label on hover seems weirdly detached from anything.
It should fit in the row under the bar (but will be quite small to fit) or should be a tooltip probably:

Screenshot 2019-10-16 at 09 55 40

bartaz:
ProgressiveBar is set to `readonly` but it seems when you click+drag on it (despite of it triggering standard revision drag'n'drop) the bar is getting some mouse move events as well as the label changes value and possition (but the bar is not updated).

static/js/libs/channels.js
Lukewh: I feel like `release/helpers.js@getChannelName` is quite specific for the release UI internals, with the AVAILABLE constant. This was meant to be a more generic builder that can build a string from either the object created by `libs/channels.js@parseChannel` or what the API returns. Thoughts?

bartaz: Ah, right, the `AVAILABLE` risk edge case. Would be good to get rid of this (now that we moved the 'available' out of the table to the tabs). But that's another significant refactor...

Maybe the `getChannelName` helper can use `getChannelString` internally for now, to have one place where we construct channel name string.

static/js/publisher/release/components/devmodeRevision.js
```diff
@@ -3,14 +3,24 @@ import PropTypes from "prop-types";
 
 import { isInDevmode } from "../helpers";
 
-export default function DevmodeRevision({ revision, showTooltip }) {
+export default function DevmodeRevision({
```
bartaz: Now I realised that the name of this component is confusing.

It seems to be more like "RevisionLabel" or something... Isn't specific to devmode revisions at all...

static/js/publisher/release/components/progressiveBar.js
```diff
@@ -0,0 +1,160 @@
...
+      mousePosition: 0
+    };
+
+    this.onMouseUpHandler = this.onMouseUpHandler.bind(this);
+    this.onMouseDownHandler = this.onMouseDownHandler.bind(this);
+    this.onMouseMoveHandler = this.onMouseMoveHandler.bind(this);
+
+    this.onWheelHandler = this.onWheelHandler.bind(this);
+
+    window.addEventListener("mouseup", this.onMouseUpHandler);
```
bartaz: Do these need to be added when component is readonly?

static/js/publisher/release/components/progressiveBar.js
```diff
@@ -0,0 +1,160 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+class ProgressiveBar extends React.Component {
```
bartaz: Looking at the evolution of release UI components and recent refactorings it would be good to split this into smaller "view-only" component and "functionality" component that renders it.

In this case I suppose there would be a ProgressiveBar component that only recieves props telling it what to render (percentage). The other component will handle mouse changes and propagate them to the view one.
This would make progressive bar view more reusable in different context and hopefully make all components more focused.

static/js/publisher/release/helpers.js
```diff
@@ -1,15 +1,22 @@
...
-  if (branch) {
-    name = `${name}/${branch}`;
+  const obj = {
+    track,
+    risk,
+    branch
+  };
+
+  if (risk === AVAILABLE) {
+    delete obj.track;
```
bartaz: Let's not overcomplicate it. The `AVAILABLE` risk/channel is artificial and we just want to return it as it is. It's never in track, never has branches.

With that we also probably don't need to make `track` optional in `getChannelString`.

```suggestion
    return AVAILABLE;
```

Lukewh: Does it not need to return the branch too though - if it's present?

bartaz: The 'AVAILABLE' channel is the row where you select revisions you want to promote. It's not a real channel, it can't have branches.

Screenshot 2019-10-21 at 12 00 15

bartaz: It seemed to make sense to make it a fake channel to render it similar to other channels. Now we make it more separate (put in the tabs), so probably makes sense to refactor it at some point and remove this fake channel altogether (replace with separate 'selected revisions').

static/js/publisher/release/components/progressiveBar.js
```diff
@@ -1,7 +1,40 @@
 import React from "react";
 import PropTypes from "prop-types";
 
-class ProgressiveBar extends React.Component {
+const ProgressiveBar = ({ percentage, targetPercentage, readonly }) => (
```
bartaz: https://youtu.be/__X6cYJDXOg?t=11

Lukewh: I don't think the people that uploaded that video know what "High res" means.

static/js/publisher/release/components/revisionLabel.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 
 import { isInDevmode } from "../helpers";
 
-export default function DevmodeRevision({
+export default function RevisionLabel({
```
bartaz: Thx!

static/js/publisher/release/selectors/index.js
```diff
@@ -261,3 +261,34 @@ export function getRevisionsFromBuild(state, buildId) {
...
+
+  if (
+    release.length > 1 &&
+    release[0] &&
+    release[0] &&
+    release[0].progressive &&
+    release[0].progressive.key
+  ) {
+    progressive = release[0].progressive;
+    progressive.from = release[1].revision;
```
bartaz: I'm not sure if I understand here.

The `from` prop is meant to be revision we are "progressing" from. But, if `revision` is passed as param to this function, then `release` array will contain only items where `item.revision` equals `revision` from the param, so... `release[1].revision` will also (?).

I don't understand why we have the optional `revision` param to this function, what does it change? When we want to use it, and when not?
Tests seem to cover only calling function without revision, but in the code we seem to pass current revision to it...

Lukewh: Ah yes, this could definitely fail in some situations - I'll update the tests.
I added the optional `revision` for the history table. Simply asking if the `channel` and `arch` matched returned true for every revision in the history, the revision is meant to be a "is this specific revision being phased". Perhaps it doesn't need to be optional. I'll take a look.

bartaz: @Lukewh
It's fine if this function is both "give me progressive state for given channel/arch" (if revision is not included) or "give me progressive state of given revision in given channel/arch" (if revision is given).

But the way the `.from` is calculated is weird (or maybe just wrong) when there will be 2 progressive releases one after another of the same revision (when % is increased for example):

```
{
          architecture: "arch2",
          branch: null,
          track: "latest",
          risk: "stable",
          revision: "3",
          progressive: {
            key: "test",
            percentage: 60,
            paused: false
          }
        },
        {
          architecture: "arch2",
          branch: null,
          track: "latest",
          risk: "stable",
          revision: "3",
          progressive: {
            key: "test",
            percentage: 50,
            paused: false
          }
        },
        {
          architecture: "arch2",
          branch: null,
          track: "latest",
          risk: "stable",
          revision: "2",
          progressive: null
        }
```

I don't think we can assume that `release[1]` is the 'previous' revision. It can be previous release of the same revision. Especially if `revision` is passed as a parameter and releases are filtered based on it (all filtered items will be about only this revision).

static/js/publisher/release/selectors/index.js
```diff
@@ -261,3 +261,29 @@ export function getRevisionsFromBuild(state, buildId) {
...
+  );
+
+  let progressive = null;
+
+  const release = allReleases[releaseIndex];
+
+  if (release && release.progressive && release.progressive.key) {
+    progressive = release.progressive;
+
+    if (allReleases[releaseIndex + 1]) {
```
bartaz: We need to verify if that's correct (if there can be multiple progressive releases with same revisoin, or do they override each other).

Maybe add a TODO so we remember about it.

PR #2309, opened 2019-10-14
WIP: Releases lp tab

bartaz:


static/js/publisher/release/components/releasesTable/index.js
```diff
@@ -67,7 +87,117 @@ class ReleasesTable extends Component {
...
+
+    if (builds.length) {
+      return (
+        <Fragment>
+          <h5>Promote from recent Launchpad builds</h5>
+          {buildsToShow.map(revisions => this.renderBuildRow(revisions))}
+
+          {builds.length > MAX_BUILDS && (
+            <div className="p-releases-table__row--show-all">
+              <a onClick={this.handleToggleShowMoreBuilds.bind(this)}>
```
Lukewh: We could maybe think about making this into a wrapping component at some point, a "Show all x (y)" and "Show only z x". We have something similar for the branches.

PR #2310, opened 2019-10-15
compress store badges

sowasred2012:
## Done

- compressed all Snapcraft store badges to ~35% of their original size

## QA

- checkout the master branch
- in a file browser, from the project root, navigate to /static/images/badges
- inspect the file sizes of any of the SVGs within, see that they are all 20-30kb
- checkout this branch
- in a file browser, from the project root, navigate to /static/images/badges
- inspect the file sizes of any of the SVGs within, see that they are all 10kb or less


bartaz:
@sowasred2012 Great, thanks!

We take badges from https://github.com/snapcore/snap-store-badges/ so we will likely also need to add something to our pipeline to redo the optimisations whenever we update those.

PR #2316, opened 2019-10-18
Remove global max-width overrides. Add docs specific override

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2312

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004
- Browse the site, make sure headings and paragraphs don't look out of place
- Specifically check the docs pages.


static/sass/_snapcraft_p-docs.scss
```diff
@@ -5,6 +5,11 @@
     display: flex;
     margin: 0 auto;
 
+    // override for full width text
+    p {
+      max-width: 100%;
```
bartaz: Is this needed? Shouldn't we just stick to Vanilla defaults?

Lukewh: 35em to 40em doesn't fix this "issue" of paragraphs being narrower than tables and code blocks, so I figured we may as well let the letters free from this dystopian constraint.

Lukewh: https://github.com/canonical-web-and-design/vanilla-framework/issues/1794

matthewpaulthomas: Benefit: avoids raggedness from being only 5.9em short of the right edge of the column. Drawback: diverges from Vanilla. ¯\_(ツ)_/¯ UX+1

lyubomir-popov: A couple of questions:

- What would happen if at some point this docs site reverts to full width? Arguably, you would also need to override line-height from 1.5rem to something bigger, as suggested by the wikipedia page, to ease finding the beginning of the next line for example.

- Even if a paragraph is full width, it still has a ragged right edge, which doesn't match the tidy edge of another block element like a card, a notification, table or image. We have the means to make text blockier by force-justifying. From the perspective of similarity to a table, that would be an improvement. From the perspective of readability, that means varying the space between letters and words, which has been shown to worsen reading experience.

Lukewh: Happy to remove the override - I was just trying to directly address the raised issue

PR #2320, opened 2019-10-22
Update eslint and node-sass to latest versions

Lukewh:
## QA

- Pull the branch
- Ensure your dev environment still works ok
- Ensure building, linting and testing still work ok


static/js/base/ga.js
```diff
@@ -72,7 +72,9 @@ if (typeof dataLayer !== "undefined") {
       return;
     }
 
-    for (let key in events) {
+    let key;
```
Lukewh: I can't find a rule that allows this - it seems to be a recommended pattern

bartaz:
Probably need to update eslint plugins as well:
`warning " > eslint-plugin-react@7.10.0" has incorrect peer dependency "eslint@^3.0.0 || ^4.0.0 || ^5.0.0"`

static/js/base/ga.js
bartaz: I think is a eslint + babel-eslint bug: https://github.com/eslint/eslint/issues/12117

Maybe we should update all eslint related deps.

bartaz:
I have some problems running prettier. Maybe prettier plugin needs updating as well?

PR #2321, opened 2019-10-24
Show a message if no snaps are returned on store

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2334

## QA
- Pull the branch
- Add `featured_snaps = []` to `webapp/store/views.py:79`
- `./run`
- Visit http://0.0.0.0:8004/store - you should receive an error page

Screenshot_2019-11-04 Maintenance mode enabled

### DIFF
--- a/webapp/store/views.py
+++ b/webapp/store/views.py
@@ -62,9 +62,8 @@ def store_blueprint(store_query=None, testing=False):
 
         try:
             categories_results = api.get_categories()
-        except ApiError as api_error:
+        except ApiError:
             categories_results = []
-            status_code, error_info = _handle_errors(api_error)
 
         categories = logic.get_categories(categories_results)
 
@@ -72,8 +71,10 @@ def store_blueprint(store_query=None, testing=False):
             featured_snaps_results = api.get_searched_snaps(
                 snap_searched="", category="featured", size=10, page=1
             )
-        except ApiError:
+        except ApiError as api_error:
             featured_snaps_results = []
+            status_code, error_info = _handle_errors(api_error)
+            flask.abort(status_code)
 
         featured_snaps = logic.get_searched_snaps(featured_snaps_results)
 
--- a/webapp/store/views.py
+++ b/webapp/store/views.py
@@ -72,11 +72,13 @@ def store_blueprint(store_query=None, testing=False):
                 snap_searched="", category="featured", size=10, page=1
             )
         except ApiError as api_error:
-            featured_snaps_results = []
             status_code, error_info = _handle_errors(api_error)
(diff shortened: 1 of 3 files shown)


bartaz:
I wonder if we shouldn't just show an error page in such case.
Anything we show on this page won't work anyway probably - neither the search nor the 'See more'

Also, pretending it's a normal page makes the error message less 'visible', because the page just looks empty/broken until you read the text. It doesn't look like an error page.

Screenshot 2019-10-29 at 12 39 31

clagom:
If neither the search nor the 'see more...' would work then yes I would say an error page makes sense @bartaz. Can we find out that those won't work? If users can still do those actions then I think the solution in the screenshot would work, otherwise a more global error without further call to actions would do.

bartaz:
@clagom 

Because it's an error state from API it's really hard to say what will work and what wont.
It may have been just a temporary issue, so even refreshing the page may work or search or 'See more'. But if API is down probably neither of those will, but we wont know until user goes to these pages.

@Lukewh how is this state possible? API didn't return an error, but list of snaps was empty?

Lukewh:
@bartaz exactly. Perhaps a better solution would be not show the featured section at all, just have the search box?

clagom:
We have two examples here:
Screenshot 2019-10-29 at 13 30 23

Screenshot 2019-10-29 at 13 30 16

I would double check if we could at least still display the search bar

Lukewh:
This is now updated to simply 503 when featured snaps are not returned.

webapp/store/views.py
```diff
@@ -62,18 +62,19 @@ def store_view():
...
 
         categories = logic.get_categories(categories_results)
 
         try:
             featured_snaps_results = api.get_searched_snaps(
                 snap_searched="", category="featured", size=10, page=1
             )
-        except ApiError:
+        except ApiError as api_error:
             featured_snaps_results = []
```
bartaz: Probably not needed anymore if we just `flask.abort` couple lines later?

webapp/store/views.py
```diff
@@ -62,18 +62,19 @@ def store_view():
...
+            status_code, error_info = _handle_errors(api_error)
+            flask.abort(status_code)
```
bartaz: Should it be `return flask.abort` (at least it seems what we are doing elsewhere).
```suggestion
            return flask.abort(status_code)
```

webapp/store/views.py
Lukewh: The API returning non-200 and there being no snaps are two different cases. I figure we should show an error case for both, so the second check is for the latter case.

PR #2327, opened 2019-10-28
Progressive release apply

### LINKED ISSUE #2239: Progressive releases: On release, add option to include a progressive release %

bartaz:
Fixes #2239 

### QA

- ./run or https://snapcraft-io-canonical-web-and-design-pr-2327.run.demo.haus/
- go to releases page of any snap you own
- prepare release of any revisions on top of other revisions
- next to apply button there should be a component allowing to change the % of affected devices
- if 100% is chosen standard full release should happen when "Apply" is clicked
- if % smaller than 100% is chosen a progressive release should happen when "Apply" is clicked
- if invalid number is entered "Apply" should not be disabled
- when there is no pending releases (only pending closes) the % input should not show up
- when release to some % is made or canceled, next time release is pending % input should reset to 100%

Screenshot 2019-10-28 at 15 10 37


Lukewh:
Annoyingly, backspacing in the percentage it goes to 0, then typing "40" produces "040".

PR #2329, opened 2019-10-29
Update progressive releases

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2241

QA:

- Pull the branch
- `./run`
- Visit a snaps release page
- Make a progressive release
- Open the history for a revision and update the percentage
- Save your changes
- Try multiple different releases at once. closing channels etc.

### DIFF
--- a/static/js/publisher/release/actions/pendingReleases.js
+++ b/static/js/publisher/release/actions/pendingReleases.js
@@ -5,6 +5,8 @@ export const SET_PROGRESSIVE_RELEASE_PERCENTAGE =
   "SET_PROGRESSIVE_RELEASE_PERCENTAGE";
 export const UPDATE_PROGRESSIVE_RELEASE_PERCENTAGE =
   "UPDATE_PROGRESSIVE_RELEASE_PERCENTAGE";
+export const PAUSE_PROGRESSIVE_RELEASE = "PAUSE_PROGRESSIVE_RELEASE";
+export const RESUME_PROGRESSIVE_RELEASE = "RESUME_PROGRESSIVE_RELEASE";
 
 export function releaseRevision(revision, channel, progressive) {
   return {
@@ -32,6 +34,20 @@ export function updateProgressiveReleasePercentage(key, percentage) {
       key,
       percentage
     }
+  };
+}
+
+export function pauseProgressiveRelease(key) {
+  return {
+    type: PAUSE_PROGRESSIVE_RELEASE,
+    payload: key
+  };
+}
+
+export function resumeProgressiveRelease(key) {
+  return {
+    type: RESUME_PROGRESSIVE_RELEASE,
+    payload: key
   };
 }
 
--- a/static/js/publisher/release/actions/pendingReleases.js
+++ b/static/js/publisher/release/actions/pendingReleases.js
@@ -3,11 +3,13 @@ export const UNDO_RELEASE = "UNDO_RELEASE";
 export const CANCEL_PENDING_RELEASES = "CANCEL_PENDING_RELEASES";
 export const SET_PROGRESSIVE_RELEASE_PERCENTAGE =
   "SET_PROGRESSIVE_RELEASE_PERCENTAGE";
+export const UPDATE_PROGRESSIVE_RELEASE_PERCENTAGE =
+  "UPDATE_PROGRESSIVE_RELEASE_PERCENTAGE";
 
-export function releaseRevision(revision, channel) {
+export function releaseRevision(revision, channel, progressive) {
   return {
     type: RELEASE_REVISION,
-    payload: { revision, channel }
+    payload: { revision, channel, progressive }
   };
 }
 
@@ -16,6 +18,16 @@ import { getPendingChannelMap } from "../selectors";
 export function setProgressiveReleasePercentage(key, percentage) {
   return {
     type: SET_PROGRESSIVE_RELEASE_PERCENTAGE,
+    payload: {
+      key,
+      percentage
+    }
+  };
+}
+
+export function updateProgressiveReleasePercentage(key, percentage) {
+  return {
+    type: UPDATE_PROGRESSIVE_RELEASE_PERCENTAGE,
     payload: {
       key,
       percentage
--- a/static/js/publisher/release/actions/pendingReleases.js
+++ b/static/js/publisher/release/actions/pendingReleases.js
@@ -10,15 +10,10 @@ export const RESUME_PROGRESSIVE_RELEASE = "RESUME_PROGRESSIVE_RELEASE";
 
 import { getPendingChannelMap, hasRelease } from "../selectors";
 
-export function releaseRevision(
-  revision,
-  channel,
-  progressive,
-  canBeProgressive
-) {
+export function releaseRevision(revision, channel, progressive) {
   return {
     type: RELEASE_REVISION,
-    payload: { revision, channel, progressive, canBeProgressive }
+    payload: { revision, channel, progressive }
   };
 }
 
--- a/static/js/publisher/release/actions/pendingReleases.js
+++ b/static/js/publisher/release/actions/pendingReleases.js
@@ -8,14 +8,19 @@ export const UPDATE_PROGRESSIVE_RELEASE_PERCENTAGE =
 export const PAUSE_PROGRESSIVE_RELEASE = "PAUSE_PROGRESSIVE_RELEASE";
 export const RESUME_PROGRESSIVE_RELEASE = "RESUME_PROGRESSIVE_RELEASE";
 
-export function releaseRevision(revision, channel, progressive) {
+import { getPendingChannelMap, hasRelease } from "../selectors";
+
+export function releaseRevision(
+  revision,
+  channel,
+  progressive,
+  canBeProgressive
+) {
   return {
     type: RELEASE_REVISION,
-    payload: { revision, channel, progressive }
+    payload: { revision, channel, progressive, canBeProgressive }
   };
 }
-
-import { getPendingChannelMap } from "../selectors";
 
 export function setProgressiveReleasePercentage(key, percentage) {
   return {
@@ -65,8 +70,19 @@ export function promoteRevision(revision, channel) {
       );
     });
 
+    const canBeProgressive = revision.architectures.some(arch =>
+      hasRelease(getState(), channel, arch)
+    );
+
     if (!isAlreadyReleased) {
-      dispatch(releaseRevision(revision, channel));
+      dispatch(
+        releaseRevision(
+          revision,
+          channel,
+          undefined,
+          canBeProgressive ? true : undefined
+        )
+      );
     }
   };
 }
--- a/static/js/publisher/release/actions/pendingReleases.js
+++ b/static/js/publisher/release/actions/pendingReleases.js
@@ -10,10 +10,15 @@ export const RESUME_PROGRESSIVE_RELEASE = "RESUME_PROGRESSIVE_RELEASE";
 
 import { getPendingChannelMap, hasRelease } from "../selectors";
 
-export function releaseRevision(revision, channel, progressive) {
+export function releaseRevision(
+  revision,
+  channel,
+  progressive,
+  canBeProgressive
+) {
   return {
     type: RELEASE_REVISION,
-    payload: { revision, channel, progressive }
+    payload: { revision, channel, progressive, canBeProgressive }
   };
 }
 
--- a/static/js/publisher/release/actions/releases.js
+++ b/static/js/publisher/release/actions/releases.js
@@ -7,6 +7,7 @@ import { hideNotification, showNotification } from "./globalNotification";
 import { cancelPendingReleases } from "./pendingReleases";
 import { releaseRevisionSuccess, closeChannelSuccess } from "./channelMap";
 import { updateRevisions } from "./revisions";
+import { closeHistory } from "./history";
 
 import {
   fetchReleasesHistory,
@@ -120,11 +121,11 @@ export function releaseRevisions() {
   return (dispatch, getState) => {
     const { pendingReleases, pendingCloses, revisions, options } = getState();
     const { csrfToken, snapName, defaultTrack } = options;
-    const releases = Object.keys(pendingReleases).map(id => {
-      const pendingRelease = pendingReleases[id];
+    const releases = Object.keys(pendingReleases).map(key => {
+      const pendingRelease = pendingReleases[key];
 
       const release = {
-        id,
+        id: pendingRelease.revision.revision,
         revision: pendingRelease.revision,
         channels: pendingRelease.channels,
         progressive: pendingRelease.progressive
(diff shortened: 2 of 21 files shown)


static/js/publisher/release/reducers/pendingReleases.js
```diff
@@ -2,39 +2,44 @@ import {
...
+  PAUSE_PROGRESSIVE_RELEASE,
+  RESUME_PROGRESSIVE_RELEASE
 } from "../actions/pendingReleases";
+
 import { CLOSE_CHANNEL } from "../actions/pendingCloses";
 
 function removePendingRelease(state, revision, channel) {
-  if (state[revision.revision]) {
-    const channels = [...state[revision.revision].channels];
+  const releaseKey = `${revision.revision}-${channel}`;
```
bartaz: Why changing the key to be per revision-channel rather than just revision?

Lukewh: We were adding to the channel array if you were releasing the same revision to 2 or more channels.
This meant there was only 1 progressive release object for both of those channels.
This in turn meant you could not independently update a progressive release on one channel, and add a new progressive release on another channel with a different key/ percentage.

I thought about changing the progressive release object to be channel keyed:
`progressive: { ${channel}: {}}` but it became complicated very fast - this seemed simpler. Open to suggestions though!

bartaz: Ah, good point. 

Previous pending releases structure was 1:1 reflecting what we send to API, each 'key' (revision) was single API call (releasing a revision to list of channels).
It would be a pity if we loose that - if there is a key per revision-channel now, does it mean even standard "100%" releases are separated when on different channels? Releasing a revision to multiple channels would mean multiple API calls now?

I also need to check if there is no other code that relied on the fact that prendingRelease key is a revision (like counting how many revisions are to release or something)...

bartaz: That's what I was afraid:

Revisions to release are not counted properly now (when single revision is released to multiple channels):
Screenshot 2019-11-05 at 10 57 11


And the releases via API now happen in multiple calls (one per channel) instead of just one:

Screenshot 2019-11-05 at 10 57 41


Changes are certainly needed to the model, but we shouldn't make most common and default experience worse by adding progressive releases.

static/sass/_snapcraft_release.scss
```diff
@@ -387,6 +387,21 @@
           background-color: $color-x-light;
         }
       }
+
+      .p-revisions-list__revision-progressive {
+        display: flex;
+        flex-direction: row;
+
+        > :nth-child(1),
```
bartaz: Would be good not to style things based on their position in DOM.

static/js/publisher/release/actions/pendingReleases.js
```diff
@@ -37,8 +65,19 @@ export function promoteRevision(revision, channel) {
...
+    );
+
     if (!isAlreadyReleased) {
-      dispatch(releaseRevision(revision, channel));
+      dispatch(
+        releaseRevision(
+          revision,
+          channel,
+          undefined,
+          canBeProgressive ? true : undefined
```
bartaz: Is it used anywhere? It seems that this function takes 3 params, not 4: `export function releaseRevision(revision, channel, progressive) {`

Lukewh: Ah thanks, missed this one... I originally added this, but then realised I could just do it in one place

static/js/publisher/release/selectors/index.js
```diff
@@ -282,19 +287,53 @@ export function getProgressiveState(state, channel, arch, revision) {
...
+    if (
+      channel === pendingMatch.channel &&
+      pendingMatch.revision.architectures.includes(arch)
+    ) {
+      pendingProgressiveStatus = Object.assign({}, pendingMatch.progressive);
     }
   }
 
-  return progressive;
+  return [progressiveStatus, previousRevision, pendingProgressiveStatus];
```
bartaz: The comment for this function could be extended to explain better what is returned, especially that it's not just a single object but 3 values now.

static/js/publisher/release/reducers/pendingReleases.js
```diff
@@ -103,7 +167,9 @@ export default function pendingReleases(state = {}, action) {
       return releaseRevision(
         state,
         action.payload.revision,
-        action.payload.channel
+        action.payload.channel,
```
bartaz: Update the comment before the function with new pending releases structure

static/js/publisher/release/reducers/pendingReleases.js
```diff
@@ -2,85 +2,149 @@ import {
...
-    const channels = [...state[revision.revision].channels];
-
-    if (channels.includes(channel)) {
-      state = { ...state };
-      channels.splice(channels.indexOf(channel), 1);
-      state[revision.revision] = {
-        ...state[revision.revision],
-        channels
-      };
+  const newState = { ...state };
```
bartaz: `{ ...state }` only makes a shallow copy of the state, so any modifications on child object will be leaked.
Better to use `JSON.parse(JSON.stringify(state))` hack. We probably could extract it to some `deepClone` utility. 
```suggestion
  const newState = JSON.parse(JSON.stringify(state));
```

static/js/publisher/release/actions/releases.js
```diff
@@ -120,18 +121,23 @@ export function releaseRevisions() {
...
-        id,
-        revision: pendingRelease.revision,
-        channels: pendingRelease.channels,
-        progressive: pendingRelease.progressive
-      };
-
-      return release;
-    });
+    const releases = Object.keys(pendingReleases)
+      .map(key => {
```
bartaz: Would be nice to name the `key` as `id` or `revisionId` to make it clearer what the value is.

static/js/publisher/release/components/releasesTable/cellViews.js
```diff
@@ -61,8 +62,8 @@ export const RevisionInfo = ({
     buildIcon = <i className="p-icon--lp" />;
   }
 
-  if (progressiveState && progressiveState.from) {
-    from = progressiveState.from;
+  if (previousRevision && previousRevision.revision) {
+    from = previousRevision.revision;
```
bartaz: Why passing whole object if we only need the id?

static/js/publisher/release/components/revisionsListRow.js
```diff
@@ -1,49 +1,74 @@
...
 
   let progressiveState;
+  let previousRevision;
+  let pendingProgressiveState;
 
-  if (revision.release) {
-    progressiveState = getProgressiveState(
-      getChannelString(revision.release),
+  // Only show the progressive release status if it's the latest revision
+  if (index === 0 && revision.release) {
```
bartaz: This component doesn't care about `index` value. If the prop is meant to define if progressive release should be shown it should be `showProgressive` flag or something like this.

static/js/publisher/release/components/revisionsList.js
```diff
@@ -40,22 +40,26 @@ class RevisionsList extends Component {
...
       <RevisionsListRow
         key={`revision-row-${revision.revision}`}
+        index={index}
         revision={revision}
         isSelectable={isSelectable}
         showChannels={showChannels}
         isPending={isPending}
         isActive={isActive}
         showBuildRequest={showBuildRequest}
+        channel={activeChannel}
```
bartaz: Is `channel` used by `RevisionsListRow` component? I don't see it in the props there

static/js/publisher/release/components/revisionsListRow.js
```diff
@@ -114,10 +184,74 @@ const RevisionsListRow = props => {
       {showProgressiveReleases && (
         <td>
           {progressiveState && (
-            <ProgressiveBar
-              percentage={progressiveState.percentage}
-              readonly={true}
-            />
+            <div
```
bartaz: It would be good to extract it to separate component to contain its functionality outside of the row.

static/js/publisher/release/components/progressiveConfirm.js
```diff
@@ -1,11 +1,35 @@
...
+            <br />
+            {Object.keys(newReleases).map(revId => {
+              const release = newReleases[revId];
+
+              return (
+                <span key={revId}>
+                  <b>{release.revision.revision}</b> ({release.revision.version}
+                  ) {release.revision.architectures.join(", ")} to{" "}
+                  {release.channel}
+                  {"\n"}
```
bartaz: I think `<br />` would do the same and feel more natural in JSX.

static/js/publisher/release/components/revisionsListRow.js
```diff
@@ -69,6 +95,50 @@ const RevisionsListRow = props => {
 
   const showProgressiveReleases = isProgressiveReleaseEnabled && !showChannels;
 
+  const handlePauseProgressiveRelease = () => {
+    releaseRevision(revision, channel, progressiveState);
```
bartaz: Why both `releaseRevision` and `pauseProgressiveRelease` actions are needed to pause? (and resume)?

static/js/publisher/release/components/progressiveBar.js
```diff
@@ -54,18 +68,32 @@ class InteractiveProgressiveBar extends React.Component {
 
     this.onWheelHandler = this.onWheelHandler.bind(this);
 
-    window.addEventListener("mouseup", this.onMouseUpHandler);
-    window.addEventListener("mousemove", this.onMouseMoveHandler);
+    // If the element isn't disabled, add event liteners
+    if (!props.disabled) {
```
bartaz: Should they also be disabled when readonly?

static/js/publisher/release/components/progressiveBar.js
```diff
@@ -54,18 +68,32 @@ class InteractiveProgressiveBar extends React.Component {
...
   }
 
   componentWillUnmount() {
     window.removeEventListener("mouseup", this.onMouseUpHandler);
     window.removeEventListener("mousemove", this.onMouseMoveHandler);
   }
 
+  componentDidUpdate(prevProps) {
+    // We really don't want to flood the browser with event listeners
+    if (prevProps.disabled && !this.props.disabled) {
```
bartaz: Again, for readonly as well?

static/js/publisher/release/components/progressiveBar.js
```diff
@@ -152,25 +182,29 @@ class InteractiveProgressiveBar extends React.Component {
       <div
         className={classes.join(" ")}
         ref={this.barHolder}
-        onMouseDown={!readonly && this.onMouseDownHandler}
-        onWheel={!readonly && this.onWheelHandler}
+        onMouseDown={
+          !readonly && !disabled ? this.onMouseDownHandler : undefined
```
bartaz: Maybe combine readonly and disabled into `isInteractive`?

static/js/publisher/release/components/releasesConfirm.js
```diff
@@ -56,26 +57,111 @@ class ReleasesConfirm extends Component {
...
 
     const isPercentageValid = +percentage > 0 && +percentage <= 100;
 
+    // These are only used locally and therefore use revId-channel as a key, rather
+    // than the usual nested [revId][channel] format of pending releases
+    const progressiveUpdates = {};
+    const newReleases = {};
+    const newReleasesToProgress = {};
+
+    Object.keys(pendingReleases).forEach(revId => {
```
bartaz: This seems like a bit of data transformation. Could it be separated to outside function? Or a selector on the state maybe?

static/js/publisher/release/components/revisionsListRow.js
Lukewh: If the release isn't already pending, this just uses the established mechanism to add it to the pendingReleases state.
I guess `pause...`, `resume...` and `...update` could take all the arguments of `releaseRevision` and add it to `pending` if they didn't exist?

static/js/publisher/release/components/progressiveBar.js
Lukewh: For the `InteractiveProgressiveBar` I figured the name suggests it's not readonly, so I removed the prop (it just passes `readonly={false}` to `ProgressiveBar`. I guess I forgot to remove the `readonly` checks from render.

static/js/publisher/release/components/revisionsListRow.js
bartaz: Yes, would feel like a more intuitive API.
From a component level we don't care if it's already pending or not, just dispatch an action with what we want to pause/update. Reducer should figure out how state needs to be changed.

static/js/publisher/release/reducers/pendingReleases.js
```diff
@@ -92,7 +156,9 @@ function updateProgressiveRelease(state, progressive) {
 // {
 //  <revisionId>: {
 //    revision: { revision: <revisionId>, version, ... },
-//    channels: [ ... ]
+//    channel: channel,
```
bartaz: There is still a `<channel>` key missing under `<revisionId>`.

static/js/publisher/release/components/revisionsListRowProgressive.js
```diff
@@ -0,0 +1,186 @@
...
+
+  const handleProgressiveChange = percentage => {
+    if (!pendingProgressiveState) {
+      releaseRevision(revision, channel, progressiveState);
+    }
+    updateProgressiveReleasePercentage(progressiveState.key, percentage);
+  };
+
+  return (
+    <td>
```
bartaz: Dropping the `<td>` from here (leaving it on row component level) would make it reusable anywhere outside of the table. Unless it depends on being in the table cell (which I don't think is the case).

static/js/publisher/release/components/releasesConfirm.js
```diff
@@ -56,26 +59,59 @@ class ReleasesConfirm extends Component {
...
-      isProgressiveReleaseEnabled
+      isProgressiveReleaseEnabled,
+      separatePendingReleases
     } = this.props;
-    const releasesCount = Object.keys(pendingReleases).length;
     const closesCount = pendingCloses.length;
 
     const isPercentageValid = +percentage > 0 && +percentage <= 100;
 
+    const {
```
bartaz: Nit picking, but if these are passed from redux we could map them to individual promps in `mapStateToProps` so component doesn't rely on the format of the data.

bartaz:
QA notes:

- disabling progressive feature branch make it break (with errors in the console)
- then there is progressive release closing channel still shows progressive bar in the cell
- when canceling progressive release previous release is added on top of history but it's missing a cell (row is shorter)
- have a pending release, update the % (without saving), pause it, resume again - updates disappears from confirm box, but % change is visible in the bar in history
- have a progressive release, cancel it (without saving) - in history the previously progressive row is not draggable, revert changes (to cancel the cancel), progressive row is back, but it's not draggable
- pending progressive updates are not visible in releases table cell (cell is not marked pending, progressive bar is not updated)

PR #2330, opened 2019-10-29
Make releases confirm always visible and consistent with other forms

bartaz:
As a preparation to making releases confirm sticky and possibly with an accordion with changes, this PR makes the releases confirm bar always visible and consistent with other forms in publisher pages.

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-2330.run.demo.haus/
- go to releases page
- Save/Revert bar should be always visible and should have styling similar to Listing/Settings page
- Button labels and order should be consistent as well (Revert+Save instead of Apply+Cancel)
- when releases or closes are pending the bar should update as before, showing a summary of changes with tooltips, making buttons enabled, etc
- it should still work (releasing/closing should work as before)

Screenshot 2019-10-29 at 16 47 30

### DIFF
--- a/static/js/publisher/release/components/releasesConfirm.js
+++ b/static/js/publisher/release/components/releasesConfirm.js
@@ -74,8 +74,8 @@ class ReleasesConfirm extends Component {
     const showProgressive = isProgressiveReleaseEnabled && releasesCount > 0;
 
     return (
-      <div className="p-releases-confirm u-vertically-center">
-        <div>
+      <div className="p-releases-confirm u-vertically-center row">
+        <div className="col-5">
           {releasesCount > 0 && (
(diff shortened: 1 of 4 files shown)


kwm14:
@bartaz looking good, just one comment...is there anything we can do on mobile with the buttons?

Screenshot 2019-10-29 at 16 48 51

Potentially knock them down as seen on the /listing page...

Screenshot 2019-10-29 at 16 50 01

bartaz:
@kwm14 Updated for better small screen treatment:

Screenshot 2019-10-30 at 10 00 23

kwm14:
Nice. Thanks @bartaz LGTM 👍

PR #2332, opened 2019-10-31
Add trending badge option to publicise page

### LINKED ISSUE #2324: Update publicise page with new options

bartaz:
Fixes #2324 

Adds option to add trending badge on publicise page

### QA
- ./run or https://snapcraft-io-canonical-web-and-design-pr-2332.run.demo.haus/
- go to publicise page of any snap (not trending)
- choose GitHub badge, an option to show trending badge should be available
- click to choose trending badge
- preview should show trending badge with a warning message that this is just a preview (because snap is not trending)
- code should update to include additional trending badge
- using the code in external service should not show the trending badge (as snap is not trending)
- go to publicise page of trending snap https://snapcraft-io-canonical-web-and-design-pr-2332.run.demo.haus/kdenlive/publicise/badges
- trending badge should be checked (visible) by default (only if snap is currently trending)
- try different options on setting badges on/off

Screenshot 2019-11-04 at 16 55 37
Screenshot 2019-11-04 at 16 55 28

### DIFF
--- a/webapp/store/snap_details_views.py
+++ b/webapp/store/snap_details_views.py
@@ -352,6 +352,7 @@ def snap_details_views(store, api, handle_errors):
 
     @store.route('/<regex("' + snap_regex + '"):snap_name>/trending.svg')
     def snap_details_badge_trending(snap_name):
+        is_preview = flask.request.args.get("preview", default=0, type=int)
         context = _get_context_snap_details(snap_name)
 
         # default to empty SVG
@@ -360,7 +361,20 @@ def snap_details_views(store, api, handle_errors):
             'xmlns:xlink="http://www.w3.org/1999/xlink"></svg>'
         )
 
-        if context["trending"]:
+        # publishers can see preview of trending badge of their own snaps
+        # on Publicise page
+        show_as_preview = False
+        if is_preview and authentication.is_authenticated(flask.session):
+            if (
+                flask.session.get("openid").get("nickname")
+                == context["username"]
+            ) or (
+                "user_shared_snaps" in flask.session
+                and snap_name in flask.session.get("user_shared_snaps")
+            ):
+                show_as_preview = True
+
+        if context["trending"] or show_as_preview:
             svg = get_badge_svg(
                 snap_name=snap_name,
                 left_text=context["snap_title"],
--- a/static/js/publisher/publicise.js
+++ b/static/js/publisher/publicise.js
@@ -225,7 +225,9 @@ const getBadgesHTML = (snapName, options) => {
 };
 
 const getBadgesPreviewHTML = (snapName, options) => {
-  return getBadgesCode(snapName, options, getBadgePreview);
+  const code = getBadgesCode(snapName, options, getBadgePreview);
+
+  return code || "<em>select a badge to display</em>";
 };
 
 const getBadgesMarkdown = (snapName, options) => {
--- a/static/js/publisher/publicise.js
+++ b/static/js/publisher/publicise.js
@@ -246,9 +246,7 @@ const getBadgesHTML = (snapName, options) => {
 };
 
 const getBadgesPreviewHTML = (snapName, options) => {
-  const code = getBadgesCode(snapName, options, getBadgePreview);
-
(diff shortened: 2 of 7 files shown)


Lukewh:
Screenshot_2019-10-31  Publicise Lukotron testing
Maybe we should just not show the code blocks if nothing is selected?


Screenshot_2019-10-31  Publicise Lukotron testing (1)
I think it would be nice to show what the trending one would look like on this page. People might not want to put something on their site they have no idea what it looks like. Can we send it through from flask in the python context using the same logic that builds the badges so it's not a public url?

bartaz:
> I think it would be nice to show what the trending one would look like on this page. People might not want to put something on their site they have no idea what it looks like. 

I agree, that it would be nice to have some 'preview' of what trending badge may look like for you.
But, we also need to make sure users understand it's a preview and they will not see it until they are trending. 

@hughesjordan Any ideas about that? Basically when your snap is not trending (which is the case for most snaps) we want to show a preview of what trending badge may look like, but also making clear that it will not show up (until snap is trending). 


> Can we send it through from flask in the python context using the same logic that builds the badges so it's not a public url?

@Lukewh Not sure what you mean by that?

hughesjordan:
@Lukewh @bartaz the initial idea was that on this page the trending badge would always display in the preview when selected, as you say, to give an idea to the user what it would look like. (Apologies not properly explained at any point)

The explanatory text next to the option would hopefully make it clear to the user that it may not be present when used on external sites (may have to revisit wording - 'Only displays on external sites when your snap is flagged as trending'). If needs be we could add further messaging underneath the badges but may be overkill.

Agreed, I think if nothing has been selected we should hide the code blocks and instead show some messaging.

For me the trending badge is not selected by default.

There's a typo in the message next to the trending option 'flaged'.

hughesjordan:
@bartaz How about these?

#### Additional messaging


#### Messaging when nothing is selected

clagom:
I would use a simpler message for the preview bit when no badges are selected, without the boxed-style, the current one is fine, maybe with a longer message (as Jordan's design):
Screenshot 2019-11-04 at 10 21 04
The reason being, otherwise that notification UI seems to be what you get when no badges are selected.

When trending badge is selected, but your snap is not trending, I still would display the trending badge, as per Jordan's design. We would still display the disclaimer message anyway.

hughesjordan:
Good point @clagom 👍

webapp/store/snap_details_views.py
```diff
@@ -352,6 +352,7 @@ def snap_details_badge(snap_name):
 
     @store.route('/<regex("' + snap_regex + '"):snap_name>/trending.svg')
     def snap_details_badge_trending(snap_name):
+        is_preview = flask.request.args.get("preview", default=0, type=int)
```
Lukewh: I don't like having a "preview" arg that anyone can use - we need to verify server-side imo.

Untested:

```suggestion
        can_preview = flast.request.headers.get("Host") == "snapcraft.io"
        is_preview = flask.request.args.get("preview", default=0, type=int) if can_preview
```

clagom:
I would still display the trending badge in this preview
Screenshot 2019-11-04 at 16 09 11
I think with that disclaimer it's pretty clear that you will see it only when your snap is trending, but now we completely hide it from the users, not sure...

bartaz:
@clagom 

The badge should be there (as preview), not sure why it would not show up for you...

Screenshot 2019-11-04 at 17 36 58

bartaz:
> Sorry added comment before I saw the authentication bit 👍

@Lukewh 

Just to be 100% clear how it works now.
For trending badge to show up even if snap is not trending 2 things need to happen: preview=1 needs to be added to URL, and person requesting the badge needs to be signed into snapcraft as an owner/collaborator of the snap.

Because I don't verify referer or anything like that it's still possible for someone to put the badge with ?preview=1 on their external site. But it will only work for signed in owners of given snap.

clagom:
I see this @bartaz, including the inspector:
Screenshot 2019-11-04 at 16 41 37

clagom:
See it now!

PR #2352, opened 2019-11-13
Ensure revision can be dragged after progressive release is can…

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2349

## QA
- Pull the branch
- `./run`
- Create a progressive release
- Save
- Cancel the progressive release - the row should be draggable in history
- Revert the changes - the row should be draggable in history


bartaz:
Would be nice if we can change the cursor when draggable is off.
There are places (like between pause and the bar) where icon is grabbing hand, while it is not draggable there.

bartaz:
Good to go, thanks!

PR #2353, opened 2019-11-13
Update progressiveBar target % when props change

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2343
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2362

## QA
- Pull the branch
- `./run`
- Progressively release a revision
- Once released, slide the progressive percentage bar
- Revert the changes
- The progressive release bar should reset to its previous value

### DIFF
--- a/static/js/publisher/release/actions/releases.js
+++ b/static/js/publisher/release/actions/releases.js
@@ -130,7 +130,11 @@ export function releaseRevisions() {
             id: pendingRelease.revision.revision,
             revision: pendingRelease.revision,
             channels: [pendingRelease.channel],
-            progressive: pendingRelease.progressive
+            progressive:
+              pendingRelease.progressive &&
+              pendingRelease.progressive.percentages < 100
+                ? pendingRelease.progressive
+                : null
           };
 
           return release;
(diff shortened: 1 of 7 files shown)


bartaz:
Something strange happens when you drag to 100% or further. Progress seems to be broken after that.

Lukewh:
@bartaz ah! I imagine because at that point it is no longer classed as a "progressive release" so some of the UI updates to reflect that.

bartaz:
There is an issue with pending update dissapearing after pausing:

progress

bartaz:
Not sure if it's intended or not, but for sure not very clear - if you update the % and cancel before saving, and then save - both things are updated (% and pause), so you end up with changed % and save.

But tooltip on confirm only tells about pause.

Probably no need to fix it in this PR, but would be good to fix it separately.

Lukewh:
> Not sure if it's intended or not, but for sure not very clear - if you update the % and cancel before saving, and then save - both things are updated (% and pause), so you end up with changed % and save.
> 
> But tooltip on confirm only tells about pause.
> 
> Probably no need to fix it in this PR, but would be good to fix it separately.

Ah minor change needed - will fix quickly here

bartaz:
@Lukewh oh, cool, close this one when you do in this case: https://github.com/canonical-web-and-design/snapcraft.io/issues/2362 as I just added it

PR #2357, opened 2019-11-14
Change `canBeProgressive` into `previousRevisions`

### LINKED ISSUE #2346: Progressive releases: investigate if `canBeProgressive` needs to be part of state or can be computed when needed

bartaz:
Fixes #2346 

### QA
- make sure everything works as before (around releasing progressive releases and not allowing progressive releases when there is no revision already released)


static/js/publisher/release/actions/pendingReleases.js
```diff
@@ -70,17 +70,25 @@ export function promoteRevision(revision, channel) {
       );
     });
 
-    const canBeProgressive = revision.architectures.some(arch =>
-      hasRelease(getState(), channel, arch)
-    );
+    const previousRevisions = {};
+
+    revision.architectures.forEach(arch => {
```
Lukewh: should we `const { channelMap } = getState()` this, so we don't need to call `getState()` twice - not sure how performant the function is in redux?

bartaz: Good call, updated.

PR #2358, opened 2019-11-15
Fetch snap builds from launchpad

### LINKED ISSUE #2356: Backend endpoint and pass payload to view

jpmartinspt:
## Done

Fetches snap builds from launchpad, based on the snap's store name.
Passes the builds to the template under `snap_builds` and the build repository under `snap_builds_git`.

## QA
TBD

Fixes #2356

### DIFF
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -20,6 +20,7 @@ from webapp.api.exceptions import (
     MacaroonRefreshRequired,
     MissingUsername,
 )
+from webapp.api.launchpad import launchpad
 from webapp.api.store import StoreApi
 from webapp.store.logic import (
     get_categories,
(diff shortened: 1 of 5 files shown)


webapp/publisher/snaps/views.py
```diff
@@ -1308,3 +1309,53 @@ def post_preview(snap_name):
...
+            builds.append(
+                {
+                    "arch_tag": build.arch_tag,
+                    "channels": build.channels,
+                    "datebuilt": build.datebuilt,
+                    "duration": build.duration,
+                    "link": build.web_link,
+                    "revision_id": build.revision_id,
+                    "title": build.title,
+                }
```
bartaz: I'm not sure what `channels` are in terms of builds, I don't know if it has anything to do with store channels, so we may not need it.

What we want for sure is build status.
If I recall correctly there are 2 fields - status of the build itself, and status of the store upload, and in BSI we combine them into one somehow - I don't recall the exact logic. Maybe we could replicate it here, or at least have both fields in the context for front-end to figure out the status.

For the `web_link` - I'm not sure if we want to link directly to LP build page. It may actually be useful to link to BSI page instead. This would require building the URL from BSI domain + github handle + build id. We can do it on front-end, but it's a good candidate for backend logic as well.

Please also include the log link.

webapp/publisher/snaps/views.py
```diff
@@ -1308,3 +1309,53 @@ def post_preview(snap_name):
...
+    except ApiError as api_error:
+        return _handle_errors(api_error)
+
+    context = {
+        "snap_id": details["snap_id"],
+        "snap_name": details["snap_name"],
+        "snap_title": details["title"],
+    }
+
+    snap = launchpad.snaps.findByStoreName(
```
bartaz: I find it this `snap` name a bit confusing. Usually our snap data comes from store API, but this one comes from LP, so is not the same as store snap. Also, it seems that it's a list rather than single snap (even if we are interested in just the first one).

Could we call this `lp_snaps` or something like that (to suggest it's a list and where it comes from)?

jpmartinspt:
@bartaz Updated most of it. Is this what you were referring to in terms of the build status?
https://github.com/jpmartinspt/build.snapcraft.io/blob/master/src/common/helpers/snap-builds.js

PR #2361, opened 2019-11-19
Progressive release cancelation UI

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2342

## QA
- Pull branch
- `./run`
- If you don't have a progressive release - create one and save
- Cancel the progressive release
- See a new item in the save bar regarding "cancelling" a progressive release

### DIFF
--- a/static/js/publisher/release/actions/pendingReleases.test.js
+++ b/static/js/publisher/release/actions/pendingReleases.test.js
@@ -230,7 +230,22 @@ describe("pendingReleases actions", () => {
 
   describe("cancelProgressiverelease", () => {
     const key = "progressive-test";
-    const previousRevision = "previous-revision";
+    const previousRevision = {
+      architectures: ["amd64"],
+      attributes: {},
+      base: "core18",
+      build_url: null,
+      channels: ["latest/edge"],
+      confinement: "strict",
+      created_at: "2019-07-16T08:58:04Z",
+      epoch: { read: null, write: null },
+      grade: "stable",
+      revision: 3,
+      "sha3-384": "test",
+      size: 4096,
+      status: "Published",
+      version: "1.8.0"
+    };
 
     it("should create an action to cancel a release", () => {
       expect(cancelProgressiveRelease(key, previousRevision).type).toBe(
--- a/static/js/publisher/release/actions/pendingReleases.test.js
+++ b/static/js/publisher/release/actions/pendingReleases.test.js
@@ -13,6 +13,7 @@ import {
   UPDATE_PROGRESSIVE_RELEASE_PERCENTAGE,
   PAUSE_PROGRESSIVE_RELEASE,
   RESUME_PROGRESSIVE_RELEASE,
+  CANCEL_PROGRESSIVE_RELEASE,
   releaseRevision,
   promoteRevision,
   promoteChannel,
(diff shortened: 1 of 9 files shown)


static/js/publisher/release/actions/pendingReleases.test.js
```diff
@@ -225,4 +227,27 @@ describe("pendingReleases actions", () => {
       expect(resumeProgressiveRelease(key).payload).toBe(key);
     });
   });
+
+  describe("cancelProgressiverelease", () => {
+    const key = "progressive-test";
+    const previousRevision = "previous-revision";
```
bartaz: Would be good for this to be the same as in real case.
Is `previousRevision` supposed to be revision id? Whole revision object? This test doesn't tell which one.

static/js/publisher/release/reducers/pendingReleases.js
```diff
@@ -154,6 +155,27 @@ function resumeProgressiveRelease(state, key) {
   return nextState;
 }
 
+function cancelProgressiveRelease(state, key, previousRevision) {
```
bartaz: Is this supposed to cancel all pending revisons with the same key?

What would happen if several revisions across different architectures (for example whole channel) was released progressively with the same key.
If `cancelProgressiveRelease` cancels all releases with the same key, will it cancel them all? How can only be just one `previousRevision`?

static/js/publisher/release/reducers/pendingReleases.js
```diff
@@ -154,6 +155,27 @@ function resumeProgressiveRelease(state, key) {
...
+    const pendingReleaseChannels = nextState[revision];
+    Object.keys(pendingReleaseChannels).forEach(channel => {
+      const pendingRelease = pendingReleaseChannels[channel];
+
+      if (
+        pendingRelease.progressive &&
+        pendingRelease.progressive.key === key
+      ) {
+        nextState = releaseRevision(state, previousRevision, channel, null);
+        nextState[previousRevision.revision][channel].replaces = pendingRelease;
```
bartaz: Please add `replaces` to the object schema in the comment below.

static/js/publisher/release/reducers/pendingReleases.test.js
```diff
@@ -776,4 +777,86 @@ describe("pendingReleases", () => {
...
+        const result = pendingReleases(
+          nonProgressiveState,
+          cancelProgressiveReleaseAction
+        );
+
+        expect(result).toEqual(nonProgressiveState);
+      });
+    });
+
+    describe("when there is a progressive pendingRelease", () => {
```
bartaz: I think there is a missing case when there are multiple releases with same progressive key

Lukewh: Added - this also clarifies the expected (if wrong, currently) behaviour a little

static/js/publisher/release/reducers/pendingReleases.js
Lukewh: I've added a comment above this function to explain a little

static/js/publisher/release/actions/pendingReleases.test.js
```diff
@@ -225,4 +227,42 @@ describe("pendingReleases actions", () => {
...
+      channels: ["latest/edge"],
+      confinement: "strict",
+      created_at: "2019-07-16T08:58:04Z",
+      epoch: { read: null, write: null },
+      grade: "stable",
+      revision: 3,
+      "sha3-384": "test",
+      size: 4096,
+      status: "Published",
+      version: "1.8.0"
```
bartaz: Wow, providing just an object with id or any necessary fields would be enough, but sure, we can be very specific ;)

PR #2367, opened 2019-11-20
Show progressive release cancel in history table

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2366

## QA
- Pull this branch
- `./run`
- Create a progressive release if one does not yet exist
- Cancel said progressive release and note the message

Screenshot_2019-11-20  Releases and revision history for lukewh-test


static/js/publisher/release/components/revisionsListRow.js
```diff
@@ -123,25 +124,30 @@ const RevisionsListRow = props => {
             )}
         </td>
       )}
+      {progressiveBeingCancelled && (
+        <td>
+          <em>Cancel &amp; release previous revision</em>
```
bartaz: Maybe I read too much into it, but it sounds to me like it's the previous revision that is canceled and released.

static/js/publisher/release/releasesState.js
```diff
@@ -126,7 +126,7 @@ function getPendingRelease(pendingReleases, arch, channel) {
         channel
       ].revision.architectures.includes(arch)
     ) {
-      pendingRelease = releasedRevision;
+      pendingRelease = pendingReleases[releasedRevision][channel];
```
bartaz: Could it just `return` immediately?

static/js/publisher/release/releasesState.test.js
```diff
@@ -0,0 +1,30 @@
+import { getPendingRelease } from "./releasesState";
```
bartaz: `releasesState` is a bunch of legacy pre-redux code that stays here because it was not yet migrated to more appropriate place.

Instead of adding test to it, we should probably move it somewhere where it belongs better.
Seems like functionality like that would work as helper or even a selector.

static/js/publisher/release/components/revisionsListRow.js
Lukewh: I've tried to make it fit on 1 line. Maybe just "Cancel"?

bartaz:
The cancel row in history is gray, while it should probably be highlighted (because it's a pending change).

Also it's draggable, which it probably shouldn't:

Screenshot 2019-11-21 at 10 04 38

PR #2370, opened 2019-11-21
Show pending progressive updates

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2350

## QA
- Pull the branch
- `./run`
- Progressively release a revision (if one does not already exist)
- Save
- Update the progressive release:
  - any action should cause the cell to be highlighted
  - updating the percentage should show the target percentage when history is closed

### DIFF
--- a/static/js/publisher/release/components/releasesTable/cellViews.js
+++ b/static/js/publisher/release/components/releasesTable/cellViews.js
@@ -47,13 +47,66 @@ EmptyInfo.propTypes = {
   trackingChannel: PropTypes.string
 };
 
+const ProgressiveState = ({
+  revision,
+  previousRevision,
+  progressiveState,
+  pendingProgressiveState
+}) => {
+  let previousRevisionInfo = "";
+  let revisionInfo = "";
+  if (progressiveState) {
+    previousRevisionInfo = ` (${100 - progressiveState.percentage}%`;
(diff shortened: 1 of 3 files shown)


bartaz:
Tooltip doesn't mention the change in % (only current values in progress), which fills inconsistent with current updated pending progress:

Screenshot 2019-11-22 at 07 52 32

Also, colours of current/pending % values are almost indistinguishable (OMG, is this a real word?).
In the interactive one it looks better because it's bigger and there is a clear white border between the two.

bartaz:
Not sure if that's related to this PR changes, but haven't noticed it before to such extend - every time I try to change the progressive release bar value I start dragging the whole row, only after I click to cancel the drag I can update the value.

static/js/publisher/release/components/releasesTable/cellViews.js
```diff
@@ -47,13 +47,66 @@ EmptyInfo.propTypes = {
   trackingChannel: PropTypes.string
 };
 
+const ProgressiveState = ({
```
bartaz: At first I thought it's rendering progress bar or something.
Name of the component doesn't suggest it's supposed to be used in tooltip - it would be good if it does. `ProgressiveTooltipMessage` maybe?

PR #2376, opened 2019-11-22
Automate publisher pages

Lukewh:
## QA

- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/publisher/jetbrains
- KDE page should have more snaps compared to production
- Jetbrains page should be the same


webapp/store/views.py
```diff
@@ -306,6 +306,38 @@ def publisher_details(publisher):
...
+            for publisher in context["publishers"]:
+                try:
+                    searched_results = api.get_searched_snaps(
+                        "publisher:" + publisher, size=500, page=1
+                    )
+                except ApiError as api_error:
+                    status_code, error_info = _handle_errors(api_error)
+
+                snaps_results = logic.get_searched_snaps(searched_results)
+                context["snaps"] = context["snaps"] + [
```
jpmartinspt: Does `context["snaps"].extend([snap for snap in snaps_results if snap["apps"]])` work?

PR #2385, opened 2019-12-02
Releases accordion

Lukewh:
Fixes: https://github.com/canonical-web-and-design/snapcraft.io/issues/2371
Fixes: https://github.com/canonical-web-and-design/snapcraft.io/issues/2373
Fixes: https://github.com/canonical-web-and-design/snapcraft.io/issues/2372

## QA
- Pull branch
- `./run`
- Use the release UI as normal
- Things that won't work properly - updating and cancelling progressive releases
- Everything else should be ok (or have bugs)

### Closed
Screenshot_2019-12-03  Releases and revision history for lukewh-test (3)
### Open
New progressive release, progressive release % update, release to a a channel+arch with no previous release, closing channel
Screenshot_2019-12-03  Releases and revision history for lukewh-test (4)

New progressive release, progressive release pause, release to a a channel+arch with no previous release, closing channel
Screenshot_2019-12-03  Releases and revision history for lukewh-test (5)

New progressive release, progressive release cancel, release to a a channel+arch with no previous release, closing channel
Screenshot_2019-12-03  Releases and revision history for lukewh-test (6)

TODO:
- [x] Update tests.
- [x] Move setting of progressive key on release


bartaz:
I see a weird case, I'm not sure why:

https://snapcraft-io-canonical-web-and-design-pr-2385.run.demo.haus/lukewh-test/releases

Open latest/beta/amd64 to see history, and drag'n'drop the revision 3 to the latest/beta/amd64 cell:
Screenshot 2019-12-03 at 15 54 20
It shows correctly that it can be released progressively:

Screenshot 2019-12-03 at 15 55 37


But open history of 'Available revisions' amd64 and take the same revision 3 from there and drop on latest/beta/amd64
Screenshot 2019-12-03 at 15 56 51

And confirm says it cannot be released progressively

Screenshot 2019-12-03 at 15 57 06

static/js/publisher/release/components/releasesConfirm.js
```diff
@@ -56,207 +53,95 @@ class ReleasesConfirm extends Component {
...
-            <button
-              className="p-button--positive is-inline u-no-margin--bottom u-no-margin--right"
-              disabled={!isApplyEnabled}
-              onClick={this.onApplyClick.bind(this)}
-            >
-              {isLoading ? "Loading..." : "Save"}
-            </button>
+      <Fragment>
+        <div className="p-releases-confirm">
+          {showDetails && <ReleasesConfirmDetails updates={updates} />}
```
bartaz: Won't comment on the design to much, but it felt a bit unexpected that the details appear on top of confirm.
They also feel a bit detached.

static/js/publisher/release/components/releasesConfirm.js
```diff
@@ -56,207 +53,95 @@ class ReleasesConfirm extends Component {
...
-                      </span>
-                    );
-                  })}
-                </span>
-              </span>{" "}
-              to release.
-            </Fragment>
-          )}{" "}
-          {progressiveUpdatesCount > 0 && (
-            <Fragment>
```
bartaz: Nice that we got rid of this stuff from here

static/js/publisher/release/reducers/pendingReleases.js
```diff
@@ -68,6 +68,14 @@ function releaseRevision(
 
   if (previousRevisions) {
     state[revision.revision][channel].previousRevisions = previousRevisions;
+
+    if (!progressive) {
```
bartaz: I think that it would be better to keep logic like this in action.

This makes the reducer depend on current time, which makes it less 'pure' and harder to test.
Sure, moving it to action would do the same to the action, but these are generally more complicated anyway.

I guess ideally we would want it out of redux completely and passed by component, but I suppose this would complicate other things.

static/js/publisher/release/components/releasesConfirmDetails.js
```diff
@@ -0,0 +1,306 @@
...
+          onChange={updatePercentage}
+          targetPercentage={percentage}
+          minPercentage={1}
+          singleDirection={type === progressiveTypes.UPDATE ? 1 : 0}
+        />
+        <span>
+          <span className="p-tooltip--btm-right">
+            <span className="p-help">{percentage}% of devices</span>
+            <span className="p-tooltip__message">
+              Releases are delivered to devices via snap refreshes, as such, is
```
bartaz: Typo, should be `it may...` or `this may...`?
```suggestion
              Releases are delivered to devices via snap refreshes, as such, it
```

static/js/publisher/release/actions/releases.js
```diff
@@ -118,16 +118,27 @@ export function handleReleaseResponse(
 }
 
 export function releaseRevisions() {
+  const progressiveKey = `ui-progressive-release-${new Date().getTime()}`;
   const mapToRelease = pendingRelease => {
+    let progressive = null;
+
+    if (
+      pendingRelease.progressive &&
+      pendingRelease.progressive &&
```
bartaz: Double checking? ;)

static/js/publisher/release/actions/pendingReleases.js
```diff
@@ -9,17 +9,38 @@ export const PAUSE_PROGRESSIVE_RELEASE = "PAUSE_PROGRESSIVE_RELEASE";
...
+      state,
+      revision.architectures,
+      channel
+    )
+      .filter(release => release.revision !== revision.revision)
+      .map(release => revisions[release.revision]);
+
+    if (!progressive) {
+      progressive = {
+        key: null,
```
bartaz: It feels a bit weird to set it to null, so maybe worth adding a comment about when the key is generated?

static/js/publisher/release/selectors/index.js
```diff
@@ -386,27 +392,15 @@ export function getSeparatePendingReleases(state) {
           });
         }
 
-        if (changes.length > 0) {
+        if (previousState.key) {
```
bartaz: Now that the key can be `null` this won't cause problems?

static/sass/_snapcraft_release.scss
```diff
@@ -17,30 +17,67 @@
...
+
+    &.is-closing {
+      grid-template-columns: auto;
+    }
+
+    &:nth-child(odd) {
+      background: $color-light;
+    }
+
+    &-join {
```
bartaz: If it contains sub-elements maybe it should be a top level pattern of its own, rather than `__row` in details?
This gets very nested.

But I wont help with naming all those things ;) So kinda understand why it happened...

static/js/publisher/release/components/releasesConfirm.js
```diff
@@ -56,207 +53,95 @@ class ReleasesConfirm extends Component {
...
-              Revert
-            </button>
-            <button
-              className="p-button--positive is-inline u-no-margin--bottom u-no-margin--right"
-              disabled={!isApplyEnabled}
-              onClick={this.onApplyClick.bind(this)}
-            >
-              {isLoading ? "Loading..." : "Save"}
-            </button>
+      <Fragment>
```
bartaz: Probably no need for that in this PR, but the 'confirm' set of components has grown enough to be a good candidate for another round of refactoring - to split pure visual stateless components from logic and data handling.

static/js/publisher/release/components/releasesConfirmDetails.js
```diff
@@ -0,0 +1,306 @@
...
+import { ProgressiveBar, InteractiveProgressiveBar } from "./progressiveBar";
+
+const progressiveTypes = {
+  RELEASE: "release",
+  UPDATE: "update",
+  CANCELLATION: "cancel"
+};
+
+const ReleaseRow = ({ type, revisionInfo, channel, progress, notes }) => (
+  <div className="p-releases-confirm-details__row">
```
bartaz: If it's a component of its own, probably good candidate for 'top level' pattern naming: `p-release-row` (?)
This would simplify deeply nested class names.

static/js/publisher/release/components/releasesConfirmDetails.js
```diff
@@ -0,0 +1,306 @@
...
+      </div>
+    );
+  });
+};
+
+CancelProgressiveRow.propTypes = {
+  release: PropTypes.object
+};
+
+const _ProgressiveRow = ({
```
bartaz: The need for such naming suggests there is too many components in this file.

static/js/publisher/release/components/releasesConfirmDetails.js
```diff
@@ -0,0 +1,306 @@
...
+
+_ProgressiveRow.propTypes = {
+  release: PropTypes.object,
+  type: PropTypes.string,
+  updateProgressiveReleasePercentage: PropTypes.func
+};
+
+const mapDispatchToProps = dispatch => {
+  return {
+    updateProgressiveReleasePercentage: (key, percentage) =>
```
bartaz: If this is the only prop passed via redux, maybe it can be passed via props instead to keep the component independent.

static/js/publisher/release/selectors/index.js
Lukewh: It actually is more precise - now `changes` contains moving from no progressive release to 100% percentage, whereas `previous.key` will be `null` for all new releases and `{string}` for any updates.

static/js/publisher/release/components/releasesConfirm.js
Lukewh: https://github.com/canonical-web-and-design/snapcraft.io/issues/2386

static/js/publisher/release/components/releasesConfirmDetails.js
Lukewh: Now I've split the component into multiple it would be a bit of a chain of props `releasesConfirm` -> `releasesConfirmDetails` -> `progressiveRow`, so I'd rather keep it in the `progressiveRow` right next to where it's used.

What do you think?

static/js/publisher/release/components/releasesConfirmDetails/releaseRow.js
```diff
@@ -0,0 +1,44 @@
...
+    </span>
+    {progress && (
+      <Fragment>
+        <span className="p-release-details-row__join">to</span>
+        <span className="p-release-details-row__progress">{progress}</span>
+      </Fragment>
+    )}
+    {!progress && (
+      <Fragment>
+        <span />
```
bartaz: Why the empty spans?

static/js/publisher/release/components/releasesConfirmDetails.js
bartaz: Makes sense. Now when the component is in separate file and `connect` is done inline with export it's ok to keep it.

static/js/publisher/release/components/releasesConfirmDetails/releaseRow.js
Lukewh: Ah yeah, I did that for the layout, but now you've pointed it out, I don't like it.

bartaz:
Couple of UXish comments maybe to address later:
- I feel like current UI is a bit too encouraging to make releases progressive - showing bar at 100% by default is just asking for users to update it, maybe it should only be enabled if user actively chooses to use progressive release (via button or something?)
- it would be nice to show "to 100% devices" (or "to all amd64 devices" ...) on rows that are not progressive as well - progressive releases show the % on both the bar and in text, standard releases (that should be default, don't give this information at all)
- it feels a bit unnecessary to show "Cannot be progressively released" on standard releases, it feels like something is 'wrong' with the release state of the application, and this is just a standard way to release to empty channel - maybe for such releases we should show the progressive bar at 100%, disabled and readonly with a tooltip that says that it "cannot be progressively released", this would make it more clear to user that this particular feature is disabled, but release will just normally release to all devices

Screenshot 2019-12-05 at 15 35 57

bartaz:
There is an issue with updates to progressive release not being propagated correctly between progressive bar on the release (in history) and in releases confirm details:

- go to releases page of snap with some progressive release in progress
- open history of progressive release and update the % to some higher value
- update will show up in the confirm box
- show details
- progressive release update will be visible in the details (and in history)
- change the value of the progressive bar in the release confirm - the bar in release history does not update
- change the value of the progressive bar in the release history - the bar in release confirm does not update
- when you close and reopen confirm details it sometimes updates to the value from release history (depending on what was updated last)

Expected:
- every place in the UI should show consistent value of progressive release %

It seems like right now we have more than one place in redux store that keeps the % of the progressive release - separate for history and separate for pending updates?
We should avoid that, every place that shows progressive release % should have the same value, they should all update in sync.

Lukewh:
Follow up for progressive bar flat right-edge at 100%:
https://github.com/canonical-web-and-design/snapcraft.io/issues/2393

Lukewh:
> Couple of UXish comments maybe to address later:
> 
>     * I feel like current UI is a bit too encouraging to make releases progressive - showing bar at 100% by default is just asking for users to update it, maybe it should only be enabled if user actively chooses to use progressive release (via button or something?)
 
Potentially, it is hidden behind the "details" accordion though?

>     * the wording "X% of devices" is not very precise, it only applies to devices in given architecture, which is not mentioned anywhere, releasing to 100% devices feels like every user will get it, when this architecture may be just a fraction of all users

Does the whole row not suggest that though if you read it all?
"Release 5 to latest/edge on arm64 to 100% of devices"

>     * it would be nice to show "to 100% devices" (or "to all amd64 devices" ...) on rows that are not progressive as well - progressive releases show the % on both the bar and in text, standard releases (that should be default, don't give this information at all)

Done

>     * it feels a bit unnecessary to show "Cannot be progressively released" on standard releases, it feels like something is 'wrong' with the release state of the application, and this is just a standard way to release to empty channel - maybe for such releases we should show the progressive bar at 100%, disabled and readonly with a tooltip that says that it "cannot be progressively released", this would make it more clear to user that this particular feature is disabled, but release will just normally release to all devices

Done

PR #2395, opened 2019-12-11
Global percentage

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2374
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2397

## QA
- Pull the branch
- `./run`
- Make some progressive releases, play with the global percentage settings

### DIFF
--- a/static/js/publisher/release/actions/releases.js
+++ b/static/js/publisher/release/actions/releases.js
@@ -128,7 +128,10 @@ export function releaseRevisions() {
     ) {
       progressive = pendingRelease.progressive;
 
-      if (progressive.key === null) {
+      if (
+        progressive.key === null ||
+        progressive.key.indexOf("ui-temp") === 0
+      ) {
         progressive.key = progressiveKey;
       }
     }
--- a/static/js/publisher/release/actions/releases.js
+++ b/static/js/publisher/release/actions/releases.js
@@ -1,6 +1,7 @@
 import {
   RISKS_WITH_AVAILABLE as RISKS,
-  DEFAULT_ERROR_MESSAGE as ERROR_MESSAGE
+  DEFAULT_ERROR_MESSAGE as ERROR_MESSAGE,
+  TEMP_KEY
 } from "../constants";
 
 import { hideNotification, showNotification } from "./globalNotification";
@@ -128,10 +129,7 @@ export function releaseRevisions() {
     ) {
       progressive = pendingRelease.progressive;
 
-      if (
-        progressive.key === null ||
-        progressive.key.indexOf("ui-temp") === 0
-      ) {
+      if (progressive.key === null || progressive.key.indexOf(TEMP_KEY) === 0) {
         progressive.key = progressiveKey;
       }
     }
--- a/static/js/publisher/release/components/releasesConfirmDetails/closeChannelsRow.js
+++ b/static/js/publisher/release/components/releasesConfirmDetails/closeChannelsRow.js
@@ -9,7 +9,7 @@ const CloseChannelsRow = ({ channels }) => {
   }
   return (
     <div className="p-release-details-row is-closing">
-      <span>Close </span>
+      <span>Close</span>
       <span>
         {group
           .map(channel => <b key={channel}>{channel}</b>)
--- a/static/js/publisher/release/components/releasesConfirmDetails/closeChannelsRow.js
+++ b/static/js/publisher/release/components/releasesConfirmDetails/closeChannelsRow.js
@@ -9,8 +9,8 @@ const CloseChannelsRow = ({ channels }) => {
   }
   return (
     <div className="p-release-details-row is-closing">
+      <span>Close </span>
       <span>
-        Close{" "}
         {group
           .map(channel => <b key={channel}>{channel}</b>)
           .reduce((acc, el) => {
--- a/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
+++ b/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
@@ -12,7 +12,7 @@ const GlobalRow = ({ useGlobal, toggleGlobal }) => {
           id="useGlobalToggle"
         />
         <label htmlFor="useGlobalToggle">
-          Use this percentage for all progressive releases
+          Use this percentage for all new progressive releases
         </label>
       </span>
     </div>
--- a/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
+++ b/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
@@ -45,7 +45,7 @@ const GlobalRow = ({
             <span className="p-tooltip--btm-center">
               <span className="p-help">{globalPercentage}% of devices</span>
               <span className="p-tooltip__message">
-                Releases are delivered to devices via snap refreshes, as such,
+                Releases are delivered to devices via snap refreshes. So
                 it may
                 <br />
                 take some time for devices to receive the new version. There is
--- a/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
+++ b/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
@@ -12,7 +12,7 @@ const GlobalRow = ({ useGlobal, toggleGlobal }) => {
           id="useGlobalToggle"
         />
         <label htmlFor="useGlobalToggle">
-          Same percentage for all releases
+          Use this percentage for all releases
         </label>
       </span>
     </div>
--- a/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
+++ b/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
@@ -51,7 +51,7 @@ const GlobalRow = ({
                 take some time for devices to receive the new version. There is
                 also no
                 <br />
-                guarentee that this release will achieve the entire target
+                guarantee that this release will achieve the entire target
                 percentage.
               </span>
             </span>
--- a/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
+++ b/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
@@ -1,0 +1,74 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import { InteractiveProgressiveBar } from "../progressiveBar";
+
+const GlobalRow = ({
+  useGlobal,
+  toggleGlobal,
+  globalPercentage,
+  updatePercentage
+}) => {
+  return (
+    <div className="p-releases-confirm__details-global row u-vertically-center">
+      <div className="col-4">
+        <label>
+          <span>
+            <span className="p-tooltip--btm-center">
+              <span className="p-help">
+                Use the same progressive release percentage
+              </span>
+              <span className="p-tooltip__message">
+                For new progressive releases
+              </span>
+            </span>
+          </span>
+          <span className="p-releases-confirm__details-switch">
+            <input
+              type="checkbox"
+              className="p-switch"
+              checked={useGlobal}
+              onChange={toggleGlobal}
+            />
+            <div className="p-switch__slider" />
+          </span>
+        </label>
+      </div>
+      {useGlobal && (
+        <div className="col-4 p-release-details-row__progress">
+          <InteractiveProgressiveBar
+            percentage={globalPercentage}
+            targetPercentage={globalPercentage}
+            minPercentage={1}
+            singleDirection={0}
+            onChange={updatePercentage}
+          />
+          <span>
+            <span className="p-tooltip--btm-right">
+              <span className="p-help">{globalPercentage}% of devices</span>
+              <span className="p-tooltip__message">
+                Releases are delivered to devices via snap refreshes, as such,
+                it may
+                <br />
+                take some time for devices to receive the new version. There is
+                also no
+                <br />
+                guarentee that this release will achieve the entire target
+                percentage.
+              </span>
+            </span>
+          </span>
+        </div>
+      )}
+    </div>
+  );
+};
+
+GlobalRow.propTypes = {
+  useGlobal: PropTypes.boolean,
+  toggleGlobal: PropTypes.func,
+  globalPercentage: PropTypes.number,
+  updatePercentage: PropTypes.func
+};
+
+export default GlobalRow;
--- a/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
+++ b/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
@@ -1,68 +1,26 @@
 import React from "react";
 import PropTypes from "prop-types";
 
-import { InteractiveProgressiveBar } from "../progressiveBar";
-
-const GlobalRow = ({
-  useGlobal,
-  toggleGlobal,
-  globalPercentage,
-  updatePercentage
-}) => {
+const GlobalRow = ({ useGlobal, toggleGlobal }) => {
   return (
-    <div className="p-releases-confirm__details-global row u-vertically-center">
-      <div className="col-4">
-        <label>
-          <span className="p-tooltip--btm-center">
-            <span className="p-help">
-              Use the same progressive release percentage
-            </span>
-            <span className="p-tooltip__message">
-              For new progressive releases
-            </span>
-          </span>
-          <span className="p-releases-confirm__details-switch">
-            <input
-              type="checkbox"
-              className="p-switch"
-              checked={useGlobal}
-              onChange={toggleGlobal}
-            />
-            <div className="p-switch__slider" />
-          </span>
+    <div className="p-release-details-row is-global">
+      <span className="p-release-details-row__global-check">
+        <input
+          type="checkbox"
+          checked={useGlobal}
+          onChange={toggleGlobal}
+          id="useGlobalToggle"
+        />
+        <label htmlFor="useGlobalToggle">
+          Same percentage for all releases
         </label>
-      </div>
-      {useGlobal && (
-        <div className="col-4 p-release-details-row__progress">
-          <InteractiveProgressiveBar
-            percentage={globalPercentage}
-            targetPercentage={globalPercentage}
-            minPercentage={1}
-            singleDirection={0}
-            onChange={updatePercentage}
-          />
-          <span>
-            <span className="p-tooltip--btm-center">
-              <span className="p-help">{globalPercentage}% of devices</span>
-              <span className="p-tooltip__message">
-                Releases are delivered to devices via snap refreshes. So it may
-                <br />
-                take some time for devices to receive the new version. There is
-                also no
-                <br />
-                guarantee that this release will achieve the entire target
-                percentage.
-              </span>
-            </span>
-          </span>
-        </div>
-      )}
+      </span>
     </div>
   );
 };
 
 GlobalRow.propTypes = {
-  useGlobal: PropTypes.boolean,
+  useGlobal: PropTypes.bool,
   toggleGlobal: PropTypes.func,
   globalPercentage: PropTypes.number,
   updatePercentage: PropTypes.func
--- a/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
+++ b/static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
@@ -12,7 +12,7 @@ const GlobalRow = ({ useGlobal, toggleGlobal }) => {
           id="useGlobalToggle"
         />
         <label htmlFor="useGlobalToggle">
-          Use this percentage for all releases
+          Use this percentage for all progressive releases
         </label>
       </span>
     </div>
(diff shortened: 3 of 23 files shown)


bartaz:
There doesn't seem to be a tooltip, but underline and cursor suggests it:

Screenshot 2019-12-11 at 16 27 29

static/sass/_snapcraft_p-release-details-row.scss
```diff
@@ -11,7 +11,7 @@
       grid-template-columns: auto;
     }
 
-    &:nth-child(odd) {
+    &:nth-child(even) {
```
bartaz: I recall some "no zebra stripes" design rule for tables, don't remember where...

kwm14: MAAS did a spike exploring zebra strips but not sure on the outcome of it... https://github.com/canonical-web-and-design/MAAS-design/issues/524#issuecomment-415693413

static/js/publisher/release/actions/releases.js
```diff
@@ -128,7 +128,10 @@ export function releaseRevisions() {
     ) {
       progressive = pendingRelease.progressive;
 
-      if (progressive.key === null) {
+      if (
+        progressive.key === null ||
+        progressive.key.indexOf("ui-temp") === 0
```
bartaz: It may be worth putting "ui-temp" into a constant - it's already in 3 places I think.

static/js/publisher/release/components/releasesConfirmDetails/index.js
```diff
@@ -27,14 +43,83 @@ const ReleasesConfirmDetails = ({ updates, isProgressiveReleaseEnabled }) => {
...
+
+  const updatePercentage = percentage => {
+    setGlobalPercentage(percentage);
+    updateProgressiveReleasePercentage(null, percentage);
+  };
+
   return (
     <div className="p-releases-confirm__details">
+      {showProgressiveReleases && (
+        <div className="p-releases-confirm__details-global row u-vertically-center">
```
bartaz: Could this whole thing be a separate component just rendered here? It adds quite a lot to the JSX size.

bartaz:
Not sure if that's a bug or an expected feature...

When you switch between global and non global with progress not being on 100% the changes you make later to progress bar are rendered as "pending" (in darker progress color with border between them).
It should be reserved to changes being made on top of already saved changes, and not happen when changes are not yet saved.

progressive-updates

static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
```diff
@@ -0,0 +1,74 @@
...
+  useGlobal,
+  toggleGlobal,
+  globalPercentage,
+  updatePercentage
+}) => {
+  return (
+    <div className="p-releases-confirm__details-global row u-vertically-center">
+      <div className="col-4">
+        <label>
+          <span>
```
bartaz: Is this additional span needed around the one with tooltip?

static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
```diff
@@ -0,0 +1,74 @@
...
+      {useGlobal && (
+        <div className="col-4 p-release-details-row__progress">
+          <InteractiveProgressiveBar
+            percentage={globalPercentage}
+            targetPercentage={globalPercentage}
+            minPercentage={1}
+            singleDirection={0}
+            onChange={updatePercentage}
+          />
+          <span>
```
bartaz: Again - why additional span?

static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
Lukewh: This one is not, removing

static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
Lukewh: This one is, because of `display: flex` of the parent, without it the tooltip doesn't align with the text.

static/sass/_snapcraft_p-release-details-row.scss
nottrobin: Is this a Mantra? :eyes: @anthonydillon

static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
```diff
@@ -0,0 +1,72 @@
...
+            <span className="p-tooltip--btm-center">
+              <span className="p-help">{globalPercentage}% of devices</span>
+              <span className="p-tooltip__message">
+                Releases are delivered to devices via snap refreshes, as such,
+                it may
+                <br />
+                take some time for devices to receive the new version. There is
+                also no
+                <br />
+                guarentee that this release will achieve the entire target
```
matthewpaulthomas:
```suggestion
                guarantee that this release will achieve the entire target
```

static/sass/_snapcraft_p-release-details-row.scss
matthewpaulthomas: The drawback, of zebra stripes for table rows, is that the table needs at least four rows for people to have a chance of realising that they _are_ zebra stripes — rather than just one item being selected and the others not. Even if some tables ~always had at least four items, this one definitely won’t.

static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
```diff
@@ -0,0 +1,72 @@
...
+
+const GlobalRow = ({
+  useGlobal,
+  toggleGlobal,
+  globalPercentage,
+  updatePercentage
+}) => {
+  return (
+    <div className="p-releases-confirm__details-global row u-vertically-center">
+      <div className="col-4">
```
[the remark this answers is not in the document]
Lukewh: Wow, this depends massively on the breakpoints. I'll have to have a fiddle. It's fine for me, but if I zoom in and out it wraps/ unwraps willy-nilly.

static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
```diff
@@ -0,0 +1,72 @@
...
+        <label>
+          <span className="p-tooltip--btm-center">
+            <span className="p-help">
+              Use the same progressive release percentage
```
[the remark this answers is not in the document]
Lukewh: I like this - I'll update and ping when done, thanks @matthewpaulthomas

static/sass/_snapcraft_p-release-details-row.scss
Lukewh: Would each row be OK with the background simply removed? Or would we need to add horizontal lines?

static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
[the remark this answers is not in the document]
Lukewh: I won't fiddle actually, I'll just implement your updated suggestion of the layout :man_shrugging:

bartaz:
@matthewpaulthomas 

Any specific reason why having the checkbox after first item instead of just at the bottom of the list?
When at the bottom it will still fix the issue with switch being the first thing user reads.

Now the checkbox seems to break the flow of releases list, making it look like release(s) below checkbox are somehow different from the one at the top:

Screenshot 2019-12-13 at 12 32 31

Also, with current spacing it's not exactly clear that the "same percentage for all releases" relates to the progress bars.

matthewpaulthomas:
I just remembered where I designed this before — Mozilla’s proxy settings, which are somehow unchanged in Firefox ~20 years later:

image

Meanwhile Microsoft put their equivalent checkbox at the bottom, as you suggest @bartaz:

image

It’s true that putting the checkbox between first and second is less visually elegant. I think it’s clearer, though, because it means you can say “this” and it’s obvious that “this” is the one immediately above, rather than saying “same” and requiring people to deduce what is the same as what.

…Ah. Oops. Let’s actually word it that way, then.

> Also, with current spacing it's not exactly clear that the "same percentage for all releases" relates to the [sliders].

Agreed. Less vertical space, and ideally the left of the checkbox should be flush with the visible left end of the slider trough. (Currently it’s flush with the left end of a hovered/pressed slider thumb at 0%.)

static/js/publisher/release/components/releasesConfirmDetails/globalRow.js
```diff
@@ -0,0 +1,29 @@
...
+    <div className="p-release-details-row is-global">
+      <span className="p-release-details-row__global-check">
+        <input
+          type="checkbox"
+          checked={useGlobal}
+          onChange={toggleGlobal}
+          id="useGlobalToggle"
+        />
+        <label htmlFor="useGlobalToggle">
+          Same percentage for all releases
```
matthewpaulthomas:
```suggestion
          Use this percentage for all releases
```

static/js/publisher/release/components/releasesConfirmDetails/progressiveRow.js
```diff
@@ -54,15 +65,21 @@ const ProgressiveRow = ({
...
-        <InteractiveProgressiveBar
-          percentage={startingPercentage}
-          onChange={updatePercentage}
-          targetPercentage={targetPercentage}
-          minPercentage={1}
-          singleDirection={type === progressiveTypes.UPDATE ? 1 : 0}
-        />
+        {globalPercentage &&
+          !updateGlobalPercentage && (
+            <ProgressiveBar percentage={globalPercentage} disabled={true} />
```
matthewpaulthomas: I realise it’s not a new problem with this change, but enabled and disabled sliders look very similar. Would be good if disabled sliders could be greyed out.

static/js/publisher/release/components/releasesConfirmDetails/releaseRow.js
```diff
@@ -27,7 +27,7 @@ const ReleaseRow = ({ type, revisionInfo, channel, progress, notes }) => (
         <span className="p-release-details-row__progress">
           <ProgressiveBar percentage={100} disabled={true} />
           <span>
-            <span className="p-tooltip--btm-right">
+            <span className="p-tooltip--btm-center">
               <span className="p-help">100% of devices</span>
               <span className="p-tooltip__message">
                 Cannot progressively release to an empty channel
```
matthewpaulthomas: When testing this, I really thought it was a bug that one of the sliders was stuck at 100%. Only by accident I triggered the tooltip and noticed that it was different from the other tooltips.

If “Releases are delivered to devices…” was static text below the table, rather than a tooltip, this value being the only one with a dotted underline would have been a hint that the difference was intentional.

Even better, this tooltip would be static text as well. It could go in the row’s rightmost cell, which would be empty otherwise.

static/js/publisher/release/components/releasesConfirmDetails/progressiveRow.js
Lukewh: Added a new issue here: https://github.com/canonical-web-and-design/snapcraft.io/issues/2398

static/js/publisher/release/components/releasesConfirmDetails/closeChannelsRow.js
```diff
@@ -9,8 +9,8 @@ const CloseChannelsRow = ({ channels }) => {
   }
   return (
     <div className="p-release-details-row is-closing">
+      <span>Close </span>
```
bartaz: Is this space needed?

static/js/publisher/release/components/releasesConfirmDetails/index.js
```diff
@@ -1,16 +1,31 @@
...
 
-const ReleasesConfirmDetails = ({ updates, isProgressiveReleaseEnabled }) => {
+const ReleasesConfirmDetails = ({
+  updates,
+  isProgressiveReleaseEnabled,
+  updateProgressiveReleasePercentage,
+  setTemporaryProgressiveReleaseKeys,
+  removeTemporaryProgressiveReleaseKeys
+}) => {
+  const [useGlobal, setGlobal] = useState(true);
```
bartaz: `useGlobal` is using hooks naming convention of `use*` which may be confusing when reading the code and may cause problems with linting tools (if we start using some hooks features in linting).

static/js/publisher/release/components/releasesConfirmDetails/progressiveRow.js
```diff
@@ -54,15 +65,21 @@ const ProgressiveRow = ({
...
     progress = (
       <Fragment>
-        <InteractiveProgressiveBar
-          percentage={startingPercentage}
-          onChange={updatePercentage}
-          targetPercentage={targetPercentage}
-          minPercentage={1}
-          singleDirection={type === progressiveTypes.UPDATE ? 1 : 0}
-        />
+        {globalPercentage &&
```
bartaz: There is a lot of conditionals in this JSX making it hard to deduce what is meant to show up when.

`globalPercentage && !updateGlobalPercentage` ... `(!globalPercentage || updateGlobalPercentage)`.

Seems like there are 2 possible states - show progress bar or show interactive progress bar. Would be nice if JSX would show this simplicity and the conditional logic would be moved to code.

bartaz:
Looking at the 'Show details' with a dropdown-like arrow I really expect the details to open below it.

Screenshot 2019-12-16 at 11 28 12

Details opening up above the confirmation and moving it down the page is quite unexpected (even when I expect it because I know it will be there...).

Additionally it doesn't seem clear which 'percentage' is meant by 'Use this percentage' checkbox.
Maybe it's because active and inactive bars look very similar in colour.

Screenshot 2019-12-16 at 11 28 05

bartaz:
When you update % of already existing progressive release we don't show 'pending' update different on a bar anymore, just regular bar is moving:

Screenshot 2019-12-16 at 11 34 25

Is that expected? Seems like it changed in this PR.

bartaz:
Maybe updates of existing progressive releases should be more separated/distinct from regular releases, it seems like 'Use this percentage' should affect ALL, but it doesn't:

Screenshot 2019-12-16 at 11 37 33

bartaz:
There is a bug (weird state you can get to) when updating and releasing existing progressive releases.

1) Have existing progressive release at some %
2) As first change, update this % to some other value
3) New 'Update...' entry will show up in confirm details
4) Take the cell of this updated progressive release and drag'n'drop it to other channel
5) New Update...' entry shows up in confirm details (at 100%) when it should be a full 'Release'

Screenshot 2019-12-16 at 11 41 33

static/js/publisher/release/components/releasesConfirmDetails/index.js
Lukewh: Why would it cause problems with linting? Do you have a link I could have a gander at?

Lukewh:
> Looking at the 'Show details' with a dropdown-like arrow I really expect the details to open below it.
> 
> Details opening up above the confirmation and moving it down the page is quite unexpected (even when I expect it because I know it will be there...).
> 
> Additionally it doesn't seem clear which 'percentage' is meant by 'Use this percentage' checkbox.
> Maybe it's because active and inactive bars look very similar in colour.

I get why it was designed this way, as it's a summary that you read and then click "save" on, but I also get why it's confusing. Is your suggestion to move the summary below the bar?


> When you update % of already existing progressive release we don't show 'pending' update different on a bar anymore, just regular bar is moving:
>
> Is that expected? Seems like it changed in this PR.

Unexpected, thanks

> Maybe updates of existing progressive releases should be more separated/distinct from regular releases, it seems like 'Use this percentage' should affect ALL, but it doesn't:

I toyed with adding a horizontal rule between each section, but it seems like overkill - maybe just one between the progressive releases section and the rest? I've also updated the text to read "Use this percentage for all progressive releases".

> There is a bug (weird state you can get to) when updating and releasing existing progressive releases.
>
> 1. Have existing progressive release at some %
> 2. As first change, update this % to some other value
> 3. New 'Update...' entry will show up in confirm details
> 4. Take the cell of this updated progressive release and drag'n'drop it to other channel
> 5. New Update...' entry shows up in confirm details (at 100%) when it should be a full 'Release'

Yay! A new edge-case!

static/js/publisher/release/components/releasesConfirmDetails/index.js
bartaz: I'm not sure exactly that it may be a problem, but I think the `use*` name convention is used to determine if given function is a hook or not: https://reactjs.org/docs/hooks-rules.html

Maybe it will not be an issue if it's not a function, but still may be confusing to read in the code.

bartaz:
> I've also updated the text to read "Use this percentage for all progressive releases".

The updated progressive releases are still progressive releases, so this updated text still seems to apply to them. It would have to say something like "Use this percentage for all new progressive releases" maybe?
But still, visual distinction will probably work better than any text we can think of.

static/js/publisher/release/components/releasesTable/droppableRow.js
```diff
@@ -50,7 +50,17 @@ const ReleasesTableDroppableRow = props => {
   const [{ isOver, canDrop, item }, drop] = useDrop({
     accept: DND_ITEM_REVISIONS,
     drop: item => {
-      item.revisions.forEach(r => promoteRevision(r, channel));
+      item.revisions.forEach(r => {
+        const revision = JSON.parse(JSON.stringify(r));
```
bartaz: It starts to be asking to become a `clone` helper or something. We use it a lot already.

Lukewh: https://github.com/canonical-web-and-design/snapcraft.io/issues/2400

bartaz: The other though I have regarding this:

The fact that we have to do this (clone and reset/strip progressive data) in drag'n'drop handling means that there is too much stuff in the `revision` object that we pass. Or the logic in the action (`promoteRevision`) should be doing this (ignore any progressive release data).
Drag'n'drop handling should ideally know as little as possible about structure of data. It should only care that on drop it's doing something with 'a thing' that was dragged.

bartaz:
Couple other small QA issues:

When updating existing progressive release the 'X% will stay on' doesn't update:
Screenshot 2019-12-16 at 16 46 21

It's quite weird when you switch between 'Use this % on all' on and off and bars jump between different values. When I click 'Use this %' I expect all bars will get the same value as the top one, not jump to last saved common position.

progressive-updates-1

Lukewh:
- [ ] Promoting a row where one item can be progressively released, causes the entire channel to be progressively released

bartaz:
Not sure how much we really need to cover in this PR, maybe some things can be fixed separately, but there are still some things to cover:

When changing % of existing progressive release there is no way to change % back to original value:
1. Have an existing progressive release
2. Update the % to some new value - 'Update' entry shows up in confirm, there is pending release to save
3. Change the % back to previous value by moving the bar as far as possible
4. 'Update' entry is still there in confirm list, even that % doesn't change, something can be saved (released) even that there is no change to any value


When creating new progressive release (by making new full release and changing % in the confirm dialog) the pending progressive state is not reflected in the table cell.
Pending changes to existing progressive releases are visible in the cells, so for consistency we should do the same with new progressive releases:

Screenshot 2019-12-18 at 14 28 23

In example above 113 is new release that is going to be progressive, but it's not visible in the cell.
We currently show all pending status in the table (table shows the state that will be there after saving), so for consistency pending new progressive releases should be shown as well (the % bar in the cell).

Lukewh:
> Not sure how much we really need to cover in this PR, maybe some things can be fixed separately, but there are still some things to cover:
> ### 1
> 
> When changing % of existing progressive release there is no way to change % back to original value:
> 
>     1. Have an existing progressive release
> 
>     2. Update the % to some new value - 'Update' entry shows up in confirm, there is pending release to save
> 
>     3. Change the % back to previous value by moving the bar as far as possible
> 
>     4. 'Update' entry is still there in confirm list, even that % doesn't change, something can be saved (released) even that there is no change to any value

I think this should be covered when the release UI is reviewed by UX - there's also the issue that you can't "undo" closing a channel, without reverting all changes.

> ### 2
> 
> When creating new progressive release (by making new full release and changing % in the confirm dialog) the pending progressive state is not reflected in the table cell.
> Pending changes to existing progressive releases are visible in the cells, so for consistency we should do the same with new progressive releases:
> In example above 113 is new release that is going to be progressive, but it's not visible in the cell.
> We currently show all pending status in the table (table shows the state that will be there after saving), so for consistency pending new progressive releases should be shown as well (the % bar in the cell).

I've updated the branch to show the bar :+1:

bartaz:
Couple of things maybe to address later:

### 1) showing full bar at 100%

In the releases table when new or existing progressive releases is updated to 100% we show a full bar. After saving this bar will disappear, because it's going to be full (not progressive) release.

On the other hand, hiding bar at 100% may be confusing? So maybe that's something to figure out later with UX.

Screenshot 2019-12-20 at 14 03 03

### 2) performance of updates

It seems that now when % bar is changed (especially on multiple revisions at once) redux is LOADED with `UPDATE_` actions. It starts to be visible that browser doesn't manage to handle all the updates and there is sometimes delays in updates.

It should be reviewed where are the biggest bottlenecks. Maybe the progress bar update events need to be throttled (if they are not currently).

Lukewh:
yay

PR #2402, opened 2019-12-16
Don't resize banners - they should look good

Lukewh:
Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2382

This will increase the page size slightly, but it's worth it for quality looking images, imo. We ask publishers to upload higher res images, then we compress them!

## QA
- Pull the branch
- `./run`
- At 1x dpi ensure the 0ad banner under the "Games" heading is nice and sharp


kwm14:
OAD looking sharp 💯

PR #2404, opened 2019-12-17
1776 save/revert are not disabled for snap with categories

### LINKED ISSUE #1776: Save/Revert are not disabled for snap with categories
Found when testing #1772

1. Go to listing page of a snap with category set
2. Change anything on listing page (add something to title)
3. Save/Revert button should get enabled
4. Without saving anything remove the change done before (change the title back)
5. Save/Revert button should get disabled (there are no changes to save), but doesn't

### Expected behaviour

Save/Revert button should get properly disabled if changes are removed.

### Debugging

It seems that when snap has categories they always show up in the diff that we use to check if anything changed.

solazio:
Fixes #1776 

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/snap_name/listing (the snap should have at least one category set)
- Change anything on listing page (add something to title)
- Save/Revert button should get enabled
- Without saving anything remove the change done before (change the title back)
- Save/Revert button should get disabled (there are no changes to save)

### DIFF
--- a/static/js/libs/arrays.js
+++ b/static/js/libs/arrays.js
@@ -18,8 +18,8 @@ function arraysEqual(oldArray, newArray) {
     return false;
   }
 
-  const _oldArray = oldArray.splice(0);
-  const _newArray = newArray.splice(0);
+  const _oldArray = [...oldArray];
+  const _newArray = [...newArray];
 
   _oldArray.sort();
   _newArray.sort();
--- a/static/js/libs/arrays.test.js
+++ b/static/js/libs/arrays.test.js
@@ -10,7 +10,7 @@ describe("arraysEqual", () => {
   });
 
   it("should return arrays of strings as equal", () => {
-    expect(arraysEqual(["test"], ["test"])).toEqual(true);
+    expect(arraysEqual(["a", "b", "c"], ["a", "b", "c"])).toEqual(true);
   });
 
   it("should return arrays of numbers as equal", () => {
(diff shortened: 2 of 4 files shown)


static/js/libs/arrays.js
```diff
@@ -18,8 +18,8 @@ function arraysEqual(oldArray, newArray) {
     return false;
   }
 
-  const _oldArray = oldArray.splice(0);
-  const _newArray = newArray.splice(0);
+  const _oldArray = [...oldArray];
```
bartaz: ES6 FTW!

static/js/libs/arrays.test.js
```diff
@@ -45,6 +45,18 @@ describe("arraysEqual", () => {
   it("should fail if an array of objects differ", () => {
     expect(arraysEqual([{ test: "test" }], [{ test: "test2" }])).toEqual(false);
   });
+
+  it("should fail if an array is modified", () => {
```
bartaz: One more test case I would add is to make sure `arraysEqual` compares 2 identical arrays longer that 1 element.
Our current test compare same 1 element arrays. The only test featuring 2 item array is the one that compares arrays with different lengths.
Current tests would still pass if our implementation would only compare first element of the array.

We should add tests for cases like:

`expect(arraysEqual(["a", "b", "c"], ["a", "b", "c"])).toEqual(true);`
`expect(arraysEqual(["a", "b", "c"], ["a", "d", "c"])).toEqual(false);`

solazio:
> I would suggest adding couple more test cases (or updating existing ones no 1 element arrays), but apart from that it all looks good!

@bartaz I updated the existing tests

PR #2405, opened 2019-12-17
Make snap test step skip-able from build snap step when on macos

### LINKED ISSUE #1298: Make greyed out future steps clickable, so you can jump ahead
From @mariadeanton. While you can't skip everything (OS selection, in particular), there are steps you could skip, like packaging and testing (when on a Mac). We should let the user jump past these, as they may be trying to find their way back to a point in the first snap flow.

screenshot 2018-11-12 at 17 52 10

solazio:
Fixes #1298 

## QA
- Pull the branch
- `./run`
- Visit http://0.0.0.0:8004/first-snap/python/macos/build and see "Publish to store" step is clickable

## Screenshot

image

### DIFF
--- a/templates/first-snap/_breadcrumb.html
+++ b/templates/first-snap/_breadcrumb.html
@@ -39,7 +39,11 @@
         </li>
       {% endif %}
       <li class="p-breadcrumbs__item{% if fsf_step == 'language' or fsf_step == 'install' or fsf_step == 'package' or fsf_step == 'build' or fsf_step == 'test' %} is-disabled{% endif %}">
-        Publish to store
+        {% if fsf_step == 'build' and os and os.startswith('macos') %}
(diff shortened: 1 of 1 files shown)


bartaz:
I think at some point there was a proposal (maybe there is even issue for that) to make all future steps clickable in the breadcrumb, at least after choosing the OS (so when the URLs of future steps are already known).

Lukewh:
> I think at some point there was a proposal (maybe there is even issue for that) to make all future steps clickable in the breadcrumb, at least after choosing the OS (so when the URLs of future steps are already known).

Indeed, I think everything after picking OS can be jumped to - we just need to make sure things don't break if a snap name isn't selected etc.

solazio:
> Indeed, I think everything after picking OS can be jumped to - we just need to make sure things don't break if a snap name isn't selected etc.

@Lukewh @bartaz but the snap name is automatically generated, or not?

PR #2447, opened 2020-01-16
GitHub Authentication

### LINKED ISSUE #2421: Register repo with snap via launchpad
Show some kind of message/ indication that the snap has been registered with a repo

jkfran:
## Done

- Allow user to authenticate on GitHub
- List all the user repositories on GitHub

## Issue / Card

Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2419

## QA

- Pull the branch
- Login as a publisher
- Login with GitHub
- See your GitHub repos from your account

### DIFF
--- a/webapp/api/github.py
+++ b/webapp/api/github.py
@@ -1,0 +1,78 @@
+from urllib.parse import urljoin
+from webapp import api
+
+
+class GitHubAPI:
+    """
+    Provides authentication for GitHub users. Helper methods are also provided
+    for checking organization access and getting user data from the Github API.
+    """
+
+    BASE_URL = "https://api.github.com/"
+
+    def __init__(
+        self, access_token=None, session=api.requests.Session(),
+    ):
+        self.access_token = access_token
+        self.session = session
+        self.session.headers["Accept"] = "application/json"
+
+        if self.access_token:
+            self.session.headers["Authorization"] = f"token {access_token}"
+
+    def _request(self, method, url_path, data={}, params={}):
+        """
+        Makes a raw HTTP request and returns the response.
+        """
+        return self.session.request(
+            method, urljoin(self.BASE_URL, url_path), params=params, json=data
+        )
+
+    def get_resource(self, url_path, per_page=100, page=1):
+        """
+        Makes a raw HTTP GET request and returns the response.
+        """
+        params = {"per_page": per_page, "page": page}
+
+        return self._request(method="GET", url_path=url_path, params=params,)
+
+    def get_user(self):
+        """
+        Sets the user to the currently authorized user if it is not already and
+        returns it. The user object is the dictionary response from the API.
+        """
+        resp = self.get_resource("user")
+
+        return resp.json() if resp.ok else None
+
+    def get_user_repositories(self, all_pages=True, page=1):
+        """
+        Lists repositories that the authenticated user has explicit permission
+        to access.
+        """
+        repos = []
+        resp = self.get_resource(f"user/repos", page=page)
+
+        if resp.ok:
+            repos.extend(resp.json())
+
+        if all_pages and "next" in resp.links:
+            repos.extend(self.get_user_repositories(page=page + 1))
+
+        return repos
+
+    def get_user_orgs(self, all_pages=True, page=1):
+        """
+        Lists organizations that the authenticated user has explicit permission
+        to access.
+        """
+        orgs = []
+        resp = self.get_resource(f"user/orgs", page=page)
+
+        if resp.ok:
+            orgs.extend(resp.json())
+
+        if all_pages and "next" in resp.links:
+            orgs.extend(self.get_user_orgs(page=page + 1))
+
+        return orgs
--- a/webapp/login/oauth_views.py
+++ b/webapp/login/oauth_views.py
@@ -1,0 +1,79 @@
+import os
+
+import flask
+from webapp.decorators import login_required
+from webapp.api.requests import Session
+from urllib.parse import urlencode
+from werkzeug.exceptions import BadRequest
+
+
+oauth = flask.Blueprint(
+    "oauth", __name__, template_folder="/templates", static_folder="/static",
+)
+
+
+@oauth.route("/github/auth", methods=["GET"])
+@login_required
+def github_auth():
+    """
+    Redirect to authorize our Github application and request
+    access to the user's data.
+    """
+
+    if flask.request.args.get("back"):
+        flask.session["github_auth_redirect"] = flask.request.headers.get(
+            "Referer"
+        )
+
+    params = {
+        "client_id": os.getenv("GITHUB_CLIENT_ID"),
+        "scope": "repo",
+        "state": flask.session["csrf_token"],
+    }
+
+    return flask.redirect(
+        f"https://github.com/login/oauth/authorize?{urlencode(params)}"
+    )
+
+
+@oauth.route("/github/auth/verify", methods=["GET"])
+@login_required
+def github_login_verify():
+    """
+    Handles response after the redirect to Github. This response determines
+    if the user has allowed this application access. If we were then we send
+    a POST request for the access_key used to authenticate requests to Github.
+    """
+    url_to_redirect = flask.session.pop(
+        "github_auth_redirect", flask.url_for("snapcraft.homepage")
+    )
+
+    state = flask.request.args.get("state")
+
+    # Avoid CSRF attacks
+    if state != flask.session["csrf_token"]:
+        flask.flash("Invalid request", "negative")
+        return flask.redirect(url_to_redirect)
+
+    data = {
+        "code": flask.request.args.get("code"),
+        "client_id": os.getenv("GITHUB_CLIENT_ID"),
+        "client_secret": os.getenv("GITHUB_CLIENT_SECRET"),
+    }
+
+    session = Session()
+    response = session.request(
+        method="POST",
+        url="https://github.com/login/oauth/access_token",
+        json=data,
+        headers={"Accept": "application/json"},
+    )
+
+    data = response.json()
+
+    if "error" in data:
+        raise BadRequest(data["error_description"], response=response)
+
+    flask.session["github_auth_secret"] = data["access_token"]
+
+    return flask.redirect(url_to_redirect)
--- a/webapp/publisher/snaps/builds.py
+++ b/webapp/publisher/snaps/builds.py
@@ -39,9 +39,11 @@ def build_link(bsi_url, snap, build):
     """ Builds the link to the build page
     """
     build_id = build.self_link.split("/")[-1]
-    github_user = snap.git_repository_url.split("https://github.com/")[-1]
 
-    return f"{bsi_url}/user/{github_user}/{build_id}"
+    # Remove GitHub hostname & split owner/repo
+    owner, repo = snap.git_repository_url[19:].split("/")
+
+    return f"{bsi_url}/user/{owner}/{repo}/{build_id}"
 
 
 def _map_upload_state(upload_state):
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -27,6 +27,7 @@ from canonicalwebteam.store_api.exceptions import (
     StoreApiTimeoutError,
     StoreApiCircuitBreaker,
 )
+from webapp.api.github import GitHubAPI
 from webapp.api.launchpad import launchpad
 from webapp.store.logic import (
     get_categories,
@@ -1350,6 +1351,12 @@ def get_snap_builds(snap_name):
         owner="https://api.launchpad.net/devel/~build.snapcraft.io",
         store_name=details["snap_name"],
     )
+
+    github = GitHubAPI(flask.session.get("github_auth_secret"))
+    context["github_user"] = github.get_user()
+
+    if context["github_user"]:
+        context["github_repositories"] = github.get_user_repositories()
 
     context["snap_builds_enabled"] = len(lp_snaps) >= 1
     if context["snap_builds_enabled"]:
--- a/entrypoint
+++ b/entrypoint
@@ -2,7 +2,7 @@
 
 set -e
 
-RUN_COMMAND="talisker.gunicorn webapp.app:create_app() --bind $1 --worker-class sync --workers 5 --name talisker-`hostname`"
+RUN_COMMAND="talisker.gunicorn.gevent webapp.app:create_app() --bind $1 --worker-class gevent --max-requests 1000 --name talisker-`hostname`"
 
 if [ "${FLASK_DEBUG}" = true ] || [ "${FLASK_DEBUG}" = 1 ]; then
     RUN_COMMAND="${RUN_COMMAND} --reload --log-level debug --timeout 9999"
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,6 +24,7 @@ requests-cache==0.5.2
 responses==0.9.0
 ruamel.yaml==0.15.72
 talisker[flask,gunicorn,raven,prometheus]==0.15.0
+gevent==1.4.0
 
 # Temporary fix to SSO redirect issue
 Werkzeug==0.14.1
--- a/static/sass/_snapcraft_p-build.scss
+++ b/static/sass/_snapcraft_p-build.scss
@@ -1,4 +1,9 @@
 @mixin snapcraft-p-build {
+  .p-icon--github-white {
+    @extend %icon;
+    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='1024' height='999' version='1'%3E%3Cpath fill='%23fff' fill-rule='evenodd' d='M512 0a512 512 0 0 0-162 998c26 4 35-11 35-25l-1-95c-142 31-172-61-172-61-23-59-57-74-57-74-46-32 4-32 4-32 51 4 78 53 78 53 46 78 120 56 149 43 5-33 18-56 33-69-114-13-234-57-234-253 0-56 20-101 53-137-5-13-23-65 5-136 0 0 43-13 141 53a488 488 0 0 1 256 0c98-66 141-53 141-53 28 71 10 123 5 136 33 36 53 81 53 137 0 197-120 240-234 253 18 16 35 47 35 95l-1 140c0 14 9 30 35 25A512 512 0 0 0 512 0z'/%3E%3C/svg%3E");
+  }
+
   .p-build__github-button {
     font-size: 1.25rem;
     margin-top: $spv-outer--regular-scaleable;
--- a/templates/publisher/builds.html
+++ b/templates/publisher/builds.html
@@ -9,12 +9,115 @@ Builds for {% if snap_title %}{{ snap_title }}{% else %}{{ snap_name }}{% endif 
   {% set selected_tab='builds' %}
   {% include "publisher/_header.html" %}
 
+  <form id="market-form" class="market-form p-form p-form--stacked" method="POST" enctype='multipart/form-data'>
+    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+    <input type="hidden" name="snap_id" value="{{ snap_id }}" />
+
+    {% if not github_repository %}
+    <div class="snapcraft-p-sticky js-sticky-bar">
+      <div class="row">
+        <div class="col-7">
+          <p class="snapcraft-p-market-message u-no-margin--bottom">
+            {% if not github_user %}
+            You need to be logged in on GitHub to make changes
+            {% else %}
+            Your GitHub account is connected.
+            {% endif %}
+          </p>
+        </div>
+        <div class="col-5">
+          <div class="u-align--right u-clearfix">
+            {% if not github_user %}
+              <a href="/github/auth?back=1" class="p-button--positive has-icon u-no-margin--bottom">
+                <span>Log in </span><i class="p-icon--github-white"></i>
+              </a>
+            {% else %}
+              <a class="p-button--neutral js-form-revert u-no-margin--bottom" href="/account/snaps/{{ snap_name }}/builds">Revert</a>
+              <button type="submit" class="p-button--positive p-button-spinner js-form-submit u-no-margin--bottom" name="submit_apply" value="Save">
+                <span class="p-button__spinner vanilla-workaround--dark">
+                  <i class="p-icon--spinner u-animation--spin"></i>
+                </span>
+                <span class="p-button__text">Save</span>
+              </button>
+            {% endif %}
+          </div>
+        </div>
+      </div>
+    </div>
+
+    <div class="row">
+      <hr class="u-no-margin--bottom" />
+    </div>
+    {% endif %}
+
+    <section class="p-strip is-shallow">
+      {% with messages = get_flashed_messages(with_categories=true) %}
+      {% if messages %}
+      <div class="u-fixed-width">
+        {% for category, message in messages %}
+        <div class="p-notification--{{ category }}">
+          <p class="p-notification__response">
+            {{ message }}
+          </p>
+        </div>
+        {% endfor %}
+      </div>
+      {% endif %}
+      {% endwith %}
+
+      <div class="u-fixed-width">
+        <h2 class="p-heading--four">
+          GitHub configuration
+        </h2>
+      </div>
+
+      <div class="row p-form__group">
+        <div class="col-2" data-tour="listing-category">
+          <label class="p-form__label">Repository:</label>
+        </div>
+        <div class="col-4">
+          <div class="p-form__control" data-tour="listing-category">
+            {% if github_repository %}
+              {{ github_repository }}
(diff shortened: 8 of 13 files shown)


nottrobin:
@jkfran this PR is pretty big. Could you try to split work up into multiple PRs if possible, so it's easier to read & review them? 'Cos large PRs sometimes sit around for ages 'cos people don't have time to do them. If you can describe the work with multiple bullet points, it could probably be split up.

You can base a second PR off the first one's branch if you like, and add a comment asking that people review the first one first - I do that a lot.

jkfran:
@nottrobin Makes sense, I will try to do it from now on.

webapp/publisher/snaps/views.py
```diff
@@ -1344,36 +1348,89 @@ def get_snap_builds(snap_name):
...
+        details = api.get_snap_info(snap_name, flask.session)
+    except ApiResponseErrorList as api_response_error_list:
+        if api_response_error_list.status_code == 404:
+            return flask.abort(404, "No snap named {}".format(snap_name))
+        else:
+            return _handle_error_list(api_response_error_list.errors)
+    except ApiError as api_error:
+        return _handle_errors(api_error)
+
+    github_repo = flask.request.form.get("github_repository")
```
jpmartinspt: You could move this two lines inside the `else` clause.

webapp/publisher/snaps/views.py
```diff
@@ -1344,36 +1348,89 @@ def get_snap_builds(snap_name):
...
+    git_url = "https://github.com/" + github_repo
+
+    # Get built snap in launchpad with this store name
+    launchpad = Launchpad()
```
jpmartinspt: Can be removed and we just use the top level one?

webapp/publisher/snaps/views.py
jkfran: You are right!

webapp/publisher/snaps/views.py
jkfran: My bad...

webapp/api/github.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return data["access_token"]
+
+    def request(self, base_url, resource, params, method, access_token=None):
+        """
+        Makes a raw HTTP request and returns the response.
+        """
+        url = base_url + resource
+
+        if params is None:
```
jpmartinspt: Why not just default this to `{}` in the method signature?

jkfran: Done!

webapp/api/github.py
```diff
@@ -0,0 +1,126 @@
...
+        response = self.request(
+            "GET", self.base_url, resource, params, access_token
+        )
+
+        if response.ok:
+            data = response.json()
+
+        return response, data
+
+    def get_user(self):
```
jpmartinspt: This will work fine the first time, but as soon as another user authenticates this function will return info about the first user?

webapp/api/github.py
```diff
@@ -0,0 +1,126 @@
...
+        Makes a raw HTTP request and returns the response.
+        """
+        url = base_url + resource
+
+        if access_token is None:
+            access_token = flask.session.get("github_auth_secret")
+
+        if access_token:
+            params.update({"access_token": access_token})
+            self.session.headers["Authorization"] = "token %s" % access_token
```
jpmartinspt: can we use f strings here?

webapp/api/github.py
```diff
@@ -0,0 +1,127 @@
...
+        return response, data
+
+    def get_user(self):
+        """
+        Sets the user to the currently authorized user if it is not already and
+        returns it. The user object is the dictionary response from the API.
+        """
+        try:
+            resp, user = self.get_resource("user")
+        except Exception:
```
jpmartinspt: I'd rather not have such a broad try in here. If we don't care about potential failures we can do it on a case by case basis for the specific function call?

webapp/api/github.py
```diff
@@ -0,0 +1,127 @@
...
+            user = None
+
+        return user
+
+    def get_user_repositories(self):
+        """
+        Lists repositories that the authenticated user has explicit permission
+        to access.
+        """
+        try:
```
jpmartinspt: Same as the above comment

webapp/api/github.py
jkfran: I removed the self.user

webapp/api/github.py
jkfran: Done

jkfran:
Hey @nottrobin, can you check now this PR? Now is only for integration with GitHub.

webapp/publisher/snaps/views.py
```diff
@@ -1351,6 +1353,13 @@ def get_snap_builds(snap_name):
         store_name=details["snap_name"],
     )
 
+    try:
+        context["github_user"] = github.get_user()
+        context["github_repositories"] = github.get_user_repositories()
+    except Exception:
```
nottrobin: You should always catch an explicit exception.

webapp/api/github.py
```diff
@@ -0,0 +1,143 @@
...
+        response = self.request(
+            "POST", self.base_auth_url, "access_token", data
+        )
+
+        try:
+            data = response.json()
+        except ValueError:
+            pass
+
+        if not response.ok or "error" in data or "access_token" not in data:
```
nottrobin: Why are you testing for 3 different types of failure here? Are you sure that these are all cases we expect? Are you sure that all failure states wouldn't be covered by `not response.ok`?

I kinda feel you're still masking here. Error messages are for developers, not for the public. If you take 3 different things that can possibly go wrong and mask them with one generic message, you reduce the chance that a developer can see an error and immediately tell what went wrong. In this case they're gonna have to run the code themselves, throw a break point in above this statement and inspect the response themselves to see what actually happened.

Even if you wanted to catch all 3 of these possible cases and raise your own custom errors like `GithubAuthException`, I feel you should either create 1 custom exception per possible error state or modify the exception object in some way to indicate what exactly happened.

Ok I'll delve into a little theory here. There are basically 2 different things you could do when you think there might be an error:

1. Catch errors yourself and re-raise them
2. Let the code error naturally (e.g. in this case you could call [`response.raise_for_status()`](https://2.python-requests.org/en/master/api/#requests.Response.raise_for_status), which will pass through silently if `response.ok` is `True`, or raise an [`HTTPError`](https://2.python-requests.org/en/master/api/#requests.HTTPError) if it's `False`)

The argument for 1. is encapsulation. In certain theories, all your code layers should be self-contained and opaque. The client of your code shouldn't care *at all* what you're using inside your module. In this case you might, at some point in the future, want to swap out your usage of `request` for e.g. using `urllib` directly. If you let your layer raise `requests.exceptions.HTTPError` errors, then client code might come to rely on those errors in their code, and so if you stop using `requests` it could end up breaking their code. And so you should catch every possible error state and raise your own custom exceptions, because that way you have complete control over whether those exceptions change or not.

The argument for .2 (and against .1) is that, for those who do understand how the module works, it becomes much less clear what actually went wrong. `requests` is a pretty universal library, and so if any Python developer ever encounters a `requests.exceptions.HTTPError` they probably know what it means and what to do with it. Your custom errors are much harder because the dev has to read your documentation or your code to work out what they mean and how to use them.

Another argument for .2 in my opinion is about efficiently of development, [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product) and [gold-plating](https://en.wikipedia.org/wiki/Gold_plating_(project_management)):

> gold plating is the phenomenon of working on a project or task past the point of diminishing returns

It is literally impossible to anticipate all the ways in which your code could fail, and so you could spend all month trying to make sure your layer of code is robust against every eventually, and you could end up creating hundreds of custom exceptions along the way, which then make it a huge task to learn the full interface of your module. And in most cases this will cover plenty of cases that might never come up.

For me, the two arguments for option 2. massively trump the benefits of option 1., especially when this code is really mostly for our benefit and not a professional package or community offering. My attitude to exception handling is - try to make sure the code will naturally fail in an obvious way (`response.raise_for_status()`), but beyond that don't bother over-engineering your error handling.

So, for example (one which came up in @carkod's work), if your website expects a `SECRET_KEY` environment variable, then use `os.environ["SECRET_KEY"]` to access it rather than `os.getenv("SECRET_KEY")`. This way, if the `SECRET_KEY` is missing, it will raise a `KeyError: 'SECRET_KEY'` error. This is easy, quick, and should provide the developer with a pretty obvious indication of what went wrong.

In conclusion, what I would do in your case is not both defining `GithubAuthException` and instead use `response.raise_for_status()` where appropriate. If you're sure that the API will sometimes return a `2xx` response but without the expected `access_token` property, then it might be a valid reason for writing and raising a custom exception. But if you're not absolutely sure that's an expected resposne, then simply access the property when you need it with `data["access_token"]` and it will raise an error naturally if ever that case comes up.

Does this make sense?

webapp/login/oauth_views.py
```diff
@@ -0,0 +1,47 @@
...
+    flask.session["github_auth_secret"] = auth_token
+
+    url = flask.session.pop(
+        "github_auth_redirect", flask.url_for("snapcraft.homepage")
+    )
+
+    return flask.redirect(url)
+
+
+@oauth.errorhandler(GithubAuthException)
```
nottrobin: Again, this flow would be quite difficult for a developer to unpick. I think it would be much better to do this logic and redirection directly in the view where you expect the authentication to fail. This has the benefit that you won't end up catching and masking errors where you didn't expect with a redirect. If something happens that you didn't expect, the best possible thing that can happen is that your application breaks with a useful error message.

webapp/api/github.py
jkfran: Okay, I understand let's say that I use `raise_for_status()` and `data["access_token"]`. Then I have:
```python

    def verify_auth(self, code):
        ...
        response.raise_for_status()
        data = response.json()
        return data["access_token"]
```
So that means different kind of exceptions could happen, right? Then I need to notify the user that something went wrong during the authorization verification. So with one should I use? Because I need to have something like this, right?:
```python
@oauth.route("/github/auth/verify", methods=["GET"])
@login_required
def github_login_verify():
    ...
    try:
        auth_token = github.verify_auth(auth_code)
    except RequestException, KeyError:
        flask.flash(f"An error has occurred linking your GitHub account", "negative")
        return flask.redirect(flask.url_for("snapcraft.homepage"))
```

But if I do this where is the benefit for a developer when debugging?

Obviously I need to notify the user that something went wrong during the authorization or is it better to let the exceptions raise and show to the user a 500 error page?

nottrobin: What I'm saying is, deliberately handle the states you expect. And write your code such that it will error clearly for states you don't expect (by using `dict["key"]` where we expect keys to exist, and `raise_for_status()` to raise errors for any unexpected statuses).

Your use of `except RequestException, KeyError` is, I'm afraid, still not doing this because it's catching a whole load of things that could happen (any non `2xx` status was returned from the API *or* a `2xx` status was returned but it didn't contain a token). It will then clobber any useful error information by just showing a "An error has occurred linking your GitHub account", when there could have been an important unexpected error that we needed to know about.

So I assume one of the cases you do expect is that the user might decline to grant permissions on the GitHub authorization page. In this case, what will happen in your code? I would expect that maybe the response will contain a `status_code` of `403 forbidden`. if this is the case, then check for that - e.g. `if response.status_code == 403:`.

Another possible case might be that the user doesn't log in at all on GitHub and returns to our site unauthenticated, in which case maybe the API returns a response with a `status_code` of `401`, so we should make sure we do something sensible there.

Now, if it is the case that that's what the API will return, then we could write it like this:

``` python3
# github.py
def verify_auth(self, code):
    ...
    response.raise_for_status()
    return response.json()["access_token"]

# views.py
from requests.exceptions import HTTPError

@oauth.route("/github/auth/verify", methods=["GET"])
@login_required
def github_login_verify():
    ...
    try:
        auth_token = github.verify_auth(auth_code)
    except HTTPError as error:
        if error.response.status_code == 403:
            flask.flash(f"We were forbidden from accessing your Github account", "negative")
            return flask.redirect(flask.url_for("snapcraft.homepage"))
        elif error.response.status_code == 401:
            flask.flash(f"*something about what happened here*", "negative")
            return flask.redirect(flask.url_for("snapcraft.homepage"))
        else:
            raise error  # Anything we didn't anticipate
```

The important point here is we're explicitly checking for the states we expect, and doing something, and for the states we don't expect (a non 401 or 403 status code) we are still raising an error. If the `else:` case is hit here, the user will see a `500` status page, which is what should happen because it was unexpected.

Do you see what I mean?

Don't just copy my solution because I have no idea if this is how the GitHub API behaves. Check how it actually does behave, either by reading docs or testing it out, and then code your solution explicitly to respond to how it's expected to behave.

nottrobin: Although, are you sure that the homepage is where you want to go if you if there's an error authenticating with GitHub? Surely you would want to remain on the build page for your snap wouldn't you?

nottrobin: > Now, if it is the case that that's what the API will return, then we could write it like this:

Another option here would use custom error messages to keep the `401` logic in the `verify_auth` method, which is arguable where it might belong:

``` python3
# github.py
from requests.exceptions import HTTPError


class Unauthorized(HTTPError):
    pass


class Forbidden(HTTPError):
    pass


def verify_auth(self, code):
    ...

    if response.status_code == 401:
        raise Unauthorized(response=response)
    elif response.status_code == 403:
        raise Forbidden(response=response)

    response.raise_for_status()

    return response.json()["access_token"]


# views.py
import github


@oauth.route("/github/auth/verify", methods=["GET"])
@login_required
def github_login_verify():
    ...
    try:
        auth_token = github.verify_auth(auth_code)
    except github.Forbidden:
            flask.flash(f"We were forbidden from accessing your Github account", "negative")
    except github.Unauthorized:
            flask.flash(f"We couldn't get authorization to GitHub", "negative")
```

This is probably a little neater by encapsulating more of the logic in the right place (`verify_auth`) and keeping the view code pretty obviously straightforward. But it still achieved the exact same logic: Explicitly check for the states we expect - 401 or 403 - and allow the code to error if any state we didn't expect occurs.

webapp/api/github.py
```diff
@@ -0,0 +1,143 @@
...
+
+class GithubAPI:
+    """
+    Provides authentication for GitHub users. Helper methods are also provided
+    for checking organization access and getting user data from the Github API.
+    """
+
+    def __init__(self):
+        self.client_id = GITHUB_CLIENT_ID
+        self.client_secret = GITHUB_CLIENT_SECRET
```
nottrobin: I think ideally it should take at least the client_ info and the session as arguments. The client stuff should be the client's responsibility to supply, so I think that should be passed through in oauth_views.py etc., and if you're not going to allow the user to override the base_url etc. (a perfectly valid choice) then they should be constants on the object:

``` python3
    BASE_URL = "https://api.github.com/"
    AUTH_URL = "https://github.com/login/oauth/"

    def __init__(
        self,
        client_id,
        client_secret,
        session=api.requests.Session()
    ):
        self.client_id = client_id
        self.client_secret = client_secret
        self.session = session
        self.session.headers["Accept"] = "application/json"
```

webapp/api/github.py
```diff
@@ -0,0 +1,143 @@
...
+    def authorize(self, scope=None, state=None):
+        """
+        Redirect to Github and request access to a user's data.
+        """
+        params = {
+            "client_id": self.client_id,
+        }
+
+        if scope:
+            params.update({"scope": scope})
```
nottrobin: isn't this the same as `param["scope"] = scope`?

jkfran: You are right!

webapp/api/github.py
jkfran: Okay, I did it!

jkfran:
The pythons tests didn't pass without the:
```
# Temporary fix to SSO redirect issue
Werkzeug==0.14.1
```
I added it back but it will be good to find out why and do a cleanup.

webapp/api/github.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(
+        self, client_id, client_secret, session=api.requests.Session()
+    ):
+        self.client_id = client_id
+        self.client_secret = client_secret
+        self.session = session
+        self.session.headers["Accept"] = "application/json"
+
+    def authorize(self, scope=None, state=None):
```
nottrobin: I don't feel this method belongs on this class. This class should be about retrieving data from the GitHub API, not redirecting the user. Redirecting the user is a view-level concern.

I think all the logic in this function should instead be moved directly to the `github_auth` view function.

webapp/publisher/snaps/views.py
```diff
@@ -51,6 +53,10 @@
 )
 
 store_api = SnapcraftStoreApi(talisker.requests.get_session(requests.Session))
+github = GithubAPI(
```
nottrobin: Hmm it is a shame to be duplicating this block between here and the `oauth_views.py` file. Can you try to think of a way around that?

webapp/api/github.py
```diff
@@ -0,0 +1,139 @@
...
+    def request(
+        self, method, base_url, resource, data={}, params={}, access_token=None
+    ):
+        """
+        Makes a raw HTTP request and returns the response.
+        """
+        url = base_url + resource
+
+        if access_token is None:
+            access_token = flask.session.get("github_auth_secret")
```
nottrobin: I think for this class to remain neatly encapsulated, it would be best to avoid any Flask-level functions if we can. I think what should happen here is that getting the secret from the session should be something that happens more at the view level and then you should pass it through as an argument wherever you're instantiating the `GithubAPI` class.

webapp/api/github.py
```diff
@@ -0,0 +1,139 @@
...
+        response.raise_for_status()
+
+        data = response.json()
+
+        if "error" in data:
+            raise GithubAPIError(data["error_description"], response=response)
+
+        return data["access_token"]
+
+    def request(
```
nottrobin: I believe this is a purely internal method, only used by other methods in this same class? If that's the case, you should prefix its name with an underscore (`def _request`) to indicate that it's not intended to form part of the class's public interface.

webapp/api/github.py
```diff
@@ -0,0 +1,139 @@
...
+from urllib.parse import urlencode
+from webapp import api
+from requests.exceptions import RequestException
+
+
+class GithubAPIError(RequestException):
+    pass
+
+
+class GithubAPI:
```
nottrobin: This is kinda nit-picky, but "GitHub" is technically spelled with a capital "H"

webapp/api/github.py
```diff
@@ -0,0 +1,139 @@
...
+
+        return data["access_token"]
+
+    def request(
+        self, method, base_url, resource, data={}, params={}, access_token=None
+    ):
+        """
+        Makes a raw HTTP request and returns the response.
+        """
+        url = base_url + resource
```
nottrobin: It's a little weird that you're asking for this as 2 separate arguments and then simply concatenating them. Why don't you just ask for one `url`?

webapp/api/github.py
```diff
@@ -0,0 +1,139 @@
...
+            params["scope"] = scope
+
+        if state:
+            params["state"] = state
+
+        auth_url = f"{self.AUTH_URL}authorize?{urlencode(params)}"
+
+        return flask.redirect(auth_url)
+
+    def verify_auth(self, code):
```
nottrobin: If we follow the logic of my point about removing the `authorize` function, I think maybe this function also doesn't belong i this class. The auth flow is a bit of a special case, and this logic again feels to me like it really belongs in the view function.

If you were to also move this logic to `oauth_views.py` then I don't think you'd have any need for the `GithubAPI` class in that file at all which would solve your duplication problem. And then this class would be simply about making authenticated calls to the API to retrieve actual data, which seem neatly encapsulated.

Do you agree?

webapp/api/github.py
nottrobin: Although if we do make this class more single-purpose about only making direct authenticated API calls, then maybe you can remove the `base_url` argument, and make this line:

```suggestion
        url = os.path.join(self.BASE_URL, resource)
```

webapp/api/github.py
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        response = self.request("POST", self.AUTH_URL, "access_token", data)
+
+        response.raise_for_status()
+
+        data = response.json()
+
+        if "error" in data:
+            raise GithubAPIError(data["error_description"], response=response)
```
nottrobin: If we do move this logic into the views in `oauth_views.py`, then `GitHubAPIError` won't any longer quite make sense as a name for this exception, as it's not in the github API file any more.

Maybe we should use a [`werkzeug.exceptions.BadRequest`](https://werkzeug.palletsprojects.com/en/0.16.x/exceptions/#werkzeug.exceptions.BadRequest) error? That might make more sense.

webapp/api/github.py
jkfran: I agree, but is it fine to have a method that only build and return the authorization URL to GitHub and then use that URL in the view to do the redirect?

webapp/api/github.py
jkfran: @nottrobin maybe this is related with your previous comment but right now if I add a property `self.access_token` that also can be set with the constructor does that could cause the users to have the same access_token? because we are instantiating the class only once?

webapp/api/github.py
jkfran: I don't see the problem in having a function to exchange the authorization_code for the access_token of the user, plus you told me before to change the `request` method to `_method`, if we need to use that method in the view this is not consistent.

jkfran: Maybe what I should change is the name of this method?

jkfran:
@nottrobin I made the changes that we agreed earlier. Please let me know.

webapp/login/oauth_views.py
```diff
@@ -0,0 +1,80 @@
...
+from webapp.api.requests import Session
+from urllib.parse import urlencode
+from werkzeug.exceptions import BadRequest
+
+
+oauth = flask.Blueprint(
+    "oauth", __name__, template_folder="/templates", static_folder="/static",
+)
+
+session = Session()
```
jkfran: @nottrobin I am now not sure about this

webapp/publisher/snaps/views.py
```diff
@@ -1351,6 +1351,12 @@ def get_snap_builds(snap_name):
         store_name=details["snap_name"],
     )
 
+    github = get_github_api(flask.session.get("github_auth_secret"))
```
nottrobin: Sorry... this is actually pointless. For some reason I thought the instantiation of `GitHubAPI` would be a long thing, and so it would be too verbose to do it in-place in every view function.

But actually, if it's just `github = GitHubAPI(flask.session.get("github_auth_secret"))`, you should just do it directly inline here, and not bother with the factory function.

jkfran: It's fine, it makes sense, at the beginning this PR was like this but without the token argument.

I just did this change.

webapp/login/oauth_views.py
nottrobin: Do you mean, should you be simply using `Session()` on line 66? Is this the only place this session is used? If that's the case, that would seem reasonable to me.

jkfran: Okay, Yes it is the only place so I have moved it there now, but I was wondering because of the problem we had with github when we instantiated here in which users use the same instance. If at the performance level, in this case, it is better or not

webapp/publisher/snaps/builds.py
```diff
@@ -35,13 +35,21 @@ class LaunchpadStoreUploadState(Enum):
     UPLOADED = "Uploaded"
 
 
+def get_github_repository(github_repo_url):
+    """ Returns a string with the GitHub owner
+        and the repository name separated by an
+        slash: <Owner>/<Repository name>
+    """
+    return github_repo_url.split("https://github.com/")[-1]
```
nottrobin: This really seems so trivial that creating its own function just obfuscates code. Putting this logic inline would make the original code literally no larger.

I would just do this inline:

``` python3
# Remove hostname & split owner/repo
owner, repo = github_repo_url[19:].split('/')
```

jkfran: Okay, I like this, I guess I did it with a function because in the next PR I need to repeat this operation.

webapp/login/oauth_views.py
nottrobin: I can't imagine that the instantiation of a session object is a particularly expensive operation, so doing it once per request shouldn't be a problem. So in this case I think moving it into the view function is definitely best.

webapp/publisher/snaps/builds.py
nottrobin: Yeah, I assumed you must be doing it in more than one place. And while we should be DRY, I think you really get diminishing returns on code re-use when it's just one or two lines. I personally think just repeating this logic in 2 or 3 places so that you can make the code more transparent is more important than DRYness in this case.

PR #2448, opened 2020-01-16
Duplicate licenses

### LINKED ISSUE #1329: Duplicate, invalid GPL license options in the drop down list
### Expected behaviour
Only unique license items in the list 

default

Only the latter option work, if the former option is selected it will return the following error:

```
 Invalid symbols sequence such as (A B) for token: "-only" at position: 18 
```

### Specs

- _URL:_ https://snapcraft.io/_snap_name_/listing
- _Operating system:_ Ubuntu 18.10 AMD64
- _Browser:_ Mozilla Firefox 63.0

Lukewh:
## Done

- Use the store whitelisted SPDX's

## Issue / Card

Fixes #1329 

## QA

- Pull the branch
- Visit http://0.0.0.0:8004/snap_name/listing and scroll down to the license section
- Add and remove licenses - you shouldn't get errors or duplicate licenses

### DIFF
--- a/webapp/helpers.py
+++ b/webapp/helpers.py
@@ -25,14 +25,8 @@ class RegexConverter(BaseConverter):
 
 def get_licenses():
     try:
-        with open("webapp/licenses.json") as f:
-            licenses_file = json.load(f)
-
-        licenses = [
-            license
-            for license in licenses_file["licenses"]
-            if not license["isDeprecatedLicenseId"]
-        ]
+        with open("webapp/store-licenses.json") as f:
+            licenses = json.load(f)
 
         def _build_custom_license(license_id, license_name):
             return {"licenseId": license_id, "name": license_name}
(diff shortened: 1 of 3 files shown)


webapp/helpers.py
```diff
@@ -25,17 +25,30 @@ def __init__(self, url_map, *items):
...
-            licenses_file = json.load(f)
-
-        licenses = licenses_file["licenses"]
+        with open("webapp/store-licenses.json") as f:
+            licenses = json.load(f)
 
         def _build_custom_license(license_id, license_name):
             return {"licenseId": license_id, "name": license_name}
 
         CUSTOM_LICENSES = [
```
jkfran: @Lukewh I am just trying to understand this, why these licenses defined here are not in the json file so we don't need to define them here?

webapp/helpers.py
```diff
@@ -25,17 +25,30 @@ def __init__(self, url_map, *items):
 
 def get_licenses():
     try:
-        with open("webapp/licenses.json") as f:
```
jkfran: @Lukewh So can we delete this file now?

webapp/helpers.py
Lukewh: The json file is a subset of https://spdx.org/licenses/, the custom licenses are ones we're adding specifically for snapcraft so this is just to keep them separate. The way this worked and how it will again one day, is that the licenses.json (spdx v3) is downloaded automatically on app start which would overwrite our custom licenses - so this is both left over from before and also prep for the future.

webapp/helpers.py
Lukewh: Ah yes - I'll comment it out with an extra comment - I don't want to remove it because I'm clinging on to the possibility that we'll soon move to v3 and this will be relevant once again

Lukewh: oh I see - I DID delete it. Yes we can delete this for now. We'll need to put it back when we can use v3

PR #2454, opened 2020-01-20
Register GitHub repo with Snap via the Launchpad API

jkfran:
## Done

- Replace launchpadlib
- Launchpad API request to create a new Snap

## Issue / Card

Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2421

## QA

**Ask @jkfran for the needed environment variables**

- Pull the branch
- Login as a publisher
- Login with GitHub, select and save
- Check that the repo is now always displayed in the GitHub configuration section

### DIFF
--- a/webapp/api/launchpad.py
+++ b/webapp/api/launchpad.py
@@ -1,6 +1,88 @@
-from launchpadlib.launchpad import Launchpad
+from hashlib import md5
+from webapp.api.requests import Session
 
 
-launchpad = Launchpad.login_anonymously(
-    "storefront", "production", version="devel"
-)
+class Launchpad:
+    """
+    Provides authentication for Launchpad.
+    """
+
+    BASE_URL = "https://api.launchpad.net"
+
+    def __init__(self, username, token, signature, session=Session()):
+        self.username = username
+        self.session = session
+        self.session.headers["Accept"] = "application/json"
+        self.session.headers["Authorization"] = (
+            f'OAuth oauth_version="1.0", '
+            f'oauth_signature_method="PLAINTEXT", '
+            f"oauth_consumer_key={username}, "
+            f'oauth_token="{token}", '
+            f'oauth_signature="&{signature}"'
+        )
+
+    def _request(self, method="GET", url=None, params={}, data={}):
+        """
+        Makes a raw HTTP request and returns the response.
+        """
+        url = f"{self.BASE_URL}/devel/{url}"
+
+        response = self.session.request(method, url, params=params, data=data)
+        response.raise_for_status()
+
+        return response.json() if response.content else None
+
+    def get_collection_entries(self, resource, params=None):
+        """
+        Return collection items from the API
+        """
+        collection = self._request(url=resource, params=params)
+
+        return collection.get("entries", [])
+
+    def get_snap_by_store_name(self, snap_name):
+        """
+        Return an Snap from the Launchpad API by store_name
+        """
+        snaps = self.get_collection_entries(
+            "+snaps",
+            {
+                "ws.op": "findByStoreName",
+                "owner": "/~build.snapcraft.io",
+                "store_name": snap_name,
+            },
+        )
+
+        # The Launchpad API only allow to perform a find by store_name
+        # but we are only interested in exactly this one
+        if snaps and snaps[0]["store_name"] == snap_name:
+            return snaps[0]
+
+        return None
+
+    def new_snap(self, snap_name, git_url):
+        """
+        Create an ISnap in Launchpad
+        """
+        data = {
+            "ws.op": "new",
+            "owner": f"/~{self.username}",
+            "name": md5(git_url.encode("UTF-8")).hexdigest(),
+            "store_name": snap_name,
+            "git_repository_url": git_url,
+            "git_path": "HEAD",
+            "auto_build": "false",
+            "auto_build_archive": "/ubuntu/+archive/primary",
+            "auto_build_pocket": "Updates",
+            "processors": [
+                "/+processors/amd64",
+                "/+processors/arm64",
+                "/+processors/armhf",
+                "/+processors/i386",
+                "/+processors/ppc64el",
+                "/+processors/s390x",
+            ],
+            "store_series": "/+snappy-series/16",
+        }
+
+        return self._request("POST", "+snaps", data=data)
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -1414,19 +1414,25 @@ def post_snap_builds(snap_name):
 
     # Get built snap in launchpad with this store name
     lp_snap = launchpad.get_snap_by_store_name(details["snap_name"])
+    github_repo = flask.request.form.get("github_repository")
+    git_url = f"https://github.com/{github_repo}"
 
-    if lp_snap:
-        flask.flash(
-            "It is not possible to change the git repository for this snap.",
-            "negative",
+    if not lp_snap:
+        result = launchpad.new_snap(snap_name, git_url)
+
+        if result:
+            flask.flash(
+                "The GitHub repository was linked correctly.", "positive"
+            )
+        else:
+            flask.flash(
+                "An error occurred linking the GitHub repository.", "negative"
+            )
+    elif lp_snap["git_repository_url"] != git_url:
+        # In the future, create a new record, delete the old one
+        raise AttributeError(
+            f"Snap {snap_name} already has a build repository associated"
         )
-    else:
-        github_repo = flask.request.form.get("github_repository")
-        git_url = f"https://github.com/{github_repo}"
-
-        launchpad.new_snap(snap_name, git_url)
-
-        flask.flash("The GitHub repository was linked correctly.", "positive")
 
     return flask.redirect(
         flask.url_for(".get_snap_builds", snap_name=snap_name)
--- a/webapp/publisher/snaps/views.py
+++ b/webapp/publisher/snaps/views.py
@@ -1,3 +1,4 @@
+import os
 from json import loads
 
 import flask
@@ -28,7 +29,7 @@ from canonicalwebteam.store_api.exceptions import (
     StoreApiCircuitBreaker,
 )
 from webapp.api.github import GitHubAPI
-from webapp.api.launchpad import launchpad
+from webapp.api.launchpad import Launchpad
 from webapp.store.logic import (
     get_categories,
     filter_screenshots,
@@ -52,6 +53,11 @@ publisher_snaps = flask.Blueprint(
 )
 
 store_api = SnapcraftStoreApi(talisker.requests.get_session(requests.Session))
+launchpad = Launchpad(
+    username=os.getenv("LP_API_USERNAME"),
+    token=os.getenv("LP_API_TOKEN"),
+    signature=os.getenv("LP_API_TOKEN_SECRET"),
+)
 
 
 def refresh_redirect(path):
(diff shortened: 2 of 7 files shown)


requirements.txt
```diff
@@ -11,7 +11,6 @@ Flask==1.0.2
 Flask-OpenID-Stateless==1.2.6
 Flask-WTF==0.14.2
 humanize==0.5.1
-launchpadlib==1.10.7
```
nottrobin: :tada:

webapp/api/launchpad.py
```diff
@@ -1,6 +1,105 @@
...
+        return None
+
+    def new_snap(self, snap_name, git_url):
+        """
+        Create an ISnap in Launchpad
+        """
+        data = {
+            "ws.op": "new",
+            "owner": f"/~{self.username}",
+            "name": md5(snap_name.encode("UTF-8")).hexdigest(),
```
nottrobin: Why do you need to hash the name?

nottrobin: This is kinda weird. Is this basically a unique ID?

jkfran: Yes, for this API we use always the same user, it is the same user that build.snapcraft.io is using. All the launchpad snaps created with build.s.i are owned by this user. Let me know when you need the environment variables.

requirements.txt
jkfran: :v: As you can see before we were using anonymous login but it became a nightmare to login with this library

webapp/api/launchpad.py
nottrobin: User? `md5(snap_name.encode("UTF-8")).hexdigest()` presumably generates a hash that's based on the snap name rather than the user's name doesn't it?

jkfran: Do you mean the reason for the "/\~" before the username? It's a pattern that launchpad follow I don't know exactly how it works, they have "~" and "+" for the different entry types. At the beginning, while I was testing with this Colin told me to use this prefix for the username and "+" for the snaps for the endpoint. We are doing the same thing currently in build.s.i:
https://github.com/canonical-web-and-design/build.snapcraft.io/blob/f3c648a1f463efc10f64bfea1b8eabb4965844e6/src/server/handlers/launchpad.js#L200

nottrobin: I'm talking about line 87, not line 86

webapp/publisher/snaps/views.py
```diff
@@ -52,6 +53,11 @@
 )
 
 store_api = SnapcraftStoreApi(talisker.requests.get_session(requests.Session))
+launchpad = Launchpad(
```
nottrobin: What do you think of the naming - `Launchpad` and `launchpad`, vs `LaunchpadAPI` and `launchpad_api`?

I would normally do the latter, but maybe the former does make sense and is more succinct - `launchpad.get_snaps` is not wrong - maybe `API` is technically redundant, since as we're programming, everything is an application programming interface in one sense.

This would imply, then, that we should be consistent with this naming everywhere and call our SnapStoreAPI simply `SnapStore` etc. (and it should probably be `SnapStore` rather than `Snapcraft`) but this is all way out of scope of this PR. So it's really just me talking to myself.

Please disregard this comment =)

webapp/api/launchpad.py
jkfran: Oh wait, there is an issue in that line, it should be the md5 of the git repo, not the snap name.

For the name property in launchpad we are using the md5 of the git repo and the actual snap name is on "store_name". I am following this pattern from build.s.i. I didn't ask why.

webapp/publisher/snaps/views.py
```diff
@@ -1345,42 +1351,90 @@ def get_snap_builds(snap_name):
...
+                    "title": build["title"],
                 }
             )
 
     return flask.render_template("publisher/builds.html", **context)
+
+
+@publisher_snaps.route("/<snap_name>/builds", methods=["POST"])
+@login_required
+def get_snap_builds_post(snap_name):
```
nottrobin: Shouldn't this be called `post_snap_builds`?

webapp/api/launchpad.py
nottrobin: Yeah that's fine I guess. It would be nice to know why, but probably not worth it here.

webapp/publisher/snaps/views.py
jkfran: :flushed: You are right!

webapp/publisher/snaps/views.py
```diff
@@ -1345,42 +1351,90 @@ def get_snap_builds(snap_name):
...
+    try:
+        details = api.get_snap_info(snap_name, flask.session)
+    except ApiResponseErrorList as api_response_error_list:
+        if api_response_error_list.status_code == 404:
+            return flask.abort(404, "No snap named {}".format(snap_name))
+        else:
+            return _handle_error_list(api_response_error_list.errors)
+    except ApiError as api_error:
+        return _handle_errors(api_error)
```
nottrobin: This function should have been called `_handle_error` or perhaps `_handle_api_error`. It shouldn't be using a plural when it takes a single error.

But you didn't create the function, and definitely not in this PR, so it's out of scope, unless you feel like changing it everywhere.

jkfran: I will create an issue for this and the other comment about the naming.

webapp/publisher/snaps/views.py
```diff
@@ -1345,42 +1351,90 @@ def get_snap_builds(snap_name):
...
+
+    # Get built snap in launchpad with this store name
+    lp_snap = launchpad.get_snap_by_store_name(details["snap_name"])
+
+    if lp_snap:
+        flask.flash(
+            "It is not possible to change the git repository for this snap.",
```
nottrobin: Huh? Really? Why not?

webapp/publisher/snaps/views.py
```diff
@@ -1345,42 +1351,90 @@ def get_snap_builds(snap_name):
...
+        git_url = f"https://github.com/{github_repo}"
+
+        result = launchpad.new_snap(snap_name, git_url)
+
+        if result:
+            flask.flash(
+                "The GitHub repository was linked correctly.", "positive"
+            )
+        else:
+            flask.flash(
```
nottrobin: Are we expecting an error to occur? If not, couldn't we let this 500 naturally rather than masking it?

webapp/publisher/snaps/views.py
jkfran: Basically this form should be done just once by the user, this is a protection to check if the user somehow is trying to do the POST request twice.

The reason behind is because we are creating a Snap in launchpad. I agreed with @Lukewh to just let the user do it once. If we want to allow the user to change the repo multiple times then we will have to delete the snap from launchpad and create a new one (or to update the snap, I don't know), but we are not sure about the implications of doing that but it could be something that we want to allow in the future.

webapp/publisher/snaps/views.py
jkfran: Yes

webapp/publisher/snaps/views.py
jkfran: Filled https://github.com/canonical-web-and-design/snapcraft.io/issues/2466

webapp/publisher/snaps/views.py
jkfran: Filled https://github.com/canonical-web-and-design/snapcraft.io/issues/2467

webapp/publisher/snaps/views.py
nottrobin: Hmm. If we were sure the user submitted the form twice, we should say that in the message, or just silently ignore the second submission.

If we're not sure that's what happened, it seems like we should expose the error a bit more clearly.

How sure are we? Or can we tell the difference?

webapp/api/launchpad.py
```diff
@@ -1,6 +1,105 @@
...
+
+        return response
+
+    def get(self, resource, params=None):
+        """
+        Makes a raw HTTP GET request and returns the response.
+        """
+        response = self._request("GET", resource, params=params)
+
+        return response.json()
```
nottrobin: Should we be doing a [`response.raise_for_status()`](https://2.python-requests.org/en/master/api/#requests.Response.raise_for_status) call in all these functions? So that we get clearer error messages when we get erroneous responses from the API.

webapp/publisher/snaps/views.py
nottrobin: Oh wait...

If we got the response from the API, couldn't we check if the repo URL against that snap is actually the same as the one being submitted? If it was, then it seems safe to just continue happily. This should cover the double-submission problem.

If the repo being submitted is different, then we could raise a real error, rather than the `flask.flash` version. Or we could simply go ahead and call `launchpad.new_snap` and let the API respond with an error, which we can hopefully leave to error naturally (see my other comment about `raise_for_status`)

webapp/api/launchpad.py
jkfran: Hey @nottrobin , but we are, right? The `_request` method is doing it

webapp/publisher/snaps/views.py
jkfran: The API will create a new snap because the md5 hash of the git repo is different.

jkfran: It will raise an error only when using the same repo

jkfran: If there is a Snap in launchpad with this store_name if because it has already a git repo so we shouldn't create another one because in launchpad the store_name is not unique. What we could do is only show the error when the git repo is different and do nothing if the git repo is the same one. WDYT?

webapp/api/launchpad.py
```diff
@@ -1,6 +1,105 @@
...
+            f"oauth_consumer_key={username}, "
+            f'oauth_token="{token}", '
+            f'oauth_signature="&{signature}"'
+        )
+
+    def _request(self, method, url, params={}, data={}):
+        """
+        Makes a raw HTTP request and returns the response.
+        """
+        if not url.startswith("http"):
```
nottrobin: When are we using a URL that doesn't start with `BASE_URL`? It would be much neater to keep this class *explicitly* to calls made to the same API...

webapp/api/launchpad.py
nottrobin: Oh I see! Wow. In that case, I will change my critique to:

These `get` and `post` methods seem really redundant since we already have a `_request` method...

Maybe the `_request` method could be improved a bit by defaulting `method` to `"GET"`? I'm also suggesting a similar thing in @carkod's PR: https://github.com/canonical-web-and-design/anbox-cloud.io/pull/45#discussion_r370720385

nottrobin:
These are my snaps:

image

As you can see, none of them are owned by me, but I should still be able to make changes to them.

If I try to go to the "build" tab for any of them, I see this:

image

I imagine this is not something you can fix in your code but a bug with the API?

webapp/api/launchpad.py
```diff
@@ -1,6 +1,105 @@
...
+
+        return None
+
+    def new_snap(self, snap_name, git_url):
+        """
+        Create an ISnap in Launchpad
+        """
+        data = {
+            "ws.op": "new",
+            "owner": f"/~{self.username}",
```
nottrobin: Is the owner necessarily `self.username`? Mightn't the owner be different from the current user?

webapp/publisher/snaps/views.py
nottrobin: > It will raise an error only when using the same repo

This is kinda why I was confused by the `name` being a hash of the repo. The way I would expect this to work would be that you would tell it 2 things - the snap name and the repository you want to build from.

What happens if you make the `name` attribute the same as the `snap_name` instead of being generated from the `repo`?

jkfran: It will create a Snap in launchpad with this name. If we want to change this behaviour what happen with all the snaps already created in launchpad using the hash of another property? It will create an inconsistency in launchpad, for this feature everything will work as expected but I don't know what could go wrong if we decide to change that property.

webapp/api/launchpad.py
jkfran: All the snaps in launchpad are created by the same user. In this case (and always unless trying with the staging username) self.username is `os.getenv("LP_API_USERNAME")`

webapp/api/launchpad.py
jkfran: You are right. The response of a snap return properties with links and I was just using this to call those URLs.

webapp/api/launchpad.py
jkfran: Okay, I thought the `get` and `post` methods will be useful.

jkfran:
Hey @nottrobin , I did some changes for the review.

About the HTTPError, I have no idea why the API is doing that. In this case, you are accessing a public endpoint. You are doing a GET request to this URL:
https://api.launchpad.net/devel/+snaps?ws.op=findByStoreName&owner=%2F~build.snapcraft.io&store_name=documentation-builder

I opened this URL with my browser and I get the expected behaviour, a JSON.

Because my user permissions I can see other people snaps as a publisher and if I go to http://0.0.0.0:8004/documentation-builder/builds I don't' get this error.

webapp/publisher/snaps/views.py
nottrobin: @jkfran please file an issue about tidying up our usage of the API

nottrobin:
My problem was that I didn't have Launchpad API credentials in my local environment. Of course.

PR #2455, opened 2020-01-21
Use the correct last page link

### LINKED ISSUE #2453: Last page button doesn't always return the correct page
### Expected behaviour
The "Last >>" button should take you to the actual last page and not the last-1 page.

On a category page, specifically the one I'm noticing is https://snapcraft.io/search?category=games - Clicking the navigation button "Last >>" at the bottom right of the webpage takes you to page 5. There's actually 6 pages to the game category, and other pages of the games category (I.E, page 2 onwards), the last page button works properly.

I assume this problem might exist with the other categories but I haven't been able to reproduce it myself.

### Specs

- _URL:_ https://snapcraft.io/search?category=games
- _Operating system:_ Win 10 1909
- _Browser:_ Firefox 72.01

solazio:
## Done

- "Last >>" button link fixed to take you to the last page if you are on the first page 

## Issue / Card

Fixes #2453 

## QA

- Pull the branch
- View the site locally in your web browser at: http://0.0.0.0:8004/search?category=games&page=1
- Click on the "Last >>" link and see that you are taken to page 6

### DIFF
--- a/webapp/store/views.py
+++ b/webapp/store/views.py
@@ -145,15 +145,11 @@ def store_blueprint(store_query=None, testing=False):
         if not snap_searched and not snap_category:
             return flask.redirect(flask.url_for(".homepage"))
 
-        # The default size should be 44 (rows of 4)
-        # it's important that this is smaller than the category page 1 size
-        # below otherwise snaps can be missed out of results
-        size = 44
+        # The default size is 44
+        # (11 rows of 4) - on search results pages
+        # or (1 + 5 rows of 3 + 7 rows of 4) - on category page 1
 
-        # Page 1 has a snap at the top, and a few rows of 3, followed by rows
-        # of 4 - so we need to offset to ensure there's no hanging snap
-        if snap_category and page == 1:
-            size = 47
+        size = 44
 
         error_info = {}
         searched_results = []
@@ -209,7 +205,7 @@ def store_blueprint(store_query=None, testing=False):
 
         # These are the hand-selected "featured snaps" in each category.
         # We don't have this information on the API, so it's hardcoded.
-        number_of_featured_snaps = 19
+        number_of_featured_snaps = 16
 
         if snap_category_display and page == 1:
             if snaps_results and snaps_results[0]:
--- a/tests/store/tests_search.py
+++ b/tests/store/tests_search.py
@@ -221,7 +221,7 @@ class GetSearchViewTest(TestCase):
             snap_list.append({"package_name": "toto" + str(i), "icon_url": ""})
 
         payload = {
-            "_embedded": {"clickindex:package": snap_list[:47]},
+            "_embedded": {"clickindex:package": snap_list[:44]},
             "total": 144,
             "_links": {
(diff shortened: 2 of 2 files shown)


templates/store/search.html
```diff
@@ -94,8 +94,10 @@ <h2>We couldn't find the category <div class="no-search-query">{{category_displa
         {% if links.prev %}<a href="{{ links.prev }}">&#8249;&nbsp;Previous</a>{% endif %}
       </div>
       <div class="u-float-right u-align--right">
-        {% if links.next %}<a href="{{ links.next }}">Next&nbsp;&#8250;</a>{% endif %}
-        {% if links.last %}&nbsp;<a href="{{ links.last }}">Last&nbsp;&#187;</a>{% endif %}
+        {% if total > 47 %}
```
Lukewh: Why does the total need to be over 47? There should be a comment explaining the number, or better yet this should be a variable.
This condition would also be better in the related `view.py` - if the total isn't over 47 `links.next` and `links.last` don't get set and therefore won't be shown in the template.
Adding the condition against the magic number in the template means that there is now 4 potential conditions:
not over 47, next and last links don't show
over 47 and next link isn't defined so doesn't show, but last link is defined and shows
over 47 and next link is defined so shows, but last link isn't defined so doesn't show
over 47 and next and last links are both undefined so don't show
These potential 4 conditions are also spread between the template and the `views.py` which means if there's an error it's harder to debug which condition is failing.

webapp/store/views.py
```diff
@@ -172,7 +172,7 @@ def search_snap():
 
         if "total" in searched_results:
             total_results_count = searched_results["total"]
-            total_pages = ceil(total_results_count / size)
+            total_pages = ceil(total_results_count / 44)
```
Lukewh: Search results and category results have a different number of results on the page (hence the `size` variable) - setting this as a static number breaks one or the other

templates/store/search.html
solazio: The problem is that currently for category pages we show 47 snaps on the first page and 44 on the rest of the pages. Because of this we show snaps 45, 46 & 47 on both first and second page. After discussing with Robin and filing an [issue](https://github.com/canonical-web-and-design/snapcraft.io/issues/2457) the easiest/best thing to do is to update the design of the first page to show 44 snaps (like the other pages). 
Regarding the condition in the code, I will  add it to `views.py` as it would be easier to manage.  Basically it is needed to hide the "Last page" navigation if we have less than 47 snaps, as the last page would be page 2 and would contain maximum 3 snaps, already shown on page 1.

webapp/store/views.py
solazio: As far as I can see, both search results and category results display 44 snaps per page, except first page of category results. Or am I missing something?

templates/store/search.html
Lukewh: I don't think we need to make that change as a second issue. 
In `views.py` you could change `19` to `16` on line 212, remove lines 148-157 (so there's no conditional changing things on category pages).
After that you should be able to remove the conditional added to the template too.

PR #2469, opened 2020-01-27
Move progressiveRow to class. Throttle onChange handler

### LINKED ISSUE #2411: Investigate release UI action bottlenecks
Specifically for progressive releases [see comment](https://github.com/canonical-web-and-design/snapcraft.io/pull/2395#issuecomment-567918676).

Lukewh:
## Done

- Move progressiveRow to standard React component (so isn't initalized and event bound every time the percentage is updated)
- Add [react-throttle-render](https://gitlab.com/ryo33/react-throttle-render) library and selectively throttle components

## Issue / Card

Fixes #2411 

## QA

- Pull the branch
- Prepare a few progressive releases
- Open the details before you save
- Scroll/ drag the first slider - the others should update smoothly (if use for all checkbox is checked)

### DIFF
--- a/static/js/publisher/release/components/progressiveBar.js
+++ b/static/js/publisher/release/components/progressiveBar.js
@@ -1,5 +1,4 @@
 import React from "react";
-import throttle from "react-throttle-render";
 import PropTypes from "prop-types";
 
 import throttle from "../../../libs/throttle";
@@ -261,18 +260,4 @@ InteractiveProgressiveBar.propTypes = {
   minPercentage: PropTypes.number
 };
 
-const DisabledProgressiveBar = throttle(250)(props => (
-  <ProgressiveBar
-    percentage={props.percentage}
-    targetPercentage={props.targetPercentage}
-    readonly={true}
-    disabled={true}
-  />
-));
-
-DisabledProgressiveBar.propTypes = {
-  percentage: PropTypes.number,
-  targetPercentage: PropTypes.number
-};
-
-export { ProgressiveBar, InteractiveProgressiveBar, DisabledProgressiveBar };
+export { ProgressiveBar, InteractiveProgressiveBar };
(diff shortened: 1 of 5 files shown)


bartaz:
Not sure if this is related to this PR or some other error, but if you have existing progressive release, change the %, you can only update the % in release history, but not in confirm box:

progressive-updates-3

static/js/publisher/release/components/progressiveBar.js
```diff
@@ -171,7 +173,9 @@ class InteractiveProgressiveBar extends React.Component {
     };
 
     this.setState(newState);
-    onChange(target);
+    if (onChange) {
+      throttle(onChange(target), 250);
```
bartaz: `throttle` argument should be a function, does `onChange(target)` return a function to call?
Or should it be something like `throttle(() => onChange(target), 250`?

Lukewh: You're right - and when it's called correctly it prevents the UI updating.

bartaz:
Updated functionality works good.

I found another issue, likely unrelated to this PR:

Its about when updates in 'confirm box' are not propagated to the cells in the releases table.

Case 1)

1) Have a prepared progressive release (update its % to less than 100%)
2) Drop another release
3) By default new release will use the same % as the previous one
4) Percentage in the confirm box correctly shows same % for both, but bar in the releases table shows 100% for new release
5) Move the bar in confirm box, everything gets updated correctly

progressive-updates-4

Case 2)

1) Have multiple releases prepared
2) Update their (shared) % to value different than 100%
3) Unselect the checkbox to use same %
4) Change values of individual %s
5) Select checkbox again (to use same %)
6) Bars in confirm box will switch to show same value, but bars in releases table will show previous different values for each cell
7) Move the bar in confirm box, all % get updated to correct value
progressive-updates-5

PR #2479, opened 2020-01-28
Check if snapcraft.yaml is present

### LINKED ISSUE #2449: Check whether snapcraft.yaml is present and show a message if not
on page load
Are there builds?

YES: 
 - load the page
 
NO: 
 - check against github repo (see build for possible snapcraft.yaml locations and names)
 - if now present - trigger the first build
 - if not present - do nothing

jkfran:
## Done

In this PR we are using the GitHub GraphQL API to:

- List all the user repositories and mark the ones with a snapcraft.yaml present:
github-graphql1

- Show an error on the builds page when the selected repo for the snap doesnt contain a snapcraft.yaml:
github-graphql2

## Issue / Card

Fixes #2449

## QA

- Pull the branch
- Login as a publisher
- Login with GitHub
- Try to link you snap with a github repo and check that the repos contains a mark when snapcraft.yaml is present
- Check if a snap with a github repo without snapcraft.yaml get the error

### DIFF
--- a/webapp/api/github.py
+++ b/webapp/api/github.py
@@ -1,5 +1,5 @@
-from urllib.parse import urljoin
 from webapp import api
+from werkzeug.exceptions import Unauthorized
 
 
 class GitHubAPI:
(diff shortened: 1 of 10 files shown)


webapp/publisher/snaps/views.py
```diff
@@ -1390,11 +1404,16 @@ def get_snap_builds(snap_name):
...
+        try:
+            context["github_user"] = github.get_username()
+            # Get snapcraft repositories sorted by snapcraft_yaml
+            context["github_repositories"] = sorted(
+                github.get_user_repositories(),
+                key=lambda k: k["snapcraft_yaml"],
+                reverse=True,
+            )
+        except Unauthorized:
+            pass
```
nottrobin: Why wouldn't we care about an unauthorized response?

jkfran: If we get an Unauthorized exception here that means the user is not logged in GitHub, that means github_user will be None in the template and we will show the login button.

nottrobin: Then maybe you could just put the line in here `context["github_user"] = None`? Just to show that we are deliberately doing something.

jkfran: Yes, it will make it easier to understand

webapp/api/github.py
```diff
@@ -18,59 +18,121 @@ def __init__(self, access_token=None, session=api.requests.Session()):
...
+        gql = (
+            '''{
+            repository(name: "'''
+            + repo
+            + '''", owner: "'''
+            + owner
+            + """") {
+                id
+                nameWithOwner
+                yamlLocation1:
```
Lukewh: is there a way to template some of this out with GraphQL? For instance the repetition of these yamlLocations?

jkfran: Luke I changed this request to use the Rest API because we might don't need the user session to verify this.

jkfran:
Hey @nottrobin , this is ready to be reviewed.

nottrobin:
@jkfran think you could consider testing this branch a bit as @Lukewh suggests? I'd be happy to pair on that with you on Monday / Tuesday if you want?

jkfran:
Hello @nottrobin ,

Thank you for your proposal, I have created a new commit in this branch with the changes for the API testing.

Let me know what do you think and if you think is better to pair for the review.

Thank you.

webapp/publisher/snaps/views.py
```diff
@@ -1356,11 +1357,29 @@ def get_snap_builds(snap_name):
...
+    except Unauthorized:
+        context["github_user"] = None
+
     if lp_snap:
         # Git repository without GitHub hostname
         context["github_repository"] = lp_snap["git_repository_url"][19:]
+        github_owner, github_repo = context["github_repository"].split("/")
+
+        # Check if this repo has a snapcraft.yaml
+        context["yaml_file_exist"] = github.is_snapcraft_yaml_present(
```
nottrobin:
```suggestion
        context["yaml_file_exists"] = github.is_snapcraft_yaml_present(
```

webapp/publisher/snaps/views.py
```diff
@@ -1356,11 +1357,29 @@ def get_snap_builds(snap_name):
     }
 
     # Get built snap in launchpad with this store name
+    github = GitHubAPI(flask.session.get("github_auth_secret"))
     lp_snap = launchpad.get_snap_by_store_name(details["snap_name"])
 
+    try:
+        context["github_user"] = github.get_username()
```
nottrobin: If this snap has `builds` lower down, do we need any information at all from their GitHub account? 'cos if not, we should move this request lower down in the logic so we're not making it if we don't need to.

jkfran: You are right, I moved it at the top when trying to use GraphQL to check the yaml file in the repo. But now that method doesn't need user authentication.

PR #2504, opened 2020-02-10
Connect to repo

### LINKED ISSUE #2484: Connect to repo page with repo filter

### LINKED ISSUE #2485: Style the build list heading "your snap is linked to"

solazio:
## Done

-  Add connect snap to repo functionality
- Update styling for Builds

## Issue / Card

Fixes #2484 
Fixes #2485 

## QA

- Pull the branch
- View the site locally in your web browser at: http://0.0.0.0:8004/[snap_name]/builds
- Check a snap with builds and notice the new position of the `Trigger new build` button
- Check a snap that is not connected to any repos (both with JS on and disabled) and see that you can connect the snap to your repo and trigger a new build

## Screenshots


static/js/publisher/builds/components/searchSelect.js
```diff
@@ -0,0 +1,408 @@
...
+      // Get the correct objects for the currentValues as selected
+      selected: this.props.values.filter(value =>
+        this.props.value.includes(value.key)
+      ),
+      values: values,
+      searchTerm: "",
+      searchResults: values,
+      showSearch: false,
+      highlightedOption: null,
+      isMulti: this.props.isMulti || false,
```
Lukewh: We should put default prop values in a `defaultProps` object. Like `SearchSelect.propTypes` you do:
```
SearchSelect.defaultProps = {
  isMulti: false,
  placeholder: ""
};
```

static/js/publisher/builds/components/searchSelect.js
```diff
@@ -0,0 +1,408 @@
...
+  sortByName(a, b) {
+    return a.name.localeCompare(b.name);
+  }
+
+  /**
+   * Remove an item from the 'selected' list.
+   * Add it back into the 'values' list.
+   */
+  removeItem(key) {
+    const toRemove = this.state.selected.filter(item => item.key === key)[0];
```
Lukewh: I know this is probably my code, but we should use destructuring for getting state and props:
```
const {selected, values} = this.state;
const {getSelectedItems} = this.props;
```

Then in the function we can just have things like:
```
const newValues = values.slice(0);
const newSelected = selected.filter(...);

... getSelectedItems(selected);
```

static/js/publisher/builds/components/searchSelect.js
```diff
@@ -0,0 +1,408 @@
...
+    }
+  }
+
+  /**
+   * Add an item to the 'selected' list.
+   * Remove it from the 'values' list.
+   */
+  addItem(key) {
+    // Get the object based on the key
+    const toAdd = this.state.values.filter(item => item.key === key)[0];
```
Lukewh: Can be destructured too

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,143 @@
...
+    });
+  }
+
+  /**
+   * Update repo list with repos of the selected organization
+   *
+   * @param value
+   */
+  updateRepoList(value) {
+    const organization = this.state.organizations.filter(
```
Lukewh: I think the value and label for repo's should be the same - whatever is shown in the url bar.
That way when you read the row it'll say "canonical-web-and-design/snapcraft.io" not "Canonical web and design/snapcraft.io". It also means this filter won't be required.

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,143 @@
...
+    alert(organization);
+    fetch("https://jsonplaceholder.typicode.com/posts")
+      .then(response => response.json())
+      .then(json => {
+        return json;
+      });
+  }
+
+  render() {
+    return (
```
Lukewh: destructure again for this class, especially for render, makes the code a lot easier to read.

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,143 @@
...
+    };
+
+    this.updateRepoList = this.updateRepoList.bind(this);
+    this.updateSelectedRepo = this.updateSelectedRepo.bind(this);
+  }
+
+  /**
+   * Update the original input value and dispatch a change event to the input and form.
+   */
+  // updateHandler(input, delimiter) {
```
Lukewh: you'll need this. The page should have a simple text input for the full repo path that get's hidden and replaced by this component. Then when this component updates with both the org and the repo the input value get's updated ready to be submitted to move to the next step. It's also a fallback for no JS or JS breaking for whatever reason. Check multiSelect's "init" function.

Lukewh:
Might need a rebase
Screenshot_2020-02-17  Builds for test-snap-lukewh

danielmutis:
@solazio  We are missing the messages telling users what has gone wrong and how to fix it. I.e. https://zpl.io/2Z71deL

Next to the github username there is a 'Change Account' link missing. 

Also, if we could centre that X next to the repo name it'd be 👌🏾

Lukewh:
> Next to the github username there is a 'Change Account' link missing.
We don't have this functionality yet

Lukewh:
The repo name list should just be the repo name - it shouldn't include the org again.
Screenshot_2020-02-17  Builds for lukewh-test-2

Lukewh:
The styling needs to be updated so it look like a normal input - you can't select multiple so having the `x` and pill styling doesn't make sense.
From
Screenshot_2020-02-17  Builds for lukewh-test-2 (1)
To
Screenshot_2020-02-17  Builds for lukewh-test-2 (2)

You could do this by setting a class on the component.

Another issue is that when something is selected - when you click on the input you don't get the list again, you have to remove the single item with the "x" to then add another.

danielmutis:
Can we have the 'Try again' as an icon like in the design?

Lukewh:
If you add a repo, then immediately remove it the check still happens and show's an error for a blank input
output
we should either cancel the check or ignore the result if the input value changes.

Lukewh:
Just to make a note:
- we decided in a meeting to use the HTML dataset component. This should solve many of the issues raised here
- @solazio will ask @waynecrosby for the retry icon again

waynecrosby:
@Lukewh We don't have a 'retry' icon per se but we do have a 'restart' and a 'change-version' option that I think will probably work ok. See what you think of the below...
Screenshot 2020-02-17 at 12 42 05
Screenshot 2020-02-17 at 12 42 22

Lukewh:
> @Lukewh We don't have a 'retry' icon per se but we do have a 'restart' and a 'change-version' option that I think will probably work ok. See what you think of the below...
> Screenshot 2020-02-17 at 12 42 05
> Screenshot 2020-02-17 at 12 42 22

Perfect thanks @waynecrosby. I think the change-version is probably the most accurate for this application Screenshot 2020-02-17 at 12 42 22

waynecrosby:
That's actually the 'restart' icon @Lukewh, my bad for the order sorry.
If you are still happy with that one let me know and I'll send the SVG to Ovi.

Lukewh:
@waynecrosby you troll. Yeah happy with that one thanks!

Lukewh:
> Might need a rebase
> Screenshot_2020-02-17 Builds for test-snap-lukewh

Still see this issue

Lukewh:
`get started with a template` link links to the username/snap-name, not the repo. For example my repo is called `Lukewh/testing123`, the snap is called `super-test-snap` and the link links to `Lukewh/super-test-snap` which isn't a repo, it should link to `Lukewh/testing123`

Lukewh:
Is there a reason the error icon is outside the input? Vanilla has it inside:
Screenshot_2020-02-19  Builds for test-lukewh-11

https://docs.vanillaframework.io/base/forms#validation

In the message "Name mismatch: the snapcraft.yaml uses the snap name "test-snap-lukewh", but you've registered the name "test-lukewh-11". Update your snapcraft.yaml to continue." can we add a link to "Update your snapcraft.yaml" that links to https://github.com/[org or user]/[repo]/edit/master/[path to snapcraft.yaml]?

It might be nice to add a tooltip to the retry button that says "Retry" or "Re-check" too. @danielmutis wdyt?

danielmutis:
@Lukewh Yeah we can add the tooltip. 'Re-check' is probably more accurate. 

I checked with @spencerbygraves  and he agrees that the best location for the icon is outside as it is.

anthonydillon:
@spencerbygraves if we are changing the form validation pattern here please propose the change to Vanilla too please

spencerbygraves:
@anthonydillon I'm not proposing a change, apparently this is a data field with an arrow icon that needs to go in that spot. @danielmutis, can you supply a screenshot please.

bartaz:
Currently for selects Vanilla just uses border color to show error:

Screenshot 2020-02-19 at 15 25 15

Icon is not rendered at all (to avoid overlapping with arrow I suppose).

danielmutis:
@spencerbygraves @anthonydillon  This is a screenshot of what currently happens. The issue is similar to the one @bartaz showed on that example: the arrow of the datalist is only visible on hover, so the placement of the error icon would be a bit odd.

I guess if the case already exists where we just don't show it, we can be consistent with that and show no icon, just the red outline.

anthonydillon:
This is how Vanilla solves this. 
Screenshot_20200219-160515

bartaz:
Ah right, @anthonydillon is correct. I haven't notice this example. There is specific class for validating selects.

danielmutis:
@anthonydillon I think this solution wouldn't work in this case because the arrow only shows on hover in most browsers and doesn't even appear on Firefox. That would make the error icon look like it's floating when the arrow doesn't show.

I think the best way to go is to have it outside which would at least be more explicit than just the colour outline.

static/js/publisher/builds/components/datalistSelect.js
```diff
@@ -0,0 +1,58 @@
+import React, { Fragment } from "react";
+import PropTypes from "prop-types";
+
+class DatalistSelect extends React.Component {
```
Lukewh: Could just be:

```
const DatalistSelect = ({props obj from line 10}) => (
  <Fragment>...</Fragment>
);
```
No need to do the full class dance

static/js/publisher/builds/components/select.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+class Select extends React.Component {
```
Lukewh: Again could be

```
const Select = ({options, updateSelection, disabled, selectedOption}) => (
  <select>...</select>
);
```

static/js/publisher/builds/components/select.js
```diff
@@ -0,0 +1,45 @@
...
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    const { options, updateSelection, disabled, selectedOption } = this.props;
+    return (
+      <div>
```
Lukewh: any reason for the extra div?

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,348 @@
...
+
+  renderMessage() {
+    const {
+      message,
+      selectedRepo,
+      selectedOrganization,
+      yamlFilePath
+    } = this.state;
+    if (message) {
+      if (message.description.slice(0, 14) === "Name mismatch:") {
```
Lukewh: if we had `status` you could base this on `NAME_MISMATCH`

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,348 @@
...
+    super(props);
+
+    this.state = {
+      selectedRepo: "",
+      repoList: [],
+      selectedOrganization: "",
+      organizations: this.props.organizations,
+      user: this.props.user,
+      isRepoListDisabled: true,
+      message: { icon: "", description: "" },
```
Lukewh: I'd feel more comfortable with this being:
```
message: <string>,
status: <type>,
```

Where we can set the "type" in some consts like:
```
const LOADING = "LOADING";
const ERROR = "ERROR";
const NAME_MISMATCH = "NAME_MISMATCH";
const SUCCESS = "SUCCESS";
```

Then in the code we can set:
- for default state
```
message: null,
status: null
```
- for any other state
```
message: "The message"|null,
status: LOADING|ERROR|NAME_MISMATCH|SUCCESS,
```

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,348 @@
...
+
+      this.setState({
+        isRepoListDisabled: true,
+        message: { icon: "spinner", description: "" }
+      });
+
+      fetch(url)
+        .then(res => res.json())
+        .then(result => {
+          this.setState({
```
Lukewh: Does the API return a result.error.type or some such that you can match against for NAME_MISMATCH so you don't have to string match later on? The result.error.message could change

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,348 @@
...
+            </a>
+          </p>
+        </div>
+      );
+    }
+  }
+
+  renderIcon() {
+    const { message } = this.state;
+    if (message.icon) {
```
Lukewh:
```
if (status) {
  let icon = "";
  switch(status) {
    case LOADING:
    default:
      icon = "spinner";
      break;
    case ERROR:
    case NAME_MISMATCH:
      icon = "error";
      break;
    case SUCCESS:
      icon = "success";
      break;
  }
}
```

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,348 @@
...
+      return (
+        <span className="p-build__icon">
+          <i
+            className={`p-icon--${message.icon}${
+              message.icon === "spinner" ? " u-animation--spin" : ""
```
Lukewh: `status === LOADING ? "u-animation--spin" : ""`

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,348 @@
...
+            }`}
+          />
+        </span>
+      );
+    }
+  }
+
+  renderButton() {
+    const { message } = this.state;
+    if (message.icon === "error") {
```
Lukewh: Using status makes this a lot cleaner - "message" means nothing for whether to render a button

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,360 @@
...
+        .then(result => {
+          if (result.success) {
+            this.setState({
+              isRepoListDisabled: false,
+              status: SUCCESS
+            });
+          } else {
+            this.setState({
+              isRepoListDisabled: false,
+              status: result.error.type,
```
Lukewh: :+1: nice using the same status as the API returns

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,360 @@
...
+    const {
+      message,
+      status,
+      selectedRepo,
+      selectedOrganization,
+      yamlFilePath,
+      snapName
+    } = this.state;
+    if (status === SNAP_NAME_DOES_NOT_MATCH) {
+      const splitDescription = message.split(". ");
```
Lukewh: We shouldn't rely on the format of the message returned by the API. At some point someone might think "Oh this should say 'Yo - this don't match'" and change it in the backend. At which point this will no longer match and break this code path. It's better to just type what we want the message to say here.

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,360 @@
...
+    const { status } = this.state;
+    if (
+      status === ERROR ||
+      status === SNAP_NAME_DOES_NOT_MATCH ||
+      status === MISSING_YAML_FILE
+    ) {
+      return (
+        <button
+          className="p-tooltip--btm-center"
+          aria-describedby="btm-cntr"
```
Lukewh: This would be better with an id like "recheck-tooltip". "btm-cntr" isn't very descriptive.

static/js/publisher/builds/components/repoConnect.js
```diff
@@ -0,0 +1,360 @@
...
+      case SUCCESS:
+        icon = " is-success";
+        break;
+    }
+    return icon;
+  }
+
+  componentDidMount() {
+    // Pre-select user's own "organization" if the user is not part of any organizations
+    if (this.state.organizations.length === 2) {
```
Lukewh: Maybe a little note of why this works if length is 2. I would have thought if the user isn't part of an org, this would be === 1.

static/js/publisher/builds/components/select.js
```diff
@@ -0,0 +1,31 @@
+import React, { Fragment } from "react";
+import PropTypes from "prop-types";
+
+const Select = ({ options, updateSelection, disabled, selectedOption }) => (
+  <Fragment>
+    <select
```
Lukewh: Only one element, Fragment not needed.

static/sass/_snapcraft_p-build.scss
```diff
@@ -21,6 +23,17 @@
     height: 30px;
     vertical-align: middle;
   }
-}
 
+  .p-icon--restart {
+    @extend %icon;
+    background-image: url("#{$assets-path}34f54cf0-restart.svg");
```
Lukewh: We should probably add a little comment that says to switch to the vanilla icon when these get added to vanilla.

static/js/publisher/builds/components/repoConnect.js
Lukewh: This still seems relevant - seems like you already had to update it based on the error in the API being different?

solazio: In a perfect world you should see the same message for both js-enabled or js-disabled situations. However, the js-enabled message has a part that is bold, has a link, etc. and therefore is slightly different. Also in case there is a name mismatch, the message contains the name of whatever it is in the yaml file. Do we want to keep the messages separate (js-enabled and js-disabled)?

PR #2505, opened 2020-02-11
Replace forEach with regular for loops for showMore elements

### LINKED ISSUE #2497: Replace forEach with for loop
There are some older browsers causing sentry error when hitting this code: 
```js
showMoreContainer.forEach(el => {
```

This could be resolved by replacing the forEach with an old fashioned for loop.

solazio:
## Done

- Replace forEach with regular for loops for showMore elements

## Issue / Card

Fixes #2497 

## QA

- Pull the branch
- Click on any language icons and then click "Show more"
- Note it works as expected

### DIFF
--- a/static/js/public/fsf-expandable-yaml-details.js
+++ b/static/js/public/fsf-expandable-yaml-details.js
@@ -1,6 +1,6 @@
 export default function initExpandableYaml() {
-  const showMoreContainer = document.querySelectorAll(
-    "[data-js='js-show-more']"
+  const showMoreContainer = [].slice.call(
+    document.querySelectorAll("[data-js='js-show-more']")
   );
 
   if (showMoreContainer) {
--- a/static/js/public/fsf-language-select.js
+++ b/static/js/public/fsf-language-select.js
@@ -18,8 +18,8 @@ function initFSFLanguageSelect(rootEl) {
 
   // reset expandable yaml files to being truncated
   const resetExpandableYaml = () => {
-    const showMoreContainer = document.querySelectorAll(
-      "[data-js='js-show-more']"
+    const showMoreContainer = [].slice.call(
+      document.querySelectorAll("[data-js='js-show-more']")
     );
 
     if (showMoreContainer) {


bartaz:
Not that I have anything against this change, but `forEach` on arrays is widely supported.
We shouldn't go back to using for loop if we can just cast nodelist to array.

squidsoup:
Is this build not babelised? Seems a shame to use old js.

solazio:
> Not that I have anything against this change, but `forEach` on arrays is widely supported.
> We shouldn't go back to using for loop if we can just cast nodelist to array.
> 
> Like we do for example here:
> https://github.com/canonical-web-and-design/snapcraft.io/pull/2505/files#diff-4a1f549679dffec0f6444c3bd19af4ffL4

@bartaz is it not better to use `Array.prototype` instead of `[ ]` as it does the same thing (as far as I know) but it's more explicit (at least in my head) even though it's more character?

squidsoup:
> > Not that I have anything against this change, but `forEach` on arrays is widely supported.
> > We shouldn't go back to using for loop if we can just cast nodelist to array.
> > Like we do for example here:
> > https://github.com/canonical-web-and-design/snapcraft.io/pull/2505/files#diff-4a1f549679dffec0f6444c3bd19af4ffL4
> 
> @bartaz is it not better to use `Array.prototype` instead of `[ ]` as it does the same thing (as far as I know) but it's more explicit (at least in my head) even though it's more character?

`[].foo` is pretty idiomatic @solazio - in some browsers `Array.prototype` performs better, but not in any meaningful way.

solazio:
@bartaz and @squidsoup does [this](https://github.com/canonical-web-and-design/snapcraft.io/pull/2509/files#diff-2ac26b14a50a30144ef8ddf5505639beR196) do the same thing? If yes, we could use that option instead...?
@steverydz ^

bartaz:
Yes, this

```js
[].forEach.call(nodes, function(node) {})
```

simply calls `forEach` from an array on your nodes object.

I kind of prefer assigning `nodes = [].slice.call(nodes)` in the first place, because it simply creates an array of nodes once for all. Without it you would need to remember to do the `[].forEach.call` dance every time you iterate on given list of nodes.

steverydz:
That makes sense. I guess it depends on how much you're likely to loop over something.

bartaz:
I would say, by convention, whenever you are doing `querySelectorAll` you should immediately turn the result into an array.

PR #2528, opened 2020-02-17
FSF expand next accordion when clicking register snap name

solazio:
## Done

- Expand next accordion when clicking register snap name and no errors

## Issue / Card

Fixes https://github.com/canonical-web-and-design/snap-squad/issues/1038

## QA

- Pull the branch
- View the site locally in your web browser at: http://0.0.0.0:8004/first-snap/python/linux/push
- Click on `Register snap name` and see the next accordion panel is expanded automatically.

### DIFF
--- a/static/js/public/first-snap-flow.js
+++ b/static/js/public/first-snap-flow.js
@@ -2,6 +2,8 @@
 
 import "whatwg-fetch";
 
+import { toggleAccordion } from "./accordion";
+
 function install(language) {
   const osPickers = Array.from(document.querySelectorAll(".js-os-select"));
   const osWrappers = Array.from(document.querySelectorAll(".js-os-wrapper"));
@@ -267,6 +269,14 @@ function initRegisterName(formEl, notificationEl, successEl) {
         } else if (json.code == "created") {
           showSuccess(`Name "${json.snap_name}" registered successfully.`);
         } else if (json.code == "already_owned") {
+          // Jump to the next accordion panel
+          const currentPanel = formEl.closest(".p-accordion__group");
+          const currentToggle = currentPanel.querySelector(".p-accordion__tab");
+          const nextPanel = currentPanel.nextElementSibling;
+          const nextToggle = nextPanel.querySelector(".p-accordion__tab");
+          toggleAccordion(currentToggle, false);
+          toggleAccordion(nextToggle, true);
+
           showSuccess(`You already own "${json.snap_name}"".`);
         }
       })
--- a/static/js/public/first-snap-flow.js
+++ b/static/js/public/first-snap-flow.js
@@ -257,6 +257,21 @@ function initRegisterName(formEl, notificationEl, successEl) {
   formEl.addEventListener("submit", event => {
     event.preventDefault();
     let formData = new FormData(formEl);
+    const submitButton = formEl.querySelector(
+      "[data-js='js-snap-name-register']"
+    );
+
+    const currentPanel = formEl.closest(".p-accordion__group");
+    const currentToggle = currentPanel.querySelector(".p-accordion__tab");
+    const nextPanel = currentPanel.nextElementSibling;
+    const nextToggle = nextPanel.querySelector(".p-accordion__tab");
+
+    // Show spinner if data fetch takes long
+    setTimeout(() => {
+      if (currentToggle.getAttribute("aria-expanded") === "true") {
+        submitButton.classList.add("has-spinner");
+      }
+    }, 400);
 
     fetch("/register-snap/json", {
       method: "POST",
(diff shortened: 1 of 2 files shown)


Lukewh:
We should probably show some kind of spinner after clicking register. I just pressed it multiple times thinking it wasn't working - maybe show the spinner in the button or something? Wdyt @danielmutis Also the goto listing button seems to do nothing too... Unrelated to this I imagine.

solazio:
> We should probably show some kind of spinner after clicking register. I just pressed it multiple times thinking it wasn't working - maybe show the spinner in the button or something? Wdyt @danielmutis?
> 
> Also the goto listing button seems to do nothing too... Unrelated to this I imagine.

Done. Also create a [new issue](https://github.com/canonical-web-and-design/snapcraft.io/issues/2533) for `Go to listing` button.

danielmutis:
I think it's a good idea to show the spinner just to signal that something is happening. When I did it it happened so quickly I didn't even see it.

anthonydillon:
@Caleb-Ellis has worked on buttons with action feedback for MAAS.

Caleb-Ellis:
We have a [react component called ActionButton](https://github.com/canonical-web-and-design/react-components/blob/master/src/components/ActionButton/ActionButton.js) that we use in maas-ui. At some point the css was going to be upstreamed to Vanilla but I don't think that's happened yet, so it's hardcoded here.

But yeah like @danielmutis said sometimes state changes/requests resolve in like 10ms, so we set a minimum of 400ms for the spinner otherwise it looks jumpy. I don't know the use-case here, but in maas-ui we also use a success icon for 2000ms.

The docs/demos for the react-components is unbeLIEVABLY broken, but you should be able to see it in action [here](http://karura.internal:5240/MAAS/r/settings/images/ubuntu) if you're using the VPN (plz don't change other settings). You can find credentials on the wiki if you search for "Design MAAS lab"

anthonydillon:
There is an example here: https://github.com/canonical-web-and-design/vanilla-framework/pull/2027#issuecomment-446153209

solazio:
@anthonydillon my button is not react, and I don't think it worth using react just to have that button. @danielmutis do you think we need to set a minimum duration for the spinner?

Lukewh:
@solazio check out https://github.com/canonical-web-and-design/snapcraft.io/blob/master/static/js/publisher/form.js#L281
The listing form show's a spinner after 2seconds. You can use the markup for that button, this JS with reduced time

anthonydillon:
Sure, borrow the way the button works in Vanilla if that helps.

solazio:
I did it this way:
- If it takes more than 400ms for the fetch request, show the spinner and jump to the next accordion section once the fetch is done
- If it takes less than 400ms, jump to the next accordion section once the fetch is done

Lukewh:
If there's an error, the spinner isn't removed

Screenshot_2020-02-19 Snapcraft - Snaps are universal Linux packages

Lukewh:
Also when the spinner is enabled, the button should be disabled, to stop people clicking it lots of times

static/js/public/first-snap-flow.js
```diff
@@ -255,22 +257,66 @@ function initRegisterName(formEl, notificationEl, successEl) {
...
     fetch("/register-snap/json", {
       method: "POST",
       body: formData
     })
       .then(response => response.json())
       .then(json => {
+        clearTimeout(timer);
         if (json.errors) {
           showError(json.errors[0].message);
+          if (submitButton.disabled) {
```
Lukewh: This is repeated a couple of times, could be a function above the fetch somewhere

```
const enableButton = () => {
  if (submitButton.disabled) {
    submitButton.classList.remove("has-spinner");
    submitButton.disabled = false;
  }
}
```

Then just call `enableButton()` in each of these if's. That way if we want to add a different class in the future we just update 1 place.

static/js/public/first-snap-flow.js
```diff
@@ -255,22 +257,66 @@ function initRegisterName(formEl, notificationEl, successEl) {
...
+    }, 400);
 
     fetch("/register-snap/json", {
       method: "POST",
       body: formData
     })
       .then(response => response.json())
       .then(json => {
+        clearTimeout(timer);
         if (json.errors) {
```
Lukewh: This could be reduced down:
```
clearTimeout(timer);
let message;
if (json.errors) {
  showError(json.errors[0].message);
  if (submitButton.disabled) {
    ...
  } 
  return;
} else if (json.code === "created") {
  message = `Name "${json.snap_name}" registered successfully.`;
} else if (json.code === "already_owned") {
  message = `You already own "${json.snap_name}"".`;
}
// Jump to the next accordion panel
toggleAccordion(currentToggle, false);
toggleAccordion(nextToggle, true);

if (submitButton.disabled) {
  submitButton.classList.remove("has-spinner");
  submitButton.disabled = false;
}

showSuccess(message); 
```

PR #2549, opened 2020-02-21
Build pages with logs

### LINKED ISSUE #2488: Build the build details page - with logs

Lukewh:
## Done

- Add single build page

## Issue / Card

Fixes #2488 

## QA

- Pull the branch
- Visit http://0.0.0.0:8004/snap_name/builds for a snap that has builds
- Click on one of the build numbers from the table
- See the build details page in all its glory

### DIFF
--- a/templates/publisher/build.html
+++ b/templates/publisher/build.html
@@ -1,0 +1,83 @@
+{% extends "publisher/_publisher_layout.html" %}
+
+{% block meta_title %}
+Builds for
+{% if snap_title %}{{ snap_title }}{% else %}{{ snap_name }}{% endif %}
+{% endblock %}
+
+{% block content %}
+<div id="main-content" class="u-no-margin--top">
+  {% set selected_tab='builds' %}
+  {% set suppress_nav=True%}
+  {% set suppress_title=True%}
+  {% set snap_link="builds" %}
+  {% include "publisher/_header.html" %}
+
+  <section class="p-strip is-shallow u-no-padding--top" id="top">
+    <div class="row">
+      <h1 class="p-heading--three">Build #{{snap_build.id}}</h1>
+      <div id="builds-wrapper" class="col-12">
+        <table>
+          <thead>
+            <tr>
+              <th>ID</th>
+              <th>Architecture</th>
+              <th class="u-hide--small">Duration</th>
+              <th class="has-icon">Result</th>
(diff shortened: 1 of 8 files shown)


webapp/publisher/snaps/build_views.py
```diff
@@ -151,6 +192,10 @@ def validate_repo(github_token, snap_name, gh_owner, gh_repo):
     return result
 
 
+def get_build_log(log_url):
```
Lukewh: I'd anticipate this isn't the right place to do this, or the right thing to do....

webapp/publisher/snaps/build_views.py
```diff
@@ -24,7 +25,7 @@
 )
 
 
-def get_builds(lp_snap, context, selection):
+def get_builds(lp_snap, context, build_id, selection):
```
Lukewh: I think this needs to be refactored so it doesn't modify the context and just returns the required fields that can be added to the context.

Lukewh: https://github.com/canonical-web-and-design/snapcraft.io/issues/2550

webapp/publisher/snaps/build_views.py
jkfran: Hey Luke, Instead of doing this here, why not modify the `webapp/api/launchpad.py` with a method that directly return the log content?

webapp/publisher/snaps/build_views.py
```diff
@@ -34,6 +35,13 @@ def get_builds(lp_snap, context, selection):
         lp_snap["builds_collection_link"][32:]
     )
 
+    if build_id:
```
jkfran: It will be better to create a new API call to launchpad to get only one build. It is not good to get all the builds and then select one.

templates/publisher/build.html
```diff
@@ -0,0 +1,82 @@
...
+{% endblock %}
+
+{% block scripts %}
+
+<script>
+  window.addEventListener("DOMContentLoaded", function () {
+    Raven.context(function () {
+      {% if snap_builds %}
+        snapcraft.publisher.initBuilds('#builds-wrapper',
+                                     {{ snap_name | tojson }},
```
solazio: the indent for lines 72 to 75 doesn't look right - at least on github...

PR #2555, opened 2020-02-21
Switch over to using canonicalwebteam.launchpad

nottrobin:
The code from webapp/api/launchpad.py was ported to
canonicalwebteam.launchpad, so it could be shared by ubuntu.com's
image builder stuff. It just means a few small interface changes.

Update this code to use that new module and interface, and remove
the old code from this codebase.

QA
--

Run the site, go to your publisher section, find a snap, go to the "builds" tab.

Try adding a repo to build from, and try triggering a build. Check everything works as expected.


Lukewh:
"SnapBuilder" is a bad name, the module does more than just build snaps.

nottrobin:
Like what?

Lukewh:
It:
 - triggers snap builds
 - gets lists of builds for a snap
 - will get an individual build by id

Something like `SnapBuild` or `SnapBuilds` would be more accurate imo. I think thr `er` suffix makes it too specific.

nottrobin:
Right fair enough.

I might just call it "Launchpad". I was trying to justify it existing alongside launchpadlib by pretending it had more specific purpose. But really it's just a collection of Launchpad API functions that we happen to find useful, and we find launchpadlib too confusing to use.

nottrobin:
@Lukewh might this be acceptable to you now?

jkfran:
@nottrobin Thank you!

PR #2564, opened 2020-02-24
Remove "Go to listing" button

### LINKED ISSUE #2533: "Go to listing" button seems to do nothing
http://0.0.0.0:8004/first-snap/python/linux/push

image

solazio:
## Done

- Update `Go to listing` button link depending if the snap is published or not
- Replace `is--disabled` class with `is-disabled` (it seems it was missed when [we upgraded to vanilla v2.0.1](https://github.com/canonical-web-and-design/vanilla-framework/pull/2320))

## Issue / Card

Fixes #2533 

## QA

- Pull the branch
- View the site locally in your web browser at: http://0.0.0.0:8004//first-snap/python/linux/push
1. Click `Register snap name`
2. See the 'Go to listing` button is enabled. Click it
3. See you are taken to the listing page. Note the message at the top
4. Go back to http://0.0.0.0:8004//first-snap/python/linux/push and click `Register snap name`
5. Push app to the store - `snapcraft push --release edge *.snap`
6. Click `Continue` or `Go to listing` button
7. See you are taken to the listing page. Note the message at the top is different this time

### DIFF
--- a/static/js/public/first-snap-flow.js
+++ b/static/js/public/first-snap-flow.js
@@ -22,7 +22,7 @@ function install(language) {
     if (!document.querySelector(".js-linux-manual")) {
       const paginationBtn = document.querySelector(`#js-pagination-next`);
       if (paginationBtn) {
-        paginationBtn.classList.remove("is--disabled");
+        paginationBtn.classList.remove("is-disabled");
         paginationBtn.href = `/first-snap/${language}/${selectedOs}/package`;
       }
     }
@@ -86,7 +86,7 @@ function install(language) {
 
     const paginationBtn = document.querySelector(`#js-pagination-next`);
     if (paginationBtn) {
-      paginationBtn.classList.remove("is--disabled");
+      paginationBtn.classList.remove("is-disabled");
       paginationBtn.href = `/first-snap/${language}/${type}/package`;
     }
   }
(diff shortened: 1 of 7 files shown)


Lukewh:
Why remove the button? It just wasn't getting updated with the snap name in the url after the user has registered the name.

solazio:
@Lukewh it was doing the same thing the other button is doing and after discussing with @danielmutis he said there is no point to have 2 buttons doing the same thing.

solazio:
@Lukewh if I enable the `Go to listing` button without pushing to the store, I'm taken to the page below, which suggests that my snap was actually pushed to the store (which is not true, because I cannot see it in my snap list). Considering this behaviour probably we should remove the `Go to listing` button?

Lukewh:
> @Lukewh if I enable the `Go to listing` button without pushing to the store, I'm taken to the page below, which suggests that my snap was actually pushed to the store (which is not true, because I cannot see it in my snap list). Considering this behaviour probably we should remove the `Go to listing` button?

Or we should allow people to edit the listing information for their snap before they've pushed a revision. The second table on the /snaps page is not very useful as it currently is... We should maybe enable clicking on the name to go to the listings page?

solazio:
@Lukewh I think this is a good idea. Will enable clicks on registered snaps.

static/js/public/first-snap-flow.js
```diff
@@ -168,19 +168,20 @@ function push() {
   }
 
   getCount(snapName => {
+    // Enable "Continue" button
     const continueBtn = document.querySelector(".js-continue");
     if (continueBtn) {
       continueBtn.href = `/${snapName}/listing?from=first-snap`;
```
Lukewh: I should have written in the issue, not just said at standup, could this link to the release UI?
```suggestion
      continueBtn.href = `/${snapName}/releases?from=first-snap`;
```

solazio: Ahhh...sorry. Will update.

PR #2565, opened 2020-02-24
Use flask-base

jkfran:
## Done

Changes to use flask-base

## QA

- Pull the branch


webapp/app.py
```diff
@@ -4,7 +4,7 @@
 The web frontend for the snap store.
 """
 
-import flask
+from canonicalwebteam.flask_base.app import FlaskBase
 from werkzeug.contrib.fixers import ProxyFix
```
nottrobin: This should be removed - it's included in flask-base:

https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/blob/master/canonicalwebteam/flask_base/app.py#L58

webapp/app.py
```diff
@@ -4,7 +4,7 @@
...
 from werkzeug.debug import DebuggedApplication
```
nottrobin: This too: https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/blob/master/canonicalwebteam/flask_base/app.py#L56

requirements.txt
```diff
@@ -7,7 +7,7 @@ canonicalwebteam.yaml-responses==1.1.1
 canonicalwebteam.image-template==1.0.0
```
nottrobin: yaml-responses on the previous line should also be removed as it's included in flask-base:

https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/blob/master/canonicalwebteam/flask_base/app.py#L62

jkfran: Done!

webapp/app.py
```diff
@@ -4,19 +4,13 @@
 The web frontend for the snap store.
 """
 
-import flask
-from werkzeug.contrib.fixers import ProxyFix
-from werkzeug.debug import DebuggedApplication
+from canonicalwebteam.flask_base.app import FlaskBase
 
 import talisker.flask
```
nottrobin: Sorry I didn't catch this in the previous review round, but these 2 lines should also be removed as they're in flask-base:

https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/blob/master/canonicalwebteam/flask_base/app.py#L85

nottrobin:
You should also make use of `favicon_url` to replace `webapp/snapcraft/views.py` line 247

webapp/app.py
```diff
@@ -33,17 +27,16 @@
...
-
     app.config.from_object("webapp.config")
     app.testing = testing
 
-    app.wsgi_app = ProxyFix(app.wsgi_app)
-    if app.debug:
-        app.wsgi_app = DebuggedApplication(app.wsgi_app)
-
     app.url_map.strict_slashes = False
     app.url_map.converters["regex"] = helpers.RegexConverter
```
nottrobin: These 2 lines are also included in flask-base:

https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/blob/master/canonicalwebteam/flask_base/app.py#L53

webapp/app.py
```diff
@@ -56,10 +49,6 @@ def create_app(testing=False):
         talisker.logs.set_global_extra({"service": "snapcraft.io"})
```
nottrobin: The two lines above here, where we configure two separate sessions with talisker.... this should be done differently but I guess this is out of scope of this PR.

Ideally we would have a single session that we create here with `talisker.requests.get_session()` and then we pass through to these API libraries.

But this is out of scope from this PR, so we should probably create an issue about it.

nottrobin:
Oh you should also remove the `robots.txt` and `humans.txt` logic from snapcraft.io in favour of what's in flask-base:

https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/blob/master/canonicalwebteam/flask_base/app.py#L115

It looks like snapcraft sends an empty response for `/robots.txt` - I think a more straightforward solution is to use flask-base's default behaviour and create an empty file at `/static/robots.txt`.

webapp/app.py
jkfran: I get an error related with this, Werkzeug is [used in Flask-base here](https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/blob/master/canonicalwebteam/flask_base/converters.py#L1) but it is not a dependency in the library:
```
  File "/home/fran/Workspace/snapcraft.io/webapp/app.py", line 9, in <module>
    from canonicalwebteam.flask_base.app import FlaskBase
  File "/usr/local/lib/python3.6/dist-packages/canonicalwebteam/flask_base/app.py", line 7, in <module>
    from werkzeug.middleware.proxy_fix import ProxyFix
ModuleNotFoundError: No module named 'werkzeug.middleware'
```

jkfran: Solved!

jkfran:
@nottrobin Can you review this again with all these changes? Also, for the `robots.txt` and `humans.txt` I got a 404, these files are expected to be in the root folder:
https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/blob/master/canonicalwebteam/flask_base/app.py#L112-L113

I moved both files to the root folder if that is fine.

PR #2566, opened 2020-02-25
Fix marketo response index error

### LINKED ISSUE #2276: IndexError: list index out of range
https://sentry.is.canonical.com/canonical/snapcraft-io/issues/2776/?referrer=github_plugin

```
IndexError: list index out of range
  File "webapp/publisher/views.py", line 90, in get_account_details
    marketo_user = marketo.get_user(flask_user["email"])
  File "webapp/api/marketo.py", line 105, in get_user
    return response["result"][0]

IndexError: list index out of range
```

solazio:
## Done

- Fix marketo response index error

## Issue / Card

Fixes #2276 
Error on sentry: https://sentry.is.canonical.com/canonical/snapcraft-io/issues/2776/

## QA

- Pull the branch
- View the site locally in your web browser at: http://0.0.0.0:8004/account/details
- I don't know how to QA this.

### DIFF
--- a/webapp/api/marketo.py
+++ b/webapp/api/marketo.py
@@ -84,7 +84,11 @@ class MarketoApi:
         response = self.request(
             method="GET", url=LEAD_BY_EMAIL, url_args=dict(email=email)
         )
-        return response["result"][0]
+        if "result" in response and len(response["result"]) > 0:
+            return response["result"][0]
+
+        else:
+            return None
 
     def get_newsletter_subscription(self, lead_id):
         response = self.request(
--- a/webapp/publisher/views.py
+++ b/webapp/publisher/views.py
@@ -87,13 +87,14 @@ def get_account_details():
     # if anything fails, just continue and don't show
     # this section
     try:
-        marketo_user = marketo.get_user(flask_user["email"])
-        marketo_subscribed = marketo.get_newsletter_subscription(
-            marketo_user["id"]
-        )
         subscribed_to_newsletter = False
-        if marketo_subscribed.get("snapcraftnewsletter"):
-            subscribed_to_newsletter = True
+        marketo_user = marketo.get_user(flask_user["email"])
+        if marketo_user:
+            marketo_subscribed = marketo.get_newsletter_subscription(
+                marketo_user["id"]
+            )
+            if marketo_subscribed.get("snapcraftnewsletter"):
+                subscribed_to_newsletter = True
 
         subscriptions = {"newsletter": subscribed_to_newsletter}
     except Exception:


webapp/api/marketo.py
```diff
@@ -84,7 +84,7 @@ def get_user(self, email):
         response = self.request(
             method="GET", url=LEAD_BY_EMAIL, url_args=dict(email=email)
         )
-        return response["result"][0]
+        return response["result"][0] if len(response["result"]) > 0 else {}
```
Lukewh: While we're here we should probably check response is there too:
`return response["result"][0] if "result" in response and len(response["result"]) > 0 else {}`?

jkfran: @solazio why an empty dict instead of `None`?

jkfran:
@solazio If you are return an empty dict then is going to create a similar exception here:
https://github.com/canonical-web-and-design/snapcraft.io/blob/master/webapp/publisher/views.py#L92

right?

solazio:
> @solazio If you are return an empty dict then is going to create a similar exception here:
> https://github.com/canonical-web-and-design/snapcraft.io/blob/master/webapp/publisher/views.py#L92
> 
> right?

@jkfran that is 100% correct. That's why I changed that bit as well, but for an unknown reason I didn't stage those changes. Will update in a sec.

webapp/api/marketo.py
solazio: @Lukewh yes, it makes sense to check if it exists as well. thanks

webapp/api/marketo.py
```diff
@@ -84,7 +84,11 @@ def get_user(self, email):
         response = self.request(
             method="GET", url=LEAD_BY_EMAIL, url_args=dict(email=email)
         )
-        return response["result"][0]
+        return (
+            response["result"][0]
+            if "result" in response and len(response["result"]) > 0
+            else {}
```
Lukewh: As @jkfran mentioned I think this can be None rather than an empty dict.

Might also be good to update the same kind of return on ~ line 99?

webapp/publisher/views.py
```diff
@@ -87,13 +87,14 @@ def get_account_details():
...
-        )
         subscribed_to_newsletter = False
-        if marketo_subscribed.get("snapcraftnewsletter"):
-            subscribed_to_newsletter = True
+        marketo_user = marketo.get_user(flask_user["email"])
+        if marketo_user:
+            marketo_subscribed = marketo.get_newsletter_subscription(
+                marketo_user["id"]
+            )
+            if marketo_subscribed.get("snapcraftnewsletter"):
```
Lukewh: though if you make the change above for the get_newsletter_subscription method, this will also need updating to check that marketo_subscribed exists AND snapcraftnewsletter is in the dict
```
if marketo_subscribed and marketo_subscribed.get("snapcraftnewlsetter"):
```
wdyt @jkfran ?

webapp/api/marketo.py
jkfran: Yeah, IMO it should return `None`, it makes it clear that there is no user. I also prefer the basic if in this case just because the condition is too long. Like:
```python
if "result" in response and len(response["result"]):
    return response["result"][0]
else:
    return None
```

webapp/publisher/views.py
jkfran: Well, I have seen that `get_newsletter_subscription` return always a dict. So that means this condition will always work. It could be like that.

solazio: In that case I leave this condition as it is.

jkfran: But you now updated the method `get_newsletter_subscription` to return `None` so now this condition could raise an exception because you can't use `get()` with None. If you want to return `None` then this needs to be like Luke mention before.

solazio: So be it.

solazio: Unless @Lukewh or you (@jkfran ) have a strong opinion about returning an empty dict on https://github.com/canonical-web-and-design/snapcraft.io/blob/master/webapp/api/marketo.py#L95 I will go with that.

jkfran: It is fine to me

PR #2642, opened 2020-03-31
Revision status

### LINKED ISSUE #2031: "Snap is not ready to be published" with no indication why
### Expected behaviour
When promoting a new build to a channel, if the snap is not ready to be published, there should be some indication _why_ this is the case and how to fix it.

I have a new snap, `magnus`, built with build.snapcraft.io from a github repository. I have completed the listing page, and the build has also completed; when I try to promote this new build to edge (which is the first promotion; the snap has not yet been published), the website shows the error:

Error:Error while releasing 1.0.1-ef41eb2 (4) to latest/edge. Snap is not ready to be published.

As far as I can tell, there is no indication anywhere about _why_ the snap is not ready to be published or how I might rectify that.

### Specs

- _URL:_ https://snapcraft.io/magnus/releases
- _Operating system:_ Ubuntu 18.04
- _Browser:_ Firefox

image

Lukewh:
## Done

- If a revision can't be released, add the reason WHY to the tooltip
- Remove the ability to drag revisions that can't be released (and the drag handle)
- In the unreleased section disable the checkbox and don't allow revisions to be dragged.

## Issue / Card

Fixes #2031 

## QA

- Pull the branch
- Visit http://0.0.0.0:8004/snap_name/releases where snap_name is the name of a snap with a blocked revision (I can share one with you if required)
- See the tooltip for the revision
- Try and stage that revision to be promoted (you shouldn't be able to)

### DIFF
--- a/static/js/publisher/release/components/releasesTable/cellViews.js
+++ b/static/js/publisher/release/components/releasesTable/cellViews.js
@@ -3,7 +3,11 @@ import PropTypes from "prop-types";
 
 import RevisionLabel from "../revisionLabel";
 
-import { isInDevmode, isRevisionBuiltOnLauchpad } from "../../helpers";
+import {
+  isInDevmode,
+  isRevisionBuiltOnLauchpad,
+  canBeReleased
+} from "../../helpers";
 import { useDragging, Handle } from "../dnd";
 
 // content of a cell when channel is closed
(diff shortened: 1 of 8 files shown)


solazio:
@Lukewh if you promote the whole row, it promotes that revision as well, even though is not released once you click save. Perhaps we should not allow it to be be promoted. Also maybe we should make it grey instead of yellow... but I'm not the best person to recommend a colour change :).

PR #2695, opened 2020-04-27
Disable categories dropdown when API call fail

### LINKED ISSUE #2693: "Category" on Listing page contains "None" as only entry
### Expected behaviour

Either I get to see a list of categories or, if for some technical reason I can't see that list (perhaps I haven't pushed a snap file, or I have pushed a snap file but it's private or unlisted in the store, or something went wrong with the API that fetches the category list) then the category dropdown should be disabled, with a message stating that I can't set a category at the moment but I can (and should) later. I can't think of any reason why I should ever be presented with a Category select element which has one entry, "None", and nothing else; that's never going to be correct.

Create a new snap in the snap store.

### Specs

- _URL:_ https://snapcraft.io/groovy-gorillas/listing
- _Operating system:_ Ubuntu 18.04
- _Browser:_ Firefox

jkfran:
## Done

Disable categories dropdown when API call fail

## Issue / Card

Fixes #2693 

## QA

- Pull the branch
- Use dotrun snap to start the site.
- Edit `.venv/lib/python3.6/site-packages/canonicalwebteam/store_api/store.py`.
- Add `raise StoreApiResponseError("error", 666)` in line 181
- Go to /<snap_name>/listing/ the dropdown should be disabled and you should see a temporary error message.

## Screenshots

Screenshot from 2020-04-27 11-30-25

### DIFF
--- a/templates/publisher/listing.html
+++ b/templates/publisher/listing.html
@@ -149,6 +149,7 @@ Listing details for {% if display_title %}{{ display_title }}{% else %}{{ snap_t
             <div class="col-8 col-start-large-3">
               <p class="p-form-help-text">
                 {% if not categories %}
+                <i class="p-icon--warning"></i>
                 There has been a problem obtaining the available categories, please try again later.
                 {% else %}
                 Include a category and your snap will be better ranked within the Snap Store.
--- a/templates/publisher/listing.html
+++ b/templates/publisher/listing.html
@@ -129,7 +129,7 @@ Listing details for {% if display_title %}{{ display_title }}{% else %}{{ snap_t
               {% if snap_categories.categories[0] %}
               value="{{ snap_categories.categories[0] }}"
               {% endif %}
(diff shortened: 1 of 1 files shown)


Lukewh:
Will QA in a bit, when I've got a few minutes

solazio:
my only suggestion is to add a warning icon like this:
image

having the icon, it is obvious to the user that something is wrong there?

templates/publisher/listing.html
```diff
@@ -148,8 +148,12 @@ <h2 class="p-heading--four">
           <div class="js-categories-category1-help-text row {% if snap_categories.categories[0] %}u-hide{% endif %}">
             <div class="col-8 col-start-large-3">
               <p class="p-form-help-text">
+                {% if not categories %}
+                There has been a problem obtaining the available categories, please try again later.
```
solazio, line 152:
```suggestion
                <i class="p-icon--warning"></i>
                There has been a problem obtaining the available categories, please try again later.
```

PR #2701, opened 2020-04-28
Fix GitHub rate limit issue

### LINKED ISSUE #2700: GitHub issue: Rate limit exceeded
We are getting some errors on Sentry about this because one of the GitHub API calls is made without authentication and it is exceeding the GitHub rate limit.

jkfran:
## Done

Fix GitHub rate limit issue

## Issue / Card

Fixes #2700

## QA

- Pull the branch
- Logout if you are logged and log in again.
- As a publisher go to any snap and click in the "Builds" tab, it should be working as expected.

### DIFF
--- a/webapp/publisher/snaps/build_views.py
+++ b/webapp/publisher/snaps/build_views.py
@@ -89,10 +89,17 @@ def get_snap_builds(snap_name):
     }
 
     # Get built snap in launchpad with this store name
-    github = GitHub(flask.session.get("github_auth_secret"))
     lp_snap = launchpad.get_snap_by_store_name(details["snap_name"])
 
     if lp_snap:
+        # In this case we can use the GitHub user account or
+        # the Snapcraft GitHub user to check the snapcraft.yaml
+        github = GitHub(
+            flask.session.get(
+                "github_auth_secret", GITHUB_SNAPCRAFT_USER_TOKEN
+            )
+        )
+
         # Git repository without GitHub hostname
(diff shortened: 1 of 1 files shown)


webapp/publisher/snaps/build_views.py
```diff
@@ -89,10 +89,17 @@ def get_snap_builds(snap_name):
...
     # Get built snap in launchpad with this store name
-    github = GitHub(flask.session.get("github_auth_secret"))
     lp_snap = launchpad.get_snap_by_store_name(details["snap_name"])
 
     if lp_snap:
+        # In this case we can use the GitHub user account or
+        # the Snapcraft GitHub user to check the snapcraft.yaml
+        github = GitHub(
+            flask.session.get(
+                "github_auth_secret", GITHUB_SNAPCRAFT_USER_TOKEN
```
Lukewh, line 99: Is there any reason that the GitHub user account wouldn't be available?

jkfran: Yes, the user could be logged out from GitHub because of multiple reasons, like being in a different device or if they lost their Snapcraft session for any reason. This is for an API call that we do even when the user is not logged in GitHub to show a message if the snapcraft.yaml is not present in the repo, obviously, this is for snaps already linked to a repo.

We were doing more requests than the rate limit for an anonymous user and this change is to use the user session on GitHub or our Snapcraft GitHub user instead of an anonymous call.

PR #2747, opened 2020-05-14
Changed build endpoint to return a JSON

### LINKED ISSUE #2745: Update Trigger Build endpoint to JSON
If request is successful - return `{"success": true}`
If request is failure - return server error (catch it with sentry)

jkfran:
## Done

Changed build endpoint to return a JSON object

## Issue / Card

Fixes #2745

## QA

- Pull the branch
- As a publisher go to a Snap using the builds feature
- Trigger a new build
- The reponse should be a json object in your browser

### DIFF
--- a/webapp/publisher/snaps/build_views.py
+++ b/webapp/publisher/snaps/build_views.py
@@ -381,35 +381,33 @@ def post_snap_builds(snap_name):
 
 @login_required
 def post_build(snap_name):
-    try:
-        details = api.get_snap_info(snap_name, flask.session)
-    except ApiResponseErrorList as api_response_error_list:
-        if api_response_error_list.status_code == 404:
-            return flask.abort(404, "No snap named {}".format(snap_name))
-        else:
-            return _handle_error_list(api_response_error_list.errors)
-    except ApiError as api_error:
-        return _handle_error(api_error)
+    # Don't allow builds from no contributors
+    account_snaps = api.get_account_snaps(flask.session)
+
+    if snap_name not in account_snaps:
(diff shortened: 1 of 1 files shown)


webapp/publisher/snaps/build_views.py
```diff
@@ -381,15 +381,13 @@ def post_snap_builds(snap_name):
...
-    except ApiResponseErrorList as api_response_error_list:
-        if api_response_error_list.status_code == 404:
-            return flask.abort(404, "No snap named {}".format(snap_name))
-        else:
-            return _handle_error_list(api_response_error_list.errors)
-    except ApiError as api_error:
-        return _handle_error(api_error)
+    details = api.get_snap_info(snap_name, flask.session)
+
+    # Don't allow builds from Admins that are no contributors
```
jkfran, line 386: @Lukewh I added this, I think we were missing it. Does make sense to check this?

webapp/publisher/snaps/build_views.py
```diff
@@ -381,15 +381,13 @@ def post_snap_builds(snap_name):
...
+    account_snaps = api.get_account_snaps(flask.session)
+
+    if snap_name not in account_snaps:
+        return flask.jsonify({"success": False})
```
Lukewh, line 390: As we know this error here, we should add an:
```
{
  "error": {
    "type": some_type,
    "message": some_message
  }
}
```
object to this response!

jkfran: Okay, I updated it

PR #2750, opened 2020-05-15
2746 create trigger new build component

### LINKED ISSUE #2746: Create new Trigger new build component
Include it in the "Builds" component.

- [ ] Onclick - POST to the trigger new build endpoint
- [ ] Onclick - Update state to "[loader] Requesting"
- [ ] When the first built result comes back from the builds component update to "Trigger new build"
- [ ] On error - display a generic "Something went wrong, please try again" notification above the table

solazio:
## Done

- `Trigger new build` on the frontend

## Issue / Card

Fixes #2746

## QA

- Pull the branch
- Click the `Trigger new build` button and see the button is disable with a spinner until the table is reloaded

### DIFF
--- a/static/js/publisher/builds/helpers.js
+++ b/static/js/publisher/builds/helpers.js
@@ -9,6 +9,10 @@ const FAILED_TO_BUILD = "failed_to_build";
 const CANCELLED = "cancelled";
 const UNKNOWN = "unknown";
 
+const ERROR = "ERROR";
+const SUCCESS = "SUCCESS";
+const IDLE = "IDLE";
+
 export const UserFacingStatus = {
   // Used only when there is no build returned from LP.
   // When build is returned from LP (scheduled) it's 'Building soon' for BSI.
@@ -80,3 +84,9 @@ export function createDuration(duration) {
   }
   return "";
 }
+
+export const TriggerBuildStatus = {
+  ERROR,
+  SUCCESS,
+  IDLE
+};
--- a/static/js/publisher/builds/index.js
+++ b/static/js/publisher/builds/index.js
@@ -26,15 +26,8 @@ class Builds extends React.Component {
       isLoading: false,
       fetchSize: 15,
       fetchStart: 0,
-      isTooSlow: false,
       builds: props.builds ? props.builds : []
     };
-
-    this.initTimer = setTimeout(() => {
-      this.setState({
-        isTooSlow: true
-      });
-    }, 35000);
 
     this.showMoreHandler = this.showMoreHandler.bind(this);
     this.triggerBuildHandler = this.triggerBuildHandler.bind(this);
@@ -84,7 +77,6 @@ class Builds extends React.Component {
           triggerBuildStatus:
             triggerBuildStatus === SUCCESS ? IDLE : triggerBuildStatus,
           isLoading: false,
-          isTooSlow: builds.length === 0 && result.snap_builds.length === 0,
           builds: fromStart
             ? result.snap_builds
             : builds.slice().concat(result.snap_builds)
@@ -150,7 +142,6 @@ class Builds extends React.Component {
     const {
       builds,
       isLoading,
-      isTooSlow,
       triggerBuildStatus,
       triggerBuildLoading
     } = this.state;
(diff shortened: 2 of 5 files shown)


static/js/publisher/builds/index.js
```diff
@@ -84,6 +101,30 @@ class Builds extends React.Component {
...
+      }
+    })
+      .then(res => res.json())
+      .then(result => {
+        if (result.success) {
+          this.setState({ triggerBuildSucccess: true });
+        }
+      })
+      .catch(() => {
+        this.setState({ triggerBuildSucccess: "ERROR" });
```
Lukewh, line 124: we should define some consts for different states
```
const ERROR = "ERROR";
const SUCCESS = "SUCCESS";
const IDLE = "IDLE"
```

And use those, rather than a mix of boolean, string and null types.

static/js/publisher/builds/index.js
```diff
@@ -84,6 +101,30 @@ class Builds extends React.Component {
...
+      headers: {
+        "Content-Type": "application/json",
+        "X-CSRFToken": csrf_token
+      }
+    })
+      .then(res => res.json())
+      .then(result => {
+        if (result.success) {
+          this.setState({ triggerBuildSucccess: true });
+        }
```
Lukewh, line 121: There's also the `else` of `FORBIDDEN` that the API could return here.

static/js/publisher/builds/helpers.js
```diff
@@ -80,3 +84,9 @@ export function createDuration(duration) {
   }
   return "";
 }
+
+export const TriggerBuildStatus = {
```
Lukewh, line 88:
```suggestion
export const TriggerBuildStatus = {
  ERROR,
  SUCCESS,
  IDLE
};
```

static/js/publisher/builds/index.js
```diff
@@ -62,13 +72,25 @@ class Builds extends React.Component {
...
       .then(res => res.json())
       .then(result => {
-        this.setState({
-          isLoading: false,
-          isTooSlow: builds.length === 0 && result.snap_builds.length === 0,
-          builds: fromStart
-            ? result.snap_builds
-            : builds.slice().concat(result.snap_builds)
-        });
+        if (triggerBuildSucccess === SUCCESS) {
```
Lukewh, line 75: sorry I meant you don't need the if/else if you just make `triggerBuildLoading` and `triggerBuildSuccess` ternary operators. That way you don't need to repeat the `isLoading`, `isTooSlow` and `builds` keys - making it more complicated to update in the future.

static/js/publisher/builds/index.js
```diff
@@ -15,6 +21,8 @@ class Builds extends React.Component {
     this.fetchTimer = null;
 
     this.state = {
+      triggerBuildLoading: false,
+      triggerBuildSucccess: false,
```
Lukewh, line 25: as this can be `ERROR`, `SUCCESS` and `IDLE` now the initial state should probably be `IDLE` and the name of the variable would be better as `triggerBuildStatus`.

static/js/publisher/builds/index.js
solazio: ahhh, ok, sorry

static/js/publisher/builds/index.js
```diff
@@ -63,6 +73,10 @@ class Builds extends React.Component {
       .then(res => res.json())
       .then(result => {
         this.setState({
+          triggerBuildLoading:
+            triggerBuildStatus === SUCCESS ? !SUCCESS : triggerBuildStatus,
+          triggerBuildStatus:
+            triggerBuildStatus === SUCCESS ? IDLE : triggerBuildStatus,
```
Lukewh, line 79: I think the logic for these two needs clarifying - I get console errors. The first return of /builds.json marks it as done. On next fetch of /builds.json I get:
```
Warning: Failed prop type: Invalid prop `isLoading` of type `string` supplied to `TriggerBuild`, expected `boolean`.
    in TriggerBuild (created by Builds)
    in Builds react_devtools_backend.js:6:5119
```

solazio: was supposed to use `triggerBuildLoading` instead on line 77

PR #2765, opened 2020-05-19
Use Vanilla layout and side navigation in docs

bartaz:
## Done

- Updates docs templates to use Vanilla grid layout and side navigation pattern
- Removes unnecessary docs layout styles
- Updates docs search to use Vanilla search box pattern

## QA

- Pull the branch
- Check [updated docs](https://snapcraft-io-canonical-web-and-design-pr-2765.run.demo.haus/docs/)
- Make sure content looks good on different pages (try different type of content, with images, tables, etc)
- Make sure side navigation looks and works as expected
  - active item should be marked
  - side nav should be sticky and scroll correctly
  - on small screens it should hide and work as a drawer with a toggle
- Make sure everything works on different screen sizes
- Make sure search works as expected

Screenshot 2020-05-19 at 13 07 00

### DIFF
--- a/static/sass/_snapcraft_p-docs.scss
+++ b/static/sass/_snapcraft_p-docs.scss
@@ -1,86 +1,8 @@
 @mixin snapcraft-pattern-docs {
   $banner-background: #106363;
 
-  .docs-container {
-    display: flex;
-    margin: 0 auto;
-
-    // override for full width text
-    // https://github.com/canonical-web-and-design/vanilla-framework/issues/1794
-    p {
-      max-width: 100%;
-    }
-
-    .lightbox-wrapper {
-      .meta {
-        display: none;
-      }
-    }
-
-    // Discourse adds "emoji" <img>s, which are huge
-    // They need to be constrained
-    .emoji {
-      height: 1rem;
-    }
-
-    .p-sidebar {
-      border-right: .0625rem solid $color-mid-light;
-      flex: 0 0 18rem;
-    }
-
-    .p-sidebar__content {
-      padding: 1.5rem;
-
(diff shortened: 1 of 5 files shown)


bartaz:
Just noticed there is an issue with global nav (on small screens):

Screenshot 2020-05-19 at 13 46 59

First of all - @Lukewh  seems like snapcraft is using old version of global nav. It now always appears on top, doesn't switch to bottom on small screens.

But most importantly, there is an issue with global nav vs side nav z-index that we haven't seen in Vanilla (because it doesn't have global nav).
@anthonydillon @solazio dqlite docs just hack around it adding top margin to side nav on mobile: https://github.com/canonical-web-and-design/dqlite.io/blob/master/static/sass/styles.scss#L15-L19

We need to find better solution for this.
Is global nav always meant to be on top? Even with side nav open? Should side nav cover the global nav?

lyubomir-popov:
@bartaz afaik we've never managed z-index in vanilla. Last week we had a similar issue with a tabstrip that comes in contact with a contextual menu. I found an article on the subject: https://www.smashingmagazine.com/2019/04/z-index-component-based-web-application/
1:23 PM What they propose seems quite straightforward to implement.

Lukewh:
Works as expected on the whole - the only "issue" that is probably an enhancement I would expect is that when you load a page towards the end of the sidenav, the sidenav should scroll to keep the users selection in view, rather than scrolling back to the top.

bartaz:
@Lukewh I updated the script to scroll to current item (based on what we use in Vanilla docs). So marking for code/qa again (but you can focus on last commit).

PR #2824, opened 2020-05-26
List Launchpad builds with status 'Needs building'

### LINKED ISSUE #2749: When builds table has "in progress" builds, show the launchpad build status
When a build is triggered and the table contains "In progress" builds we should make a call to the new build status endpoint.
It should use a separate timer to get the information (as we'll be moving the builds table away from polling to use webhooks at some point) and can update less frequently - 30s should be ok.

jkfran:
## Done

- List builds with status 'Needs building'.
- Include estimated time for 'Needs building' builds.
- Fix pagination.

## Issue / Card

Fixes #2749

## QA

- Pull the branch
- Go to any snap, trigger a build. Check that builds with the status "Needs building" on Launchpad are listed as "Building soon" in the table.

### DIFF
--- a/webapp/publisher/snaps/build_views.py
+++ b/webapp/publisher/snaps/build_views.py
@@ -55,7 +55,7 @@ def get_builds(lp_snap, selection):
             "queue_time": None,
         }
 
-        if status == "building_soon":
+        if build["buildstate"] == "Needs building":
             if not builders_status:
                 builders_status = launchpad.get_builders_status()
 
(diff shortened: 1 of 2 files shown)


webapp/publisher/snaps/build_views.py
```diff
@@ -247,7 +247,7 @@ def get_snap_builds_json(snap_name):
 
     start = flask.request.args.get("start", 0, type=int)
     size = flask.request.args.get("size", 15, type=int)
-    build_slice = slice(start, start + size)
+    build_slice = slice(start, size)
```
Lukewh, line 250: Is this not needed? The two cases from the frontend:
1) Clicking show more: start = 15, size = 15
2) Refreshing the status of all visible rows (after clicking show more 30+) which means: start = 0, size = 30+

jkfran: Hey @Lukewh , it doesn't match with my experience, this are the requests made in the frontend:

Show more: builds.json?start=15&size=30
Table refresh: builds.json?size=30

Lukewh: I just QA'd and it works fine - no sure what the `start + ` was for :shrug:

PR #2885, opened 2020-06-15
Enable captions by default

### LINKED ISSUE #2842: Enable youtube captions by default
### Expected behaviour
When a video starts playing I'd like the captions to be on to begin with.

E.g. go on https://snapcraft.io/maas - video starts playing, I need to manually enable the captions

solazio:
## Done

- Enable captions by default

## Issue / Card

Fixes #2842 
## QA

- Pull the branch

- There is no easy way to test this, unless you know a snap that has a video and the video has a subtitle/cc (not auto-generated). If you do know `that snap` go to http://0.0.0.0:8004/<that_snap_name> and see the video has the captions `on` by default
If you don't know any snap that has that, just trust me. Here is an example that [has subtitle](https://www.youtube.com/embed/up21mvokyQ4?cc_load_policy=1) and [one that doesn't have](https://www.youtube.com/embed/0FC94Tgz2ho?cc_load_policy=1). Both of these have the caption turned on by default but only the one with subtitle will actually show it.

**Info**
It turns out hat you can only see the [subtitles/cc](https://developers.google.com/youtube/player_parameters#cc_load_policy) by default if the video has one. If it doesn't have one, it will show you the auto-generated one if you click the button but it will not show it by default.

### DIFF
--- a/templates/partials/_video.html
+++ b/templates/partials/_video.html
@@ -1,6 +1,6 @@
 {% if video.type == "youtube" %}
   <iframe id="ytplayer" type="text/html" width="818" height="460"
-          src="{{ video.url }}?autoplay=1&mute=1&modestbranding=1&rel=0" frameborder="0"></iframe>
+          src="{{ video.url }}?autoplay=1&mute=1&cc_load_policy=1&modestbranding=1&rel=0" frameborder="0"></iframe>
 {% elif video.type == "vimeo" %}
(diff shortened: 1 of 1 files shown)


lilyvidenova:
Hmmm captions don't seem to be enabled for me

solazio:
> Hmmm captions don't seem to be enabled for me

You need to make sure the video you are checking has a subtitle, otherwise will not show it. See my note above ^

PR #2896, opened 2020-06-16
Show warning about unstable channels

### LINKED ISSUE #1247: Non-stable channels which are publicly exposed don't explain risk levels
I've seen a few complaints online about various snaps which are not production quality. On further inspection, the snaps in question are not even available in the stable channel. Normal users, who just want software, and who are not invested in our platform (which is a significant majority of snap users) do not appreciate the risks inherent with non-stable snaps. We do not make it clear that these snaps are not ready for prime-time. Let's take the use case of someone being told there's a copy of steam in the snap store and given this link https://snapcraft.io/steamforwindows (but it also works if they search as it's the first hit).

image

Here's the store page. Note I'm picking on a snap *we* (snapcrafters) owns, but this could happen with *any* edge channel snap.

image

I see no install button but a clue is the "All versions" button, which is where I might expect an Install button.

image

Still no install option, but if I somehow know to hover my mouse over it (which seems obtuse) then I see an "Install >" link. Click that and the UI changes.

image

The description is "latest/edge". In my persona of non-snap advocate and end user, I have *no* *clue* what that means, but "latest" sounds awesome, just what I need!

I have two options. One asks if I want to open the Desktop Store (which doesn't work on KDE Neon 18.04 with the snap-store snap installed, but I'll save that for another bug). 

The other option is to paste a command in a terminal, which I do, because I'm a user and I know no different. 

```
$ sudo snap install steamforwindows --edge
[sudo] password for alan: 
Download snap "steamforwindows" (11) from channel "edge"                            16% 3.75MB/s 30.3s
```

I now have an edge channel, bleeding edge, probably broken snap installed, with nothing explained to me along the way about these risks.

We should absolutely not be pointing normal end users at the edge channel with no details about what their expectations are.

Lukewh:
## Done

- Added a couple of messages to the channel map dropdown, where appropriate.

## Issue / Card

Fixes #1247

## QA

- Pull the branch
- Visit http://0.0.0.0:8004/deno and see the channel map has warning
- Visit a few other snaps from /store or search and see what it looks like!

## Screenshots

image
image

### DIFF
--- a/static/js/public/snap-details/channelMap.js
+++ b/static/js/public/snap-details/channelMap.js
@@ -31,7 +31,7 @@ class ChannelMap {
     let numberTracks = [];
     let stringTracks = [];
     let latestTracks = [];
-    rows.forEach((row) => {
+    rows.forEach(row => {
       // numbers are defined by any string starting any of the following patterns:
       //   just a number – 1,2,3,4,
       //   numbers on the left in a pattern – 2018.3 , 1.1, 1.1.23 ...
@@ -46,7 +46,7 @@ class ChannelMap {
     });
 
     // Ignore case
-    stringTracks.sort(function (a, b) {
+    stringTracks.sort(function(a, b) {
       return a[0].toLowerCase().localeCompare(b[0].toLowerCase());
     });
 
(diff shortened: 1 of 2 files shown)


lilyvidenova:
Hmm looking at it, I think I was wrong @Lukewh can we remove the message from the candidate install modal and update the beta one to the following:

Snaps on the /beta channel may have unfinished parts, breaking changes may occur. 

I'd like to come up with something useful for candidate too but I'll need a bit more time for it, so no message will be better than a pretty wrong message.

PR #2913, opened 2020-06-25
Remove logout redirect through BSI

### LINKED ISSUE #2906: Remove signout redirect to logout of BSI too
Currently when you logout of snapcraft we redirect to BSI logout -> REMOVE THIS.

Lukewh:
## Done

- Remove redirect to BSI on logout

## Issue / Card

Fixes #2906 

## QA

- Pull the branch
- Logout
- Be redirected to /

### DIFF
--- a/tests/publisher/tests_account_logout.py
+++ b/tests/publisher/tests_account_logout.py
@@ -17,29 +17,4 @@ class LogoutRedirects(BaseTestCases.BaseAppTesting):
 
         self.assertEqual(302, response.status_code)
 
-        self.assertEqual(
-            (
-                "https://login.ubuntu.com/+logout"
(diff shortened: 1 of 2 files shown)


tbille:
@Lukewh I think we shouuldn't logout anymore from SSO. This is not a pattern that we really have anymore. So you can update the tests to just redirect to `/`

Lukewh:
> @Lukewh I think we shouuldn't logout anymore from SSO. This is not a pattern that we really have anymore. So you can update the tests to just redirect to `/`

We had reports from advocacy and others that use multiple accounts that it was a pain that they'd click logout, go to the homepage, click login and be automatically logged in to the previous account, when what they really wanted was to log in with a different account. That's why we added the full SSO logout.

It also raises the question, if you don't logout of SSO is there any point in logging out at all?

PR #2941, opened 2020-07-10
Update docs search navigation

### LINKED ISSUE #2933: Improve pagination for docs search
image

https://snapcraft.io/docs/search?q=snap&start=21&siteSearch=snapcraft.io/docs

solazio:
## Done

- Update pagination for docs search

## Issue / Card

Fixes #2933

## QA

- Pull the branch
- View the site locally in your web browser at: http://0.0.0.0:8004/docs/search
- Search for a term and see the pagination buttons look good

## Screenshots

image

### DIFF
--- a/static/sass/_patterns_pagination.scss
+++ b/static/sass/_patterns_pagination.scss
@@ -17,4 +17,9 @@
       }
     }
   }
+
+  // 10.07.2020 - Ovi: This can be remove once this issue is closed
+  .p-pagination__link--previous {
+    margin-inline-end: .5rem;
+  }
 }
--- a/templates/docs/search.html
+++ b/templates/docs/search.html
@@ -33,23 +33,19 @@
     </div>
   </div>
   {% endfor %}
-  <div class="p-strip p-pagination">
-    {% if  results.queries and results.queries.previousPage %}
-    <a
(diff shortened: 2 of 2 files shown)


anthonydillon:
I would prefer this to use the Vanilla pattern used in all other instances of this search. It is much easier to update a pattern instance when its similarly used in all cases. Clearly the index pattern would work better here but until we can achieve that we should use the other pattern.

tbille:
@anthonydillon can I suggest we land that PR since at the moment it's an issue that we have on the live site. It seems that having the index pattern implies work also in the `canonicalwebteam.search` module ([link](https://developers.google.com/custom-search/v1/reference/rest/v1/cse/list#response) to what @solazio pointed ).

Once fixed we can implement pattern on snacpraft.io.

nottrobin:
@tbille, @anthonydillon is off today. If this is the only way to fix this live issue quickly, go ahead and approve and merge - and add a high priority / critical issue to the board to address Ant's concern.

lilyvidenova:
Hm, I can't search for anything in the demo I'm getting a "canonicalwebteam.search.views.NoAPIKeyError" but I'm +1 based on the screenshot :)

PR #2953, opened 2020-07-14
Remove hardcode of default track for node

tbille:
## Done

**This pull request can be merged once Node has setup the default track to be the prefered track**

Now that default tracks are handled by the API we can enter in contact with node so they can set their default track to their latest LTS 12.

At the moment this was coded in the backend as an exception for node and is one LTS late.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/snapcraft.io/issues/2952

## QA

- Pull the branch
- Go on /node 
- **For now there is no default track so the track is the latest (15)**
- Check that the listing/metrics/releases pages are working

### DIFF
--- a/tests/tests_templates_utils.py
+++ b/tests/tests_templates_utils.py
@@ -74,14 +74,6 @@ class TemplateUtilsTest(unittest.TestCase):
         )
         self.assertTrue(result, "sudo snap install skype --classic")
 
-    def test_install_snippet_with_classic_and_default_track(self):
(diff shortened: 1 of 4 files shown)


jkfran:
@tbille The QA works fine, but I didn't change the QA label to do another one after the set the default track. :+1: ?

jkfran:
QA +1

Lukewh:
:raised_hands:

PR #2973, opened 2020-07-24
Upgrade to babel 7

### LINKED ISSUE #2930: Upgrade babel
As it turns out to be very complicated to do this in a renovate/all PR because are so many things updated, I decided to try to remove babel 6 and add babel 7 from scratch.

solazio:
## Done

- Upgrade to babel 7

## Issue / Card

Fixes #2930 

## QA

- Pull the branch
- Just click around and see everything works as expected

## Screenshots


static/js/public/first-snap-flow.test.js
```diff
@@ -6,6 +6,9 @@ describe("initChooseName", () => {
   let nameInput;
   let submitButton;
 
+  const _window = window;
+  window.location = _window.location;
```
jkfran, line 10: This looks really hacky, is this fine?

anthonydillon: @solazio what the reason for this? It seems it's just reassigning the value from itself?

solazio: the "good" way to do this would be `const { location } = window;`, but  babel-eslint module doesn't like that because further I I only use `window.location` and not `location` so it throws an error. To fix it I either hack it this way or I would probably need to find and install another module which under the hood will probably do the same thing.

jkfran: But I still don't understand the reason. if _windows = windows, then when you do `window.location = _window.location;` is exactly as it was before, right?

solazio: fixed it now. thanks

jkfran: @solazio Thank you for the explanation in the call!

PR #2984, opened 2020-08-03
Show queue time only for the 'Building Soon' builds

### LINKED ISSUE #2983: Queue time is added to the released builds
Go to `/<snap_name>/builds` and trigger a new build. After the page reloads, if you have any builds in the queue, when you hover over `Building Soon` should show the estimated time in the queue. If you hover over a `Released` build should say nothing, however it shows some queue time.

solazio:
## Done

- Show queue time only for the 'Building Soon' builds

## Issue / Card

Fixes #2983 

## QA

- Pull the branch
- Trigger a new build and hover over items in the `Result` column. You should see a `Queue time ...` tooltip only when you hover over a `Building soon` build and not older `Released` builds as currently happens

## Screenshots

### DIFF
--- a/static/js/publisher/builds/index.js
+++ b/static/js/publisher/builds/index.js
@@ -256,9 +256,10 @@ class Builds extends React.Component {
                 </Fragment>
               ),
               className: "has-icon",
-              title: queueTime[build.arch_tag]
-                ? `Queue time: up to ${queueTime[build.arch_tag]}`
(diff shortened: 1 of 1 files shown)


tbille:
I don't think this is the solution here.
The snap is already built. So it shouldn't be built again. The issue seems more a state issue, it seems that the built date is not being shared.

PR #3001, opened 2020-08-10
Update supported releases

popey:
Remove the EOL releases and add 20.04

## Done

Checked which releases are supported. Removed unsupported ones, and added in 20.04 LTS.

## Issue / Card

Reported at https://forum.snapcraft.io/t/outdated-snapcraft-info-at-every-ubuntu-package/19316

## QA

Did all this! dotrun is rather neat!

- Pull the branch

Made sure I didn't put a duff link in there, or break existing links.

## Screenshots

Screenshot from 2020-08-10 17-21-26

### DIFF
--- a/webapp/store/content/distros/ubuntu.yaml
+++ b/webapp/store/content/distros/ubuntu.yaml
@@ -6,7 +6,7 @@ logo-mono: https://assets.ubuntu.com/v1/20ba8b71-Distro_Logo_Ubuntu_White.svg
 install:
   -
     action: |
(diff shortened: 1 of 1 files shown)


diddledan:
While I understand we don't want to list out-of-support distros, do we really need to list every supported one in addition to "and later"? Surely the "and later" covers the "including" distros listed thereafter?

popey:
For normal people, I think, yes. Note these instructions need to apply to all people who install snaps. It's possible the version numbers are a confusing mess to some people. So I don't think it hurts to be super explicit about which versions are supported here.

diddledan:
Roger that :-) good thinking, Batman.

PR #3075, opened 2020-09-25
Fix 500 error after webhook creation

jkfran:
## Done

- The first call from the webhook doesn't contain the branch. This PR fix the error 

## QA

- Ping me for the QA process


tbille (approved):
Code looks good
QA'd together. Let's merge it.

PR #3093, opened 2020-10-01
Fetch builds only after the previous fetch finished

### LINKED ISSUE #2889: Snapcraft builds: Only check for new builds if the previous fetch has finished
We're currently triggering a check every x seconds, whether or not Launchpad has returned. We should only fetch again once the previous fetch has either succeeded or failed.

solazio:
## Done

- Fetch new builds only after the previous fetch finished

## Issue / Card

Fixes #2889 

## QA

- Pull the branch
- I don't know how you can QA this. But basically it should only trigger a new builds fetch 30s after the previous one has finished. Previously we were making requests every 30s, even if the last request was not finished

## Screenshots

### DIFF
--- a/static/js/publisher/builds/index.js
+++ b/static/js/publisher/builds/index.js
@@ -81,7 +81,7 @@ class Builds extends React.Component {
       triggerBuildLoading,
       shouldUpdateQueueTime,
     } = this.state;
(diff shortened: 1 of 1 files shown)


jkfran (approved):
Works as expected 👍🏻

PR #3095, opened 2020-10-01
HACKING.md - Snap automated builds

jkfran:
Update HACKING.md file to include instructions for the Snap automated builds feature.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -30,7 +30,7 @@ GITHUB_CLIENT_SECRET=<Your client secret>
 
 #### Use custom GitHub Webhook URL
 
-When you run the project locally, you probably want to link GitHub repos with your local environment; however, GitHub wouldn't be able to reach your localhost. You can use something like [ngrok](https://ngrok.com/) to proxy your machine and update the following variable in your `.env.local` file:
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -13,14 +13,52 @@ DASHBOARD_API_V2=https://dashboard.staging.snapcraft.io/api/v2/
...
+Once you've created the application, you should be given the client id and the client secret. Update your `.env.local` file with the your secrets:
+
+```bash
+GITHUB_CLIENT_ID=<Your client id>
+GITHUB_CLIENT_SECRET=<Your client secret>
+```
+
+#### Use custom GitHub Webhook URL
+
+When you run the project locally, you probably want to link GitHub repos with your local environment; however, GitHub wouldn't be able to reach your localhost. You can use something like [ngrok](https://ngrok.com/) to proxy your machine and update the following variable in your `.env.local` file:
```
tbille, line 33:
```suggestion
When you run the project locally, you probably want to link GitHub repos with your local environment; otherwise GitHub won't be able to reach your localhost. You can use a tool like [ngrok](https://ngrok.com/) to proxy to your machine and update the following variable in your `.env.local` file:
```

tbille (approved):
Small changes, but seems good to me :+1:

Issue #2777, opened 2020-05-21
Store pagination links are tiny and hard to spot


fitojb:
### Expected behaviour

Up to now, whenever I visited the Games category of the Store, I always found the same few results. It took me a while to realize I could load more of them by changing the page.

### Steps to reproduce the problem
0. Have failing eyesight or tired eyes.
1. Go to https://snapcraft.io/search?category=games
2. Scroll to the bottom.

![image](https://user-images.githubusercontent.com/554953/82551837-9212f580-9b26-11ea-964b-64cad516b69f.png)

For the benefit of discoverability by dumbasses like meself, please enclose these links in a box (like in `/first-snap` — cf. #1969), or add icons to them.

RCMainak:
I would like to working on issue.

fitojb:
Thanks @RCMainak !

PR #3096, opened 2020-10-01
Changed pagination links to buttons

RCMainak:
Fixed issue #2777. Changed pagination links to buttons.

### DIFF
--- a/templates/store/search.html
+++ b/templates/store/search.html
@@ -89,14 +89,19 @@
   {% if (searched_snaps or featured_snaps) and links %}
   <section class="p-strip is-shallow u-no-padding--top">
     <nav class="u-fixed-width u-clearfix">
-      <div class="u-float-left u-align--left">
-        {% if links.first %}<a href="{{ links.first }}">&#171;&nbsp;First</a>&nbsp;{% endif %}
-        {% if links.prev %}<a href="{{ links.prev }}">&#8249;&nbsp;Previous</a>{% endif %}
-      </div>
-      <div class="u-float-right u-align--right">
-        {% if links.next %}<a href="{{ links.next }}">Next&nbsp;&#8250;</a>{% endif %}
-        {% if links.last %}&nbsp;<a href="{{ links.last }}">Last&nbsp;&#187;</a>{% endif %}
-      </div>
+      <ol class="p-pagination">
+        {% if links.prev %}
+        <li class="p-pagination__item">
+          <a class="p-pagination__link--previous" href="{{ links.prev }}" title="Previous page"><i class="p-icon--contextual-menu"></i><span>Previous page</span></a>
+        </li>
+        {% endif %}
+
+        {% if links.next %}
+        <li class="p-pagination__item u-align-text--right">
+          <a class="p-pagination__link--next" href="{{ links.next }}" title="Next page"><span>Next page</span><i class="p-icon--contextual-menu"></i></a>
+        </li>
+        {% endif %}
+      </ol>
     </nav>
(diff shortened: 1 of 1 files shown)


lilyvidenova:
Is there a demo for this @solazio?

anthonydillon:
@lilyvidenova here you go: https://snapcraft-io-3096.demos.haus/search?q=Next&page=2

Although I think it looks better I would recommend using the Vanilla pattern for this.
https://vanillaframework.io/docs/patterns/pagination#previous--next-buttons-only

lilyvidenova:
I agree, could we please use the pagination component Ant linked above @RCMainak :)

RCMainak:
Sure. I will make the changes. Do you want to keep the first & last links or should I remove those keeping only previous & next?

lilyvidenova:
@RCMainak just like in the Vanilla example, please: 
image

Thanks for your contribution!

anthonydillon:
@lilyvidenova it is not trivial to add the index pagination and I would say out of scope for issue as a "Good first issue". I originally linked to the button only pattern:
image

I would say "Last" is not needed as it is unlikely the user wants to jump to the last page. @lilyvidenova do you agree?

lilyvidenova:
Yep, completely agree. Previous and next will suffice.

RCMainak:
I can try to implement the index pagination, hopefully I can do that by modifying [def store_view():](https://github.com/canonical-web-and-design/snapcraft.io/blob/master/webapp/store/views.py#L65) @anthonydillon

RCMainak:
@anthonydillon If we are going with previous & next pagination, I think the updated changes will be sufficient. All though I have a question, do you think I should add `p-pagination__item--right` class to Vanilla in a separate PR?

RCMainak:
Ops! The check failed because I removed first & last links in the [search_snap() method](https://github.com/canonical-web-and-design/snapcraft.io/blob/3597687798bb99262fc51a9a7ee0e42d2becba41/webapp/store/views.py#L130). Are those safe to remove or should I leave me as it was before?

If it's safe to remove the first & last links then I will update the tests.

static/sass/_patterns_pagination.scss
```diff
@@ -22,4 +22,9 @@
   .p-pagination__link--previous {
     margin-inline-end: .5rem;
   }
+
+
+  .p-pagination__item--right {
```
anthonydillon, line 27: Could you try and use the existing alignment utilities instead of adding a new pattern please. https://vanillaframework.io/docs/utilities/align

anthonydillon (changes requested):
Thank you for updating your PR. 

I have added a suggestion to use existing utilities for the alignment. From a quick look it appears `store.logic.py` also use `links.last`. I think it would be safer to leave it in the logic but only display next and prev in the template.

If you would like to follow up with the index based pagination that would be great. Let us know if you need any help.

static/sass/_patterns_pagination.scss
RCMainak: I was looking for something like this. Seems like I was looking at the wrong places. :sweat_smile:

RCMainak:
Thank you @anthonydillon for your support. I will try to implement the index pagination in a separate PR.

anthonydillon (approved):
Great thank you :+1:

RCMainak:
Hello, if you think I have made a positive impact you can help me by adding `hacktoberfest-accepted` label to this PR. Thanks.

anthonydillon:
Sure done, thank you

PR #3102, opened 2020-10-05
Increase API requests timeout

jkfran:


solazio (approved):
Looks good :+1:. Thanks

PR #3180, opened 2020-11-19
Add to sitemap: tutorials, docs, install pages

tbille:
## Done

Add missing pages to the sitemap

Add tutorials, docs and distro install pages.

## QA

- Pull the branch
- View the site locally in your web browser at: http://0.0.0.0:8004/sitemap.xml
- This page takes a while but is heavily cached
- In there there should be the docs, tutorials and install pages

### DIFF
--- a/templates/sitemap.xml
+++ b/templates/sitemap.xml
@@ -21,6 +21,15 @@
     <changefreq>weekly</changefreq>
     <priority>0.5</priority>
   </url>
+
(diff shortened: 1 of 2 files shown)


tbille:
Actually this is not a good solution.
We should make it smaller and replace the sitemap to make it an index and have a sitemap per section of the site (store, docs, tutorials, blog, main)

Closing the PR

PR #3228, opened 2020-11-26
Get brand store snaps

jkfran:
## Done

- Get list of snaps from a brand store

## QA

- You have to be an store administrator
- Snaps names should appear in `/admin`


steverydz (approved):
LGTM :+1:

PR #3257, opened 2020-12-04
Extract verified developer into partial

tbille:
## Done

Refactor duplicate code into partial

## QA

- Pull the branch
- http://0.0.0.0:8004/eks
- http://0.0.0.0:8004/store
- Test an embedded card 
- Navigate in the site to make sure that we don't have icons going crazy


steverydz (approved):
LGTM :+1:

PR #3299, opened 2020-12-16
Add no script message to publisher pages

### LINKED ISSUE #2850: Add messaging for people without JavaScript turned on
@nottrobin commented on [Fri Nov 30 2018](https://github.com/canonical-websites/snapcraft.io/issues/1380)

Currently, if I visit https://build.snapcraft.io/user/nottrobin with JavaScript disabled, I see this:

image

I actually have 11 repos set up in build.snacraft.io. I can click "Add repos", in which case I see this:

image

Both are very misleading. We should add a message to explain to people that they can't use the app with JavaScript turned off.

steverydz:
## Done
Added a message for users with JS disabled

## How to QA
- Disable JS in your browser
- Go to https://snapcraft-io-3299.demos.haus/snaps
- Check that there is a message about having JS disabled
- Click through to a snap page
- Check that there is a message about having JS disabled

## Issue / Card
Fixes #2850

### DIFF
--- a/templates/publisher/_header.html
+++ b/templates/publisher/_header.html
@@ -1,3 +1,5 @@
+{% include "publisher/_noscript.html" %}
+
 <section class="p-strip is-shallow u-no-padding--bottom">
   {% if from in ["first-snap", "first-snap-unpublished"] %}
   <div class="row">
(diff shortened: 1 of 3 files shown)


tbille (approved):
Works as expected

PR #3318, opened 2021-01-07
Add build status column

### LINKED ISSUE #3309: JS Fetch JSON endpoint to show status

steverydz:
## Done
Added a build status column to the snaps listing

## How to QA
- Go to https://snapcraft-io-3318.demos.haus/snaps
- Check that the table has a build status column

## Issue / Card
Fixes #3309, #3310

### DIFF
--- a/static/js/publisher/build-status.js
+++ b/static/js/publisher/build-status.js
@@ -1,0 +1,32 @@
+import { UserFacingStatus } from "./builds/helpers";
+
+function getStatuses() {
+  return fetch("/snap-builds.json").then((r) => r.json());
+}
+
+function addBuildStatus(row, data) {
+  const snapName = row.dataset.snapName;
+  const releaseData = data.find((d) => d.name === snapName);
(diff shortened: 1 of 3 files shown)


jkfran:
@steverydz I'm not getting the column value for my snaps. I can see the following error in the JS console:
```
e.querySelector(...) is null
```

jkfran (approved):
It works fine, but I think before merging to master it needs a UX review? @danielmutis 

This is how it looks right now:
image

danielmutis:
I tried it and it looks fine to me. It's strange that on demo most of my charms with builds don't show the latest builds (as if it wasn't build with github) and the status is unkown rather than Not Released (which is what we should show for charms we don't know where they come from).

The label in the table should be "Status" not "Build status" @steverydz

steverydz:
@danielmutis I've made the changes we discussed earlier. If the build status is unknown it uses the release status.

https://snapcraft-io-3318.demos.haus/snaps

PR #3353, opened 2021-01-26
Load YouTube video only when needed

### LINKED ISSUE #3341: Replace YouTube embed with thumbnail which loads video on click

steverydz:
## Done
Show YouTube thumbnail by default then load the video on hover to improve performance

## QA
- Go to https://snapcraft-io-3353.demos.haus/code
- Hover over the video section and see that the YouTube video loads

## Issue
Fixes #3341

### DIFF
--- a/templates/partials/_video.html
+++ b/templates/partials/_video.html
@@ -1,6 +1,10 @@
 {% if video.type == "youtube" %}
-  <iframe id="ytplayer" type="text/html" width="818" height="460"
-          src="" frameborder="0" data-src="{{ video.url }}?autoplay=0&cc_load_policy=1&modestbranding=1&rel=0"></iframe>
+  <div class="youtube-container" data-youtube>
+    <div class="youtube-thumbnail">
+      <img src="https://i3.ytimg.com/vi/{{ video.id }}/maxresdefault.jpg" alt="" width="1029" height="588">
+      <img src="{{ static_url('images/yt_play_btn.svg') }}" alt="" width="68" height="48" class="youtube-play-button">
+    </div>
+  </div>
 {% elif video.type == "vimeo" %}
   <iframe id="vimeoplayer" width="818" height="460" frameborder="0"
           webkitallowfullscreen mozallowfullscreen allowfullscreen
@@ -11,13 +15,8 @@
 
 <script>
   document.addEventListener("DOMContentLoaded", function() {
-    const youtubeVideoFrame = document.getElementById("ytplayer");
     const vimeoplayerFrame = document.getElementById("vimeoplayer");
     const asciicastplayerFrame = document.getElementById("asciicastplayer");
-
-    if (youtubeVideoFrame) {
-      youtubeVideoFrame.src = youtubeVideoFrame.dataset.src;
-    }
 
     if (vimeoplayerFrame) {
       vimeoplayerFrame.src = vimeoplayerFrame.dataset.src;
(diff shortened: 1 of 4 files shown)


templates/partials/_video.html
```diff
@@ -31,5 +28,25 @@
...
+
+    if (document.querySelector("[data-youtube]")) {
+      const videoFrame = document.createElement("iframe");
+      videoFrame.id = "ytplayer";
+      videoFrame.type = "text/html";
+      videoFrame.width = "818";
+      videoFrame.height = "460";
+      videoFrame.src = "{{ video.url }}?autoplay=0&cc_load_policy=1&modestbranding=1&rel=0";
+
+      const videoWrapper = document.getElementById("youtube-video-wrapper");
```
anthonydillon, line 40: You are mixing your lookups by checking for [data-youtube] then selecting the element with the `id`. You would set one as a variable above and use everywhere you need the wrapper. Protects from someone changing one and breaking this and saves traversing the dom twice.

anthonydillon (changes requested):
Nice work, works well!

Huge gains on the performance score:
| Before | After |
| --- | --- |
| image | image

templates/partials/_video.html
```diff
@@ -1,6 +1,8 @@
 {% if video.type == "youtube" %}
-  <iframe id="ytplayer" type="text/html" width="818" height="460"
-          src="" frameborder="0" data-src="{{ video.url }}?autoplay=0&cc_load_policy=1&modestbranding=1&rel=0"></iframe>
+  <div id="youtube-video-wrapper" class="youtube-container" data-youtube>
+    <img src="http://i3.ytimg.com/vi/{{ video.id }}/maxresdefault.jpg" alt="" width="818" height="460">
```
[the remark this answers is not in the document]
steverydz: Done

templates/partials/_video.html
steverydz: Of course. What was I thinking? Fixed.

steverydz:
@anthonydillon I've made the requested changes

tbille:
Really nice. 

The of the thumbnail and the video (loaded on hover) is too small compared to the orginal one. Also is there a way to avoid the blank that happens on hover, like some kind of smoother swapping? 

Your modification:
image

The orginal

image

tbille:
@steverydz would it make sense to keep the image loaded, and put on top of it the youtube video? This will help avoiding this second with the blank transition

anthonydillon:
It takes a bit to load the video so I clicked the image a few times before the video landed and kicked off. I think its good but the blank might be clearer to the user?

anthonydillon:
Maybe trigger on click and have it autoplay?

steverydz:
> Maybe trigger on click and have it autoplay?

It doesn't autoplay when loading this way. I can't find anything saying it's deliberate. Maybe it's triggered by the parent load event?

anthonydillon:
I get "Feature Policy: Skipping unsupported feature name “autoplay”." in the console and no image or video now.

steverydz:
@anthonydillon Fixed. Firefox and Chrome handle iframe load event differently. I have to leave `allow="autoplay"` otherwise it won't autoplay in Chrome. Firefox is just a warning so should be fine.

anthonydillon:
@steverydz thanks, that works. Maybe the wrapper should be a button or something so tabbing works and you get cursor: pointer on hover?

steverydz:
@anthonydillon Good idea! Done

anthonydillon (approved):
Good stuff @steverydz, thank you :+1:

PR #3364, opened 2021-01-28
Update channel map

jkfran:
## Done
- Update canonicalwebteam.store-api and replaced ["_embedded"]["clickindex:package"] code
- Update channel map to show "released-at" date instead of "created-at"

## How to QA
- Check all the consumer part works fine

## Issue / Card
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/1759


jkfran:
@tbille The channel map reshape is very similar between the two stores right now:

https://github.com/canonical-web-and-design/charmhub.io/blob/master/webapp/store/logic.py#L165-L222

The main difference is that we are ordering the channels on the backend in charmhub and we here we are ordering then in the front, the order is to show the channel like in this order:
```
latest/stable
latest/candidate
latest/edge
```

Doing it in the backend is better for clients without JS but it probably slows a bit the response time, I didn't change it, what do you think?

tbille:
@jkfran I think this is a great start. I opened an issue to bring this logic to snapcraft: https://github.com/canonical-web-and-design/snapcraft.io/issues/3365

tbille (approved):
LGTM

PR #3381, opened 2021-02-01
Add collaborators col to snaps table

### LINKED ISSUE #3370: Add "Collaborators" column to snaps table

steverydz:
## Done
Added collaborators column to the snaps table in the admin pages

## QA
- Go to https://snapcraft-io-3381.demos.haus/admin
- Check that the snaps table has a collaborators column which has a comma delimited list of users

## Issue
Fixes #3370


solazio (approved):
Looks good, however I have a few comments not necessarily related to this PR:

- On mobile I cannot close the navigation
- On mobile the table looks very bad. Perhaps you can use the `p-table--mobile-card`
- Is there a need no handle the case when no `users` exists? Or it will always be at least 1 user?

PR #3395, opened 2021-02-02
Allow adding and removing of snaps from store

### LINKED ISSUE #3374: Mange snaps: Include/remove snaps

steverydz:
## Done
Add manage snaps UI

## QA
- Go to https://snapcraft-io-3395.demos.haus/admin
- Click on "Manage snaps"
- Check that the table has checkboxes for each snap
- Check that the form works (although there will be an error because no snaps are being sent)

## Issue
Fixes #3374

### DIFF
--- a/static/js/public/manage-snaps.js
+++ b/static/js/public/manage-snaps.js
@@ -1,0 +1,57 @@
+function formatSelectedSnapData(snapCheckboxes) {
+  let snaps = [];
+
+  snapCheckboxes.forEach((snap) => {
+    snaps.push({
+      name: snap.dataset.snapName,
+    });
+  });
+
+  return snaps;
+}
+
+function updateSnaps() {
+  const manageSnapsForm = document.querySelector("#manage-snaps-form");
+  const saveButton = document.querySelector(".js-save-button");
+  const snapsTable = document.querySelector("#snaps-table");
+  const snapCheckboxes = Array.prototype.slice.call(
+    snapsTable.querySelectorAll(":checked")
+  );
+
+  const originalSnapState = formatSelectedSnapData(snapCheckboxes);
+  let currentSnaps = formatSelectedSnapData(originalSnapState);
+  let dirtyData = false;
+
+  snapCheckboxes.forEach((snap) => {
+    snap.addEventListener("change", () => {
+      currentSnaps = formatSelectedSnapData(snapCheckboxes);
+
+      if (
+        JSON.stringify(originalSnapState.sort()) !==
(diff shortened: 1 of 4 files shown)


static/js/public/manage-snaps.js
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function updateSnaps() {
+  const manageSnapsForm = document.querySelector("#manage-snaps-form");
+  const saveButton = document.querySelector(".js-save-button");
+  const snapCheckboxes = Array.prototype.slice.call(
+    document.querySelectorAll(".js-snap-checkbox")
+  );
+
+  const originalSnapState = getSelectedSnaps(snapCheckboxes);
```
anthonydillon, line 22: Couldn't you just querySelectorAll with `:checked` to get the checked issue?

static/js/public/manage-snaps.js
```diff
@@ -0,0 +1,58 @@
...
+  let currentSnaps = getSelectedSnaps(snapCheckboxes);
```
anthonydillon, line 23: You could save the compute here by just assigning the original variable above

static/js/public/manage-snaps.js
```diff
@@ -0,0 +1,58 @@
...
+  let dirtyData = false;
+
+  snapCheckboxes.forEach((snap) => {
+    snap.addEventListener("change", () => {
+      currentSnaps = getSelectedSnaps(snapCheckboxes);
```
anthonydillon, line 28: Maybe replace with a querySelectorAll with `:checked` as the function is doing DOM look ups anyway

jkfran (approved):

anthonydillon (dismissed):
Few questions in the code

static/js/public/manage-snaps.js
steverydz: Good point. I've had to scope it to the table though to be on the safe side.

static/js/public/manage-snaps.js
steverydz: Good point. Done.

static/js/public/manage-snaps.js
steverydz: In this case we want data rather than DOM nodes. I've renamed the function so its intent is clearer.

static/js/public/manage-snaps.js
anthonydillon: My question was can we replace formatSelectedSnaoData with a `querySelectorAll('.input:selected')`

PR #3398, opened 2021-02-03
Filter members

### LINKED ISSUE #3241: Add search and filter component to "Manage members" table
Design: https://app.zeplin.io/project/5f3170425822f41fc299a330/screen/5fb6913708e04b418dfbb8a3

This will require the table to be built in JS.

steverydz:
## Done
Added a filter to the members table

## QA
- Go to https://snapcraft-io-3398.demos.haus/admin/ahnuP3quahti9vis8aiw/members/manage
- Use the filter to search by display name, username or email
- Check that only the appropriate table rows are present
- Use the filter to search by role
- Check that only the appropriate table rows are present
- Change some roles and save
- Check that the form still works

## Issue
Fixes #3241

### DIFF
--- a/static/js/public/manage-members.js
+++ b/static/js/public/manage-members.js
@@ -1,3 +1,5 @@
+import debounce from "../libs/debounce";
+
 function inviteMember() {
   const inviteMemberForm = document.querySelector("#invite-member-form");
   const emailField = inviteMemberForm.querySelector("#invite-member-email");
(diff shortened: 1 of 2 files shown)


static/js/public/manage-members.js
```diff
@@ -138,4 +140,77 @@ function updateMembers(membersData) {
...
+
+  const ROLES = {
+    admin: "admin",
+    reviewer: "review",
+    viewer: "view",
+    publisher: "access",
+  };
+
+  const membersTableBody = document.querySelector("#members-table tbody");
+  const filterMembersField = document.querySelector("#filter-members");
```
solazio, line 155: In order to avoid confusion and bugs, lately I've been using the `data` attribute to target elements. So in this case you could use:
```suggestion
  const filterMembersField = document.querySelector("[data-js='filter-members']");
```
In my view this will make you code more future proof and avoid issues with people changing class or id names. 
This is more a recommendation, rather than anything.

solazio (approved):
Looks good, just one small comment in the code.

static/js/public/manage-members.js
steverydz: In cases like these I think an ID is better as it indicates there will only be one of these elements. I do agree with using data attributes in place of classes though.

PR #3404, opened 2021-02-09
Add manage member invites form

### LINKED ISSUE #3375: Manage members: Revoke, resend and reopen invites

steverydz:
## Done
Added the ability to resend, reopen and revoke member invites

## QA
- Go to https://snapcraft-io-3404.demos.haus/admin
- Click "Members" on the left navigation
- Click the "Invites" tab above the table
- Use the search box to filter the table by each invites email address
- Check that invites are grouped by status
- Use the action buttons to revoke, resend and reopen invites
- Check that the status changes appropriately

## Issue
Fixes #3375

### DIFF
--- a/templates/admin/invites.html
+++ b/templates/admin/invites.html
@@ -1,0 +1,198 @@
+{% set current_tab = "invites" %}
+
+{% extends "/admin/admin_layout.html" %}
+
+{% block meta_title %}Manage invites in {{ store.name }} | Snapcraft{% endblock %}
+
+{% block admin_content %}
+
+<ul class="p-inline-list--stretch">
+  <li class="p-inline-list__item u-no-margin--right">
+    <div class="p-button-group">
+      <div class="p-button-group__inner">
+        <div class="p-button-group__buttons">
+          <a href="/admin/{{ store.id }}/members" class="p-button-group__button">Members</a>
+          <a href="/admin/{{ store.id }}/members/invites" class="p-button-group__button is-selected">Invites</a>
+        </div>
+      </div>
+    </div>
+  </li>
+  <li class="p-inline-list__item u-align--right">
+    <label for="filter-invites" class="u-hide">Search for user</label>
+    <input type="text" name="filter-invites" id="filter-invites" placeholder="Search for user" style=" max-width: 320px; min-width: 0;">
+  </li>
+</ul>
+
+<table id="invites-table" class="p-table--mobile-card">
+  <thead>
+    <tr>
+      <th>Status</th>
+      <th>User</th>
+      <th>Expires</th>
+      <th>Roles</th>
+      <th>&nbsp;</th>
+    </tr>
+  </thead>
+  <tbody>
+
+    {% if pending_invites %}
+      <tr>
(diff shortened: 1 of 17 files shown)


steverydz:
@danielmutis Can you take a look at this and let me know what you think?

https://snapcraft-io-3404.demos.haus/admin/ahnuP3quahti9vis8aiw/members/manage

I've made the radio buttons disabled if an invite is expired or revoked because you aren't able to resend or revoke in either of these cases.

The data in the columns is everything we have available.

solazio:
2 `Menu` buttons is a bit confusing

image

solazio:
@danielmutis @steverydz  perhaps we can add an `title` tooltip to the buttons (especially the `Reopen` one) to explain in more detail what they don. I literally have no idea what `Reopen` will do if I click it. Maybe is just me, not the majority of users.

templates/admin/invites.html
```diff
@@ -0,0 +1,194 @@
...
+{% endblock %}
+
+{% block scripts_includes %}
+<script src="{{ static_url('js/dist/modal.js') }}" defer></script>
+<script src="{{ static_url('js/dist/manage-invites.js') }}" defer></script>
+{% endblock %}
+
+{% block scripts %}
+<script>
+  window.addEventListener("DOMContentLoaded", function () {
```
solazio, line 189: Perhaps we should use Raven to get Sentry errors when it occurs.

solazio (commented):
Just small comment in the code.

steverydz:
> @danielmutis @steverydz perhaps we can add an `title` tooltip to the buttons (especially the `Reopen` one) to explain in more detail what they don. I literally have no idea what `Reopen` will do if I click it. Maybe is just me, not the majority of users.

I'm not sure that's necessary. It's a page of invites and the buttons are tied to each invite. Is it really that unclear that reopen would reopen the invite?

steverydz:
> 2 Menu buttons is a bit confusing

@solazio I agree. @danielmutis do you have a suggestion for different wording here? How about something like "Back to stores"?

steverydz:
> Perhaps we should use Raven to get Sentry errors when it occurs.

Done

steverydz:
@solazio 
- I've added Raven
- I don't think the title/tooltip is necessary on the buttons but I'll leave that for @danielmutis to decide
- I'll also leave @danielmutis to decide about the double "Menu" labelling. I'd suggest "Back to stores"

solazio (approved):

danielmutis:
A few comments:

- The Manage invites title is too far from the "back to members" link. It should be similar to snapcraft. 

- We can change each of the messages when a user does an action (reopen, resend...) as an explanation of what it will do, instead of "do you really want to perform this action". This could be "Resending your invite will send a reminder email to this user. Do you still want to do it?"; "Reopening your invite will send a new invite to this email. Do you still want to do it?"; and "Revoking your invite will prevent this user from accepting your invite. Do you still want to do it?". Let's also align the buttons here to the right. The green one should be the first from the right. 

- About the menu, I really don't think this should be a button, it should be a white link. It could say "< Toggle side navigation" like we do on docs.

PR #3406, opened 2021-02-11
Count accept-encoding instances in prometheus

nottrobin:
Here's why:

The cache has to vary by encoding, as in, if a browser requests "gzip" compression of a page, it clearly can't use the same cache as for a browser requesting "br" (brotli) compression. The cache can't actually do this perfectly, because when a request comes in, it doesn't know which compression the app will respond with, so it has to check with the app. This means it uses the distinct value of the "accept-encoding" header instead. So "accept-encoding: gzip, br" will lead to a different cache from "accept-encoding: br, gzip". It does this by setting the header "vary: accept-encoding".

We [are currently looking into](https://github.com/canonical-web-and-design/maas.io/issues/559) how to change our cache-control headers to ensure pages are effectively cached. To understand this, we need to know how many cache variations to expect per page. This in turn requires us to understand how many variations of the "accept-encoding" header we receive, and how many of each. So this is fundamentally what I'm trying to count.

I added in the user-agent, not because I strictly need it, but this will then allow us to see which browsers are sending which accept-encoding strings, which could be useful for a future blog post.

## QA

I don't know how to QA it properly, but if you whack a debugger at the end of the `prometheus_metrics` function, and then run the site and visit it in your browser, you can see that it runs without error.

### DIFF
--- a/webapp/handlers.py
+++ b/webapp/handlers.py
@@ -1,11 +1,11 @@
 import socket
 from urllib.parse import unquote, urlparse, urlunparse
 
-from canonicalwebteam import image_template
 import flask
-
 import prometheus_client
+import user_agents
 import webapp.template_utils as template_utils
+from canonicalwebteam import image_template
 from webapp import authentication
 
 from datetime import datetime
@@ -17,6 +17,12 @@ badge_counter = prometheus_client.Counter(
 badge_logged_in_counter = prometheus_client.Counter(
     "badge_logged_in_counter",
     "A counter of badges requests of logged in users",
+)
+
+accept_encoding_counter = prometheus_client.Counter(
+    "accept_encoding_counter",
+    "A counter for Accept-Encoding headers, split by browser",
+    ["accept_encoding", "browser_family"],
 )
 
 
@@ -114,6 +120,21 @@ def set_handlers(app):
 
     @app.before_request
     def prometheus_metrics():
+        # Accept-encoding counter
+        # ===
+        agent_string = flask.request.headers.get("User-Agent")
+
+        # Exclude probes, which happen behind the cache
+        if not agent_string.startswith(("kube-probe", "Prometheus")):
+            agent = user_agents.parse(agent_string or "")
+
+            accept_encoding_counter.labels(
+                accept_encoding=flask.request.headers.get("Accept-Encoding"),
+                browser_family=agent.browser.family,
+            ).inc()
+
+        # Badge counters
+        # ===
         if "/static/images/badges" in flask.request.url:
             if flask.session:
                 badge_logged_in_counter.inc()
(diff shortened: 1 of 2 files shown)


webapp/handlers.py
```diff
@@ -114,6 +120,13 @@ def clear_trailing():
 
     @app.before_request
     def prometheus_metrics():
+        # Accept-encoding counter
+        accept_encoding_counter.labels(
+            accept_encoding=flask.request.headers.get("Accept-Encoding"),
+            user_agent=flask.request.headers.get("User-Agent"),
```
tbille, line 126: so actually the user agent will be an issue. 

The way we can gather prometheus metrics is with a  determined number of labels, otherwise it becomes a nightmare to aggregate values and we will fill the database very fast.

By doing a quick check on graylog currently snapcraft has <316,287 different user agent values just in the last 7 days.

instead of storing the user agent like this, could we determine/calculate what browser is being used from the user agent and store the browser name instead as a label?
As an example 
```
accept_encoding_counter{accept_encoding="br",browser="Chrome"} 12
accept_encoding_counter{accept_encoding="br",browser="Firefox"} 9
```

I will make it much easier for us to do some analytics.

nottrobin: Good point, thanks!

I've changed it to only send "browser_family" and "browser_version". Hopefully this is a small enough variation to handle. Or do you need me not to include version? I just thought there's a good chance that the user agent string changed between versions of browser, and it would be good to detect that so we can inform the world.

webapp/handlers.py
```diff
@@ -114,6 +120,16 @@ def clear_trailing():
 
     @app.before_request
     def prometheus_metrics():
+        # Accept-encoding counter
+        user_agent = user_agents.parse(flask.request.headers.get("User-Agent"))
+
+        accept_encoding_counter.labels(
+            accept_encoding=flask.request.headers.get("Accept-Encoding"),
+            browser_family=user_agents.browser.family,
```
tbille, line 128: It seems that this make the k8s check fail.
You might want to check if the user agent is defined

https://sentry.is.canonical.com/canonical/snapcraft-io/issues/16317/?query=is%3Aunresolved

nottrobin:
@tbille this is now awesome. +1 please? ;)

webapp/handlers.py
```diff
@@ -114,6 +120,21 @@ def clear_trailing():
 
     @app.before_request
     def prometheus_metrics():
+        # Accept-encoding counter
+        # ===
+        agent_string = flask.request.headers.get("User-Agent")
+
+        # Exclude k8s probes, which happen behind the cache
+        if not agent_string.startswith("kube-probe"):
```
tbille, line 128: sorry another one. We are also getting this user agent from behind the cache: `Prometheus/2.9.2`

:sweat:

tbille (approved):
Let's merge it! :rocket:

PR #3412, opened 2021-02-18
Add changes count to members save button

### LINKED ISSUE #3410: Add number of changes to save button on the manage members page
On the manage members page, when changes have been made, the number of changes should be shown in the "Save" button.

- If there are no changes, the button should read "Save changes" and be disabled
- If there are changed, the button should read "Save 10 changes"

steverydz:
## Done
Added a change count to the save button for the manage members admin page

## QA
- Go to https://snapcraft-io-3412.demos.haus/admin/ahnuP3quahti9vis8aiw/members/manage
- Check and uncheck roles
- If there are no changes from the original state the save button should be disabled and read "Save changes"
- If there is one change from the original state the save button should be enabled and read "Save 1 change"
- If there are more than one changes from the original state the save button should be enabled and read "Save x changes", with x reflecting the number of changes from the original state

## Issue
Fixes #3410

### DIFF
--- a/static/js/public/manage-members.js
+++ b/static/js/public/manage-members.js
@@ -110,9 +110,37 @@ function updateMembers(membersData) {
         member.dirty = false;
       }
 
-      const dirtyState = newData.filter(function (data) {
-        return data.dirty;
+      const dirtyState = newData.filter((data) => data.dirty);
+
+      let changeCount = 0;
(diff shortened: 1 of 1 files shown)


solazio (approved):
Looks good :+1:. Thanks

PR #3464, opened 2021-03-15
Fix sentry error

### LINKED ISSUE #3463: TypeError: n.contentWindow.document.body is null
https://sentry.is.canonical.com/canonical/snapcraft-io/issues/1692/?referrer=github_plugin

```
TypeError: n.contentWindow.document.body is null
  at updateCardSize (./static/js/publisher/publicise.js:156:13)
  at d (/static/js/modules/raven.min.js:2:4689)
```

steverydz:
## Done
Fix sentry error

##  QA
- Go to https://snapcraft-io-3464.demos.haus/lukewh-test/publicise/cards
- Open the browser console
- Resize and check that there are no console errors

## Issue
Fixes #3463

### DIFF
--- a/static/js/publisher/publicise.js
+++ b/static/js/publisher/publicise.js
@@ -149,7 +149,7 @@ function initEmbeddedCardPicker(options) {
   function updateCardSize() {
     // calulate frame height to be a bit bigger then content itself
     // to have some spare room for responsiveness
-    if (previewFrame.offsetParent) {
+    if (previewFrame.offsetParent && previewFrame.contentWindow.document.body) {
       const height =
         Math.floor(
           (previewFrame.contentWindow.document.body.clientHeight + 20) / 10


anthonydillon:
I don't get errors but I can't seem to replicate in production. Have you managed to trigger this in production?

steverydz:
> I don't get errors but I can't seem to replicate in production. Have you managed to trigger this in production?

No, but my thinking here is to make the code more defensive so it's less likely to error. It could be related to a specific snap but sentry doesn't give the URL so it's hard to say.

anthonydillon:
I think I found the underlying issue. [Every second we are updating the content](https://github.com/canonical-web-and-design/snapcraft.io/blob/master/static/js/publisher/publicise.js#L177) in the card (I assume to address resizing). If you switch the options:
image
It rebuilds the preview and if that happens at the same time as the internal triggers the preview does not exist. 

I think the true fix would be to rearchitect how this works and not just trigger every second but this would hide the error for now.

static/js/publisher/publicise.js
```diff
@@ -149,7 +149,12 @@ function initEmbeddedCardPicker(options) {
   function updateCardSize() {
     // calulate frame height to be a bit bigger then content itself
     // to have some spare room for responsiveness
-    if (previewFrame.offsetParent) {
+    if (
+      previewFrame.offsetParent &&
+      previewFrame.contentWindow &&
+      previewFrame.document &&
+      previewFrame.document.body
```
anthonydillon, line 156: You just need to check the preview has a body to cover this error: `previewFrame.contentWindow.document.body`

anthonydillon (changes requested):
You can simplify this patch

static/js/publisher/publicise.js
steverydz: That's some good debugging! I obviously wasn't clicking around fast enough. Anyway, fixed.

anthonydillon (approved):
LGTM :+1:

PR #3479, opened 2021-03-30
Add action to check docs links

anthonydillon:
## Done
Add action to check docs links.

## How to QA
- Check CI passes


anthonydillon:
So @tbille this thing is crawling all over the place so will need to find a way to limit the crawl to just docs first.

PR #3496, opened 2021-04-16
Add sticky footer

steverydz:
## Done
- Added a sticky footer when changes are present and adding functionality to revert changes
- Removed global footer

## QA
- Go to https://snapcraft-io-3496.demos.haus/admin/ahnuP3quahti9vis8aiw/members
- Make changes and check that a sticky footer appears at the bottom of the page
- Check that the page count is correct
- Use the filter and make changes and check that the changes persist
- In the footer click the revert button and check that all changes have been returned to their original state

## Issue
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/1993


solazio:
@danielmutis are you happy with the position of the notification bar?
image

static/js/public/manage-members.js
```diff
@@ -1,194 +1,247 @@
...
-      if (dirtyState.length) {
-        if (changeCount === 1) {
-          saveButton.innerText = "Save 1 change";
-        } else {
-          saveButton.innerText = `Save ${changeCount} changes`;
-        }
-      } else {
-        saveButton.innerText = "Save changes";
+    const roleCheckboxes = clone.querySelectorAll("[data-js-role-checkbox]");
+    roleCheckboxes.forEach((checkbox) => {
```
solazio, line 114: You should do:
```suggestion
    Array.from(roleCheckboxes).forEach((checkbox) => {
```

or something similar, so you don't do `forEach` on DOM nodes.
This comment applies to other places where you use `queryselectorAll`.

templates/_layout-no-footer.html
```diff
@@ -0,0 +1,126 @@
+<!doctype html>
```
solazio, line 1: Can you not use one base layout and add/or not the footer on the template, so we don't have to maintain 2 templates with duplicate html?

solazio (commented):
Looks good @steverydz , just a few comments in the code.

solazio (approved):
Looks good :+1:. Thanks

PR #3510, opened 2021-04-28
Update snaps table

steverydz:
## Done
Added rows to table for snaps in current store, and snaps in subsequent stores

## QA
- Go to https://snapcraft-io-3510.demos.haus/admin/ahnuP3quahti9vis8aiw/snaps
- Check that the snaps in the current store are in their own section
- Check that any snaps in other stores are in their own sections (in this case there will only be the global store)

## Issue
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/2030


solazio (approved):
Looks good :+1:. Thanks

solazio:
Good job @steverydz ! I really like how it looks on smaller screens.

PR #3527, opened 2021-05-04
Add checkbox for uploading metadata for snap

### LINKED ISSUE #3322: Add checkbox to disable/enable Upload metadata in settings page

steverydz:
## Done
Added a checkbox to the snap settings page which enables or disables the uploading of metadata

## QA
- You need a snap on staging
- Go to https://snapcraft-io-3527.demos.haus/SNAP_NAME/settings
- Check that there is a checkbox for uploading metadata
- Check the box and save the form
- Check that the box is still checked
- Uncheck the box and save the form
- Check that the box is still unchecked
- Go to https://snapcraft-io-3527.demos.haus/SNAP_NAME/listing
- If the upload metadata box on the settings page is checked you should see a notification
- If the upload metadata box on the settings page is not checked you should not see a notification

## Issue
Fixes #3322

### DIFF
--- a/templates/publisher/settings.html
+++ b/templates/publisher/settings.html
@@ -173,6 +173,29 @@ Settings for {% if display_title %}{{ display_title }}{% else %}{{ snap_title }}
       </div>
 
       <div class="row p-form__group">
+        <hr class="u-no-margin--bottom">
+      </div>
+
+      <div class="row p-form__group">
+        <div class="col-2">
+          <label class="p-form__label">Upload metadata on release:</label>
+        </div>
+        <div class="col-8">
+          <div class="p-form__control">
+            <input
+              type="checkbox"
+              name="update_metadata_on_release"
+              id="update_metadata_on_release"
+              {% if update_metadata_on_release|tojson == "true" %}checked{% endif %}
+            />
+            <label for="update_metadata_on_release">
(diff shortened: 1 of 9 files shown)


steverydz:
@tbille I'm not sure I have the right link here? I made a best guess.

templates/publisher/settings.html
```diff
@@ -172,6 +172,31 @@
...
+        <hr class="u-no-margin--bottom">
+      </div>
+
+      <div class="row p-form__group">
+        <div class="col-2">
+          <label class="p-form__label">Upload metadata on release:</label>
+        </div>
+        <div class="col-8">
+          <div class="p-form__control">
+            <input
```
tbille, line 185: On check/uncheck update the state of the save button

steverydz:
@tbille I've fixed the enable/disable save button issue

tbille:
@steverydz it seems that when I save with the checkbox  checked, then I reload I don't get the checkbox checked

steverydz:
> @steverydz it seems that when I save with the checkbox checked, then I reload I don't get the checkbox checked

@tbille that's because there is no value for that field returned from the backend - is this using staging?

jkfran:
I don't think this message is being shown on the right action.
image

Right now this message is displayed whenever the publisher save the listing form with or without the new checkbox enabled.

IMO it should appear on the listing page without having to save the changes when the developer has this new checkbox enabled on the settings page.

jkfran (approved):
Nice!

PR #3529, opened 2021-05-05
Empty invites state

### LINKED ISSUE #3517: Admin pages: invites screen looks incomplete when no invites are available
### Expected behaviour

There is clear message stating that there are no invites in a state that is actionable by the user (no revoked and no pending invites).

See for example in this image, it feels like the UI has not completed the loading of the content:

Screenshot_2021-04-28 Manage invites in Foo Bar Baz Zaraza Foo Bar Baz Zaraza Snapcraft

Browse the manage invites page for a store that has no pending nor revoked invites.

steverydz:
## Done
Added a notification if there are no invites

## QA
- Go to https://snapcraft-io-3529.demos.haus//admin/test1LaNg1xi6eonae5R/members/invites
- Check that there is a notification explaining that there are no invites

## Issue
Fixes #3517

### DIFF
--- a/templates/admin/invites.html
+++ b/templates/admin/invites.html
@@ -171,6 +171,12 @@
   </tbody>
 </table>
 
+{% if not pending_invites and not expired_invites and not revoked_invites %}
+  <div class="p-notification--information">
+    <p class="p-notification__response">There are currently no pending, expired or revoked invites.</p>
+  </div>
+{% endif %}
+
 <div class="p-modal" id="invite-modal" style="display: none; z-index: 110;">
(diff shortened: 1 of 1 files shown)


AnnSoMuller:
looking at Vanilla, That looks good!
I would just correct a bit the message itself: "There a**re** currently no pending, expired or revoked invites**.**"
- replace "a" by "are"
- add a dot at the end

steverydz:
@AnnSoMuller thanks - fixed

jkfran (approved):

PR #3592, opened 2021-06-30
Fix navigation highlight color

steverydz:
## Done
Fixed the navigation highlight color for the current page

## QA
- Go to https://snapcraft-io-3592.demos.haus/
- Click on any of the pages in the top navigation
- Check that there is an orange line underneath the link for the page that you are on

## Issue
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/2115

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -409,3 +409,9 @@ code {
 .u-text-muted {
   color: $color-mid-dark;
 }
+
+.p-navigation {
+  .p-navigation__link.is-selected > a::before {
+    background-color: $ubuntu-orange;
+  }
+}


static/sass/styles.scss
```diff
@@ -409,3 +409,9 @@ code {
 .u-text-muted {
   color: $color-mid-dark;
 }
+
+.p-navigation {
+  .p-navigation__link.is-selected > a::before {
+    background-color: $ubuntu-orange;
```
sowasred2012, line 415: I think you could achieve this by updating the value of [`$color-accent`](https://github.com/canonical-web-and-design/snapcraft.io/blob/master/static/sass/styles.scss#L4) to be `$ubuntu-orange`. This is the variable used by default in the navigation pattern, and the only other pattern it affects is the `is-ticked` list icon. Snapcraft does use that pattern on the [docs search](https://snapcraft-io-3592.demos.haus/docs/search?q=ooooooo), but if we could live with those being Ubuntu orange it would mean less overrides.

Including these links to show what changing the `$color-accent` variable would impact:
https://github.com/canonical-web-and-design/vanilla-framework/search?q=color-accent
https://github.com/canonical-web-and-design/snapcraft.io/search?q=color-accent

steverydz: @long-chung @AnnSoMuller do either of you have any objections to the ticks on [the docs search](https://snapcraft-io-3592.demos.haus/docs/search?q=ooooooo) being Ubuntu orange?

long-chung:
Are there any other pages across canonical products that use orange circled ticks?
@waynecrosby @marcushas

static/sass/styles.scss
sowasred2012: Just had a chat with @lyubomir-popov about this - it's a little weird that `$color-accent` ties the nav highlight and is-ticked list items like this, so we're gonna look at whether that's something we can address in Vanilla directly.

sowasred2012 (approved):
LGTM

waynecrosby:
@long-chung Not to my knowledge

PR #3597, opened 2021-07-07
Fix hidden tooltip on IoT page

### LINKED ISSUE #3596: snapcraft.io/iot verified account green tick bug
on the snapcraft.io/iot there seems to be a bug with the certified publishers. 

### Hover on green tick

**current behaviour:**

On hover, all you can see its a little black arrow:

**expected behaviour:**

On the snap details page you can see there is a hover on the green tick that says "verified account":

steverydz:
## Done
Fixed the hidden tooltip on the verified badge on the IoT page

## QA
- Go to https://snapcraft-io-3597.demos.haus/iot
- Hover over any of the ticks in the green circles in the "IoT tools and server" section
- Check that you see the tooltip

## Issue
Fixes #3596

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -415,3 +415,7 @@ code {
     background-color: $ubuntu-orange;
   }
 }
+
+.u-overflow-visible {
+  overflow: visible !important;
+}
--- a/templates/partials/_verified_developer.html
+++ b/templates/partials/_verified_developer.html
@@ -1,4 +1,4 @@
-<span class="p-verified p-tooltip p-tooltip--top-center" aria-describedby="{{ package_name }}-tooltip">
-    <img src="https://assets.ubuntu.com/v1/75654c90-rosette.svg" width="12" height="12" />
+<span class="p-verified p-tooltip p-tooltip--top-center">
+    <img src="https://assets.ubuntu.com/v1/75654c90-rosette.svg" width="12" height="12" alt="Verified account" />
     <span class="p-tooltip__message u-align--center" role="tooltip" id="{{ package_name }}-tooltip">Verified account</span>
 </span>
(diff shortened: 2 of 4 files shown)


sowasred2012:
This is technically out of scope for this issue, but also technically not because the tooltip info is "hidden" to some users, and it would be a fairly simple drive by update to include here. 

The verified icon needs alt text, because at the moment screen readers are just reading out the filename of the icon ("75654c90-rosette.svg"). 

The wrapping span has an `aria-labelledby` attribute that is being ignored (maybe spans are ignored by screen readers?), but the effect that was clearly intended by adding it can be achieved with img alt text:

[_verified_developer.html](https://github.com/canonical-web-and-design/snapcraft.io/blob/master/templates/partials/_verified_developer.html):

Current:
```html
<span class="p-verified p-tooltip p-tooltip--top-center" aria-describedby="{{ package_name }}-tooltip">
    <img src="https://assets.ubuntu.com/v1/75654c90-rosette.svg" width="12" height="12" />
    <span class="p-tooltip__message u-align--center" role="tooltip" id="{{ package_name }}-tooltip">Verified account</span>
</span>
```

Suggestion:
```html
<span class="p-verified p-tooltip p-tooltip--top-center">
    <img src="https://assets.ubuntu.com/v1/75654c90-rosette.svg" width="12" height="12" alt="Verified account" />
    <span class="p-tooltip__message u-align--center" role="tooltip" id="{{ package_name }}-tooltip">Verified account</span>
</span>
```
You can verify this fix in Chrome with the [Screen Reader extension](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en), or with the [WAVE extension](https://wave.webaim.org/extension/), currently it shows 7 errors in Chrome, and this issue covers 6 of them.

sowasred2012 (approved):
LGTM, thanks!

steverydz:
@sowasred2012 thanks - that's a good catch! I've fixed it. I also fixed the other error.

PR #3615, opened 2021-07-30
Use original Discourse topic

jkfran:
## Done
- Use original Discourse topic for docs

## How to QA
- https://snapcraft-io-3615.demos.haus/docs should work as usual


steverydz (approved):
LGTM 👍

PR #3637, opened 2021-08-24
Temp Fix: Avoid authenticated API calls to Discourse

jkfran:
## Done
- Avoid authenticated API calls to Discourse

## How to QA
- /docs should work as usual

## Issue / Card
We are waiting on IS to update the discourse configuration for authenticated calls RT#132688


steverydz (approved):
LGTM 👍

PR #3638, opened 2021-08-24
Add store search filter to brand store

### LINKED ISSUE #2168: Publisher listing page markup and SCSS is too custom
The way the page is currently put together is not ideal. Some of the custom styles are modifying vanilla components that create inconsistencies.

Possibly look at cleaning up sticky header styles as well and turn it into a pattern?

steverydz:
## Done
Added a search filter to the brand store side navigation

## QA
- Go to https://snapcraft-io-3638.demos.haus/admin/test1LaNg1xi6eonae5R/snaps
- Type in store names in the side navigation search
- Check that the list is filtered correctly
- Clear the search
- Check that the list is as before

## Issue
Fixes #2168

### DIFF
--- a/static/js/public/brand-store-search.js
+++ b/static/js/public/brand-store-search.js
@@ -1,0 +1,46 @@
+import debounce from "../libs/debounce";
+
+function init(stores) {
+  const searchInput = document.querySelector(".js-store-search-input");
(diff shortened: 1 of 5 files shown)


jkfran (approved):
LGTM

PR #3683, opened 2021-09-10
Write brand store settings form tests

steverydz:
## Done
Write tests for the brand store settings view

## QA
- Run `yarn test-brand-store`
- Check that the tests pass

### DIFF
--- a/static/js/brand-store/Settings/Settings.test.js
+++ b/static/js/brand-store/Settings/Settings.test.js
@@ -44,7 +44,9 @@ test("the 'is public' checkbox should not be checked when the current store is s
     </Provider>
   );
 
-  expect(document.querySelector("#is_public")).not.toBeChecked();
+  expect(
+    screen.getByLabelText("Include this store in public lists")
+  ).not.toBeChecked();
 });
 
 test("the 'is public' checkbox should be checked when the current store is not set to private", () => {
@@ -57,7 +59,9 @@ test("the 'is public' checkbox should be checked when the current store is not s
     </Provider>
   );
 
-  expect(document.querySelector("#is_public")).toBeChecked();
+  expect(
+    screen.getByLabelText("Include this store in public lists")
+  ).toBeChecked();
 });
 
 test("the correct value is given to the store ID field", () => {
--- a/static/js/brand-store/Settings/Settings.test.js
+++ b/static/js/brand-store/Settings/Settings.test.js
@@ -1,6 +1,6 @@
 import React from "react";
 import { Provider, useSelector } from "react-redux";
-import { render } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
 import "@testing-library/jest-dom";
 import Settings from "./Settings";
 import store from "../stores/store";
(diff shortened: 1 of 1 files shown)


static/js/brand-store/Settings/Settings.test.js
```diff
@@ -15,59 +15,105 @@ jest.mock("react-router-dom", () => ({
...
+test("the save button is enabled when the data changes", () => {
+  setupMockSelector(initialState);
+
+  render(
+    <Provider store={store}>
+      <Settings />
+    </Provider>
+  );
+
+  fireEvent.click(screen.getByRole("checkbox"));
```
hatched, line 117: It's preferential to use `userEvent` vs `fireEvent` however in this case you can probably just do...
```suggestion
  screen.getByRole("checkbox").click();
```

static/js/brand-store/Settings/Settings.test.js
```diff
@@ -15,59 +15,105 @@ jest.mock("react-router-dom", () => ({
...
+test("the save button is disabled by default", () => {
+  setupMockSelector(initialState);
+
+  render(
+    <Provider store={store}>
+      <Settings />
+    </Provider>
+  );
+
+  expect(screen.getByText("Save changes").disabled).toEqual(true);
```
hatched, line 105: Up to you but you can use [toBeDisabled()](https://github.com/testing-library/jest-dom#tobedisabled) instead of the generic `toEqual`

steverydz:
@hatched thanks - I've made the changes

static/js/brand-store/Settings/Settings.test.js
```diff
@@ -15,59 +15,105 @@ jest.mock("react-router-dom", () => ({
...
+  initialState.currentStore.currentStore.private = false;
+  setupMockSelector(initialState);
+
+  render(
+    <Provider store={store}>
+      <Settings />
+    </Provider>
+  );
+
+  expect(document.querySelector("#is_public").checked).toEqual(true);
```
hatched, line 60: You can use `toBeChecked` here https://github.com/testing-library/jest-dom#tobechecked

steverydz: Done

nottrobin (approved):
LGTM :+1:

But a couple of questions out of scope for this PR:

- Why `test-brand-store` (rather than `test`)?
- Why are so many checks failing? Is that true of `master` too?

steverydz:
@nottrobin thanks

> Why `test-brand-store` (rather than `test`)?

It will be added to the `test` script once we're closer to completion - we just want to keep it separate for simplicity whilst working on it

> Why are so many checks failing? Is that true of `master` too?

I'll take a look at the `test-js` check as I hadn't noticed that. The dotrun and python checks are because of the issue with the docker snap - they're failing everywhere.

static/js/brand-store/Settings/Settings.test.js
```diff
@@ -15,59 +15,105 @@ jest.mock("react-router-dom", () => ({
...
 
-  test("the 'is public' checkbox should not be checked when the current store is set to private", () => {
-    render(
-      <Provider store={store}>
-        <Settings />
-      </Provider>
-    );
-    expect(document.querySelector("#is_public").checked).toEqual(false);
-  });
+  expect(document.querySelector("#is_public")).not.toBeChecked();
```
petermakowski, line 47: Would you consider focusing less on implementation details and making this more user-centric?
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text

```suggestion
  expect(screen.getByLabelText("Include this store in public lists")).not.toBeChecked();
```

static/js/brand-store/Settings/Settings.test.js
```diff
@@ -15,59 +15,105 @@ jest.mock("react-router-dom", () => ({
...
+  initialState.currentStore.currentStore.private = false;
+  setupMockSelector(initialState);
+
+  render(
+    <Provider store={store}>
+      <Settings />
+    </Provider>
+  );
+
+  expect(document.querySelector("#is_public")).toBeChecked();
```
petermakowski, line 60: As above:
```suggestion
  expect(screen.getByLabelText("Include this store in public lists")).toBeChecked();
```

static/js/brand-store/Settings/Settings.test.js
petermakowski: If you want to be more specific you could use `getByRole`:
```suggestion
  expect(
    screen.getByRole("checkbox", { name: "Include this store in public lists" })
  ).not.toBeChecked();
```

steverydz: I'll do it in this case as that text isn't likely to change anytime soon, but in general I don't think it's a good idea to query by the text as in a lot of cases in our products the text can change a lot which shouldn't break tests (IMO).

When writing tests I keep the following in mind:

> Tests should be sensitive to changes in the behaviour of the code under test. If the behaviour changes, the test result should change.
> Source: https://medium.com/@kentbeck_7670/test-desiderata-94150638a4b3

I don't think text changes fall into this category. Something to discuss at dev catchup 😁

PR #3684, opened 2021-09-12
Fix navigation brand color for contrast and visibility

fitojb:
## Done
- Changed the definition of “brand color” to use a value that is actually from the Snap logo, which makes the site integrate better with that brand (unlike previously, when Ubuntu orange was used, visually tying the product with Ubuntu in detriment of other distros) and adds the benefit that the active tab is actually distinguishable now (the previous color was too dark and too similar to the background of the navigation bar).

## How to QA

## Issue / Card
N/A

## Screenshots
Before: 
before

After: 
imatge

### DIFF
--- a/static/sass/styles-embedded.scss
+++ b/static/sass/styles-embedded.scss
@@ -1,5 +1,5 @@
 // colour definitions
-$color-brand: #00302f;
+$color-brand: #fa6340;
 $color-accent: $color-brand;
 $color-navigation-background: #252525;
 
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -1,6 +1,6 @@
 // colour definitions
 $ubuntu-orange: #e95420;
-$color-brand: #00302f;
+$color-brand: #fa6340;
 $color-accent: $color-brand;
 $color-navigation-active-bar: $ubuntu-orange;
 


danielmutis:
@fitojb thanks for looking at this issue. I think you are right that the current implementation uses a colour for the active page that is almost impossible to distinguish from the background. However, I think this should be consistent with the implementation we have of the navigation on other Canonical websites. So, I think the colour of this highlight should be our brand orange and not the green from the logo.

If you could replace the colour by that orange I'd be happy to approve this PR.

fitojb:
@danielmutis As I said, IMO using Ubuntu orange everywhere diminishes its distinctive quality for the Ubuntu brand, and Snap technology isn’t Ubuntu-specific, but distro-agnostic. (And the orange highlight in Snapcraft’s logo isn’t the same orange as Ubuntu’s: `#fa6340` vs. `#e95420`. I can use Snapcraft’s orange instead, but I don’t think it should be Ubuntu’s.)

fitojb:
Updated patch not to be green anymore.

danielmutis:
Thank you for your input. Nevertheless, our brand team is currently working on a rebrand for our products and a new navigation will come in due course. I insist that it's best not to add a new shade of orange to our brand colours which is also not even so different from the Ubuntu one.

fitojb:
If the navigation is going to be replaced again, there’s no point in me patching something that’s not going to last. Team, I’d advise you not to reinvent the wheel so often.

PR #3693, opened 2021-09-15
Add snaps table

### LINKED ISSUE #3650: Build snaps table
Snaps should not be shown/available in the global store

steverydz:
## Done
- Added the basic snaps table with a filter
- Added the tabbed navigation to be shared between all views

**Note**: Tests will come in a follow-up PR

## QA
- Go to https://snapcraft-io-3693.demos.haus/admin/ahnuP3quahti9vis8aiw/snaps
- Check that there is a table of snaps on the page
- Check that the search snaps filter filters snaps by name

## Issue
Fixes #3650, #3651

### DIFF
--- a/static/js/brand-store/SectionNav/SectionNav.js
+++ b/static/js/brand-store/SectionNav/SectionNav.js
@@ -1,0 +1,47 @@
+import React from "react";
+import { Link, useParams } from "react-router-dom";
+import PropTypes from "prop-types";
+
+function SectionNav({ sectionName }) {
+  const { id } = useParams();
+
+  return (
+    <div className="p-tabs">
+      <ul className="p-tabs__list">
+        <li className="p-tabs__item">
+          <Link
+            to={`/admin/${id}/snaps`}
+            className="p-tabs__link"
+            aria-selected={sectionName === "snaps"}
+          >
+            Snaps
+          </Link>
+        </li>
+        <li className="p-tabs__item">
+          <Link
+            to={`/admin/${id}/members`}
+            className="p-tabs__link"
+            aria-selected={sectionName === "members"}
+          >
+            Members
+          </Link>
+        </li>
+        <li className="p-tabs__item">
+          <Link
+            to={`/admin/${id}/settings`}
+            className="p-tabs__link"
+            aria-selected={sectionName === "settings"}
+          >
+            Settings
+          </Link>
+        </li>
+      </ul>
+    </div>
+  );
+}
+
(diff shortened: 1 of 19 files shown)


static/js/brand-store/SectionNav/SectionNav.test.js
```diff
@@ -0,0 +1,94 @@
...
+test("active state is set on snaps when on members section", () => {
+  render(
+    <Router>
+      <SectionNav sectionName="members" />
+    </Router>
+  );
+
+  expect(screen.getByText("Members").getAttribute("aria-selected")).toEqual(
+    "true"
+  );
```
petermakowski, line 49: It's good to see more testing-library-style queries 👍 

This is fine as is, but in case of fairly generic strings such as this it might better to be more specific and use `getByRole`.

Also helps the reader of this test understand what you're testing exactly.

```suggestion
  expect(screen.getByRole('link', {name: "Members" }).getAttribute("aria-selected")).toEqual(
    "true"
  );
```

static/js/brand-store/SectionNav/SectionNav.js
```diff
@@ -0,0 +1,47 @@
...
+  const { id } = useParams();
+
+  return (
+    <div className="p-tabs">
+      <ul className="p-tabs__list">
+        <li className="p-tabs__item">
+          <Link
+            to={`/admin/${id}/snaps`}
+            className="p-tabs__link"
+            aria-selected={sectionName === "snaps"}
```
petermakowski, line 15: adding `role="tab"` to each link will make for a better experience for screen reader users (e.g. voiceover doesn't mention the status "selected" without it at all).

steverydz:
@petermakowski I've made the changes - if you can approve then I can merge :)

static/js/brand-store/SectionNav/SectionNav.test.js
```diff
@@ -0,0 +1,94 @@
+import React from "react";
+import { BrowserRouter as Router } from "react-router-dom";
+import { render, screen } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import SectionNav from "./SectionNav";
+
+jest.mock("react-router-dom", () => ({
```
hatched, line 7: Instead of mocking this we've instead used `MemoryRouter` from `react-router` when we don't need to interact with the route and used `createMemoryHistory` from the `history` package passed to the `Router` when we do. It allows us to be a bit more explicit in our tests.

petermakowski (approved):

PR #3703, opened 2021-09-22
Add members filter

### LINKED ISSUE #3655: Build members table filter

steverydz:
## Done
Added the members table filter

## QA
- Go to https://snapcraft-io-3703.demos.haus/admin/ahnuP3quahti9vis8aiw/members
- Use the search and filter input to filter members by name or email address

## Issue
Fixes #3655

### DIFF
--- a/static/js/brand-store/Members/Members.js
+++ b/static/js/brand-store/Members/Members.js
@@ -1,4 +1,4 @@
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
 import { useSelector, useDispatch } from "react-redux";
 import { format } from "date-fns";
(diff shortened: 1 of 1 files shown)


jkfran (approved):
LGTM 👍🏽

static/js/brand-store/Members/Members.js
```diff
@@ -71,6 +76,10 @@ function Members() {
     dispatch(fetchInvites(id));
   }, [id]);
 
+  useEffect(() => {
+    setFilteredMembers(members);
```
hatched, line 80: Typically you want to avoid having two sources of truth for a dataset.

What can potentially happen here is that you filter a set of members below, but then the member list updates which will then trigger this and wipe out the filter. What you probably want to do is have a conditional that checks if there is a filter and if not then use the full list, else use the filtered results.

PR #3707, opened 2021-09-23
Remove usabilla code

steverydz:
## Done
Removed usabilla code as it is injected via GTM

## QA
Check that the usabilla code is not in the template


anthonydillon (approved):
LGTM :+1:

PR #3709, opened 2021-09-24
Brandstore react

steverydz:
## Done
- _A changelist description explaining what the change achieves and why you’re making it._

## How to QA

## Issue / Card
Fixes #

## Screenshots


long-chung:
Looking really good Steve. Thanks for doing this. It feels so much more polished than before.
Just a few points from my end

- [ ] Can we left align all the content rather than centred 
- [ ] For the rows that don't have checkboxes, i'm assuming they can't be removed? if so can we make them non links like we have done on the "web & design" section
- [ ] Snaps tab on mobile - need to group "published in" with a heading as only the first card is showing this
- [ ] In the members tab, could you reduce the column widths of all the checkboxes 
- [ ] members tab mobile - are you able to keep the checkboxes all on the same row?
- [ ] Probably out of scope but would be good to implement in the future is a full width top nav

PR #3715, opened 2021-09-29
Add member role management

steverydz:
## Done
Added the ability to manage members roles in the brand store

## QA
- Go to https://snapcraft-io-3715.demos.haus/admin/ahnuP3quahti9vis8aiw/members
- Make changes to member roles
- Check that the save bar appears at the bottom of the page
- Save the form and the changes should persist
- Make more changes, then undo those changes
- Check that the save bar disappears

## Issue
Fixes #3715

### DIFF
--- a/static/js/brand-store/Members/MembersTable.js
+++ b/static/js/brand-store/Members/MembersTable.js
@@ -1,6 +1,7 @@
 import React, { useState, useEffect } from "react";
 import PropTypes from "prop-types";
 import { MainTable, CheckboxInput } from "@canonical/react-components";
+import ROLES from "./memberRoles";
 
 function MembersTable({ filteredMembers, changedMembers, setChangedMembers }) {
   const [members, setMembers] = useState(filteredMembers);
@@ -67,7 +68,7 @@ function MembersTable({ filteredMembers, changedMembers, setChangedMembers }) {
             <>
               <i
                 className="p-icon--information"
-                title="Admins manage the store's users and roles, and control the store's settings."
+                title={ROLES.admin.description}
               >
                 Role description
               </i>{" "}
@@ -80,7 +81,7 @@ function MembersTable({ filteredMembers, changedMembers, setChangedMembers }) {
             <>
               <i
                 className="p-icon--information"
-                title="Reviewers can approve or reject snaps, and edit snap declarations."
+                title={ROLES.review.description}
               >
                 Role description
               </i>{" "}
@@ -91,10 +92,7 @@ function MembersTable({ filteredMembers, changedMembers, setChangedMembers }) {
         {
           content: (
             <>
-              <i
-                className="p-icon--information"
-                title="Viewers are read-only users and can view snap details, metrics, and the contents of this store."
-              >
+              <i className="p-icon--information" title={ROLES.view.description}>
                 Role description
               </i>{" "}
               Viewer
@@ -106,7 +104,7 @@ function MembersTable({ filteredMembers, changedMembers, setChangedMembers }) {
             <>
               <i
                 className="p-icon--information"
-                title="Publishers can invite collaborators to a snap, publish snaps and update snap details."
+                title={ROLES.access.description}
               >
                 Role description
               </i>{" "}
--- a/static/js/brand-store/Members/MembersTable.js
+++ b/static/js/brand-store/Members/MembersTable.js
@@ -1,8 +1,61 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import PropTypes from "prop-types";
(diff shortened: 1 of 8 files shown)


static/js/brand-store/Members/MembersTable.js
```diff
@@ -1,8 +1,61 @@
...
 
-function MembersTable({ filteredMembers }) {
+function MembersTable({ filteredMembers, changedMembers, setChangedMembers }) {
+  const [members, setMembers] = useState(filteredMembers);
+
+  const handleRoleChange = (member, role) => {
+    const updatedMembers = members.map((item) => {
+      let updatedItem = {};
+
+      if (item.id === member.id) {
```
hatched, line 12: To make the code easier to follow you can invert this conditional and return early. This will remove an indent and others won't have to try and figure out what happens if it's not true (in this case, nothing)

static/js/brand-store/Members/MembersTable.js
```diff
@@ -1,8 +1,61 @@
...
+        if (item.roles.includes(role)) {
+          updatedItem = {
+            ...item,
```
hatched, line 15: I'm not sure what the structure of `item` is but this is doing a shallow copy so you will want to be careful.

static/js/brand-store/Members/MembersTable.js
```diff
@@ -1,8 +1,61 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import PropTypes from "prop-types";
 import { MainTable, CheckboxInput } from "@canonical/react-components";
 
-function MembersTable({ filteredMembers }) {
+function MembersTable({ filteredMembers, changedMembers, setChangedMembers }) {
+  const [members, setMembers] = useState(filteredMembers);
+
+  const handleRoleChange = (member, role) => {
```
hatched, line 8: If you change `member` here to something like `currentMember` then below you can change `item` to `member` and it'll read better :)

static/js/brand-store/Members/MembersTable.js
```diff
@@ -1,8 +1,61 @@
...
+        } else {
+          updatedItem = {
+            ...item,
+            roles: [].concat(item.roles, [role]),
+          };
+        }
+
+        if (changedMembers.find((m) => m.id === member.id)) {
+          if (
+            JSON.stringify([...member.roles].sort()) !==
```
hatched, line 27: While this is a tricky way of comparing two arrays it's not very resilient to values that won't stringify. In fact it'll make the application 🔥 . So I'd recommend using an existing library for this or creating a little util like...

untested...
```js
function checkArrayEqual(array1, array2) {
  if (array1.length !== array2.length) {
    return false
  }

  const array2Sorted = array2.roles.slice().sort();
  return array1.slice().sort().every((val, idx) => val === array2Sorted[idx]);
}

checkArrayEqual(member.roles, updatedItem.roles);
```

hatched (changes requested):
Looks good, just some small tweaks to make the code a bit more easier to read and robust.

static/js/brand-store/Members/memberRoles.js
```diff
@@ -0,0 +1,24 @@
+const ROLES = {
+  admin: {
+    name: "Admin",
+    description:
```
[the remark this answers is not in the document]
steverydz: This is used in the add/invite member form but also should have been used for the tooltips on the members table, which I missed, so have now updated.

static/js/brand-store/Members/MembersTable.js
steverydz: Done

static/js/brand-store/Members/MembersTable.js
steverydz: I get your point. In this case `item` is a shallow object (is that a term 😉 ).

static/js/brand-store/Members/MembersTable.js
steverydz: Done

static/js/brand-store/Members/MembersTable.js
steverydz: Done

steverydz:
Thanks @hatched. I've made those changes now.

static/js/brand-store/Members/MembersTable.js
hatched: Haha, I'm not sure, but now it is! You trendsetter you!

hatched (approved):
Looks good, thanks for those updates 👍

hatched:
During the QA I noticed that the Toast never disappeared unless I explicitly closed it. I think this is out of scope for this PR though so 👍

PR #3776, opened 2021-11-22
Setup typescript for brandstore

steverydz:
## Done
Added TypeScript to the project for brand store (and ignored so can be converted gradually)

**Note**: This PR contains a bug fix from [another PR](https://github.com/canonical-web-and-design/snapcraft.io/pull/3775).

## QA
- Go to https://snapcraft-io-3776.demos.haus/admin
- Check that the app still works

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,12 @@
+{
+  "compilerOptions": {
+    "outDir": "./static/js/dist",
+    "noImplicitAny": true,
+    "module": "es6",
+    "target": "es5",
+    "jsx": "react",
+    "allowJs": true,
+    "moduleResolution": "node",
+    "strict": true
+  }
+}
--- a/package.json
+++ b/package.json
@@ -81,6 +81,8 @@
     "style-loader": "2.0.0",
     "swiper": "6.5.1",
     "topojson-client": "3.1.0",
+    "ts-loader": "9.2.6",
+    "typescript": "4.5.2",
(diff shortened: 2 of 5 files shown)


tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+  "compilerOptions": {
+    "outDir": "./static/js/dist",
+    "noImplicitAny": true,
+    "module": "es6",
+    "target": "es5",
+    "jsx": "react",
+    "allowJs": true,
+    "moduleResolution": "node"
+  },
+  "exclude": [
```
hatched, line 11: Instead of excluding, you should be able to only have it include ts files. That way you don't have to update it as you convert the files over. Simply changing the extension will be sufficient.

steverydz: OK, I've removed `exclude`. I tried adding `include` but it errors if it's empty but I'll add it once I start converting files.

tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
```
hatched, line 2: I'd like to see `strict: true` here here....It's best to start with `strict` so you don't have to go back and fix issues when you want to enable it later.

You can see the config we use for the [Juju Dashboard here](https://github.com/canonical-web-and-design/jaas-dashboard/blob/main/tsconfig.json)

hatched (approved):
Looks good, I'd just like to see `strict`mode enabled.

PR #3779, opened 2021-11-24
remove general use of sudo from snap install commands

philclifford:
## Removed sudo from install instructions
- This should just replace all calls to `sudo snap` with just `snap` in order to let snap elevate privileges where necessary and to stop encouraging universal unnecessary use of `sudo`

- I have changed the calls under `./webapp` but not those under `.gitub/workflows/`

### DIFF
--- a/static/js/publisher/release/actions/defaultTrack.js
+++ b/static/js/publisher/release/actions/defaultTrack.js
@@ -38,7 +38,7 @@ export function clearDefaultTrack() {
         showNotification({
           status: "success",
           appearance: "positive",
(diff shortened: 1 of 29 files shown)


jkfran:
Hi @philclifford , thank you for your contribution, unfortunately, `snap` might not be able to elevate privileges when necessary in all the different environments, it is reasonable for us to let the users know that the command needs privileges by adding `sudo`.

We can remove `sudo` in cases where privileges are not needed but all the changes here are to the `snap install` instruction witch does need privileges.

Thank you.

PR #3782, opened 2021-11-26
Setup typescript for brandstore

### LINKED ISSUE #3767: Convert brand store settings view to TypeScript

steverydz:
## Done
Converted the settings view and related tests to TypeScript

## QA
- Go to https://snapcraft-io-3782.demos.haus/admin/test1LaNg1xi6eonae5R/settings
- Check that the view still works as before (compare to live)

## Issue
Fixes #3767


static/js/brand-store/Settings/Settings.test.tsx
```diff
@@ -1,31 +1,19 @@
...
-
-const setupMockSelector = (state) => {
-  useSelector.mockImplementation((callback) => {
-    return callback(state);
-  });
-};
-
-beforeEach(() => {
-  initialState = {
+function getInitialState() {
```
hatched, line 15: Can you type the return of this function...
```suggestion
function getInitialState(): State {
```

static/js/brand-store/Settings/Settings.test.tsx
```diff
@@ -44,6 +32,35 @@ beforeEach(() => {
       loading: false,
     },
   };
+}
+
+interface Member {}
```
hatched, line 37: This type doesn't appear to be being used.

static/js/brand-store/Settings/Settings.tsx
```diff
@@ -19,30 +19,58 @@ import NotAuthorized from "../NotAuthorized";
 import PasswordToggle from "../shared/PasswordToggle";
 import SectionNav from "../SectionNav";
 
+declare global {
```
hatched, line 22: These types of things I like to move to a shared global definition (`d.ts`) file as it'll be valid for the whole application.

static/js/brand-store/Settings/Settings.tsx
```diff
@@ -19,30 +19,58 @@ import NotAuthorized from "../NotAuthorized";
...
+  interface Window {
+    CSRF_TOKEN: string;
+  }
+}
+
+interface RootState {
```
hatched, line 28: This type appears to be very similar to the one in the test file above...I'm assuming they are the same. What you'll want to do is have a single place where these are typed....typically it would be in this file (or in a separate `types` file). Then the type is exported from that file. You'd then import it using `import type {RootState} from "./Settings"` in the test files (or other files that need it).

static/js/brand-store/Settings/Settings.tsx
```diff
@@ -19,30 +19,58 @@ import NotAuthorized from "../NotAuthorized";
...
+  currentStore: {
+    loading: Boolean;
+  };
+
+  members: {
+    loading: Boolean;
+  };
+}
+
+interface RouteParams {
```
hatched, line 38: Unless there is a reason to use `interface` we typically use `type`. There are some legitimate reasons to use `interface` but I don't see them here.

static/js/brand-store/Settings/Settings.tsx
```diff
@@ -83,8 +111,8 @@ function Settings() {
     setIsPrivateStore(!isPrivateStore);
   };
 
-  const handleRadioButtonChange = (e) => {
-    setManualReviewPolicy(e.target.value);
+  const handleRadioButtonChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setManualReviewPolicy((e.target as HTMLInputElement).value);
```
hatched, line 115: Typically when you have to use `as` it's a point to stop and take a second glance....it shouldn't be required in the majority of cases.

In this case I think you want `e.currentTarget` instead.

static/js/brand-store/Settings/Settings.tsx
```diff
@@ -19,30 +19,58 @@ import NotAuthorized from "../NotAuthorized";
...
-  const { id } = useParams();
+  const { id } = useParams<RouteParams>();
 
   const [isPrivateStore, setIsPrivateStore] = useState(true);
   const [manualReviewPolicy, setManualReviewPolicy] = useState("");
   const [isSaving, setIsSaving] = useState(false);
   const [showSuccessNotification, setShowSuccessNotification] = useState(false);
   const [showErrorNotification, setShowErrorNotification] = useState(false);
-  const [currentMember, setCurrentMember] = useState(null);
+  const [currentMember, setCurrentMember] = useState({ roles: Array() });
```
hatched, line 63: Can you type this `useState` plz.

static/js/brand-store/Settings/Settings.tsx
hatched: Also this type will very likely need to be shared across a number of components, you'll probably want to move this out into the file where the routes are defined (then exported), or a shared types file.

hatched (changes requested):
Good first pass @steverydz lemme know if you have any questions about my comments here.

steverydz:
Thanks @hatched, that's really helpful. I've addressed all of those comments.

hatched (approved):
Looks good thanks for the updates 👍

PR #3798, opened 2021-12-08
Alias publisher names to IDs in store search

### LINKED ISSUE #3797: Alias search for publishers name to ID in the backend
If search is `publisher:snapcrafters` then convert that to the publisher ID in the backend.

Do the same for `publisher:kde` and `publisher:jetbrains`.

steverydz:
## Done
Aliased the publisher names to their IDs for the search

## QA
- Go to https://snapcraft-io-3798.demos.haus/
- Search for `publisher:jetbrains`
- Check that the search results are correct
- Search for `publisher:kde`
- Check that the search results are correct
- Search for `publisher:snapcrafters`
- Check that the search results are correct

## Issue
Fixes #3797

### DIFF
--- a/templates/store/search.html
+++ b/templates/store/search.html
@@ -2,7 +2,7 @@
 
 {% block meta_title %}
   {% if query %}
-    Snap search results for '{{ query }}'
+    Snap search results for '{{ display_query }}'
     {% if category_display %}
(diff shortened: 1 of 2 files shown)


jkfran (approved):
Nice!

PR #3801, opened 2021-12-13
Convert brand store section nav to TypeScript

### LINKED ISSUE #3774: Convert brand store SectionNav view to TypeScript

steverydz:
## Done
Converted brand store `SectionNav` component to Typescript

## QA
- Go to https://snapcraft-io-3801.demos.haus/admin
- Check that the section nav works

## Issue
Fixes #3774


static/js/brand-store/SectionNav/SectionNav.test.tsx
```diff
@@ -5,7 +5,7 @@ import "@testing-library/jest-dom";
 import SectionNav from "./SectionNav";
 
 jest.mock("react-router-dom", () => ({
-  ...jest.requireActual("react-router-dom"),
+  ...jest.requireActual<any>("react-router-dom"),
```
petermakowski, line 8: Could you give more details on the TS error message here please?

If you don't want to create a proper type/interface now, might be worth creating a type to be updated in the future as more files get moved over.

As an example we use the following in maas-ui:
`export type TSFixMe = any;`

petermakowski (approved):
LGTM, one comment below.

static/js/brand-store/SectionNav/SectionNav.test.tsx
steverydz: The error was `Spread types may only be created from object types`. I've removed the type now and there is no error so I think my editor might have not updated.

petermakowski: Awesome 👍

PR #3861, opened 2022-01-19
Reintroduce upload metadata release field

steverydz:
## Done
- _A changelist description explaining what the change achieves and why you’re making it._

## How to QA

## Issue / Card
Fixes #

## Screenshots


jkfran:
I think there is something not working as expected or the message is not right.

image

The warning message says:
> If you manually edit the Listing page, the automatic updates will be turned off.

This is not happening when I modify the listing page.

jkfran:
image

And this one is saying the opposite.

Roadmaster:
I concur with Fran that the messaging and behaviour are somewhat inconsistent.

"If you manually edit the Listing page, the automatic updates will be turned off." is a lie - I tried it on the demo and the flag (back in the settings tab) was not turned off. So in that sense, the warning that "any changes you make here will be overwritten" is correct because editing metadata doesn't turn off the flag. However, as I remember the definition of the feature, it should indeed turn the flag off if any field in the "listing" page is edited. So I think that should be implemented, at which point "any changes you make here will be overwritten" becomes redundant.

Also, in the listing page, when I edit any metadata, I get a message at the top saying:

"Information here was automatically updated to the latest version of the snapcraft.yaml released to the stable channel"

This is confusing - this appeared when I edited the metadata manually, so it sounds incorrect. Maybe the logic around this should also be reviewed. Or the message can be changed to (bold to emphasize the changes) "Information here **will be** automatically updated to the latest version of the snapcraft.yaml released to the stable channel. **If this is undesirable, turn off 'upload metadata on release' in the Settings tab**". (though the last part is unneeded if we implement the "flag turned off magically on manual metadata update on the web" behaviour).

jkfran (approved):
LGTM! Thank you @steverydz

PR #3862, opened 2022-01-19
Update macOS install instructions

owenjones:
`brew cask` commands were deprecated in 2.6.0 in favour of using the `--cask` flag: https://github.com/Homebrew/brew/pull/8899

## Done
- Replaced deprecated `brew cask install multipass` command with `brew install --cask multipass`


jkfran (approved):
Thank you!

PR #3865, opened 2022-01-21
Test with gevent.Timeout

jkfran:
## Done
- Remove pybreaker, since it is not being useful.
- Added `gevent.Timeout` for HTTP requests with a limit close to the gunicorn timeout limit. This should avoid the restarts of the gunicorn worker and we should get the `GeventGreenletTimeout` exception on Sentry every time one was about to happen.

## QA
- Demo should work as usual

**This is a temporary change until we analyze the results**

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,6 @@ bleach==3.3.0
 humanize==3.2.0
 mistune==0.8.4
 pybadges==2.2.1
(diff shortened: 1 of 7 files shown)


nottrobin (approved):
LGTM

steverydz (approved):
LGTM 👍

PR #3869, opened 2022-01-25
Upgrade flask-base and set session cookie to SameSite=None Secure

jkfran:
## Done
- Upgrade flask-base to v1
- Set session cookie to `SameSite=None Secure`. To avoid issues with Firefox, see:
https://bugs.launchpad.net/canonical-identity-provider/+bug/1895734/

## How to QA
- `dotrun` and login


steverydz (approved):
LGTM 👍

PR #3872, opened 2022-02-02
Revert debugging changes

jkfran:
## Done
- Remove extra logging DEBUG line (START)
- Remove gevent.Timeout

### DIFF
--- a/webapp/api/requests.py
+++ b/webapp/api/requests.py
@@ -1,6 +1,5 @@
 import os
 
-from gevent import Timeout as GeventTimeout
 from requests import Session as RequestsSession
 from requests.exceptions import Timeout, ConnectionError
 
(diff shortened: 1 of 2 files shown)


webapp/api/requests.py
```diff
@@ -30,15 +29,9 @@ def __init__(self, *args, **kwargs):
...
-            ):
-                return super().request(
-                    method=method, url=url, timeout=timeout, **kwargs
-                )
+            return super().request(
+                method=method, url=url, timeout=timeout, **kwargs
+            )
         except Timeout:
             raise ApiTimeoutError(
                 "The request to {} took too long".format(url)
```
nottrobin, line 37: Why do we want to raise a custom timeout error rather than letting the natural requests.exceptions.Timeout bubble up? It seems like you'd get more context from the native error wouldn't you?

jkfran: Good question, if you look at the blame these changes are very old, at that time without content-cache looks like they had to override the Session class to cache responses, and also they decided to raise custom exceptions.

This session is used for Snapcraft store API calls and I think that's why someone decided to raise custom exceptions. On Charmhub for instance we are not doing anything like this just the regular talisker.get_session(), I think this issue can be solved with the alignment that we need to do on the codebase between the two stores and related as well with this issue https://github.com/canonical-web-and-design/snapcraft.io/issues/3230

nottrobin (approved):

PR #3889, opened 2022-02-21
Add views for publishers and reviewers

### LINKED ISSUE #3854: "My Stores" redirect loop between old dashboard and snapcraft.io
For the brand store admin pages migration, there will be a set of redirects from the old dashboard to snapcraft.io.

In particular, the current store list from dashboard will redirect to the "My stores" in dashboard. The problem is that, if the user has only publisher/reviewer access to stores, snapcraft.io would say:

```
No stores

Currently this section can only show stores if you are a viewer or an admin. If you are a publisher or reviewer you will find your stores on the dashboard.
```

I believe that the My Stores in snapcraft.io page should be extended so if the user has NOT any admin/viewer role in any store, a more specific message is shown, for example:

* If the user has a publisher role in any store, links to register a snap name directly in that store can be offered
* If the user has reviewer role in any store, links to the reviewer pages in the old dashboard can be offered. These are:
   * for staging data, https://dashboard.staging.snapcraft.io/stores/<store-id>/reviews/
   * for production, https://dashboard.snapcraft.io/stores/<store-id>/reviews/

This way we can safely redirect from dashboard to snapcraft.io without causing a redirect loop.

steverydz:
## Done
Added views for stores where the user is only a reviewer, publisher or both

## QA
- Make sure you are a publisher only in a store, a reviewer only in a store, and a reviewer and publisher in a store
- Go to https://snapcraft-io-3889.demos.haus/admin
- On the stores that you are only a reviewer you should see a reviewer page
- On the stores that you are only a publisher you should see a publisher page
- On the stores that you are a reviewer and publisher you should see a reviewer and publisher page

## Issue
Fixes #3854

## Screenshots
Publisher

Reviewer and publisher

Reviewer

### DIFF
--- a/static/js/brand-store/Publisher/Publisher.tsx
+++ b/static/js/brand-store/Publisher/Publisher.tsx
@@ -1,0 +1,15 @@
+import React from "react";
+
+function Publisher() {
+  return (
+    <div className="u-fixed-width">
+      <h1>Publisher</h1>
+      <p>
+        As a publisher you can{" "}
+        <a href="/snaps">register a snap name on the Snap store</a>.
+      </p>
+    </div>
+  );
+}
+
+export default Publisher;
--- a/static/js/brand-store/Publisher/index.ts
+++ b/static/js/brand-store/Publisher/index.ts
@@ -1,0 +1,1 @@
+export { default } from "./Publisher";
--- a/static/js/brand-store/Reviewer/Reviewer.tsx
+++ b/static/js/brand-store/Reviewer/Reviewer.tsx
@@ -1,0 +1,18 @@
+import React from "react";
+
+function Reviewer() {
+  return (
+    <div className="u-fixed-width">
+      <h1>Reviewer</h1>
+      <p>
+        As a reviewer you can{" "}
+        <a href={`${window.API_URL}stores/reviews/`}>
+          manage your snaps on the dashboard
+        </a>
+        .
+      </p>
+    </div>
+  );
+}
+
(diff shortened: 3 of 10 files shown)


steverydz:
@AnnSoMuller could you please take a look at this?

Essentially, if a user is only a reviewer they should see a specific page, if they are only a publisher they should see another specific page, and if they are both they see a different page. See screenshots.

AnnSoMuller:
@steverydz I'm not sure to well understand all the needs here.
As a Publisher only, if you haven't had any snap published yet, you can have the first screen as an empty state => makes sense :)
However, that should belong to a dedicated tab e.g "My snaps".
So for a reviewer, you have now a tab "Reviews"
etc.

Or is this temporary while we haven't done the migration?

steverydz:
@AnnSoMuller Sorry, I think I made it sound more complicated than it is 😆 

This is a temporary measure which helps Natalia prevent some redirect issues when migrating from the brand store.

She has requested that we show the links in each view to the user depending on their role. Really I just wanted a review of the pages themselves in the screenshots.

jkfran (approved):
I am not sure if looks good to display their roles (Publisher, Reviewer, ...) in a big title.

But it has my approval. Thank you Steve.

cristinadresch:
@AnnSoMuller could you have a look at this one before we merge pls?

AnnSoMuller:
As a temporary measure, I give my approval too :)

PR #3895, opened 2022-02-28
Replace password toggle component

### LINKED ISSUE #3716: Replace PasswordToggle component in the brand store with the one from react-components

steverydz:
## Done
- Removed PasswordToggle component in favour of the React components library one
- Drive by: Fixed a warning about an unsupported CSS property value when building the project

## QA
- Go to https://snapcraft-io-3895.demos.haus/admin/ahnuP3quahti9vis8aiw/settings
- Check that there is a password field for the store ID which is read only

## Issue
Fixes #3716


jkfran (approved):
LGTM!

PR #3949, opened 2022-04-05
Progressive releases table cell

steverydz:
## Done
- Add progressive release data to releases table cell

## QA
- Go to https://snapcraft-io-3949.demos.haus/core/releases
- Check that the progressive releases cell has the arrow between versions

## Issue / Card
Fixes #2331

## Screenshots
progressive_releases_table_cell

### DIFF
--- a/static/js/publisher/release/components/releasesTable/cellViews.js
+++ b/static/js/publisher/release/components/releasesTable/cellViews.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 import React, { Fragment } from "react";
 import PropTypes from "prop-types";
 
--- a/static/js/publisher/release/components/releasesTable/cellViews.js
+++ b/static/js/publisher/release/components/releasesTable/cellViews.js
@@ -144,6 +144,12 @@ export const RevisionInfo = ({
     </Fragment>
   );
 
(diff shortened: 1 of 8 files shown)


steverydz:
@anasereijo This is the PR that adds the version to version info to the release table cells. The example here has large revision numbers which means the destination revision is truncated. Not sure what would be the best solution for this?

static/js/publisher/release/components/releasesTable/cellViews.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
Lukewh, line 1: :eyes:

Lukewh (approved):
:+1: LGTM - I assume some commented out parts are going to make a reappearance in upcoming PRs, so shouldn't be deleted completely?

Lukewh:
@jkfran can you QA this - I don't have access to the core snap - thanks

static/js/publisher/release/components/releasesTable/cellViews.js
steverydz: Yeah that's right. We will need it in the next couple of tasks so didn't want to get rid of it entirely

PR #3954, opened 2022-04-12
Add progressive release tooltip

steverydz:
## Done
Added progressive release tooltip to releases table

## QA
- Go to https://snapcraft-io-3954.demos.haus/steve-test-snap/releases
- Check that there is a tooltip showing the progressive release status

## Issue
Fixes https://github.com/canonical-web-and-design/marketplace-tribe/issues/2332

### DIFF
--- a/static/js/publisher/release/components/releasesTable/cellViews.js
+++ b/static/js/publisher/release/components/releasesTable/cellViews.js
@@ -1,4 +1,3 @@
-/* eslint-disable */
 import React, { Fragment } from "react";
 import PropTypes from "prop-types";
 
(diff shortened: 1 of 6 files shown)


anasereijo:
> @anasereijo There is a scenario where a progressive release can happen without a previous revision. In the screenshot you can see this as `null → 19`. In the CLI it shows as `? → 19`. Do you have a preference as to what we do here?
> 
> progressive_releases_tooltip

At the first sight I would think '?' conveys data missing or some sort of error, but if it's what we use in the CLI we should match it to keep consistency.

@Roadmaster do you have any thoughts on this?

Lukewh:
I also think we should put something in the tooltip for unkown / none:
Revision:                    None / Unknown     19 (progressive)
Release Progress:     100% -> 77%          0% -> 23%
Version:                      None / Unknown     100
Confinement:              None / Unknown     strict

WDYT @anasereijo ?

anasereijo:
> I also think we should put something in the tooltip for unkown / none: Revision: None / Unknown 19 (progressive) Release Progress: 100% -> 77% 0% -> 23% Version: None / Unknown 100 Confinement: None / Unknown strict
> 
> WDYT @anasereijo ?

Sounds good to me! UX +1 :D

PR #3963, opened 2022-04-19
Use ID from URL in brand store settings

### LINKED ISSUE #3959: New "stores" view can show the wrong store ID
### Expected behaviour

When looking at a store's settings, the store id field should always show the ID for the store currently in focus in the stores list.

1. Go to e.g.  https://snapcraft.io/admin/<A_STORE_ID>/settings
2. Confirm the store's ID is the one for the store you are actually looking for (the ID is hidden, click on the "eye" icon to reveal it).
3. Click on a different store on the left-side 
4. The ID sometimes isn't that of the store you had actually selected.

I showed this behavior to Steve on a screensharing session. What we think might be affecting:

- If the page is reloaded, this gets "fixed" - this suggests it's something with the state in the frontend code, API responses (for e.g. a fresh load) are OK.
- It only happens when switching between stores that don't need a reload, meaning stores where you have "view" and/or "admin" permission. e.g. for stores where I have "access", the settings page is not accessible unless I type it directly in the address bar, which implies a full page reload, so I don't see the behavior.

### Specs

- _URL:_ https://snapcraft.io/admin/ANY_STORE_YOU_HAVE_ACCESS_TO/settings
- _Operating system:_ Ubuntu 20.04
- _Browser:_ Firefox somethingsomething but it's probably not browser-dependent.

This was first reported in a [private Launchpad bug](https://pad.lv/1968370/)

steverydz:
## Done
In the brand store settings form, use the store ID from the URL to prevent the wrong ID being used when state has changed

## QA
- Go to https://snapcraft-io-3963.demos.haus/admin
- Go to a number of different stores settings pages e.g. /admin/<store-id>/settings
- Check that the ID in the "Store ID" field always matches the one in the URL

## Issue
Fixes #3959

### DIFF
--- a/static/js/brand-store/Settings/Settings.test.tsx
+++ b/static/js/brand-store/Settings/Settings.test.tsx
@@ -10,7 +10,7 @@ import { RootState } from "./Settings";
 
 jest.mock("react-router-dom", () => ({
   ...jest.requireActual("react-router-dom"),
-  useParams: jest.fn().mockReturnValue({ id: "test" }),
+  useParams: jest.fn().mockReturnValue({ id: "test-id" }),
 }));
 
 function getInitialState(): RootState {
--- a/static/js/brand-store/Settings/Settings.tsx
+++ b/static/js/brand-store/Settings/Settings.tsx
@@ -181,7 +181,7 @@ function Settings() {
                   />
 
                   <PasswordToggle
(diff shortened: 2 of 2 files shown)


Lukewh (approved):
:+1: LGTM

PR #3964, opened 2022-04-19
Add progressive release column to history table

### LINKED ISSUE #3947: Build the release progress chart in the releases history table
See design: https://app.zeplin.io/project/61b3231e21194c4a48005226/screen/61b34d62712c9d4b57a6a3fd

steverydz:
## Done
Added the progressive releases column with chart to the releases history table

## QA
- Go to https://snapcraft-io-3964.demos.haus/notion-snap/releases
- Expand the progressive release table row
- Check that there is a "Release progress" table column
- Check that there is a chart that accurately represents the progressive release

## Issue
Fixes #3947, https://github.com/canonical-web-and-design/marketplace-tribe/issues/2333


jkfran (approved):
LGTM

Lukewh (approved):
:+1: LGTM

anasereijo:
This is more a visual thing but I wonder if we can adjust the table a bit better when we only have 2 cols. We have a [design for 3 cols](https://app.zeplin.io/project/61b3231e21194c4a48005226/screen/61b323cbd636e63ec9684f97) but I think we can also make it better for 2. @spencerbygraves any thoughts?

**how it looks like**




**design for 3 cols**

spencerbygraves:
@anasereijo @steverydz the text should all be aligned to the column header, with icons to the left. You can use the 3 col example above as a guide. Also, there doesn't seem to be a need to wrap "Tracking latest/stable".

Thanks

anasereijo:
- [ ] tooltip looks good but the build is missing?

how it looks like:


how it should look like:


- [ ] should the channel mention 'multiple versions' when we have a progressive release? see below

how it looks like:


how design looks like:


- [ ] expanded cell design

the table looks a bit different from the [design proposal](https://app.zeplin.io/project/61b3231e21194c4a48005226/screen/61b34d62712c9d4b57a6a3fd). could you please have another look?

how it looks like:



how it should look like:


- [ ] expanded cell missing radio buttons

- [ ] there's only 1 graph displayed - for the progressive. we should display the other one as well

how it looks like:


how it should look like:


- [ ] revision legend for the progressive release is missing

how it looks like:


how it should look like:

steverydz:
@spencerbygraves I've addressed your design feedback

steverydz:
@anasereijo 

- For the tooltip and the table, build will show when there is a build. This snap doesn't have any builds. Should we show an empty column in that case?
- To show the "Multiple versions" text I need to find a way to determine when a channel contains progressive releases so I've made [a separate issue](https://github.com/canonical-web-and-design/snapcraft.io/issues/3972) for that.
- The radio buttons only show in the revisions table as they can be changed. The releases history table is just a list of previous releases.
- The data isn't there to show the chart for the previous revision although we might be able to figure it out somehow. I've made [a separate issue](https://github.com/canonical-web-and-design/snapcraft.io/issues/3973) for that.
- Other than the above I'm not sure what you mean that there are differences from the design? Can you be more specific?

anasereijo:
> * Other than the above I'm not sure what you mean that there are differences from the design? Can you be more specific?

the cols spacing is a bit off. for instance, the graph is too close to the next col is there is a lot of space available around


thanks for checking the other stuff 👍

anasereijo:
so much better, thanks @steverydz

PR #3968, opened 2022-04-20
Fix release table row hover state

### LINKED ISSUE #3898: Release UI: Table rows remain grey after hovered
When hovering over table rows in the release UI they are highlighted with light grey. In some cases this highlight remains even when the row is no longer hovered.

steverydz:
## Done
Fixed the hover state on the release table

## QA
- Go to https://snapcraft-io-3968.demos.haus/danieltest-snap/releases
- Hover over the "Channel" cell and check that the whole row is highlighted
- Hover over any of the other cells and only that cell should be highlighted

## Issue
Fixes #3898


jkfran (approved):

Lukewh (approved):
:+1: LGTM

PR #3980, opened 2022-04-26
Add security.txt

jkfran:
## Done
- Added security.txt. The expiration date is a mandatory field so put 2024 to match with the next LTS 🤷🏽‍♂️ 

## QA

Check https://snapcraft-io-3980.demos.haus/.well-known/security.txt


Lukewh (approved):
:+1: LGTM

Issue #4045, opened 2022-06-24
Add tests for date time conversion


Lukewh:
General testing that the output is as expected, but specifically for daylight saving time etc. ensure that the date shows correctly during these cross-over periods.

@jkfran can you help Kevin with getting Python tests sorted?

Follow up to https://github.com/canonical-web-and-design/snapcraft.io/issues/3543

jkfran:
Hi @knotuteye , 

I think we can add the tests here:
https://github.com/canonical-web-and-design/snapcraft.io/blob/9d390386f668c330a194d1cf98f1b2671cc89a3a/tests/store/tests_public_logic.py#L317-L327

The test should fail before your PR https://github.com/canonical-web-and-design/snapcraft.io/pull/4046 and pass now after this change.

PR #4046, opened 2022-06-24
Published field inconsistency - yesterday vs date #3543

### LINKED ISSUE #3543: Published field inconsistency - yesterday vs date
### Expected behaviour

The "Published" field of a snap (snapcraft.io/glimpse-editor for example) should either say "Yesterday" or the date which in this case is May 11. For some reason I see both.

Go to snapcraft.io/glimpse-editor and expand the released drop-down menu and see that "yesterday" is listed for the beta channel and "May 11" for the edge channel (see attached screenshot). Yesterday is May 11 and both were built yesterday.

yesterday-is-may-11

knotuteye:
## Done
Upon investigation, this issue would only occur with a timezone-specific "disturbance" (like daylight savings). The fix was to change the datetime objects used in the convert_date function from timezone-naive ones to timezone-aware ones. 

## How to QA
- The best way to test this would be with automated testing as described in #4045

## Issue / Card
Fixes #3543 

## Screenshots
No UI changes


steverydz:
The demo still shows both yesterdays date and the word "Yesterday" as it does on production:
- https://snapcraft-io-4046.demos.haus/deno
- https://snapcraft.io/deno

Screenshot 2022-06-24 at 13 18 53

steverydz (approved):
LGTM 👍

hellsworth:
Thanks for fixing this bug!

PR #4056, opened 2022-06-30
Refactor some method calls for Release UI performance

Lukewh:
## Done
- Refactor code that looped through every single revision (possibly 20k) for each cell that was rendered
- Move architecture calculations from each cell render to immediately after data is fetched - and store them in redux.

## How to QA
- Visit https://snapcraft.io/juju/releases - wait for the page to load
- Visit demo/juju/releases - see how much quicker it loads
- Click around the UI on both and notice the demo is quicker

## Issue / Card
Fixes #

## Screenshots
Before:
image

After:
image


jkfran (approved):
LGTM

steverydz (approved):
LGTM 👍

PR #4097, opened 2022-09-07
Fix image upload aspect ratio bug

### LINKED ISSUE #4095: Icon upload shows error message but uploads correctly
### Expected behaviour

If the icon is valid and uploads, don't show an error.

See: https://chat.canonical.com/canonical/pl/eiogxseicigcbpbcce8twrk3se

> Icon fails saying it needs to be 1:1 ratio but it already is

image

> so even though that apparently failed, after adding an image to the listing it is now showing the icon

### Specs

- _URL:_ https://snapcraft.io/amberol/listing
- _Operating system:_
- _Browser:_

/cc @alexmurray

steverydz:
## Done
Fixed a bug uploading a snap icon on the listing page

## How to QA
- Go to https://snapcraft-io-4097.demos.haus/SNAP_NAME/listing
- Upload a square icon
- It should work and not show an error message

## Issue / Card
Fixes #4095

### DIFF
--- a/static/js/publisher/listing/components/ImageUpload/ImageUpload.tsx
+++ b/static/js/publisher/listing/components/ImageUpload/ImageUpload.tsx
@@ -73,8 +73,8 @@ function ImageUpload({
     renderedImage.onload = () => {
       if (
         !validateAspectRatio(renderedImage.width, renderedImage.height, {
-          width: 3,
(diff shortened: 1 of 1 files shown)


Lukewh (approved):
:+1: LGTM

PR #4099, opened 2022-09-08
Update mistune

jkfran:
## Done
Upgrade mistune library keeping the same customization, except for supporting "•" to start bullet lists, because with the latest version we need to override several library pieces. In my personal opinion, it is better this way in case someone wants to skip the regular bullet list using this symbol and having more control.

## How to QA
Check https://snapcraft-io-4099.demos.haus/fran-snap

## Issue / Card
Fixes https://github.com/canonical/marketplace-tribe/issues/2828


steverydz (approved):
LGTM 👍

PR #4102, opened 2022-09-12
Add a caption to the snaps table

### LINKED ISSUE #4094: Add captions to brand store tables
For accessibility reasons tables should have captions. A caption is a description or a title explaining to the user what the purpose of the table is. In the case of the brand store this will not need to be visible to sighted users.

steverydz:
## Done
Added a caption element to the snaps table

## How to QA
- Go to https://snapcraft-io-4102.demos.haus/admin
- Check that there is a `caption` element as the first child of the snaps table
- Check that the `caption` element is not visible on the screen

## Issue / Card
Fixes #4094

### DIFF
--- a/static/js/brand-store/Snaps/SnapsTable.js
+++ b/static/js/brand-store/Snaps/SnapsTable.js
@@ -27,7 +27,7 @@ function SnapsTable({
   return (
     <table className="p-table--mobile-card u-no-margin--bottom">
       <caption className="u-screenreader-only">
-        The snap which are published in and available to the {storeName} store.
+        The snaps which are published in and available to the {storeName} store.
       </caption>
       <thead>
         <tr>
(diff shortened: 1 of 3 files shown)


static/js/brand-store/Snaps/SnapsTable.js
```diff
@@ -26,6 +26,9 @@ function SnapsTable({
 
   return (
     <table className="p-table--mobile-card u-no-margin--bottom">
+      <caption className="u-screenreader-only">
+        The snap which are published in and available to the {storeName} store.
```
Lukewh, line 30:
```suggestion
        The snaps which are published in and available to the {storeName} store.
```

Lukewh (approved):
:+1: LGTM, just one change

PR #4125, opened 2022-09-26
Ensure we're counting *all* published snaps

Lukewh:
## Done
- Use counts of both the list of snaps from the API and the hardcoded list of featured snaps from the yaml file.

## How to QA
- Visit the https://snapcraft-io-4125.demos.haus/publisher/kde - the snap count should not be 1

## Issue / Card
Fixes #


steverydz (approved):
LGTM 👍

PR #4129, opened 2022-09-29
Show a no store page is user isn't in store

### LINKED ISSUE #4112: No page to show user that they do not have access to a store
### Expected behaviour
If a user does not have access to a store then they should be taken to a page which informs them of this and prompting them to contact their admin, rather than showing an empty store page

- Go to the URL of a store that you don't have access in or the store URL is wrong, e.g. https://snapcraft.io/admin/asdfasdfsaf/members
- See that you have an empty store page

steverydz:
## Done
Show a no stores page to a user if they are not in a store or it does not exist

## How to QA
- Go to https://snapcraft-io-4129.demos.haus/admin/madeupstore/snaps
- Check that you see a "Store not found" page
- Go to https://snapcraft-io-4129.demos.haus/admin/madeupstore/members
- Check that you see a "Store not found" page
- Go to https://snapcraft-io-4129.demos.haus/admin/madeupstore/settings
- Check that you see a "Store not found" page

## Issue / Card
Fixes #4112


Lukewh (approved):
:+1: LGTM

anasereijo:
This is looking good @steverydz ! Wondering if we do need the same message here https://snapcraft-io-4129.demos.haus/admin/madeupstore

steverydz:
@anasereijo I did have a go at that but it's not so straightforward because we don't know if the store exists or not on that route. I think it's an edge case as someone is most likely to be given a link directly to /snaps, /members or /settings

PR #4133, opened 2022-09-30
Remove flash messages from published snaps page

### LINKED ISSUE #4132: Flash message appearing multiple times on /snaps when it shouldn't show at all
This seems to happen after changes have been saved on a snap listing page. For each individual save there is another flash message.

Image Pasted at 2022-9-30 09-21

steverydz:
## Done
Removed flash messages from /snaps

## How to QA
- Go to https://snapcraft-io-4133.demos.haus/SNAP_NAME/listing
- Make some changes and save them
- Go to https://snapcraft-io-4133.demos.haus/snaps
- There should not be any "Changes have been saved" messages

## Issue / Card
Fixes #4132

### DIFF
--- a/templates/publisher/account-snaps.html
+++ b/templates/publisher/account-snaps.html
@@ -5,22 +5,6 @@ My published snaps — Linux software in the Snap Store
 {% endblock %}
 
 {% block content %}
-{% with messages = get_flashed_messages() %}
-{% if messages %}
-<section class="p-strip is-shallow">
-  {% for message in messages %}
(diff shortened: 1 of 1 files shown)


Lukewh (approved):
:+1: LGTM

jkfran:
This is not the cause of the issue, the problem is that we are setting these messages somewhere and not consumed, these messages will be revealed in any other part. So the fix should be to either consume them in the right place or stop setting them.

PR #4149, opened 2022-10-07
Remove check as the API will handle it

jkfran:
## Done

There is no need to check if the user is the owner or has permission because the API will do it.


steverydz (approved):
LGTM 👍

jkfran:
Ah, Python checks..., let's wait until Daniel gets back to us

PR #4169, opened 2022-12-05
WD-720 display correct store on the dispute page

### LINKED ISSUE #4151: Trying to file a dispute on a non-global store "forgets" the target store, gives potentially confusing error
### Expected behaviour

When filing a name dispute on a store other than Global, it should work and request the name in the targeted store.

1. https://snapcraft.io/register-snap
2.  Select a non-global store, and choose a name that is reserved in that store (in our case it was because the target store requires a prefix for name auto-approval. Let me know if you need concrete examples).
2. The next page is `https://snapcraft.io/register-snap?snap_name=<THE_NAME>&is_private=False&store=<THE_STORE_ID>&reserved=True`
3. There is a "request a reserved name" link at the top. Click on that, so we can file a dispute and claim the name anyway.
4. The next page is `https://snapcraft.io/request-reserved-name?snap_name=<THE_NAME>&store=THE_STORE_ID&is_private=False`. Note the URL does contain the target store ID. Note that the page itself says "Store: Global" - so the `store` query parameter is apparently ignored.
5. Type anything in the comment field and click on "yes I am sure"
6. Redirected to `https://snapcraft.io/register-name-dispute`. The dispute is not actually recorded; instead a "The name <THE_NAME> is not available for disputing."

That response is sent when the snap name *is* available and free to be registering, indeed meaning it cannot be disputed - instead the user can simply just register **in the Global store**. But recall the original intent was to register it in a brand store in which this particular name is reserved for some reason.

Looking at the code, templates/publisher/request-reserved-name.html hard-codes the "Global" store name, and the `get_request_reserved_name` view does not seem to get the "store" name from the query and put it in the context for the template. It seems that the functionality to file disputes in brand stores was never implemented.

### Specs

- _URL:_ per above
- _Operating system:_ Ubuntu 20.04
- _Browser:_ Firefox

codeEmpress1:
## Done
- Fix wrong store displayed on dispute page
- It shows the store selected by user now
## How to QA
- Go to /register-snap
- Select a non-global store
- Select a reserved snap name
- Submit, this should display the error containing the link to file a dispute
- Click on `file a dispute`
- The dispute page should show the correct store selected by the user
## Issue / Card
Fixes #4151 
Fixes https://warthogs.atlassian.net/browse/WD-720

### Before
image

### After
image


steverydz (approved):
LGTM 👍

Lukewh (approved):
:+1: LGTM

Issue #3838, opened 2022-01-07
Login to Developer Account gives redirect loop


JordanMG13:
A customer reported a problem being able to login to their Developer Account page.

### Expected behavior
- Click on Developer Account
- Login
- After successful authentication, be brought to https://snapcraft.io/snaps

### Steps to reproduce the problem
- Have an account with the customer
- Go to https://snapcraft.io
- Click on Developer Account
- Get an infinite redirect login loop

### Attempted fixes
- Opening browser in private window to see if issue was cookie/cache related (Result: same problem as above)

### Specs

- _URL:_ https://snapcraft.io
- _Operating system:_ Ubuntu 16.04
- _Browser:_Firefox and Safari

jkfran:
Hi @JordanMG13 

Can you confirm if the customer is still having this issue? We can't reproduce it

JordanMG13:
@jkfran Checking in with them now.

jkfran:
Hi @JordanMG13, did they get back to you?

jkfran:
There is a similar bug reported recently for Ubuntu SSO:
https://bugs.launchpad.net/canonical-identity-provider/+bug/1895734/

I'll try to keep you posted.

JordanMG13:
@jkfran They have not yet, I can ping them again. And thanks for the info!

JordanMG13:
@jkfran To confirm, they are still experiencing this issue.

jkfran:
@maxiberta, do you know when [your fix](https://code.launchpad.net/~maxiberta/canonical-identity-provider/+git/canonical-identity-provider/+merge/414316) is going to land on production?

maxiberta:
@jkfran we're planning to do it today.

maxiberta:
The Ubuntu SSO update is now in production.
Please delete all related cookies so that they're updated correctly.

JordanMG13:
I will tell them to try this, thanks!

JordanMG13:
It looks like the customer is still having this issue, even after wiping their cookies.

JordanMG13:
@jkfran Will do!

jameinel:
I also got the redirect from one of my browsers running Firefox 96.0.1. With 96.0.2 it doesn't redirect and does login. login-beta did work and didn't have the strange redirect that you see with login. Both still took ~30s before it loaded a page, without any other indication that it was working.

jkfran:
[...]

PR #4178, opened 2022-12-14
fix(login): Clear all session data when login fails

Lukewh:
## Done
- Empty session data when use is not authenticated

Fixes: #3838


jkfran (approved):
lgtm

PR #4180, opened 2022-12-14
Wd 856 refactor try except

### LINKED ISSUE #3230: Refactor try/except in all the routes
Currently we have this code duplicated multiple times in the application for all the routes using the API

https://github.com/canonical-web-and-design/snapcraft.io/blob/1cf85f9f3da332d5ec55bcbce48626554ac22925/webapp/publisher/snaps/listing_views.py#L49-L57

I think we can improve it by using flask `errorhandler` decorators like we are doing on Charmhub:
https://github.com/canonical-web-and-design/charmhub.io/blob/master/webapp/handlers.py

codeEmpress1:
## Done
- Register `StoreAPI`, `API` and `PublisherAPI` error handlers
- Remove all try... except blocks whose exceptions have been handled in the error handlers
## How to QA
-  Visit https://snapcraft-io-4180.demos.haus/
- This PR does not add or update any feature, neither does it fix a bug, it is a chore that can simply be validated by ensuring all routes still works as expected
- Also, by triggering an error in the application and checking to ensure the app does not crash but rather catches and handles the error(s) appropriately
## Issue / Card
Fixes #3230

### DIFF
--- a/webapp/handlers.py
+++ b/webapp/handlers.py
@@ -24,8 +24,11 @@ from canonicalwebteam.store_api.exceptions import (
 
 from webapp.api.exceptions import (
     ApiError,
-    ApiTimeoutError,
+    ApiConnectionError,
     ApiResponseErrorList,
+    ApiTimeoutError,
+    ApiResponseDecodeError,
+    ApiResponseError,
 )
 
 from datetime import datetime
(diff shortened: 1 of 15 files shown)


steverydz (approved):
LGTM 👍

webapp/handlers.py
```diff
@@ -103,13 +122,99 @@ def internal_error(error):
...
 
         return (
-            flask.render_template("50X.html", error_name=error_name),
+            render_template("50X.html", error_name=error_name),
             return_code,
         )
 
-    @app.errorhandler(503)
-    def service_unavailable(error):
-        return flask.render_template("503.html"), 503
```
jkfran, line 112: There is no need to delete this one, this is to show a nice page whenever a 503 happens in the entire application.

Did you remove it for any reason in particular?

jkfran:
Well done, Alimot. I added a small comment. 

I guess fixing the python tests will be a bit painful now, but I also think this refactor shouldn't change the behaviour much, so I understand why they are failing.

This refactor will improve the project a lot.

webapp/handlers.py
codeEmpress1: I removed it because I though the same 503 exception has been handled with this. 
`    @app.errorhandler(StoreApiConnectionError)
      def connection_failed(error):
        return render_template("503.html"), 503`

jkfran: I see. It's a different exception, `StoreApiConnectionError` will be raised only for the store API module, whoever the `503` exception can happen by many different things. But we indeed want to show the same template for both exceptions, so one solution could be:

```python
    @app.errorhandler(503)
    @app.errorhandler(StoreApiConnectionError)
    def connection_failed(error):
        return render_template("503.html"), 503
```

codeEmpress1: Ok, Great! Thank you Fran. I will resolve that

jkfran (approved):
Well done!

PR #4240, opened 2023-04-18
Rebrand homepage

steverydz:
## Issue / Card
Fixes #


bartaz:
I'm not sure if new grey "#f2f2f3" and old "p-strip--light" are supposed to be mixed on the same page, especially next to each other:

image

bartaz:
And as far as I understand new background is supposed to be `#f2f2f2`.

steverydz:
@bartaz Thanks. I've fixed the node headings. In regards to the backgrounds, I've based this on the design. The colour value (#f4f4f3) is from the background image on that section. I think @camillet16 or @lyubomir-popov might be best placed to comment on if this is correct?

camillet16:
@steverydz yeah could we work with the #f2f2f2 for the background grays so it would look seamless. Thank you!

steverydz:
@camillet16 I've just updated the strip colour. Is there as asset for the gradient/suru element that matches because the current one is lighter?

Screenshot 2023-04-19 at 14 01 30

steverydz:
Thanks @lyubomir-popov. Done

PR #4300, opened 2023-06-22
Move brand store components

steverydz:
## Done
- Moved all components to their own directory to separate them from types, slices, hooks etc.

## How to QA
- Go to https://snapcraft-io-4300.demos.haus/admin
- Everything should render as it does in production


chillkang (approved):
LGTM 👍

PR #4321, opened 2023-07-05
Add create new model view

steverydz:
## Done
- Added the "Create new model" view

## How to QA
- Go to https://snapcraft-io-4321.demos.haus/admin/<STORE_ID>/models
- Check that clicking the "Create new model" button opens the "Create new model" side panel
- Check that clicking the "Create new model" button appends "/create" to the URL
- Check that reloading the page with the side panel open keeps the panel open
- Check that the "Add model" button is disabled by default
- Check that the "Add model" button is enabled if the "Name" field has a value
- Check that there is an error message if the "Name" field value matches an existing model name
- Check that clicking the "Generate key" button populates the "API key" field with a 50 character value consisting of upper and lowercase letters and numbers
- Check that clicking cancel, or outside of the panel, closes it and when you reopen it the form is cleared

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-4736


Lukewh (approved):
:+1: LGTM

PR #4332, opened 2023-07-11
Add model

codeEmpress1:
## Done
- Add the following endpoints:
    - `get_model`: `[GET] /admin/store/<store_id>/models`
    - `create_model`: `[POST] /admin/store/<store_id>/models`
    - `update_model`:  `[PATCH] /admin/store/<store_id>/models/<model_name>`
 
## How to QA
- Login with `/login-beta`
- `get_model`: Go to https://snapcraft-io-4332.demos.haus/admin/store/njwQYXFnS7ppo7LaGxoh7aqVZc1CPi26/models and ensure that a response object containing `message, success, and data`  is returned
- `post_model`: the `post` endpoint can be tested by sending the request from the frontend client with payload `{"name": string, "api_key": string}`
- `update_model`: similar to the post endpoint should also be tested from the frontend client. payload: `{"api_key": string}`
## Issue / Card https://warthogs.atlassian.net/browse/WD-4657

# NOTE
THIS ENDPOINT ONLY WORKS WHEN USER AUTHENTICATE WITH LOGIN BETA FOR NOW.


webapp/admin/views.py
```diff
@@ -226,3 +230,133 @@ def update_invite_status(store_id):
...
+        store_id (int): The ID of the store for which to retrieve models.
+
+    Returns:
+        dict: A dictionary containing the response message, success status,
+        and data.
+    """
+    res = {}
+    try:
+        models = admin_api.get_store_models(flask.session, store_id)
+        res["message"] = "Models retrieved"
```
Lukewh, line 252: If the request is successful we don't need a message :)

Lukewh (changes requested):
1 small change plz

webapp/admin/views.py
codeEmpress1: For all requests? Or for the get requests only?

Lukewh: All successful requests

Lukewh (approved):
:+1: LGTM

PR #4349, opened 2023-07-24
Model service tests

steverydz:
## Done
- Added tests for the models list filter and the policies list filter
- Drive by: Removed unused variables in policies table

## How to QA
The `test-js` check should pass


chillkang (approved):
LGTM 👍

PR #4352, opened 2023-07-25
Model service hooks

steverydz:
## Done
Converted queries for the model service to custom hooks so logic doesn't need to be duplicated

## How to QA
- Go to https://snapcraft-io-4352.demos.haus/login-beta and login
- Go to https://snapcraft-io-4352.demos.haus/admin/njwQYXFnS7ppo7LaGxoh7aqVZc1CPi26/models
- Check that models data loads as expected
- Go to a model page and check that the data loads as expected
- Go to a model policies page and check that the data loads as expected


chillkang (approved):
LGTM 👍

PR #4382, opened 2023-08-16
Enable delete signing keys

steverydz:
## Done
Made sure that signing keys can be deleted and added some minor UX enhancements around loading tables

## How to QA
- Go to https://snapcraft-io-4382.demos.haus/admin/njwQYXFnS7ppo7LaGxoh7aqVZc1CPi26/signing-keys
- Check that you can delete a signing key if it has no models or policies

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-5592

### DIFF
--- a/static/js/brand-store/components/Model/Policies.test.tsx
+++ b/static/js/brand-store/components/Model/Policies.test.tsx
@@ -70,9 +70,4 @@ describe("Policies", () => {
       mockFilterQuery
     );
   });
-
-  it("displays a table of policies", () => {
-    renderComponent();
-    expect(screen.getByTestId("policies-table")).toBeInTheDocument();
-  });
 });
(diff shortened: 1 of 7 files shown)


Lukewh:
Hmm, I think it's fine but I wonder if there's a workaround. When I ctrl+clicked all of the items in the list and deleted the revisions in different tabs, when I tried to "deactivate" again, it showed the same message. I guess this is a limitation of recoil

image

Lukewh (approved):
:+1: LGTM

steverydz:
> Hmm, I think it's fine but I wonder if there's a workaround. When I ctrl+clicked all of the items in the list and deleted the revisions in different tabs, when I tried to "deactivate" again, it showed the same message. I guess this is a limitation of recoil

Maybe this would be a use case for `refetchOnWindowFocus` in React Query?

PR #4441, opened 2023-11-09
Update HTTP method for settings in brand store

steverydz:
## Done
Changed the HTTP method of settings updates in the brand store from `POST` to `PUT` as it is only updating existing data rather than creating new data

## How to QA
- Go to https://snapcraft-io-4441.demos.haus/admin/ahnuP3quahti9vis8aiw/settings
- Make a change and save it
- Check that the change is made successfully


Lukewh (approved):
:+1: LGTM

PR #4471, opened 2023-12-07
If call to account.json fails - show sign in button

Lukewh:
## Done
If the call to /account.json fails - show sign in button

## Issue / Card
Fixes #


steverydz (approved):
LGTM 👍

PR #4477, opened 2023-12-08
Remove last_login_method cookie. Extra logging for account endpoint

Lukewh:
## Done
- Removed `last_login_method` cookie on `/account.json` call as it's no longer used
- Removed all references to `last_login_method`
- Wrapped `get_account_json` in a `try: except` to attempt to log issues
- Return a 502 from `account.json` if there is an error and log it

## How to QA
- Visit the demo
- For most users everything should work as usual.
  - Logout
  - Login
  - Account menu should work
  - Going to `/snaps` and individual snap publisher pages should work

## Issue / Card
Fixes #

### DIFF
--- a/webapp/handlers.py
+++ b/webapp/handlers.py
@@ -206,13 +206,8 @@ def set_handlers(app):
             "macaroon-permission-required",
             "macaroon-authorization-required",
         ]:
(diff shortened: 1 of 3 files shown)


steverydz (approved):
LGTM 👍

PR #4481, opened 2023-12-12
Fix settings page error messages

Lukewh:
## Done
- Updated settings page to work the same as listings page.
- Make error messages a little more friendly on both
- Drive-by: Updated eslintrc to remove linting warning

## How to QA
- Visit the settings page for a snap without a description and or summary
- Try and update visibility
- ERRORS 🥳 
- The changes made should not be reverted on the page

## Issue / Card
Fixes #


steverydz (approved):
LGTM 👍

PR #4486, opened 2023-12-13
Add show more to beta store categories

steverydz:
## Done
Added "See more" categories

## How to QA
- Go to https://snapcraft-io-4486.demos.haus/beta-store
- Check that only 10 categories (inc. featured) are visible
- Check that there is a "See more" button that shows the rest of the categories


chillkang (approved):
LGTM👍

PR #4499, opened 2024-01-09
Fix search pagination

steverydz:
## Done
Fixed the pagination for beta store searches

## How to QA
- Go to https://snapcraft-io-4499.demos.haus/beta-store?categories=games&page=4
- Search for "code"
- Check that the `page` query parameter has been removed
- Check that the results count above the search results is correct


Lukewh (approved):
:+1: LGTM

PR #4507, opened 2024-01-18
update ros2 first snap with core22 humble snap

giusebar:
## Done
- Update ros2 first snap example with core22 humble snap
- Fix typo in ros2 first snap test instructions


steverydz (approved):
LGTM 👍

PR #4524, opened 2024-02-06
fix: remove extra list on search results

Lukewh:
## Done
- Removed extra ul below the card when searching for a snap

## How to QA
- Visit a brand store admin
- Click "Include snap"
- Search and click on a snap
- Only see the result once

## Screenshots
Before:
image

After:
image


steverydz (approved):
LGTM 👍

PR #4528, opened 2024-02-13
fix: markdownit initialisation update

Lukewh:
## Done
Updated how markdownit is imported and initialised.

## How to QA
- Visit any snaps build page / preview
- Everything should work as expected
- Compare with the live site.


chillkang (approved):
LGTM👍

PR #4550, opened 2024-03-05
Application layout

steverydz:
## Done
Applied the new application layout design to the brand store

## How to QA
- Go to https://snapcraft-io-4550.demos.haus/admin

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-8954


steverydz:
@AnnSoMuller This is ready for a UX review. Bear in mind this is only for the brand store at the moment and other than the section titles I haven't made any changes to the main content (tables etc.) as that will come later.

Just two things:
1. For the header/side navigation I've used the pin icon for small screens rather than the arrow icon in the design. This is because that's what this action actually does and is recommended in the Vanilla application layout docs. I can change it if you wish.
2. Can we make the stores selector in the side navigation full width? Without an icon I think it looks a bit off and it could do with the extra space.

steverydz:
@lyubomir-popov Me and @AnnSoMuller were talking about the application layout for the brand store and were thinking about making the store selector full width as it could do with the extra space given the length of some of the store names. I think we should align the "My stores" title with it if we go that way.

What did you think?

I've attached a screenshot in case you don't have access to any stores:

Screenshot 2024-03-05 at 14 40 50

AnnSoMuller:
Good on the UX.
I would like @lyubomir-popov advice for the icons choices as It would make sense to have different ones for "My stores" and "Store snaps" to avoid confusion when the navigation is collapsed.

Lukewh (approved):
:+1: LGTM

PR #4567, opened 2024-03-27
Add publisher data endpoint

steverydz:
## Done
Added an endpoint for publisher data

## How to QA
- Go to https://snapcraft-io-4567.demos.haus/account/publisher
- Check that you can see publisher data which includes `has_stores` and `subscriptions`
- Go to https://snapcraft-io-4567.demos.haus/account/details
- Check that it works as expected ([compare to live](https://snapcraft.io/account/details))

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-9991

### DIFF
--- a/webapp/publisher/views.py
+++ b/webapp/publisher/views.py
@@ -7,6 +7,7 @@ from canonicalwebteam.store_api.exceptions import StoreApiResponseErrorList
 # Local
 import webapp.api.marketo as marketo_api
 from webapp.helpers import api_publisher_session
+from webapp import helpers
 from webapp.decorators import login_required
 
 account = flask.Blueprint(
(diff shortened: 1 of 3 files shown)


webapp/publisher/views.py
```diff
@@ -64,6 +64,48 @@ def get_account_details():
     return flask.render_template("publisher/account-details.html", **context)
 
 
+@account.route("/publisher", methods=["GET"])
+@login_required
+def get_publisher_details():
```
Lukewh, line 69: this seems to be the same function body as `get_account_details` above (apart from the response. Can we reuse the method? Maybe put it in a separate function that both `get_account_details` and `get_publisher_details` both call? Alternatively the endpoint could be `/details.json` with the `.json` signifying to return a json response instead of the html template?

steverydz: @Lukewh We don't have a precedent for using `.json` for API endpoints, other than `account.json` which seems to only be as a workaround too. I agree about the reusing logic though.

Lukewh (approved):
:+1: LGTM

PR #4603, opened 2024-04-23
Show notification and set private to true

codeEmpress1:
## Done
- show notification to inform user that registering a snap requires manual review when a user selects global store (notification is hidden for other stores)
- set the default private value to true when global store is selected
- Disable the privacy buttons when global store is selected
## How to QA
- Go to https://snapcraft-io-4603.demos.haus/register-snap
- Select the global store
- Check that the notification displays
- Check that the privacy options are disabled
- Select any other store and check that notification is not displayed and you can select any privacy option of your choice
- Also, QA with an account that has access to only global store.
## Issue / Card 
https://warthogs.atlassian.net/browse/WD-10620
https://warthogs.atlassian.net/browse/WD-10618
Fixes #

### DIFF
--- a/templates/publisher/register-snap.html
+++ b/templates/publisher/register-snap.html
@@ -70,6 +70,16 @@ Register new Snap name
       {% endfor %}
     </div>
   {% endif %}
+ 
+  <div class="u-fixed-width toggle-nofication">
+    <div class="p-notification--information">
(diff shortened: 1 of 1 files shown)


steverydz (approved):
LGTM 👍

PR #4607, opened 2024-04-24
hide privacy section when user has only global store

codeEmpress1:
## Done
- Hide privacy section when user has only global store
## How to QA
- Login to  https://snapcraft-io-4607.demos.haus with an  account that has access to only global store (you can use a new account)
- Go to https://snapcraft-io-4607.demos.haus/register-snap and check that the privacy section is hidden
## Issue / Card https://warthogs.atlassian.net/browse/WD-10619
Fixes #


Lukewh (approved):
:+1: LGTM

PR #4634, opened 2024-05-21
ci: Remove unused workflows

Lukewh:
## Done
- Remove docs-links - seems to be skipped every run so isn't useful
- Remove main - this runs **after** a PR is merged. We have the PR workflow that seems to do the same thing
- Remove run-percy - this succeeds, but we never check the results

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #


steverydz (approved):
LGTM 👍

PR #4715, opened 2024-06-25
fix: add missing return types and avoid using any in /static/js/public/snap-details

chillkang:
## Done
- Added missing return types and avoided using any type in /static/js/public/snap-details

## How to QA
- All tests should pass

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): no behavioral changes

## Issue / Card
Fixes #https://warthogs.atlassian.net/browse/WD-12598

### DIFF
--- a/static/js/public/snap-details/videos.ts
+++ b/static/js/public/snap-details/videos.ts
@@ -1,4 +1,8 @@
-function vimeo() {
+interface HolderElement {
+  querySelector: (selector: string) => HTMLIFrameElement | null;
+}
+
+function vimeo(): void {
   const vimeoPlayerScript = document.createElement("script");
   vimeoPlayerScript.src = "https://player.vimeo.com/api/player.js";
   const firstScript = document.getElementsByTagName("script")[0] as HTMLElement;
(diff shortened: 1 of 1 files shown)


static/js/public/snap-details/videos.ts
```diff
@@ -29,35 +35,34 @@ function vimeo() {
...
+  const iframe = holderEl.querySelector("iframe") as HTMLIFrameElement | null;
   if (iframe && iframe.src && iframe.src.indexOf("http://") !== -1) {
     iframe.src = iframe.src.replace("http://", "https://");
   }
 
   if (videoType === "vimeo") {
     vimeo();
   } else if (videoType === "asciinema") {
-    asciinema(holderEl);
+    asciinema(holderEl as HolderElement);
```
Lukewh, line 65: Can this just be a built-in `HTMLElement`? Or do we need to be more explicit with the return type?

chillkang: Yes, I agree. It will simplify the type handling and I don't think we need any additional type constraints!

Lukewh (approved):
:+1: LGTM

PR #4736, opened 2024-07-03
fix: Fix form state bug on listing page

steverydz:
## Done
Fixed a bug where the save and revert buttons weren't disabled once a change had been reverted

## How to QA
- Go to https://snapcraft-io-4736.demos.haus/<snap_name>/listing
- Check that you can change fields, change them back to their original value and the "Save" and "Revert" buttons go back to being disabled
  - It is known that it doesn't work for the licences field due to its complexity

## Testing
- [x] This PR has tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-12782

### DIFF
--- a/static/js/publisher/listing/components/App/App.tsx
+++ b/static/js/publisher/listing/components/App/App.tsx
@@ -7,6 +7,7 @@ import {
   getFormData,
   getListingData,
   shouldShowUpdateMetadataWarning,
+  getDefaultValues,
 } from "../../utils";
 import { initListingTour } from "../../../tour";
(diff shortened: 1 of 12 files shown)


abbiesims (approved):
LGTM :+1:

abbiesims (changes requested):
When I manually delete text from a field, the Revert and Save buttons are still enabled. It is only when I click Revert that the buttons are disabled again

abbiesims (approved):
Looks good :+1:

PR #4823, opened 2024-08-28
fix: self serve track creation optional params

M7mdisk:
## Done
- Self serve track creation was giving 400 error when the optional parameters were not provided. This ensures optional params are optional
 
## How to QA
- make sure you can create tracks (with and without) optional parameters, and that they're being passed correctly

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): bug fix

## Issue / Card
Fixes WD-14532


abbiesims (approved):
LGTM :+1:

PR #4834, opened 2024-09-03
WIP: chore: Move publicise page into publisher app

steverydz:
## How to QA
- Go to https://snapcraft-io-4834.demos.haus/steve-test-snap/publicise

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #


steverydz:
@AnnSoMuller This is part of the work that will enable the publisher pages to use the application layout, then eventually lead to us being able to include brand store data.

In this case I've just moved the "Publicise" page. There's a few things I wanted to draw attention to and see what you think:
1. The publisher pages navigation can only be shown in the sidebar if the user is on a snap page, e.g. links to listing, builds etc. wouldn't work from `/validation-sets` because no snap is specified
2. I've changed the title of the publicise pages to just "Promote your snap" as the tabbed navigation should indicate where the user is
3. I have used the tabbed navigation pattern for the sections within "Publicise" as that seems to fit better

You can see this here: https://snapcraft-io-4834.demos.haus/steve-test-snap/publicise

Are you happy to keep going in this direction or have any comments?

steverydz:
@anasereijo I've made two new versions:

1. Just the changes we discussed https://snapcraft-io-4834.demos.haus/steve-test-snap/publicise
2. A mix of the application layout and the old publisher page (I think this is better): https://snapcraft-io-4838.demos.haus/steve-test-snap/publicise

PR #4921, opened 2024-12-04
fix: Fix code linting errors in publisher app

steverydz:
## Done
Fixed TypeScript and lint errors in the publisher pages.

**Note**:
- The releases section is ignored because it is legacy code and large structural changes will be required to pass these checks which is dangerous to do in the scope of this project
- Some parts around where React Query and React Form interact has been ignored as there seems to be type clashes between the two
- The lint check will still fail as there are a11y issues to be addressed which will happen in a separate PR

## How to QA
The lint errors should be related to TypeScript


chillkang (approved):
LGTM👍

PR #4925, opened 2024-12-05
fix: Fix publisher a11y errors in publicise buttons section

steverydz:
## Done
Converted pseudo labels to paragraphs

## How to QA
- Go to https://snapcraft-io-4925.demos.haus/<SNAP_NAME>/publicise
- Check that "Language", "HTML" and "Markdown" still have titles
- **Note**: `lint-js` will still fail as there are errors in other files that need to be addressed separately


chillkang (approved):
LGTM👍

PR #4926, opened 2024-12-05
fix: Fix publisher a11y errors in settings section

steverydz:
## Done
- Converted pseudo labels to paragraphs
- Fixed unsafe optional chaining

## How to QA
- Go to https://snapcraft-io-4926.demos.haus/<SNAP_NAME>/settings
- Check that "Collaboration", "Store" and "Status" rows all have titles
- Check that the "Status" row has a value
- **Note**: `lint-js` will still fail as there are errors in other files that still need to be addressed


chillkang (approved):
LGTM👍

PR #4972, opened 2025-01-13
chore: Refactor brand store state management

steverydz:
## Done
Refactored the brand store state management. The rationale is to keep relevant state together, for example all model atoms and selectors in one file rather than separated as they interact with each other, it keeps concerns in one place, and the components don't need to make the distinction between atoms and selectors.

## How to QA
- Go to https://snapcraft-io-4972.demos.haus/admin
- Click around the brand stores UI and check that it works as expected

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-18026


ilayda-cp:
- In the [global store](https://snapcraft-io-4972.demos.haus/admin/ubuntu/snaps)  brand endpoint returns 404.  
- in the Fully included stores section there is an empty element
Screenshot from 2025-01-14 12-14-20
- looks like there is a small cut off in the select all checkbox and maybe buttons need more margin but idk if this is related with this PR.
Screenshot from 2025-01-14 12-17-32

ilayda-cp:
ok i realized that none of those things related with your PR. They've already on main.

ilayda-cp (approved):

PR #5040, opened 2025-02-26
chore: Update design of published snaps page

steverydz:
## Done
Update design of published snaps page

## How to QA
- Go to https://snapcraft-io-5040.demos.haus/snaps
- Check that each section is within an accordion [as per design](https://app.zeplin.io/project/659574cda0730e724877bf7d/screen/65ccf576441ff5de65c8de7c)

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-19670

### DIFF
--- a/static/js/publisher/components/PublishedSnapList/PublishedSnapList.tsx
+++ b/static/js/publisher/components/PublishedSnapList/PublishedSnapList.tsx
@@ -79,14 +79,8 @@ function PublishedSnapList({
 
   return (
     <Strip element="section" shallow>
-      <div className="u-fixed-width u-clearfix">
+      <div className="u-fixed-width">
         <h2 className="p-heading--4 u-float-left">My published snaps</h2>
-        <a
-          href="/account/register-snap"
-          className="p-button u-float-right p-snap-list__register u-no-margin--top"
-        >
-          Register a snap name
-        </a>
       </div>
       {isEmpty && <EmptySnapList />}
 
--- a/static/js/publisher/components/PublishedSnapSection/PublishedSnapSection.tsx
+++ b/static/js/publisher/components/PublishedSnapSection/PublishedSnapSection.tsx
@@ -1,4 +1,5 @@
 import { useState } from "react";
+import { Accordion } from "@canonical/react-components";
 import { ISnap } from "../../types";
 import PublishedSnapList from "../PublishedSnapList";
 import PublisherMetrics from "../PublisherMetrics";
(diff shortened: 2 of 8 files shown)


abbiesims:
Can you add me to the project on Zeplin? I can't see the design

steverydz:
> Can you add me to the project on Zeplin? I can't see the design

@AnnSoMuller Could you help Abbie with this? ^

AnnSoMuller:
> Can you add me to the project on Zeplin? I can't see the design
@abbiesims 
can you try https://zpl.io/DpG18Qr
(it seems you are already part of the projecy)

AnnSoMuller:
1. I see some layout/header jumps when I click on "snap installs" when something opened and on a listing of a snap.
2. I'm seing the same jump btw when on a listing page of a snap and clicking back and forth "listing" and any other tab

bartaz:
> When I close "Snap Installs" and "My Published Snaps" accordions, I can't click on "Unregister" or "Publish to this name" for the top two snaps in the Register snap names section. Looks like the Snap installs is still being rendered under the hood somehow

Can't reproduce it. This seems to work fine for me (checked Chrome and Firefox on macOS)

steverydz:
@abbiesims I was able to reproduce the issue you raised and have fixed it

@bartaz The background colour for the metrics chart was added in this design. It was previously white. Should we revert back to white @AnnSoMuller?

The spacing has been overridden for these accordions (and those in /admin) because the designs use a modified version of the Vanilla accordion pattern. Could we perhaps just use the Vanilla pattern?

@AnnSoMuller I don't understand what you mean about the layout and header jumps? Can you be more specific?

AnnSoMuller:
https://github.com/user-attachments/assets/71844337-957c-4cea-931c-420747fb719b

abbiesims (approved):
LGTM code-wise! :+1:

bartaz:
The "my published snaps" title and the button seem duplicated now with accordion.

image


It seems design only has accordion one.

bartaz:
> @steverydz if you look at the top headers "My snaps/...)" 😅

The "jumping" (left/right shift) seems to happen when you have wide enough screen and when you switch between a view with scrollbar to a view without scrollbar.

Although not sure why it happens to you @AnnSoMuller on a mac, which should have scrollbars invisible by default and not affecting layout - unless you changed that.

Anyway, maybe that's a bug that could be handled separately, as it doesn't seem to be caused by this PR but the centred content in the application layout in general.

steverydz:
> The "my published snaps" title and the button seem duplicated now with accordion.

@bartaz Good catch - fixed

AnnSoMuller:
@bartaz, I didn't change the default behaviour (or didn't remember that ^^) but yes it is when the scrollbar appear or disappear, only on big resolution. Don't have the "bug" on small resolution.
And yes, discussed with @steverydz it can be on another PR :)

PR #5077, opened 2025-04-03
chore: Prepare types for React 19 upgrade

steverydz:
## Done
Update types to prepare for the upgrade to React 19

## How to QA
- All checks should pass

## Testing
- [ ] This PR has tests


bartaz:
There still seems to be `ReactNode` used in `useActiveDeviceMetrics.test.tsx` - not sure if it was intended omission?

bartaz (approved):
LGTM

steverydz:
> There still seems to be `ReactNode` used in `useActiveDeviceMetrics.test.tsx` - not sure if it was intended omission?

Yes, `ReactNode` is needed in that particular file due to the way the `createWrapper` function works. It shouldn't cause any problems with React 19 as far as I can tell.

PR #5090, opened 2025-04-08
chore: Update security.txt for SSDLC

steverydz:
## Done
Updates security.txt

### DIFF
--- a/security.md
+++ b/security.md
@@ -1,0 +1,7 @@
+Contact: mailto:security@ubuntu.com
+Expires: 2025-12-31T00:00:00.000Z
+Preferred-Languages: en
+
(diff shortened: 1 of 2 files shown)


security.md
```diff
@@ -0,0 +1,7 @@
+Contact: mailto:security@ubuntu.com
+Expires: 2025-08-01T00:00:00.000Z
```
bartaz, line 2: Not sure how is this date defined, but it's not too far into the future. Can we move it further ahead?

steverydz: I've extended it to the end of the year. It will be reviewed by the security team too so if it needs changing beyond that it shouldn't be an issue.

PR #5100, opened 2025-04-16
feat(CVEs): Adds the API endpoint to check CVEs availability for given snap

bartaz:
## Done

Adds an API endpoint to check if given snap has CVE data available.
If the user is allowed to see the CVE data (is Canonical publisher and has access to said snap), and there is CVE data available in the snap-cves repo, it should return true, otherwise it should error.

## How to QA

- sign in
- assuming you have access to docker snap (https://snapcraft-io-5100.demos.haus/docker/listing)
- go to API endpoint for snap that has CVES: https://snapcraft-io-5100.demos.haus/api/snaps/cve/docker - it should return succecss true
- go to any other snap you have access to that doesn't have CVEs: https://snapcraft-io-5100.demos.haus/api/snaps/cve/firefox - should return error that CVEs data is not available, with 404 status code
- go to API endpoint for non-existent snap: https://snapcraft-io-5100.demos.haus/api/snaps/cve/blabla - should also return error with 404 status code

## Testing
- [x] This PR has tests

## Issue / Card
Fixes WD-20926

### DIFF
--- a/webapp/publisher/cve/cve_views.py
+++ b/webapp/publisher/cve/cve_views.py
@@ -54,13 +54,12 @@ def has_cves(snap_name):
 
     snap_has_cves = CveHelper.has_cve_data(snap_name)
     if snap_has_cves:
-        return flask.jsonify({"success": True, "has_cves": True})
+        return flask.jsonify({"success": True})
     else:
         return (
             flask.jsonify(
                 {
-                    "success": True,
-                    "has_cves": False,
+                    "success": False,
                     "error": f"No CVEs data available for '{snap_name}' snap.",
                 }
             ),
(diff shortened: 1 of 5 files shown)


webapp/publisher/cve/cve_views.py
```diff
@@ -1,15 +1,87 @@
...
+    else:
+        return (
+            flask.jsonify(
+                {
+                    "success": False,
+                    "error": f"CVEs data for '{snap_name}' snap not found.",
+                }
+            ),
+            404,
+        )
```
M7mdisk, line 71: nitpick:
```suggestion
    try:
        snap_has_cves = CveHelper.has_cve_data(snap_name)
        return flask.jsonify({"success": True})
    except NotFound:
        return (
            flask.jsonify(
                {
                    "success": False,
                    "error": f"CVEs data for '{snap_name}' snap not found.",
                }
            ),
            404,
        )
```

M7mdisk (approved):
LGTM

webapp/publisher/cve/cve_views.py
bartaz: Thanks for suggestion. I updated it a bit differently (moving error handling inside has_cve_data helper, so it always returns boolean instead of throwing error.

PR #5108, opened 2025-04-23
Add API endpoint to get revisions with CVEs data

bartaz:
## Done

- renames previous `/api/cve/[SNAP]` endpoints to `/api/[SNAP]/cves` for consistency with other snap based endpoints
- replaces previous endpoint that only checked for CVEs availability with an endpoint that returns list of revisions that have CVEs data

Current CVEs endpoints are:

- `/api/[SNAP]/cves` - returns list of revisions that have CVEs data, or error if not found
- `/api/[SNAP]/[REVISION/cves` - returns list of revisions for given revision of a snap

## How to QA

- go to [demo](https://snapcraft-io-5108.demos.haus/), and sign in (ideally with canonical account with superuser access)
- check cves endpoint for a snap that doesn't have data, it should error: https://snapcraft-io-5108.demos.haus/api/firefox/cves
- check CVEs endpoint for a snap that has CVE data: https://snapcraft-io-5108.demos.haus/api/docker/cves
- take one of the revisions from the list in previous step, and check if it returns CVEs: https://snapcraft-io-5108.demos.haus/api/docker/3213/cves
## Testing
- [x] This PR has tests

## Issue / Card

Fixes [WD-21559](https://warthogs.atlassian.net/browse/WD-21559)

[WD-21559]: https://warthogs.atlassian.net/browse/WD-21559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webapp/publisher/cve/cve_views.py
+++ b/webapp/publisher/cve/cve_views.py
@@ -48,21 +48,21 @@ def get_revisions_with_cves(snap_name):
     )
     if not has_access:
         return (
-            flask.jsonify({"success": False, "error": error_message}),
+            flask.jsonify({"success": False, "message": error_message}),
             status_code,
         )
 
     revisions_with_cves = CveHelper.get_revisions_with_cves(snap_name)
     if len(revisions_with_cves) > 0:
         return flask.jsonify(
-            {"success": True, "revisions": revisions_with_cves}
+            {"success": True, "data": {"revisions": revisions_with_cves}}
         )
     else:
         return (
             flask.jsonify(
                 {
                     "success": False,
-                    "error": f"CVEs data for '{snap_name}' snap not found.",
+                    "message": f"CVEs data for '{snap_name}' snap not found.",
                 }
             ),
             404,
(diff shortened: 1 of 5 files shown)


webapp/publisher/cve/cve_views.py
```diff
@@ -52,9 +52,11 @@ def has_cves(snap_name):
             status_code,
         )
 
-    snap_has_cves = CveHelper.has_cve_data(snap_name)
-    if snap_has_cves:
-        return flask.jsonify({"success": True})
+    revisions_with_cves = CveHelper.get_revisions_with_cves(snap_name)
+    if len(revisions_with_cves) > 0:
+        return flask.jsonify(
+            {"success": True, "revisions": revisions_with_cves}
```
steverydz, line 58: In other APIs we use `data` for the main data in the request, so in this case it would be `"data": { "revisions": ... }`

There are places we don't do that but they are older endpoints from before we started doing this.

steverydz:
In the API response where there is an error, we usually use `message` instead of `error`, although I think `error` makes more sense, so perhaps we should continue with that going forwards?

steverydz (approved):
LGTM 👍 

Just one small comment and a question/statement

PR #5140, opened 2025-05-09
reorder redirects that are not working in prod

abbiesims:
## Done
- Reorder redirects that don't work in prod

## How to QA
- Can't QA until merged (these are working locally and in demos but not in production)


steverydz (approved):
LGTM 👍

PR #5144, opened 2025-05-19
remove trailing slash from broken redirects

abbiesims:
## Done
Removes optional trailing slash from redirects that are currently not working in prod. 

## QA
Impossible to QA until merged as these are currently working locally and on staging.


codeEmpress1 (approved):

codeEmpress1:
LGTM

PR #5154, opened 2025-05-28
redirect snapcraft root

medubelko:
## Done

Now that the old Snapcraft docs are removed from the index, the root page must be redirected to the RTD root.

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): Public redirect

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -8,6 +8,7 @@ tutorials(.*?): /docs/snap-tutorials
 search(.*?): /store
 
 # documentation.ubuntu.com/snapcraft redirects
+docs/snapcraft/?: https://documentation.ubuntu.com/snapcraft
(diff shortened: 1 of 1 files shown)


abbiesims (commented):
Thanks for the contribution, Michael! 

Should this just be `/docs/?` that we redirect? `docs/snapcraft` gives me a 404 - the page does not exist (we can also add both paths if people are visiting `docs/snapcraft`.

medubelko:
@abbiesims The snapd docs aren't ready for migration yet, so that half of the documentation will remain for a while yet. Therefore, the root `/docs` is still a live page.

`docs/snapcraft` represents the root of the Snapcraft portion of the Discourse docs. It was de-indexed, hence the 404. But it should point to the new RTD site.

abbiesims (approved):
Ahh, makes sense, thanks for the clarification! Looks good to me then :)

PR #5160, opened 2025-06-05
fix: Use brand account ID rather than store ID in model service requests

steverydz:
## Done
- Extracts call to the store endpoint to a custom hook
- Uses the brand ID rather than the store ID for model service requests

## How to QA
- Go to https://snapcraft-io-5160.demos.haus/admin/ahnuP3quahti9vis8aiw/settings
- Check that you are able to make and save changes
- Go to https://snapcraft-io-5160.demos.haus/admin/ahnuP3quahti9vis8aiw/snaps
- Check that the side navigation has links for "Models" and "Signing keys"
- With the network tab open, click the "Models" link
- Check that the models table loads and is populated
- In the network tab, check that calls to the policies endpoint are successful (not all models have them so won't all show in the table)
- Check that you can successfully create a new model
- Click through to a model with policies (e.g. https://snapcraft-io-5160.demos.haus/admin/ahnuP3quahti9vis8aiw/models/test_model_without_api_key)
- Check that you can update the API key of a model and save it successfully
- Go to the "Policies" tab
- Check that the policies table loads and is populated
- Check that you can successfully create a policy
- Check that you can successfully delete a policy
- Click the "Signing keys" link in the navigation
- Check that the signing keys table loads and is populated
- Check that you can successfully create a new signing key
- Check that you can successfully deactivate a new signing key

## Testing
- [x] This PR has tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-22645

### DIFF
--- a/static/js/publisher/components/Navigation/Navigation.tsx
+++ b/static/js/publisher/components/Navigation/Navigation.tsx
@@ -32,6 +32,8 @@ function Navigation({
   useEffect(() => {
     if (brand) {
       setBrandId(brand?.["account-id"]);
+    } else {
+      setBrandId("");
     }
   }, [brand]);
 
--- a/static/js/publisher/hooks/__tests__/usePolicies.test.tsx
+++ b/static/js/publisher/hooks/__tests__/usePolicies.test.tsx
@@ -33,21 +33,21 @@ const policiesResponse = [
 ];
 
 const handlers = [
-  http.get("/api/store/test-brand-id/models/test-id-success/policies", () => {
+  http.get("/api/store/test-store-id/models/test-id-success/policies", () => {
     return HttpResponse.json({
       data: policiesResponse,
       message: "",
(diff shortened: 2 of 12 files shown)


steverydz:
@Lukewh I believe I've fixed the issue with the models list and navigation showing in stores and not being cleared. Could you have a click around to confirm?

Lukewh:
> @Lukewh I believe I've fixed the issue with the models list and navigation showing in stores and not being cleared. Could you have a click around to confirm?

A lot better! I still see the signing-keys from one brand store if I switch to a different one (that doesn't have keys) and click directly on the signing keys page

steverydz:
@Lukewh I only have one store that shows signing keys so can't replicate that, but I *think* this may be fixed now

Lukewh:
@steverydz yeah seems to work for me now too 🥳

PR #5173, opened 2025-06-19
feat: Add explore page recommendation sections

steverydz:
## Done
Adds a recommendations section component to show recommendation categories

## How to QA
- Go to https://snapcraft-io-5173.demos.haus/explore
- Check that there is a section for "Recently updated snaps"
- Check that there is a section for "Most popular snaps"
- Check that there is a section for "Trending snaps"

## Testing
- [x] This PR has tests

## Issue / Card
Fixes:
- https://warthogs.atlassian.net/browse/WD-22662
- https://warthogs.atlassian.net/browse/WD-22663
- https://warthogs.atlassian.net/browse/WD-22664

### DIFF
--- a/static/js/store/pages/Explore/Explore.tsx
+++ b/static/js/store/pages/Explore/Explore.tsx
@@ -1,14 +1,86 @@
 import { useRef } from "react";
+import { useQueries } from "react-query";
 
 import Banner from "../../components/Banner";
+import RecommendationsSection from "./RecommendationsSection";
 
-function Explore(): React.JSX.Element {
+import type { UseQueryResult } from "react-query";
+import type { RecommendationData } from "../../types";
+
+function Explore(): JSX.Element {
   const searchRef = useRef<HTMLInputElement | null>(null);
   const searchSummaryRef = useRef<HTMLDivElement | null>(null);
+  const categories: string[] = ["popular", "recent", "trending"];
+
+  const recommendations: UseQueryResult<{
+    name: string;
+    snaps: RecommendationData[];
+  }>[] = useQueries(
+    categories.map((category) => ({
+      queryKey: ["recommendations", category],
+      queryFn: async () => {
+        const response = await fetch(
+          `https://recommendations.snapcraft.io/api/category/${category}`,
+        );
+
+        if (!response.ok) {
+          throw Error(`Unable to fetch ${category} data`);
+        }
+
+        const responseData = await response.json();
+
+        return {
+          name: category,
+          snaps: responseData,
+        };
+      },
+    })),
+  );
+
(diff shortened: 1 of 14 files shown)


static/js/store/pages/Explore/Explore.tsx
```diff
@@ -1,14 +1,73 @@
...
+import RecommendationsSection from "./RecommendationsSection";
 
-function Explore(): React.JSX.Element {
+import type { UseQueryResult } from "react-query";
+import type { RecommendationData } from "../../types";
+
+function Explore(): JSX.Element {
   const searchRef = useRef<HTMLInputElement | null>(null);
   const searchSummaryRef = useRef<HTMLDivElement | null>(null);
+  const categories: string[] = ["popular", "recent", "trending"];
```
edisile, line 13: I think it would make sense to have this data as an array of objects with the id for the category and its title. Something along the lines of
```tsx
const CATEGORIES = [
  { id: "recent", title: "Recently updated snaps" },
  { id: "popular", title: "Most popular snaps" },
  { id: "trending", title: "Trending snaps" },
] as const; // read-only constant outside the component, it will never change at runtime
```
This way all rendering logic for the various sections could be abstracted to rely on this single source of data that would handle the order of the sections and their title. E.g.
```tsx
{
  CATEGORIES.map(c => (
    <RecommendationsSection
        title={c.title}
        snaps={snaps[c.id]}
      />
  )
}
```

static/js/store/pages/Explore/Explore.tsx
```diff
@@ -1,14 +1,73 @@
...
+      {snaps.popular && (
+        <RecommendationsSection
+          snaps={snaps.popular}
+          title="Most popular snaps"
+        />
+      )}
+
+      {snaps.trending && (
+        <RecommendationsSection snaps={snaps.trending} title="Trending snaps" />
+      )}
```
edisile: There should be some skeletons while data is being fetched (like on the search page) and placeholders in case the data isn't available or the query throws an error

static/js/store/pages/Explore/Explore.tsx
steverydz: I like this idea, but those sections will be split up with different, varying component between them so these components won't be sequential

static/js/store/pages/Explore/Explore.tsx
steverydz: I agree - I'll add that in

steverydz:
@edisile I've added a loading cards component. Can you please re-review?

edisile (approved):
Looks good, thanks for addressing my comments!

PR #5175, opened 2025-06-23
feat: update progressive release

Lukewh:
## Done
This PR covers some broken states in the release UI, and adds the ability to update progressive releases to a higher percentage.

Things that have changed:
- Update each revision to include a list of the last releases and removed the "prog_channels" key.
  - This is then used to work out whether the current revision's release is progressive or not.
  - Previously we relied on the "progressive" key added to the revision, this functionality was broken because a revision could be released on multiple channels, meaning the progressive key would not be right for all but 1 of those channels.
  - Many refactors around the codebase to use the latest release information to determin if a release is progressive and what the actual progressive state is.
- Use the channelMap information from the API to set the current percentage of devices (this information is only available from the channelMap data (see code comment) so all "percentage" fields were null. This is now fixed).
- Update the updateReleasesData to pull all information used to reinitialize all state (see channelMap issue above). This also ensures that the whole UI get's rerendered, fixing edge-cases where certain elements were not updated due to using selectors, which are memoized and do not always cause a rerender.
- Related to the point above, used the main `/api/<snap_name>/releases` endpoint instead of `/<snap_name>/releases/json` endpoint in `fetchReleasesHistory` when a release has been made. Again, this ensure all state is up to date.
- Updated the `ProgressiveReleaseProgressChart` to have rounded corners if the current percentage is lower than the target percentage - avoiding the visual vomit of having a flat edge. Minor, but it irritated me.
- Added a "minPercentage" to the ProgressChart, this ensures that when you are doing a progressive release over a previous progressive release, you cannot submit a lower percentage than the highest percentage of the currently released revisions (the store does not allow you to do a progressive release to 50% of devices, then lower it to 20%, it always has to be higher). Word soup, drink it up.
  - Included in this is some messaging for users telling them that there is a minimum, with a red section to the progressive bar. Could do with some UX.
- Fixed progressive tooltips not showing the correct progressive information.
- Fixed the channel+arch history not showing the correct progressive information.
- Updated the logic to allow a revision to be released over the same revision IF the currently released revision is part of a progressive release.
- Fixed some deprecation warnings regarding `defaultProps`. 
Is this all perfect? No, I doubt it, but you have to ask yourself - are you perfect?

## How to QA
Settle in.

Note: when promoting revisions, try the different methods: dragging the revision, dragging the channel, promoting a revision, promoting a channel, thanks.

1. Visit the demo.
2. Go to one of your test snaps that has multiple revisions.
3. Close all the channels.
4. Promote an older revision to a channel - when saving you should not see any progressive release bar.
5. Try and promote the same revision to the same channel - you should not be allowed.
6. Take a newer revision and promote it to the same channel - when saving you should see the progressive bar set to 100. If you release it now, it should be a normal release (not progressive).
7. Close the channel, and repeat step 4, then: promote a newer revision, and set the progressive release to 20% or something. Save the changes.
8. Hover over the release, you should see the correct information.
9. Open the history for the release, you should also see the correct information.
10. Drag THE SAME revision onto the channel again (it should let you). When reviewing changes, the progressive bar should be set to 100%, but part of the bar should be red, and there should be some text below telling you what the minimum is. Try to drag the bar below that minimum - denied? You should be. Set the percentage to somewhere higher and save that changes.
11. Again, hover to confirm the correct percentages show. Open the history, confirm the correct percentages show.
12. All good? Fun times. Now drag the same revision onto the channel again, but this time just release to 100%. This should remove the progressive information from the UI.
13. If you're feeling extra spicy, and you have multple architectures, manually release each architecture to a channel, one at a time, setting different percentages.
14. Then attempt to do a progressive release again - you should only be able to release to a percentage equal to or higher than which ever you set to be higher.

## Testing
- [x] This PR has tests

## Issue / Card
Fixes #

## Screenshots
https://github.com/user-attachments/assets/f8df3ad7-3d25-48b0-a906-65673f2b847a


steverydz:
Space needed between the action and the channel name in the review changes panel

Screenshot 2025-06-25 at 14 52 55

static/js/publisher/pages/Releases/actions/releases.ts
```diff
@@ -16,26 +20,49 @@ import {
...
-    const revisionsMap = getRevisionsMap(releasesData.revisions);
-    initReleasesData(revisionsMap, releasesData.releases);
-    dispatch(updateRevisions(revisionsMap));
-    dispatch(updateReleases(releasesData.releases));
-    dispatch(updateArchitectures(releasesData.revisions));
+    const revisionsList = releasesData.revisions;
+
+    getReleaseDataFromChannelMap(channelMap, revisionsList, snapName).then(
+      ([transformedChannelMap, revisionsListAdditions]) => {
+        Array.prototype.push.apply(revisionsList, revisionsListAdditions);
```
edisile:
```suggestion
        revisionsList.push(...revisionsListAdditions);
```
This is more readable, it took me a minute to recall the arcane knowledge of what Function.prototype.apply() does

static/js/publisher/pages/Releases/components/ProgressiveReleaseProgressChart.tsx
```diff
@@ -9,6 +9,14 @@ function ProgressiveReleaseProgressChart({
...
   isPreviousRevision,
 }: ReleaseProgress) {
+  const roundedEnd = currentPercentage > targetPercentage;
+  const innerStyle: { width: string; borderRadius?: string } = {
+    width: `${currentPercentage ? currentPercentage : 0}%`,
+  };
+
+  if (roundedEnd) {
+    innerStyle.borderRadius = "10px";
+  }
```
edisile:
```suggestion
  const innerStyle: React.CSSProperties = {
    width: `${currentPercentage ? currentPercentage : 0}%`,
    borderRadius: roundedEnd ? "10px" : undefined
  };
```
erm akshually 🤓☝️ React.CSSProperties is the correct interface to use for style props

static/js/publisher/pages/Releases/components/releasesConfirmDetails/index.js
```diff
@@ -16,6 +16,24 @@ import CloseChannelsRow from "./closeChannelsRow";
...
+      const channel = u[1];
+      if (
+        channel.revision &&
+        channel.revision.releases &&
+        channel.revision.releases[0] &&
+        channel.revision.releases[0].progressive &&
+        channel.revision.releases[0].progressive.percentage > min
+      ) {
+        min = channel.revision.releases[0].progressive.percentage;
+      }
```
edisile:
```suggestion
      const percentage = channel?.revision?.releases?.[0]?.progressive?.percentage;
      if (percentage > min) {
        min = percentage;
      }
```
Why not use optional chaining? Even if it fails and `percentage === undefined`, `percentage > min` is always false

static/js/publisher/pages/Releases/reducers/pendingReleases.ts
```diff
@@ -76,7 +76,11 @@ function releaseRevision(
     state[revision.revision][channel].previousRevisions = previousRevisions;
   }
 
-  if (progressive && !state[revision.revision][channel].progressive) {
+  if (
+    progressive &&
+    (!state[revision.revision][channel].progressive ||
+      state[revision.revision][channel].progressive !== progressive)
+  ) {
```
edisile, line 84: Is the type of `progressive` in the function declaration wrong (it's declared as `null`)? Otherwise mean this whole check is pointless

static/js/publisher/pages/Releases/releasesState.js
```diff
@@ -29,19 +29,30 @@ function initReleasesData(revisionsMap, releases) {
             rev.channels.push(channel);
           }
 
-          if (!rev.prog_channels) {
-            rev.prog_channels = [];
+          if (release.progressive && release.progressive.percentage) {
+            release.isProgressive = true;
+          } else {
+            release.isProgressive = false;
           }
```
edisile, line 36:
```suggestion
          release.isProgressive = release.progressive && release.progressive.percentage;
```

edisile (approved):
Looks good, just some small nitpicks

static/js/publisher/pages/Releases/actions/releases.ts
bartaz: Yes, it seems that revisionsList is an Array, so apply doesn't seem necessary.
The `apply` trick was useful for some iterables that didn't implement whole Array interface (like NodeList). Doesn't seem needed here. Unless there is a reason for it?

static/js/publisher/pages/Releases/releasesState.js
```diff
@@ -29,19 +29,30 @@ function initReleasesData(revisionsMap, releases) {
...
-            rev.prog_channels.push(channel);
+          // Based on the note at the top of
+          // https://dashboard.snapcraft.io/docs/reference/v1/snap.html#progressive-releases
+          // We need to get the current channelMap to hydrate the current percentage
+          if (release.isProgressive && channelMap) {
+            const currentChannel = channelMap.find(
+              (c) => c.channel === channel && c.revision === release.revision,
+            );
+            if (currentChannel) {
+              release.progressive["current-percentage"] =
```
bartaz, line 45: As much as I remember working with Redux, I feel this kind of patching of data is going a bit against Redux-way, as we update one part of the state (releases) based on another part of state (channelMap). Are we also making sure that this gets updated properly when channel map changes?

I kind of feel that to make it more in the spirit of Redux, both `isProgressive` and `current-percentage` could be selectors (?), like `isReleaseProgressive(release)`, `getCurrentPercentage(release)`, that would implement the logic of checking the state, being calculated based of state, rather than being stored in state (as kind of duplication of other information in the state - risking inconsistencies).

It would be quite a stretch to refactor it, so I'm kind of fine with leaving it as-is, as kind of a shortcut. 
I'm just worried, that, while it fixes the issues we have now, it makes the whole state likely harder to reason about.

static/sass/_snapcraft_p-progressive-bar.scss
```diff
@@ -42,6 +43,12 @@
     background: $target;
   }
 
+  .progressive-bar__min {
+    @extend %bar;
+    
+    background: $min;
```
bartaz: Nitpicking, but having a color variable named `$min` makes this like a tad confusing without the context.
I see it follows the previous convention of colors named `inactive` `target` etc… but they beg to at least be namespaced by `$color-…`

static/js/publisher/pages/Releases/releasesState.js
bartaz: I think it's fine to leave it as explicit conditional statement. As a one-liner it will end up being a number, not a bool. Obviously we could cast it as  `!!(release.progressive && release.progressive.percentage)`, or even 
`!!(release.progressive?.percentage)`, but it seems to reduce the readability.

static/js/publisher/pages/Releases/helpers.ts
```diff
@@ -47,6 +47,21 @@ export function getRevisionsArchitectures(revisions: any[]) {
   return archs;
 }
 
+export function getLatestChannelRevisionRelease(
```
bartaz: For me the name suggest the function would return "latest channel", why not naming it `getLatestRelease` (the attributes should be enough to imply the filtering) or `getLatestReleaseForRevisionInChannel` (if we want to be explicit about attributes).

static/js/publisher/pages/Releases/components/releasesTable/row.tsx
```diff
@@ -6,19 +6,24 @@ import { getRevisionsArchitectures } from "../../helpers";
...
+}: {
+  canDrag: boolean;
   risk?: any;
   branch?: any;
   revisions?: any;
   canDrop?: any;
   children?: any;
 }) => {
-  const { risk, branch, revisions, canDrop, children } = props;
+  const _canDrag = !!revisions && canDrag;
```
bartaz: Can `canDrag` be just overridden if needed? Is the separate `_canDrag` needed?

```suggestion
  canDrag = !!revisions && canDrag;
```

PR #5179, opened 2025-06-26
WIP: Explore page

steverydz:
## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #

### DIFF
--- a/static/js/store/pages/Explore/Explore.tsx
+++ b/static/js/store/pages/Explore/Explore.tsx
@@ -1,0 +1,16 @@
+import { useRef } from "react";
+
+import Banner from "../../components/Banner";
+
+function Explore(): React.JSX.Element {
+  const searchRef = useRef<HTMLInputElement | null>(null);
+  const searchSummaryRef = useRef<HTMLDivElement | null>(null);
+
+  return (
+    <>
+      <Banner searchRef={searchRef} searchSummaryRef={searchSummaryRef} />
+    </>
+  );
+}
+
+export default Explore;
--- a/static/js/store/pages/Explore/Explore.tsx
+++ b/static/js/store/pages/Explore/Explore.tsx
@@ -1,14 +1,86 @@
 import { useRef } from "react";
+import { useQueries } from "react-query";
 
 import Banner from "../../components/Banner";
+import RecommendationsSection from "./RecommendationsSection";
 
-function Explore(): React.JSX.Element {
+import type { UseQueryResult } from "react-query";
+import type { RecommendationData } from "../../types";
+
+function Explore(): JSX.Element {
   const searchRef = useRef<HTMLInputElement | null>(null);
   const searchSummaryRef = useRef<HTMLDivElement | null>(null);
+  const categories: string[] = ["popular", "recent", "trending"];
+
+  const recommendations: UseQueryResult<{
+    name: string;
+    snaps: RecommendationData[];
+  }>[] = useQueries(
+    categories.map((category) => ({
+      queryKey: ["recommendations", category],
+      queryFn: async () => {
+        const response = await fetch(
+          `https://recommendations.snapcraft.io/api/category/${category}`,
+        );
+
+        if (!response.ok) {
+          throw Error(`Unable to fetch ${category} data`);
+        }
+
+        const responseData = await response.json();
+
+        return {
+          name: category,
+          snaps: responseData,
+        };
+      },
+    })),
+  );
+
+  const isLoading: boolean =
+    !recommendations || recommendations.some((r) => r.isLoading);
+
+  const snaps: Record<string, RecommendationData[]> = {};
+
+  if (recommendations) {
+    recommendations.forEach((recommendation) => {
+      if (recommendation.data) {
+        snaps[recommendation.data.name] = recommendation.data.snaps.slice(0, 6);
+      }
+    });
+  }
 
   return (
     <>
       <Banner searchRef={searchRef} searchSummaryRef={searchSummaryRef} />
+
+      {recommendations && (
+        <>
+          {snaps.recent && (
+            <RecommendationsSection
+              snaps={snaps.recent}
+              title="Recently updated snaps"
+              isLoading={isLoading}
+            />
+          )}
+
+          {snaps.popular && (
+            <RecommendationsSection
+              snaps={snaps.popular}
+              title="Most popular snaps"
+              isLoading={isLoading}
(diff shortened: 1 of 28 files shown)


static/sass/styles.scss
```diff
@@ -511,10 +511,143 @@ html {
   position: relative;
 }
 
+@media (width < $breakpoint-navigation-threshold) {
```
bartaz: Are the navigation changes relevant here? I guess these are from previous work by @alvaromateo ?
So would this be some rebase/merge issue to see them added here?

bartaz:
I guess it's too late for that, but feels that this page is static enough that it could have been rendered on server side instead of being almost fully rendered by client-side React.

It's now quite visible with page jumping when each section loads separately and jumps into view.
I wonder if it would help to have all the cards as placeholders while loading (like on Store page).

static/sass/styles.scss
steverydz: @bartaz I've been rebasing regularly as this branch has been around for a while. These aren't intentional changes in this PR. I did have conflicts at one point so this is likely due to that.

bartaz: Looks like these styles have been moved to another file, so this looks like a leftover from a rebase that can be removed.

https://github.com/canonical/snapcraft.io/pull/5348

steverydz:
> I guess it's too late for that, but feels that this page is static enough that it could have been rendered on server side instead of being almost fully rendered by client-side React.
> 
> It's now quite visible with page jumping when each section loads separately and jumps into view. I wonder if it would help to have all the cards as placeholders while loading (like on Store page).

@bartaz It's not necessarily too late. I agree that there's no real benefit of it being client-side rendered. It probably wouldn't be that much work to translate it to server-side rendered. I'll wait until I get some visual design feedback and maybe do that.

It does use the loading cards, but they're still dependent on the successful request as we don't want to show those sections if there are no results. I think rendering it server-side is ultimately the better solution.

static/js/store/pages/Explore/Explore.tsx
```diff
@@ -0,0 +1,191 @@
...
+            </Strip>
+          )}
+        </>
+      ) */}
+
+      {recommendations && (
+        <>
+          {snaps.recent && (
+            <Strip shallow className="u-no-padding--bottom">
+              <RecommendationsSection
```
M7mdisk, line 112: Just a note, from my understanding, @0-luke wants this to be more dynamic (i.e, if I want to change the title, I don't need to open a PR). The service itself still isn't ready for this, but thought I'd mention it so that its in mind when the code is changed to be server side

M7mdisk (commented):
Its very nice. but I agree with you both that this should be server side rendered 👍

leiaru-c:
**Here are points I think need updates:**

- Overall layout:
The spacing between elements seems to be a bit tight after titles.
image

- Search bar empty state: 
reviewsnapcraft1
The empty state should say "Search snap store" to stay aligned with [charmhub](https://charmhub.io/) or "Search thousands of apps" to stay true to designs. I think that searching Snapcraft is misleading, there is nothing to search on snapcraft, only on snapcraft.io or the snap store, and snapstore.io could refer to other marketing pages, not apps.

- Categories section:
image
1/ Without any illustration, it looks quite bare. I can check with the visual team on what to do here if we cannot use icons.
2/ The Featured category is missing in the demo (Yellow).

- Top rated vs most popular, most popular vs trending snaps:
The wording in both these categories is confusing. 
In the design, the titles say "top rated" versus "most popular", so my guess is that it means most liked/high rated snaps versus most downloaded snaps. 
In the demo, "most popular" versus "trending snaps" sound too similar, it makes me think that they both are based on the number of downloads.
image
image
 We need to clarify what criteria we use to show results for these categories.

- Game section:
The game spotlight is very good, can't we have it? 
image

**These are not show stoppers, but worth raising:**

- Banners:
The banners with selection of apps for specific use cases are nice, they add some color and spotlight useful apps. I am assuming it is not in the demo because it is not feasible short term,  but would be good to have.
image

- See all link for "recently updated snaps":
image
I don't know if seeing all recently updated snap is possible, but there was a link in the intial designs. 

- Categories section:
image
3/ The copy style guide specifies that we should use small caps in titles, not capitalise words. That being said, I know that the categories themselves in filters are capitalised, so imo, there are 3 options:
- Leave as is: if we go for consistency and can't modify the filters, it's ok to keep the capitalisation as is.
- Only modify the copy on the page, leave the filters as is: if we consider that the category titles are not the same as filters, then removing the caps would be good. 
- Modify the copy and update the filters: And if we want consistency with copy guidelines and the way that filters are formatted, it would be good to remove the second word capitalisation in the filters as well.

- Top rated vs most popular, most popular vs trending snaps:
In addition to feedback earlier, I think that these two concepts (most downloads, and best ratings) overlap quite a bit and might not provide as much value. In the demo, two apps overlap, which is not the end of the world but is not all that great either. A selection of popular snaps per top categories might be more interesting in the future?
image

-  Ubuntu community section:
Also really nice, what would need to be done to have that section in asap?
image

steverydz:
@leiaru-c I've addressed the key points we talked about with the design. Could you please take another look?

- Fixed spacing between sections and headings
- Updated the placeholder text of the search box
- The capitalisation for the categories comes from the API so would need to be changed at that level
- The titles, e.g. Most popular etc. come from the recommendations service so can be updated there along with the chosen snaps

leiaru-c:
Checked and it's good to go for me. Looking forward to the next iteration :)

static/js/store/pages/Explore/Explore.tsx
abbiesims, line 1: The loading on the main Explore page is a little scattered - on first render, things move about a lot before the full page is rendered. Is this something that will be better when slices are introduced? Or could we already create a skeleton with the "Recently updated snaps" loading cards visible right from the start?

https://github.com/user-attachments/assets/13a06794-ca87-4e77-8a7f-af5067ce2aab

steverydz: @abbiesims This is known. The loading cards are actually there but they don't get triggered until the JS loads, and as those sections are only shown if there is data for them, we don't know to show the loading cards in time. The next phase of this is to build a server side rendered version which will solve these issues (https://warthogs.atlassian.net/browse/WD-28151)

abbiesims: Ahh nice, makes sense

PR #5187, opened 2025-06-27
fix(release ui): update all data on successful release

Lukewh:
## Done
- Update the `updateReleasesData` to pull all information to reinitialize all state . This also ensures that the whole UI get's rerendered, fixing edge-cases where certain elements were not updated due to using selectors, which are memoized and do not always cause a rerender.
- Related to the point above, used the main `/api/<snap_name>/releases` endpoint instead of `/<snap_name>/releases/json` endpoint in `fetchReleasesHistory` (renamed to `fetchSnapReleaseStatus`) when a release has been made. Again, this ensure all state is up to date.

## How to QA
- Use the release ui as normal - there should be no obvious changes
- The tests should continue to pass

## Testing
- [x] This PR has tests - updated tests

## Issue / Card
Fixes #

### DIFF
--- a/static/js/publisher/pages/Releases/api/releases.js
+++ b/static/js/publisher/pages/Releases/api/releases.js
@@ -21,8 +21,8 @@ export function fetchReleases(onComplete, releases, csrfToken, snapName) {
   return queue;
 }
 
-export function fetchReleasesHistory(csrfToken, snapName) {
-  return fetch(`/${snapName}/releases/json`, {
+export function fetchSnapReleaseStatus(csrfToken, snapName) {
+  return fetch(`/api/${snapName}/releases`, {
     method: "GET",
     mode: "cors",
     cache: "no-cache",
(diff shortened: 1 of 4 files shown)


static/js/publisher/pages/Releases/api/releases.js
```diff
@@ -21,8 +21,8 @@ export function fetchReleases(onComplete, releases, csrfToken, snapName) {
   return queue;
 }
 
-export function fetchReleasesHistory(csrfToken, snapName) {
-  return fetch(`/${snapName}/releases/json`, {
+export function fetchSnapReleaseStatus(csrfToken, snapName) {
```
steverydz, line 24: This could probably just use `window.CSRF_TOKEN`

steverydz (approved):
LGTM 👍 

Just one suggestion in the code

static/js/publisher/pages/Releases/api/releases.js
Lukewh: I'll do a quick follow-up to this PR for this :+1:

PR #5188, opened 2025-06-27
fix: use global csrf token

Lukewh:
## Done
- Removed csrfToken from the store
- Removed code that was passing csrfToken around as a parameter
- Where csrf token is needed, change to `window.CSRF_TOKEN`

## How to QA
- Load the release ui and use it as normal.
- You should be able to release revision without CSRF issues.

## Testing
- [x] This PR has tests - updated tests

## Issue / Card
https://warthogs.atlassian.net/browse/WD-23299


steverydz (approved):
LGTM 👍

PR #5189, opened 2025-06-27
fix: refactor revisions to include releases

Lukewh:
## Done
- Removal of `prog_channels` array on a revision
- Addition of `releases` array on a revision
- Addition of `isProgressive` boolean on a release.
- Added the `["current-percentage"]` from the channel map information, as described in the [store documentation](https://dashboard.snapcraft.io/docs/reference/v1/snap.html#progressive-releases), to the `release.progressive["current-percentage"]` so the release has accurate information for the current number of devices on a release.
- Refactor all rendering code that used either `prog_channels` or did basic checks against the `progressive`  key of a revision to use  `revision.releases[0].isProgressive` to determine if the current release is progressive.

## How to QA
- If you compare a progressive release in production to this PR you will notice differences in data shown.
- You can use the deno snap to see an example (screenshots below).
- Everything else should work as before in production.

## Testing
- [x] This PR has tests - updated tests and added a couple of extra data points

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-23300

### Hover
image
Production

image
This PR

### History
image
Production (ignore the bar styling, it was fixed in a different PR)

image
This PR


edisile (approved):
Thank you for this delectable contribution

static/js/publisher/pages/Releases/selectors/__tests__/selectors.test.js
```diff
@@ -846,6 +846,7 @@ describe("getProgressiveState", () => {
         track: "latest",
         risk: "stable",
         revision: "1",
+        isProgiressive: false,
```
bartaz: typo
```suggestion
        isProgressive: false,
```

static/js/publisher/pages/Releases/components/releasesTable/cellViews.js
```diff
@@ -68,13 +68,15 @@ EmptyInfo.propTypes = {
...
+  const { progressive } = revision?.releases?.[0];
+
   const previousRevisionData = (
     <>
       <strong>{previousRevision?.revision || "Unknown"}</strong>
       <br />
       <strong>
-        {Math.round(100 - revision?.progressive?.["current-percentage"]) || 0}%
-        → {Math.round(100 - revision?.progressive?.percentage)}%
+        {Math.round(100 - progressive?.["current-percentage"]) || 0}% →{" "}
```
bartaz: It's not wrong, but is it necessary to have the condition at revision and progressive?
Would this component be ever called without a revision that is progressive?

I feel like it should be safe to do `revision.progressive.` without `?`, and if we need a failsafe it could be eariler, as the whole tooltip would not make sense if there is no progressive revision to show?

PR #5201, opened 2025-07-07
feat: Improved progressive release experience

Lukewh:
## Done
- When updating a progressive release, ensure you can't release to a lower percentage than is currently set.
- Added some extra text to the Review panel around minimum progressive releases.
- Updated the way we get the previous release, to ensure that if you:
  1. create a progressive release rev 1 -> rev 2
  2. overwrite the release with rev 1
  3. Create another progressive release rev 1 -> rev 2

  The minimum percentage from step 1. isn't shown in step 3.
- Renamed `previousRevisions` to `previousReleases` throughout (as this is what is actually stored in the variable).
- Fixed defaultProp deprecation warning.

## How to QA
- Pull the branch or view the demo
- Play with progressive releases. Create them, remove them, overwrite them. Things should work as expected.
- Try doing some strange things, like described above in the "Done" section.

## Testing
- [x] This PR has tests - updated current tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-23301

## Screenshots
image

### DIFF
--- a/static/js/publisher/pages/Releases/actions/pendingReleases.js
+++ b/static/js/publisher/pages/Releases/actions/pendingReleases.js
@@ -19,6 +19,9 @@ export function releaseRevision(revision, channel, progressive) {
     const { revisions, pendingReleases } = state;
 
     const previousReleases = getReleases(state, revision.architectures, channel)
+      // Find all revision releases for this channel and architecture
+      // that do not share the same revision number as the previous release.
+      // for example [1, 1, 2, 2, 3, 2, 2, 2, 1] will return [1, 2, 3, 2, 1]
       .reduce((acc, release) => {
         if (!acc.length || acc[acc.length - 1].revision !== release.revision) {
           acc.push(release);
--- a/static/js/publisher/pages/Releases/actions/pendingReleases.js
+++ b/static/js/publisher/pages/Releases/actions/pendingReleases.js
@@ -65,7 +65,7 @@ export function releaseRevision(revision, channel, progressive) {
         revision: revisionToRelease,
         channel,
         progressive,
-        previousReleases: previousReleases,
+        previousReleases,
       },
     });
   };
--- a/static/js/publisher/pages/Releases/actions/pendingReleases.js
+++ b/static/js/publisher/pages/Releases/actions/pendingReleases.js
@@ -18,17 +18,19 @@ export function releaseRevision(revision, channel, progressive) {
     const state = getState();
     const { revisions, pendingReleases } = state;
 
-    const previousRevisions = getReleases(
-      state,
-      revision.architectures,
-      channel,
-    )
-      .filter((release) => release.revision !== revision.revision)
+    const previousReleases = getReleases(state, revision.architectures, channel)
+      .reduce((acc, release) => {
+        if (!acc.length || acc[acc.length - 1].revision !== release.revision) {
+          acc.push(release);
+        }
+
+        return acc;
+      }, [])
       .map((release) => revisions[release.revision]);
 
     let revisionToRelease = revision;
 
-    if (!progressive && previousRevisions.length > 0 && previousRevisions[0]) {
+    if (!progressive && previousReleases.length > 0 && previousReleases[0]) {
       revisionToRelease = revisions[revision.revision];
 
       let percentage = 100;
@@ -60,7 +62,7 @@ export function releaseRevision(revision, channel, progressive) {
         revision: revisionToRelease,
         channel,
         progressive,
-        previousRevisions,
+        previousReleases: previousReleases,
       },
     });
   };
--- a/static/js/publisher/pages/Releases/components/progressiveBarControl.js
+++ b/static/js/publisher/pages/Releases/components/progressiveBarControl.js
@@ -7,6 +7,7 @@ import { updateProgressiveReleasePercentage } from "../actions/pendingReleases";
 import progressiveTypes from "./releasesConfirmDetails/types";
 
 import { InteractiveProgressiveBar } from "./progressiveBar";
+import { min } from "date-fns";
 
 class ProgressiveBarControl extends React.Component {
   constructor(props) {
@@ -26,7 +27,7 @@ class ProgressiveBarControl extends React.Component {
   }
 
   render() {
-    const { release, type, globalPercentage } = this.props;
+    const { release, type, globalPercentage, minPercentage } = this.props;
 
     if (!release.progressive) {
       return false;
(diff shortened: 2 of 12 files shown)


static/js/publisher/pages/Releases/components/progressiveBarControl.js
```diff
@@ -83,7 +84,16 @@ class ProgressiveBarControl extends React.Component {
         <div className="p-release-details-row__notes">
           <small>
             {100 - targetPercentage}% of devices will stay on the current
-            version
+            version.
+            {minPercentage !== 1 ? (
```
bartaz: Can it be 0? If not maybe having this as `minPercentage > 1` would be slightly clearer?

static/js/publisher/pages/Releases/actions/pendingReleases.js
```diff
@@ -18,17 +18,19 @@ export function releaseRevision(revision, channel, progressive) {
     const state = getState();
     const { revisions, pendingReleases } = state;
 
-    const previousRevisions = getReleases(
-      state,
-      revision.architectures,
-      channel,
-    )
-      .filter((release) => release.revision !== revision.revision)
+    const previousReleases = getReleases(state, revision.architectures, channel)
```
bartaz, line 21: Some code comment could help understanding what is being filtered here.
Pure `.filter(…)` (like it was before) is quite easy to understand by itself, but wrapping your head around `.reduce` is a bit harder. I'm not exactly sure what happens here, and more importantly why are we filtering releases like that?

static/js/publisher/pages/Releases/actions/pendingReleases.js
```diff
@@ -60,7 +62,7 @@ export function releaseRevision(revision, channel, progressive) {
         revision: revisionToRelease,
         channel,
         progressive,
-        previousRevisions,
+        previousReleases: previousReleases,
```
edisile: Nitpick
```suggestion
        previousReleases,
```

static/js/publisher/pages/Releases/components/releasesConfirmDetails/index.js
```diff
@@ -32,6 +32,25 @@ const ReleasesConfirmDetails = ({ updates, isProgressiveReleaseEnabled }) => {
   const showNewReleases = Object.keys(newReleases).length > 0;
   const showPendingCloses = pendingCloses.length > 0;
 
+  const lowestPercentage = Object.entries(updates.newReleasesToProgress).reduce(
+    (min, update) => {
+      const release = update[1];
```
bartaz: How can I find what is the shape of the `update` object/array? Or what is `newReleasesToProgress`? Do we have types for it or something?
Accessing data by arbitrary array index `update[1]` is a bit confusing and fragile without a proper context of the data shape.

static/js/publisher/pages/Releases/actions/pendingReleases.js
steverydz: I second this

Lukewh: Added some commenting below this line! Hope that helps and justifies the use of reduce :)

static/js/publisher/pages/Releases/components/releasesConfirmDetails/index.js
Lukewh: For some reason I was using "Object.entries" instead of "Object.values". entries provides an array of [key, value], hence the `update[1]`. Updated now!

static/js/publisher/pages/Releases/components/progressiveBarControl.js
```diff
@@ -83,7 +84,16 @@ class ProgressiveBarControl extends React.Component {
...
+            version.
+            {minPercentage > 1 ? (
+              <>
+                <br />
+                {minPercentage}% is the lowest percentage all revisions can be
+                released to.
+              </>
+            ) : (
+              ""
+            )}
```
edisile: The "else" branch is empty so might as well use `&&`
```suggestion
            version.
            {minPercentage > 1 && (
              <>
                <br />
                {minPercentage}% is the lowest percentage all revisions can be
                released to.
              </>
            )}
```

steverydz (approved):
LGTM 👍 

The renaming definitely helps make this easier to reason about

static/js/publisher/pages/Releases/components/releasesConfirmDetails/cancelProgressiveRow.js
```diff
@@ -5,7 +5,7 @@ const CancelProgressiveRow = ({ release }) => {
   const revisionInfo = release.revision;
 
   return revisionInfo.architectures.map((arch) => {
```
edisile, line 7: Do I need a multi-arch snap to test this component? I only have arm64 snaps and I can't get it to show up no matter what

static/js/publisher/pages/Releases/components/releasesConfirmDetails/index.js
```diff
@@ -32,6 +32,24 @@ const ReleasesConfirmDetails = ({ updates, isProgressiveReleaseEnabled }) => {
...
+
+      if (percentage > min) {
+        return percentage;
+      }
+
+      return min;
+    },
+    1,
+  );
+
```
edisile: This might need a comment or a variable name change (or both): the accumulator variable is called `min` but we're actually looking for the max among the previous release percentages

edisile: Maybe breaking it down into a map and a reduce would be even clearer
```suggestion
  // Find the highest release target percentage among previous progressive releases:
  // this will be the lowest percentage for the new release
  const lowestPercentage = Object.values(updates.newReleasesToProgress).map(
    (release) => {
      let percentage = 1;
      const previousRelease = release.previousReleases?.[0];
      if (previousRelease?.revision === release.revision.revision) {
        const p = previousRelease?.releases?.[0]?.progressive?.percentage;
        if (p) percentage = p;
      }

      return percentage;
    }
  ).reduce((a, b) => Math.max(a, b);

```

static/js/publisher/pages/Releases/components/releasesConfirmDetails/cancelProgressiveRow.js
bartaz: I believe this is a legacy backwards compatibility. It used to be possible to have a single revision built for multiple architectures, but it's not possible now (from what I understand from @Lukewh). Not sure if it's possible to build a new snap to have multiple archs.

static/js/publisher/pages/Releases/actions/pendingReleases.js
Lukewh: good catch!

static/js/publisher/pages/Releases/components/releasesConfirmDetails/cancelProgressiveRow.js
Lukewh: I don't think it's possible anymore no!

static/js/publisher/pages/Releases/components/releasesConfirmDetails/index.js
Lukewh: Personally, I prefer keeping 1 loop and reducers are fairly straightforward. I know we're not talking about tons of data, but every ms is important.
I do agree the variable should probably be "max" within the reducer function, so I'll update that!

PR #5202, opened 2025-07-07
chore: Install Jotai and add Provider

steverydz:
## Done
Installs Jotai and wraps brand store and publisher applications in a [Provider](https://jotai.org/docs/core/provider)

## How to QA
- Go to any brand store and publisher pages on the demo
- They should load and work as expected

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): No functionality changes

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-23491

### DIFF
--- a/static/js/publisher/publisher.tsx
+++ b/static/js/publisher/publisher.tsx
@@ -1,6 +1,7 @@
 import { createRoot } from "react-dom/client";
 import { createBrowserRouter, RouterProvider } from "react-router-dom";
 import { RecoilRoot } from "recoil";
+import { Provider as JotaiProvider } from "jotai";
 import { QueryClient, QueryClientProvider } from "react-query";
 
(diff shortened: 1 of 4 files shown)


static/js/publisher/publisher.tsx
```diff
@@ -102,8 +103,10 @@ const queryClient = new QueryClient({
 
 root.render(
   <RecoilRoot>
-    <QueryClientProvider client={queryClient}>
-      <RouterProvider router={router} />
-    </QueryClientProvider>
+    <Provider>
```
edisile: Just a nitpick, but I'm not a big fan of the `Provider` name, it's too generic and forces me to scroll to the top to understand what it does. Importing it as an alias would improve readability
```ts
import { Provider as JotaiProvider } from "jotai";
```

edisile (approved):

PR #5203, opened 2025-07-07
chore: Convert brandIdState to a Jotai atom

steverydz:
## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #

### DIFF
--- a/static/js/publisher/components/Navigation/Navigation.tsx
+++ b/static/js/publisher/components/Navigation/Navigation.tsx
@@ -1,5 +1,6 @@
 import { useState, useEffect } from "react";
 import { useRecoilState, useRecoilValue } from "recoil";
+import { useAtom as useJotaiState } from "jotai";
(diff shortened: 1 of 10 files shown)


static/js/publisher/components/Navigation/Navigation.tsx
```diff
@@ -1,5 +1,6 @@
 import { useState, useEffect } from "react";
 import { useRecoilState, useRecoilValue } from "recoil";
+import { useAtom as useJotaiState } from "jotai";
```
M7mdisk, line 3: question: can I ask why these renames are necessary?

steverydz: @M7mdisk To make it clearer that it's been converted to Jotai. It can be cleaned up once the conversion is complete, but this way I think it's easier to keep track of what has been updated.

edisile (approved):

PR #5223, opened 2025-07-18
Snap details header abstracted to Jinja macros

bartaz:
## Done

This converts previous Jinja partial template used to render snap details header into Jinja macros. Data parameters are now passed as arguments to macros, rather than being accessed from global template context.

This is just a first step of the potential refactoring. Further amendments are possible to clear our and restructure the parameters passed into macros. This initial update was made without changing the underlying structure of previous partial templates to avoid too many changes.

Drive-by: removed dependency on `package_name` to render unnecessary id attr on a tooltip.

## How to QA

- Verify header still renders as expected on different snaps, check both snaps detail page and distro install page
- example snaps:
- with star developer: https://snapcraft-io-5223.demos.haus/josm https://snapcraft-io-5223.demos.haus/install/josm/kubuntu
- verified account: https://snapcraft-io-5223.demos.haus/artikulate https://snapcraft-io-5223.demos.haus/install/artikulate/kde-neon
- developer + publisher https://snapcraft-io-5223.demos.haus/discord https://snapcraft-io-5223.demos.haus/install/discord/rhel
- standard publisher: https://snapcraft-io-5223.demos.haus/terminal-parrot https://snapcraft-io-5223.demos.haus/install/terminal-parrot/raspbian

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): just refactoring of templates

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-23595

## Screenshots

image


steverydz (approved):
LGTM 👍

templates/store/snap-distro-install.html
```diff
@@ -99,15 +100,13 @@ <h1 class="p-heading--2 u-no-margin is-light">Install {{ snap_title }}<br/> on <
       <div class="p-card--highlighted" style="position: relative;">
         <div class="row">
           <div class="details-block">
-            <div class="p-snap-heading">
-              {% include 'store/snap-details/_snap_heading_data.html' %}
-
-              <div class="p-snap-install-buttons">
+            {%- call(slot) package_header() -%}
+              {%- if slot == 'buttons' -%}
```
alvaromateo, line 104: Nitpick: Are you planning on later adding other options for the slot?
Cause for now I only saw 'buttons' and maybe you could get away with just a comment and using `caller()` without any parameter, always rendering what is passed and getting rid of this condition.

edisile (approved):

templates/store/snap-distro-install.html
bartaz: Good question. Indeed, with it being the only slot it would make sense to keep it unnamed.
But I wanted to keep it consistent with Vanilla that uses named slots, and I am thinking about extending it to multiple slots as a next step (one for buttons, the other for publisher's meta-data).

PR #5231, opened 2025-07-24
Update TiCS action to use self-hosted runners

edisile:
## Done
Updated TiCS action with the following:
1. use a self-hosted runner
2. use `tiobe/tics-github-action` rather than rolling our own script
3. enable manual triggering
4. fix wrong indentation on some blocks

## How to QA
There's no real way to QA it except by triggering the action and letting it run

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): Here's a successful run of the action https://github.com/canonical/snapcraft.io/actions/runs/16493890465

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -2,6 +2,7 @@ name: Test coverage
 on:
   schedule:
     - cron: "0 22 * * *"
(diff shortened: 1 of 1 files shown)


steverydz (approved):
LGTM 👍

PR #5247, opened 2025-07-31
fix: avoid exposing internal errors

ilayda-cp:
## Done
log stack traces internally, return generic error messages to avoid exposing stack traces

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-24170

### DIFF
--- a/webapp/publisher/snaps/release_views.py
+++ b/webapp/publisher/snaps/release_views.py
@@ -88,10 +88,14 @@ def post_release(snap_name):
         if api_response_error_list.status_code == 404:
             return flask.abort(404, "No snap named {}".format(snap_name))
         else:
-            response = {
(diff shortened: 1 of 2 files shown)


webapp/publisher/snaps/release_views.py
```diff
@@ -88,10 +88,14 @@ def post_release(snap_name):
...
+            return (
+                flask.jsonify(
+                    {
+                        "errors": api_response_error_list.errors,
+                    }
+                ),
+                400,
+            )
 
     return flask.jsonify(response)
```
terameta, line 100: Do we still need this line here?

ilayda-cp: Yes, it's for returning successful responses

PR #5313, opened 2025-08-20
chore: use Vite bundle for cookie-policy.js

edisile:
## Done
Changed cookie-policy.js bundle to the one build by Vite

## How to QA
- open demo site
- you should see the cookie modal
- accept cookies
- the modal should close

## Testing
- [x] This PR has tests -> run-cypress tests cookie modal behavior

## Issue / Card
Fixes [WD-24758](https://warthogs.atlassian.net/browse/WD-24758)

## Screenshots

[WD-24758]: https://warthogs.atlassian.net/browse/WD-24758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,10 +31,12 @@ ADD static/js static/js
 ADD webpack.config.js .
 ADD webpack.config.entry.js .
 ADD webpack.config.rules.js .
+ADD vite.config.js .
 ADD tsconfig.json .
 ADD babel.config.json .
 RUN yarn install
 RUN yarn run build-js
+RUN yarn run build-js-vite
 
 # Build the production image
 # ===


templates/_base-layout.html
```diff
@@ -79,7 +79,7 @@
 
     {% block header %}
     <!-- JS that is needed right away -->
-    <script src="{{ static_url('js/dist/cookie-policy.js') }}"></script>
+    <script type="module" src="{{ static_url('js/dist/vite/cookie-policy.js') }}"></script>
```
bartaz, line 82: So, Vite bundle exposes a module, not a global?

edisile: Yes, Vite bundles into ES modules by default. This module specifically doesn't expose anything, but the bundle might contain `import` statements depending on the code splitting Vite does, so we need the `type="module"` regardless.

Some other modules will expose variables in the global scope (it's something we will fix later when looking into bundle optimization), but will still need the module annotation for the same reason.

PR #5318, opened 2025-08-21
feat: Add theme switcher to brand store and publisher pages

steverydz:
## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #


static/sass/styles.scss
```diff
@@ -505,3 +505,7 @@ html {
 .l-main {
   position: relative;
 }
+
+.read-only-dark {
+  color: var(--vf-color-text-inactive)
```
bartaz: While both work, these CSS variables are kind of internal to Vanilla, and are exposed as SCSS variables as well.

So it may be a bit safer to use their SCSS equivalents, such as `$colors--theme--text-inactive`.

All theme colors are listed here:
https://github.com/canonical/vanilla-framework/blob/main/scss/_settings_colors.scss#L237-L293

edisile (changes requested):
Looks good code-wise, but there's a small visual bug in the publisher view

When hovering an image in dark mode, the "Delete" button appears with a white icon on a white background:
Image
 
-------

It might be useful to get a review from a designer as well, but here are a couple other nitpicks on things I'm not a fan of UI-wise:
- These placeholders should probably be a dark color when using dark mode:
Image

- I feel that the snap icon section should match the current theme by default when you first open the page, it looks quite jarring otherwise
Image

steverydz:
@bartaz @edisile I've addressed both of your comments/concerns so can you please re-review. I'll request a design review also.

edisile (approved):
LGTM 👍 Thank you!

leiaru-c:
Tested on Chrome + Windows 11 (I know --') and saw the following:

- When selecting dark mode and going through the tutorial bit on the store, the tooltips seem to have an issue where the arrow is still white instead of dark background color
image

- When selecting system, it follows the browser preference, not the system preference (not a bug, just a thing of note I assumed).
System - Dark / Store - System / Browser - Light
image

System - Light / Store - System / Browser - Dark
image

Besides the first mentioned issue, it seems to work as intended :)

PR #5326, opened 2025-08-21
chore: use Vite bundle for publisher-details.js

edisile:
## Done
Changed publisher-details.js bundle to the one built by Vite
also fixed one of the Cypress tests for this bundle removing a typo

## How to QA
There's no real way to QA this: the bundle is only loaded in templates/store/publisher-details.html when publisher is one of ["kde", "snapcrafters", "jetbrains"], and when there's a blog_slug attribute in the publisher details context. At the moment none of these publishers have the blog_slug attribute set, so the bundle won't be loaded.

I forced it to be loaded in order to get a run of the Cypress tests to demonstrate the bundle is working as intended (the run in question is [HERE](https://github.com/canonical/snapcraft.io/actions/runs/17131569168/job/48596973463?pr=5326#step:7:164)). Once the test run is completed the condition on blog_slug was enabled again.

To test this locally you should move the <script> tag outside of the `if blog_slug` condition, load a publisher page (e.g. /publisher/kde) and check in the console that window.snapcraft.public.publisherDetails.snapDetailsPosts is a function

## Testing
- [x] This PR has tests -> run-cypress tests that the window.snapcraft.public.publisherDetails.snapDetailsPosts function exists (successful run [HERE](https://github.com/canonical/snapcraft.io/actions/runs/17131569168/job/48596973463?pr=5326#step:7:164))

## Issue / Card
Fixes WD-24769

### DIFF
--- a/templates/store/publisher-details.html
+++ b/templates/store/publisher-details.html
@@ -178,9 +178,8 @@
 {% endblock %}
 
 {% block scripts_includes %}
-  <script type="module" src="{{ static_url('js/dist/vite/publisher-details.js') }}"></script>
   {% if blog_slug %}
-    <!-- TODO: move the script back in this if clause -->
+    <script type="module" src="{{ static_url('js/dist/vite/publisher-details.js') }}"></script>
   {% endif %}
 {% endblock %}
 


steverydz (approved):
LGTM 👍

PR #5331, opened 2025-08-22
feature: Vite dev server integration (part 1)

edisile:
This is the endgame of our Vite migration, specifically pt. 1 of the trilogy of PRs for Vite dev server integration.
This PR only adds the base code for interfacing with Vite but doesn't actually use it. Look forward to the sequel PRs for that (pt. 2 will be actually replacing the <script> tags with the appropriate `vite_import` calls, pt. 3 will be about CSS)

## Done
- created Vite dev server integration helpers
- created Vite static build integration for production
- created template helper function for importing Vite resources
- added env vars for managing Vite configuration
- added unit tests for dev and prod mode integrations

## How to QA
no need to QA, code added in this PR does nothing at the moment

## Testing
- [x] This PR has tests

## Issue / Card
Fixes WD-25632


templates/_base-layout.html
```diff
@@ -23,6 +23,8 @@
 
     <link rel="stylesheet" href="{{ static_url('css/styles.css') }}" />
 
+    {{ vite_dev_tools() }}
```
bartaz: Should this be limited to dev environment only, or does it have support for production (and some use there) as well?

webapp/handlers.py
```diff
@@ -130,6 +130,13 @@
     "'unsafe-hashes'",
 ]
 
+# Vite integration
+if ENVIRONMENT == "devel":
+    CSP["script-src-elem"].append("localhost:5173")
```
bartaz: I guess the hardcoded port is default Vite port? Is there any chance it can be different in some environments?

templates/_base-layout.html
edisile: In production mode it does nothing, it just returns an empty string. I'll put it into an `if is_development` clause to make clearer

webapp/handlers.py
edisile: Yes by default it's 5173, but it can be set to custom values if needed. Should I add an env variable to do this?

templates/_base-layout.html
edisile: Fixed

webapp/handlers.py
edisile: Added a new VITE_PORT env var and related code that handles custom/missing values

edisile:
Rebased the branch on main, should be finally ready for review and merging

steverydz (approved):
LGTM 👍

PR #5335, opened 2025-08-27
fix: load unprefixed env vars explicitly

edisile:
Fix an issue where the `webapp.config` module's variables do not reflect those available in the Flask `app.config` object created by flask_base

## Done
call `load_plain_env_variables` explicitly in the `webapp.config` module

## How to QA
This can only be tested locally
- checkout this branch
- in the .env.local file, add an env variable prefixed by FLASK_
  - a good candidate for this is FLASK_SENTRY_DSN as it is copied in a `<script>` tag on each page
- `dotrun` it
- visit http://localhost:8004
- in the inspector, look for a script that contains `Raven.config`
- the first argument of the function call should be the string that you set as FLASK_SENTRY_DSN

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -9,8 +9,11 @@ import webapp.api
 
 from talisker import logging
 
-from canonicalwebteam.flask_base.app import FlaskBase
+# We import the config module before anything else to make sure env vars are
+# loaded properly and the FLASK_* prefix is stripped before they are parsed
 import webapp.config
+
+from canonicalwebteam.flask_base.app import FlaskBase
 from webapp.blog.views import init_blog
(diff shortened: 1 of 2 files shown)


webapp/app.py
```diff
@@ -10,6 +10,7 @@
 from talisker import logging
 
 from canonicalwebteam.flask_base.app import FlaskBase
+import webapp.config
```
bartaz, line 14: I kinda agree with copilot here, maybe good to leave some comment for why the config needs to be imported here.

bartaz (approved):
LGTM, with a small suggestion for clarity.

PR #5342, opened 2025-09-02
feature: Vite dev server integration (part 2)

edisile:
Part 2 of the Vite dev server integration, this PR enables the actual Vite dev server in dev mode and replaces the JS script tags pointing to the _compiled files_ with calls to the new `vite_import` helper that point to the _source files_.
For production mode, since now all entry points use `vite_import` we don't need to declare them in a separate file anymore, we can parse them at build time via a Vite plugin and inject them in the config automatically.

## Done
- added Vite dev server scripts in dev mode
- changed script tags to `vite_import` calls
- added automatic entry point detection plugin

### PROD mode
most prod mode checks are covered by the `run-cypress` action, but some manual testing is still needed for publisher and brand store pages
- Check interactive functionality on Demo (try out featured snaps category, first snap flow, etc.)
- double check that publisher and brand store pages load and render correctly

### DEV mode
- checkout this branch
- `dotrun` it
> If you're running dotrun on macOS or Windows, you will have to use a different command to launch dotrun with an additional argument in order to get JavaScript code working:
```bash
dotrun -p 5004:5004
```
- wait for server to be up
- run `docker run --network="host" -v .:/app -w /app cypress/base:22.18.0 bash "-c" "npx cypress install && yarn run test-e2e"`
- check publisher and brand store pages render correctly

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes WD-25632

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,11 @@ dotrun
 
 Once the server has started, you can visit `http://127.0.0.1:8004` in your browser. You stop the server using `<ctrl>+c`.
 
+> If you're running dotrun on macOS or Windows, you will have to use a different command to launch dotrun with an additional argument in order to get JavaScript code working:
+```bash
+dotrun -p 5004:5004
+```
+
 For more detailed local development instructions, see [HACKING.md](HACKING.md).
 
 ## Brand stores


steverydz:
I'm getting `Uncaught ReferenceError: snapcraft is not defined` in the browser console and it seems to be requesting `.ts` files that don't exist

pr

edisile:
@steverydz could it be a CORS issue? I just realized that visiting 0.0.0.0 rather than localhost triggers this error. I just pushed a patch that fixes it (the .ts files should exist, they're the source files we build the bundles from, it's just the GET request to Vite that's failing for some reason)

steverydz:
I'm getting the same issue with the latest patch and using any host (localhost, 0.0.0.0, localhost)

edisile:
Rebased on main and added a partial fix for the issue blocking connections to the dev server

@steverydz the issue was caused by dotrun not using Docker host networking on macOS and not exposing the 5004 port. To properly test this PR (and to run the project in dev mode if this branch gets merged) you would have to run `dotrun -p 5004:5004` to explicitly map the port Vite's server runs on. I updated the README and this PR's description with a note to reflect this issue.

I'm aware this isn't the nicest way to do it, but it's the cleanest one I found as a temporary solution while I'm looking for a way to automatically map other ports via dotrun

PR #5358, opened 2025-09-09
chore: cleanup package.json

edisile:
## Done
- Fixed transitive dependency resolution for `minimatch` (it was fixed to 3.0.5 for all packages which broke some transitive dependencies that needed v5 or v9+)
- Removed unused dependencies from package.json:
	- `@tarekraafat/autocomplete.js`
	- `clipboard` (bundle would get copied to `static/js/modules`, but is never linked in HTML so it's unused)
	- `fast-average-color`
	- `markdown-it`
	- `polylabel`
	- `vite-plugin-css-injected-by-js` (leftover from Vite migration phase 1)
	- `watch-cli` (leftover from old watch scripts)
- Remove unused `clipboard.js` code in "first snap flow"
- Remove script that copies `d3.min.js` to `static/js/modules` (it's never linked in HTML)

## How to QA
If tests pass and the demo site works it means the JS build was ok

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): cleanup

## Issue / Card
Fixes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,9 @@
   },
   "resolutions": {
     "lodash": "4.17.21",
-    "minimatch": "3.1.2 || 5.1.6 || 9.0.5"
+    "eslint/**/minimatch": "3.1.2",
+    "eslint-*/minimatch": "3.1.2",
+    "@typescript-eslint/**/minimatch": "9.0.5",
+    "ts-jest/ejs/jake/filelist/minimatch": "5.1.6"
   }
 }
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test-python": "yarn run lint-python && FLASK_DEBUG=0 python3 -m unittest discover tests",
     "test-python-job": "SECRET_KEY=insecure_secret_key coverage run  --source=. -m unittest discover tests",
     "build": "yarn run copy-3rd-party-js && vite build",
-    "copy-3rd-party-js": "mkdir -p static/js/modules && cp node_modules/d3/dist/d3.min.js static/js/modules && cp node_modules/clipboard/dist/clipboard.min.js static/js/modules && cp node_modules/raven-js/dist/raven.min.js static/js/modules",
+    "copy-3rd-party-js": "mkdir -p static/js/modules && cp node_modules/raven-js/dist/raven.min.js static/js/modules",
     "serve": "./entrypoint 0.0.0.0:${PORT}",
     "start": "concurrently --raw 'yarn run copy-3rd-party-js' 'yarn run watch' 'yarn run serve'",
     "watch": "vite",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -128,6 +121,9 @@
   },
   "resolutions": {
     "lodash": "4.17.21",
-    "minimatch": "3.0.5"
+    "eslint/**/minimatch": "3.1.2",
```
bartaz, line 124: Do we still need resolutions for minimatch?

static/js/public/first-snap-flow.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
-/* globals ClipboardJS, ga */
+/* globals ga */
```
bartaz, line 2: Is this file still used anywhere? First snap flow have been removed a while ago (replaced by docs).

bartaz: Ah, I think we still have the (likely unused) old templates around. There is a backlog issue to clean it. https://warthogs.atlassian.net/browse/WD-18896

So I guess the current change is fine for the scope of this PR.

package.json
```diff
@@ -11,7 +11,7 @@
     "test-python": "yarn run lint-python && FLASK_DEBUG=0 python3 -m unittest discover tests",
     "test-python-job": "SECRET_KEY=insecure_secret_key coverage run  --source=. -m unittest discover tests",
     "build": "yarn run copy-3rd-party-js && vite build",
-    "copy-3rd-party-js": "mkdir -p static/js/modules && cp node_modules/d3/dist/d3.min.js static/js/modules && cp node_modules/clipboard/dist/clipboard.min.js static/js/modules && cp node_modules/raven-js/dist/raven.min.js static/js/modules",
+    "copy-3rd-party-js": "mkdir -p static/js/modules && cp node_modules/raven-js/dist/raven.min.js static/js/modules",
```
bartaz, line 14: Out of scope of this PR, but seems that `raven-js` is deprecated and have been replaced by `@sentry/browser` https://www.npmjs.com/package/raven-js

static/js/public/first-snap-flow.ts
edisile: Yes, the script is still used for the "Learn how to snap an app in 30 minutes" section on the [home page](https://snapcraft.io/#language) and [/about/publish](https://snapcraft.io/about/publish#language)

package.json
edisile: I'll open an issue for it

package.json
edisile: I'm not sure, I kept them just to get pinned versions and be safe. The resolution to `3.0.5` was added as a fix to a RegEx DoS vulnerability (but we probably didn't need it anyway, this lib is never shipped/exposed to users and is only used for linting and tests, so the vulnerability wasn't impacting us)

package.json
edisile: opened #5361 ([WD-26706](https://warthogs.atlassian.net/browse/WD-26706))

PR #5364, opened 2025-09-11
redirects: route plugin index and .net plugin pages

medubelko:
More pages that weren't indexed in the original docs.

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): Production change


medubelko:
@abbiesims Please review. :)

PR #5370, opened 2025-09-17
chore: Migrate from raven-js to sentry browser

steverydz:
## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #


static/js/base/sentry.ts
```diff
@@ -0,0 +1,15 @@
...
+Sentry.init({
+  allowUrls: ["staging.snapcraft.io/static/js", "snapcraft.io/static/js/"],
+  denyUrls: [
+    "staging.snapcraft.io/static/js/modules",
+    "snapcraft.io/static/js/modules",
+  ],
+  dsn: window.SENTRY_DSN,
+  environment: "{{ ENVIRONMENT }}",
+  ignoreErrors: ["AbortError"],
+  release: "{{ COMMIT_ID }}",
```
edisile: Are these attributes being populated correctly? I guess this is the equivalent of the script we had in `_base-layout.html`, but since we're outside of Jinja `{{ }}` has no effect

templates/_base-layout.html
```diff
@@ -27,20 +27,6 @@
...
-    {% block scripts_modules %}{% endblock %}
-    <script src="{{ static_url('js/modules/raven.min.js') }}" defer></script>
-    <script>
-      window.addEventListener("DOMContentLoaded", function() {
-        Raven.config('{{ SENTRY_DSN }}', {
-          whitelistUrls: ['staging.snapcraft.io/static/js', 'snapcraft.io/static/js/'],
-          ignoreUrls: ['staging.snapcraft.io/static/js/modules', 'snapcraft.io/static/js/modules'],
-          ignoreErrors: ['AbortError'],
-          release: '{{ COMMIT_ID }}',
-          environment: '{{ ENVIRONMENT }}'
```
edisile, line 39: Related to the comment above, setting `SENTRY_DSN`, `COMMIT_ID` and `ENVIRONMENT` in the `window` scope here and reading them from `window` in the Sentry init script should to the trick

steverydz:
@edisile Good call. Fixed.

edisile:
@steverydz I think you should also set `window.SENTRY_DSN` in `_base-layout.html`, at the moment it's being set only inside `templates/admin/admin.html` if I'm not mistaken

steverydz:
@edisile Good catch. Fixed

edisile (approved):
LGTM

PR #5374, opened 2025-09-18
Unify SPA routing

edisile:
## Done

- refactored the Brand stores and Publisher SPAs and merged them into a single root file
- refactored the React router declarations
- refactored the base application layout component for the Publisher part of the application
  - future work: migrate Brand store pages to this layout as well
- merged the two separate side navigation implementations we had previously
  - changed all `<a>` links to `NavLink`, no more useless page refreshes
  - refactored CSS code
  - added hover effects to the store selector list items
  - added QoL features:
    - support for light theme side nav (currently disabled)
    - preserve side nav `isCollapsed` and `isPinned` state between refreshes
  - implemented the [more recent side navigation guidelines](https://docs.google.com/document/d/1AeX_4NtUHxXDJgKqrZEpTXPJBIU9pyXxp0HJGcitX5g/edit?tab=t.0#heading=h.yij8xaij9lcy)

## How to QA
- open the demo site and log in
  - the /snaps page should load correctly
- look for any graphical glitches in the side nav
  - change browser window size, play around with the collapsed and pinned states
- from the /snaps page, open a brand store
  - there should be no page refresh
- refresh the page while in a brand store sub-section
  - the page should load correctly

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes WD-23713


static/js/publisher/components/StoreSelector/StoreSelector.tsx
```diff
@@ -10,14 +10,31 @@ type Props = {
   nativeNavLink?: boolean;
 };
 
+// TODO: refactor this and use Downshift to have proper a11y
```
steverydz, line 13: Perhaps make a ticket for this and link to it here so it can be tracked?

static/js/publisher/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,29 @@
...
+  const [storedValue, setStoredValue] = useState<T>(() => {
+    const item = window.localStorage.getItem(key);
+    return item ? JSON.parse(item) : initialValue;
+  });
+
+  // when state changes update local storage
+  useEffect(() => {
+    setTimeout(() => {
+      window.localStorage.setItem(key, JSON.stringify(storedValue));
+    }); // delay the write to let React do its job first
```
steverydz: I think it's good to put a numeric value of `0` here to make it clear that there is no time length intentionally

static/js/publisher/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+              <Route
+                path="models/:model_id/policies/create"
+                element={<Policies />}
+              />
+              <Route path="signing-keys" element={<SigningKeys />} />
+              <Route path="signing-keys/create" element={<SigningKeys />} />
+            </Route>
+          </Route>
+          {/* END brand store routes */}
+          {/* TODO: merge the two layouts */}
```
steverydz, line 115: Perhaps make a ticket for this and link to it here so it can be tracked

static/js/publisher/hooks/useSideNavigationData.ts
```diff
@@ -0,0 +1,58 @@
...
+      setBrandId(brandData?.["account-id"]);
+    } else {
+      setBrandId("");
+    }
+  }, [brandData]);
+
+  useEffect(() => {
+    if (brandStoresData) {
+      setBrandStores(brandStoresData);
+    }
```
steverydz: Should we be clearing this the brand stores like we are with the brand ID?

static/js/publisher/hooks/useSideNavigationData.ts
```diff
@@ -0,0 +1,58 @@
...
+  }, [brandStoresData]);
+
+  useEffect(() => {
+    if (publisherData) {
+      setPublisher(publisherData.publisher);
+    }
```
steverydz: Should we be clearing this the publisher like we are with the brand ID?

static/js/publisher/hooks/useSideNavigationData.ts
```diff
@@ -0,0 +1,58 @@
...
+  }, [publisherData]);
+
+  useEffect(() => {
+    if (validationSetsData) {
+      setValidationSets(validationSetsData);
+    }
```
steverydz: Should we be clearing this the validation sets like we are with the brand ID?

static/js/publisher/components/Navigation/Navigation.tsx
```diff
@@ -1,295 +1,95 @@
...
-      <header className="l-navigation-bar">
-        <div className="p-panel is-dark">
-          <div className="p-panel__header">
-            <Logo />
-            <div className="p-panel__controls">
-              <button
-                className="p-panel__toggle u-no-margin--bottom"
+      <AppNavigationBar>
+        <Panel
+          // dark
```
steverydz: Can this be removed if not needed?

static/js/publisher/index.tsx
edisile: [WD-27090](https://warthogs.atlassian.net/browse/WD-27090)

static/js/publisher/components/StoreSelector/StoreSelector.tsx
edisile: [WD-27089](https://warthogs.atlassian.net/browse/WD-27089)

static/sass/_snapcraft_l-application.scss
```diff
@@ -127,64 +107,85 @@
...
       &:last-child {
         border-bottom: 0;
       }
 
       a {
-        color: $color-x-light;
+        color: var(--vf-color-text-default);
         display: block;
         overflow: hidden;
+        padding: calc(0.4rem - 1px) 0.5rem;
```
bartaz: Such specific calculations could use a line comment with context on what are they calculating and why these specific values. Also, if possible, it should use Vanilla variables instead of hardcoded values (https://vanillaframework.io/docs/settings/spacing-settings#vertical-spacing).

bartaz: Oh, just noticed that this is not a new line, just moved. But overall, the comment still stands (at least in the scope you can provide context on).

static/js/publisher/components/Navigation/Navigation.tsx
edisile: Good catch! This is the reason why the top nav is light on mobile, I forgot to uncomment it 😅

static/sass/_snapcraft_l-application.scss
```diff
@@ -22,7 +9,7 @@
       }
 
       .logo-text {
-        color: #fff;
+        color: var(--vf-color-text-default);
```
bartaz: The CSS variables are kind of Vanilla internals, and were not meant for direct external consumption originally.

While technically there is nothing wrong with such approach, it was meant to be used via SCSS variables (that are processed into CSS variables on build time).
The list of all theme variables is in Vanilla color definitions: https://github.com/canonical/vanilla-framework/blob/main/scss/_settings_colors.scss#L237-L293

static/sass/_snapcraft_l-application.scss
```diff
@@ -127,64 +107,85 @@
...
           text-decoration: none;
         }
       }
     }
   }
 
-  // Add collapsible nav in application layout
-  @include app-p-icon;
+  .l-navigation {
+    &.is-collapsed {
```
bartaz: What's the reason for those overrides we are introducing? If we need to have them, they should be clearly commented in the code to understand what they are needed for.
With any override like this we risk that the upstream updates will not work as expected, so it's important to understand why each override is needed.

Maybe overrides can be avoided if we add some custom variant class name that is specific to our case? This could potentially also avoid using `!important` that is generally a bad practice and something that prevents CSS to work in native way.

static/sass/_snapcraft_l-application.scss
```diff
@@ -127,64 +107,85 @@
...
-      background-color: #444 !important;
+    @media (min-width: $breakpoint-small) and (max-width: $breakpoint-large) {
+      &:not(.is-pinned, :hover, :focus-within) {
+        .hide-collapsed {
+          display: none !important;
+        }
+      }
+    }
+
+    .p-panel {
```
bartaz: This seems like a variant of a panel. It could be cleaner to implement it as such (as `.p-panel--vertical` or `.p-panel.is-vertical`) and avoid dependancy on nesting inside navigation.

static/sass/_snapcraft_l-application.scss
```diff
@@ -228,51 +229,8 @@
...
-            }
-
-            &::before {
-              display: none !important;
-            }
-          }
-        }
-      }
-    }
+  .p-tooltip--right {
```
bartaz: Another very specific override. If we happen to change tooltip to any other variant it will stop working. 

If there are issues with z-index it would be good to identify them at source, or make this override better documented and applicable to any tooltip in the side nav.

bartaz (changes requested):
Some comments on SCSS changes.

static/sass/_snapcraft_l-application.scss
edisile: Thanks for pointing this out! I kept looking for the proper variables and I couldn't find them in the docs, I didn't think to look at the source directly

static/sass/_snapcraft_l-application.scss
edisile: Managed to get rid of it using a combination of utility classes on the panel component

static/sass/_snapcraft_l-application.scss
edisile: Changed selector to `[class^="p-tooltip"]` to cover all tooltip variants

static/sass/_snapcraft_l-application.scss
edisile: Replaced the magic numbers with Vanilla variables

static/sass/_snapcraft_l-application.scss
edisile: I was tricked into thinking `is-collapsed` was for desktop rather than mobile-only, in the end I replaced most of these with a custom `is-collapsed--desktop` class.
Also I made sure to split the React state associated with collapsing behavior into mobile and desktop variables

static/js/publisher/hooks/useSideNavigationData.ts
edisile: Updated this and all the others to use some form of `setBrandStores(brandStoresData || brandStoresState.init)`; it's pretty much the same as above, it's just more explicit about what happens if the result is empty

edisile:
Thank you for the thorough reviews!

@steverydz I agree there's a performance regression because of the additional bundle size, but as you said it's somewhat negligible. The idea behind this PR was to first merge the two bundles into one and then use React.lazy to split it by route. I decided to split this work in multiple PRs because otherwise it would have been much more difficult to review

steverydz (approved):
Thanks for the changes. LGTM 👍

bartaz:
The issue was likely not introduced here, but it may be a good place to fix it. It seems that there is a bit of misalignment in Snapcraft logo in the app layout.

The official spacing is what we have on home page:
image

That would look like this with just Snapcraft in the name:
image


In application layout it's all over the place:

On mobile, it's quite large and too far from orange tag:
image

On desktop, it's ok (not sure if ideal size and spacing, but it doesn't stand out at least):
image

But on very large screens (above 1680px) it gets misaligned, because fonts are scaled up, but logo is not.

image

static/js/publisher/hooks/useBrand.ts
```diff
@@ -4,6 +4,8 @@ function useBrand(id: string | undefined) {
   return useQuery({
     queryKey: ["brand", id],
     queryFn: async () => {
+      if (!id) return null;
```
bartaz: Do we expect it to be called without id (so that such silent returning of null is OK), or would it be better as an error?

static/js/publisher/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,33 @@
+import { useState, useEffect } from "react";
+
+function useLocalStorage<T = unknown>(
```
bartaz, line 3: Nice addition, but is it used anywhere? I can't seem to find the usage in the PR.

static/js/publisher/layouts/BrandStoreLayout.tsx
```diff
@@ -0,0 +1,83 @@
...
+            </div>
+          </div>
+        </div>
+      </main>
+    </div>
+  );
+}
+
+function BrandStoreLayout() {
+  useSideNavigationData();
```
bartaz, line 35: Not sure if I understand how it works here. We are not using any data that it returns?

static/js/publisher/hooks/useLocalStorage.ts
edisile: it's being used in `static/js/publisher/components/Navigation/Navigation.tsx`, the diff for that file is quite big so GitHub doesn't load it by default

static/js/publisher/hooks/useBrand.ts
edisile: It's called by the `useSideNavigationData` hook when loading the data for side navigation; if the route isn't a brand store this hook still gets called, and I can't put it in an if/else because of the hook rules. I tried thinking about other approaches but I couldn't come up with anything else...

static/js/publisher/layouts/BrandStoreLayout.tsx
edisile: Good catch, I refactored some things and stopped using the data returned from that hook altogether, so at this point it's pretty much just a helper for making multiple API calls and load their results into Jotai state vars

static/js/publisher/hooks/useBrand.ts
edisile: Added a comment to explain this behavior

static/js/publisher/layouts/BrandStoreLayout.tsx
edisile: I removed the useless return values and added a comment to explain what the hook does

static/js/publisher/hooks/useLocalStorage.ts
bartaz: Ah, sneaky GH hiding so many changes.

bartaz (approved):
Thanks for the updates. Looks good.

As for the logos inconsistencies I checked and the logo behaves exactly the same way now, so it's not any kind of regression related to this PR, so I think it's fine to address it separately, not to increase the scope of this PR.

edisile:
@bartaz thank you for the review! I added a patch for the logo issue in this PR, now it should have more consistent spacing on big screens. If you feel that it still needs some work I'll track the issue in a ticket and fix it in a separate PR

PR #5379, opened 2025-09-24
fix: Fix footer position on docs page

steverydz:
## Done
Fixes issue where the footer isn't fixed to the bottom of the screen

## How to QA
- Run locally (`dotrun` or `dotrun -p 5004:5004` on MacOS)
- Go to http://localhost:8004/docs/network-bind-interface
- Check that the footer is fixed to the bottom of the screen

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): Visual change

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-27344
Fixes https://github.com/canonical/snapcraft.io/issues/5378

### Before
before

### After
after

### DIFF
--- a/static/sass/_snapcraft_p-docs.scss
+++ b/static/sass/_snapcraft_p-docs.scss
@@ -9,4 +9,8 @@
   .emoji {
     height: 1rem;
   }
+
+  .l-docs {
+    flex: 1;
+  }
 }


static/sass/_snapcraft_p-docs.scss
```diff
@@ -9,4 +9,8 @@
   .emoji {
     height: 1rem;
   }
+
+  .l-docs {
```
bartaz, line 13: Is the bug happening in Vanilla as well? If the fix is needed, maybe worth upstreaming it. Unless it's caused by some local custom stuff.

bartaz: Well, too late for that, but it seems that we have `l-site` class on body which allows for a sticky footer, if we use `l-footer--sticky` class on it (alongside the one from documentation layout). While not a documented use of those, at least it avoids the custom CSS fix.

steverydz: @bartaz Good catch. I've updated this in [this PR](https://github.com/canonical/snapcraft.io/pull/5380)

PR #5394, opened 2025-10-01
chore: Update flask-base to version 3.0.0

steverydz:
## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #


webapp/app.py
```diff
@@ -4,15 +4,8 @@
...
 
-import talisker.requests
-import webapp.api
-
-from talisker import logging
-
 # We import the config module before anything else to make sure env vars are
 # loaded properly and the FLASK_* prefix is stripped before they are parsed
-import webapp.config
-
```
edisile, line 15: This shouldn't be removed as it would break the app in charmed environments, there's a comment in `config.py` that explains exactly why this happens, but maybe I should have added it here as well to make it more clear, sorry

To fix the issue we can either revert this change, or call `load_plain_env_variables` at the top of this file

steverydz: @edisile Fixed. I've had to add a comment for it to pass the lint check. Not sure how it passed before?

edisile: No idea tbh 🤷‍♂️

webapp/handlers.py
```diff
@@ -49,20 +48,6 @@
 
 from datetime import datetime
 
-badge_counter = prometheus_client.Counter(
```
bartaz, line 52: Are we completely losing those metrics, or will they now be automatically part of 12 factor app o11y?

HACKING.md
```diff
@@ -82,15 +82,6 @@ Install the [`dotrun`](https://snapcraft.io/dotrun) snap.
 dotrun test
 ```
 
-## Status checks and prometheus metrics
-
-[Talisker](https://talisker.readthedocs.io/en/latest/) provides a bunch of useful status checks and metrics about the running application. Some of this information is sensitive and so to access it you need to run the site with your IP address mentioned in the `TALISKER_NETWORKS` variable.
-
-Now visit http://127.0.0.1:8004/_status to see the endpoints provided by Talisker. Useful ones include:
```
bartaz, line 89: I wonder if we have any external checks or jobs that relied on the _status endpoint.

But I guess we will soon know if we do ;)

requirements.txt
```diff
@@ -24,6 +24,8 @@ ruamel.yaml==0.18.5
 vcrpy-unittest==0.1.7
 user-agents==2.2.0
 dnspython==2.6.1
+werkzeug==2.3.8
```
bartaz, line 27: Why do we need this as separate dependency?

webapp/publisher/snaps/views.py
```diff
@@ -356,11 +355,7 @@ def get_user_snaps():
 def get_snap_build_status():
     try:
         account_info = dashboard.get_account(flask.session)
-    except (StoreApiError, ApiError) as api_error:
-        logging.getLogger("talisker.wsgi").error(
```
bartaz, line 360: Is there any alternative to talisker logging now? It seems that we may lose a bit of info in the logs if we ditch that completely?

alvaromateo: With the upgrade to flask-base 3.1.0 we will get the missing logs.
I'm still working on finishing the PR with the logging changes and the new release, but the upgrade to this new version should be straightforward.

webapp/blog/views.py
```diff
@@ -13,7 +13,7 @@
 
 
 def init_blog(app, url_prefix):
-    session = talisker.requests.get_session()
+    session = requests.Session()
```
alvaromateo, line 16: Instead of creating a new session maybe we could get the current one from flask request object (to inherit all the attributes of the current session - which I don't know if they might be useful to the blog view).

webapp/docs/views.py
```diff
@@ -14,7 +14,7 @@
 
 
 def init_docs(app, url_prefix):
-    session = talisker.requests.get_session()
+    session = requests.Session()
```
alvaromateo, line 17: Same as previous comment: why not use the current session taken from flask request?

webapp/publisher/snaps/views.py
alvaromateo: Nevertheless, we should still keep some logging. The new way will be:
```
import logging
logger = logging.get_logger(__name__)
logger.error("msg")
```
And flask-base will take care of formatting and all the rest.
Instead of deleting completely the log maybe you can leave it like this to prepare the app for the future flask-base 3.1.0 version.

webapp/snapcraft/views.py
```diff
@@ -45,11 +36,7 @@ def get_account_json():
             response.headers["Cache-Control"] = "no-store"
 
             return response
-        except Exception as e:
-            logging.getLogger("talisker.wsgi").error(
```
alvaromateo, line 49: You can leave the log according to the comment here:
https://github.com/canonical/snapcraft.io/pull/5394/files#r2401178196

webapp/store/views.py
```diff
@@ -19,7 +18,7 @@
 import os
 from webapp.extensions import csrf
 
-session = talisker.requests.get_session(requests.Session)
+session = requests.Session()
```
alvaromateo, line 21: You can use flask's request session inside each view.

alvaromateo (commented):
Just a few comments/questions, but nice work getting rid of Talisker!

requirements.txt
steverydz: @bartaz Without it we get the following error: `ImportError: cannot import name 'url_quote' from 'werkzeug.urls'`

bartaz: Are we importing it directly, or is it some internal flask-base thing? I thought werkzeug was supposed to be included in flask-base.

webapp/handlers.py
steverydz: @bartaz I believe this is something we'll be able to get with COS Lite. This doesn't work once talisker is removed.

requirements.txt
steverydz: It's related to our use of `OpenID` in the login view so I think specific to our implementation as it doesn't seem to be an issue in charmhub. Given this PR is meant to only be updating this package I didn't want to mess with logic, especially around auth. I think further investigation is needed but I think out of scope for this PR.

webapp/blog/views.py
steverydz: @alvaromateo I'm not sure I understand. Using `flask.session` doesn't return the response in the same way that using `requests` does? Do you have an example of what you mean?

alvaromateo: Yes, I meant flask.session (https://flask.palletsprojects.com/en/stable/api/#flask.session).
I see that it is a proxy object and not the same type you need; my bad, I didn't check the documentation properly.
You can ignore this comment (and the others like it) :sweat_smile: Sorry!

alvaromateo (approved):
LGTM

PR #5402, opened 2025-10-08
feat: updates rss feed

M7mdisk:
## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #

### DIFF
--- a/tests/tests_feeds.py
+++ b/tests/tests_feeds.py
@@ -1,0 +1,80 @@
+import unittest
+from unittest.mock import patch, Mock
+from webapp.app import create_app
+
+
+class TestFeeds(unittest.TestCase):
+    def setUp(self):
+        self.app = create_app(testing=True)
+        self.client = self.app.test_client()
+
+    @patch('webapp.feeds.feeds.session.get')
+    def test_feeds_updates_success(self, mock_get):
+        """Test successful RSS feed generation using feedgen"""
+        # Mock API response
+        mock_response = Mock()
+        mock_response.json.return_value = {
+            "page": 1,
+            "size": 2,
+            "snaps": [
+                {
+                    "name": "test-snap",
+                    "title": "Test Snap",
+                    "summary": "A test snap",
+                    "description": "This is a test snap for testing",
+                    "publisher": "Test Publisher",
+                    "license": "MIT",
+                    "version": "1.0.0",
+                    "last_updated": "Thu, 02 Oct 2025 22:07:58 GMT",
+                    "icon": "https://example.com/icon.png",
+                    "snap_id": "test-snap-id-123",
+                    "media": [
+                        {
+                            "type": "screenshot",
+                            "url": "https://example.com/screenshot.png"
+                        }
+                    ]
+                }
+            ]
+        }
+        mock_response.raise_for_status.return_value = None
+        mock_get.return_value = mock_response
+
+        # Make request to RSS endpoint
+        response = self.client.get('/feeds/updates')
+        
+        # Check response
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content_type, 'application/rss+xml; charset=utf-8')
+        
+        # Check RSS content
+        content = response.get_data(as_text=True)
+        self.assertIn('<?xml version="1.0" encoding="UTF-8"?>', content)
+        self.assertIn('<rss version="2.0">', content)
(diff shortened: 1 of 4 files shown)


webapp/feeds/feeds.py
```diff
@@ -0,0 +1,139 @@
...
+
+    fg = FeedGenerator()
+    fg.title("Snapcraft - recently updated snaps")
+    fg.link(href="https://snapcraft.io/store", rel="alternate")
+    fg.description("Recently updated snaps published on Snapcraft")
+    fg.language("en")
+    fg.docs("http://www.rssboard.org/rss-specification")
+    fg.generator("python-feedgen")
+
+    size = int(flask.request.args.get("size", "50"))
```
edisile, line 90: I haven't thought about this while reading the spec, but what is the risk of someone using a very large value for the `size` parameter? Could it crash or significantly slow down the recommendations service?

webapp/feeds/feeds.py
```diff
@@ -0,0 +1,139 @@
...
+    if snap.get("media"):
+        for media in snap["media"]:
+            if (
+                media.get("type") == "screenshot"
+                and media.get("url")
+                and is_safe_url(media["url"])
+            ):
+                media_url = escape(media["url"])
+                description_parts.append(
+                    f'<img src="{media_url}" alt="Screenshot">'
```
edisile:
```suggestion
                    f'<img src="{media_url}" alt="">'
```
For a11y reasons it would be better to leave this empty if there isn't an actually meaningful alternative text value that we can use, otherwise screen reader users will get a bunch of hints saying "screenshot" that don't help them and might even be confusing when there are many images

webapp/feeds/feeds.py
```diff
@@ -0,0 +1,139 @@
...
+        return datetime.now(timezone.utc)
+
+
+def create_snap_description(snap):
+    """Create HTML description for RSS item."""
+    description_parts = []
+
+    if snap.get("icon") and is_safe_url(snap["icon"]):
+        icon_url = escape(snap["icon"])
+        description_parts.append(f'<img src="{icon_url}" alt="Snap icon">')
```
bartaz: Not sure what are the best practices for including images in RSS feed content, but I feel that we should at least try to limit the max size.

With the icon it should be straightforward, because I think it has to be a square (?). So we could do `` (or whatever size feels fine for an icon.

webapp/feeds/feeds.py
M7mdisk: It is capped at 50 by the recommendation service https://github.com/canonical/snap-recommendation-service/blob/415b957b34ed1e03eb1360ba9b43a39cb605ffc1/snaprecommend/api.py#L326

webapp/feeds/feeds.py
```diff
@@ -0,0 +1,139 @@
...
+def parse_snap_date(date_string):
+    """Parse date string from API to datetime object."""
+    try:
+        dt = datetime.strptime(date_string, "%a, %d %b %Y %H:%M:%S %Z")
+        return dt.replace(tzinfo=timezone.utc)
+    except ValueError:
+        return datetime.now(timezone.utc)
+
+
+def create_snap_description(snap):
```
bartaz, line 38: Probably a bit out of scope for this PR, but would be interesting if we could leverage Flask templating for this, instead of concatenating arbitrary HTML strings.

tests/tests_feeds.py
```diff
@@ -0,0 +1,62 @@
...
+        # Make request to RSS endpoint
+        response = self.client.get("/feeds/updates")
+        # Check response
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.content_type, "application/rss+xml; charset=utf-8"
+        )
+        # Check RSS content
+        content = response.get_data(as_text=True)
+        self.assertIn("<?xml version='1.0' encoding='UTF-8'?>", content)
```
bartaz, line 48: This test seems to be purely text-based.

Would it be possible to test XML more directly. At least validate returned XML, or parse XML and assert the resulting data?

M7mdisk: Done

webapp/feeds/feeds.py
M7mdisk: its a bit weird because only part of it is HTML, the rest is in a format I'd like `feedgen` to take care of (I don't want to mess up the RSS format)

PR #5417, opened 2025-10-13
fix: Show repo owner in builds repo selector

steverydz:
## Done
Adds the user or organisation name to the list of repo suggestions to avoid confusion over the repo owner when there are two repos with the same name.

## How to QA
- Run locally (builds don't work on demos)
- Go to /<snap_name>/builds
- In the "Select a repository" field in the autocomplete, the org/username for each snap should appear under the repo name

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-27256

## Screenshots
example

### DIFF
--- a/webapp/api/github.py
+++ b/webapp/api/github.py
@@ -216,7 +216,11 @@ class GitHub:
             repositories.extend(next_page)
 
         repos = [
-            {**repo, "owner": repo.get("nameWithOwner", "").split("/")[0]} if "nameWithOwner" in repo and repo.get("nameWithOwner") else {**repo, "owner": None}
+            (
+                {**repo, "owner": repo.get("nameWithOwner", "").split("/")[0]}
+                if "nameWithOwner" in repo and repo.get("nameWithOwner")
+                else {**repo, "owner": None}
+            )
             for repo in repositories
         ]
 


static/js/publisher/pages/Builds/RepoSelector.tsx
```diff
@@ -332,7 +334,9 @@ function RepoSelector({ githubData, setAutoTriggerBuild }: Props) {
             ))}
           <datalist id="repo-list">
             {repos.map((repo: Repo) => (
-              <option value={repo.name} key={repo.name} />
+              <option value={repo.name} key={repo.name}>
+                {repo.nameWithOwner}
```
edisile: small nitpick: I feel that this makes search a bit clunky if you're only looking through repos you're the owner of. Would it make sense to only display `nameWithOwner` for repos whose owner isn't the current user/org?

steverydz: @edisile Agreed. I think that would be better. I've updated it so if the owner is different from the selected user/org it says "Owned by <owner>", otherwise it says nothing.

PR #5426, opened 2025-10-20
chore: Update text on verified domain tooltip to account for verified publisher

steverydz:
## Done
Amends the text for verified domains so verified publishers don't show the disclaimer which non-verified publishers have.

## How to QA
- Go to https://snapcraft-io-5426.demos.haus/gimp
- Hover over the icon next to the ownership verified domain under "Websites" and check that the tooltip says: "The publisher has verified that they own this domain"
- Go to https://snapcraft-io-5426.demos.haus/steve-test-snap
- Hover over the icon next to the ownership verified domain under "Websites" and check that the tooltip says: "The publisher has verified that they own this domain. It does not guarantee the Snap is an official upload from the upstream project."

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-30124

### DIFF
--- a/templates/store/snap-details/_details.html
+++ b/templates/store/snap-details/_details.html
@@ -109,10 +109,9 @@
     (Ownership verified) 
     <span class="p-tooltip--btm-right" aria-describedby="verified-explanation">
       <i class="p-icon--information"></i>
-      <span class="p-tooltip__message" role="tooltip" id="verified-explanation">The publisher has verified that they own this domain. 
-  It does not guarantee the Snap is an official upload from the 
-  upstream project.</span>
-    </span>
+      <span class="p-tooltip__message" role="tooltip" id="verified-explanation">The publisher has verified that they own this domain{%- if developer_validation != VERIFIED_PUBLISHER -%}.
+It does not guarantee the Snap is an official upload
+from the upstream project.{%- endif -%}</span>
   </template>
 
   <script>


templates/store/snap-details/_details.html
```diff
@@ -109,9 +109,13 @@ <h2 class="p-modal__title" id="modal-title"><i class="p-icon--warning" style="wi
     (Ownership verified) 
     <span class="p-tooltip--btm-right" aria-describedby="verified-explanation">
       <i class="p-icon--information"></i>
-      <span class="p-tooltip__message" role="tooltip" id="verified-explanation">The publisher has verified that they own this domain. 
-  It does not guarantee the Snap is an official upload from the 
-  upstream project.</span>
+      {% if developer_validation == "verified" %}
```
bartaz: I think I've seen this using global const instead of string comparison in other places, like `{%- if developer_validation == VERIFIED_PUBLISHER -%}`, would that work?

steverydz: @bartaz You're right, that works. Thanks. I've updated it.

templates/store/snap-details/_details.html
```diff
@@ -109,9 +109,13 @@ <h2 class="p-modal__title" id="modal-title"><i class="p-icon--warning" style="wi
...
     <span class="p-tooltip--btm-right" aria-describedby="verified-explanation">
       <i class="p-icon--information"></i>
-      <span class="p-tooltip__message" role="tooltip" id="verified-explanation">The publisher has verified that they own this domain. 
-  It does not guarantee the Snap is an official upload from the 
-  upstream project.</span>
+      {% if developer_validation == VERIFIED_PUBLISHER %}
+        <span class="p-tooltip__message" role="tooltip" id="verified-explanation">The publisher has verified that they own this domain</span>
+      {% else %}
+        <span class="p-tooltip__message" role="tooltip" id="verified-explanation">The publisher has verified that they own this domain.
+It does not guarantee the Snap is an official upload
```
bartaz, line 113: With just this sentence being the difference, would it work to just wrap just this sentence in "if publisher is not verified" (instead of whole tooltip being difference inside if/else statement?

steverydz: @bartaz I did try that, but with the text inside a tooltip being `pre` it messed up the spacing and indentation and looked broken

bartaz: Ah, true.

I guess you could try doing `{%- if … -%}…{%- endif -%}` as dashes prevent Jinja from rendering empty lines in place of its commands, but it may still we fragile in `pre`.

PR #5434, opened 2025-10-23
Flask-Vite integration refactor before upstreaming

edisile:
Refactoring the Vite integration before splitting it off into its own package

## Done
- refactored the external interface exposed by the integration, now it's a Flask extension
- removed dependency on `webapp.config`, config values are now loaded from `app.config` (which in our case still comes from `webapp.config`)
  - `VITE_MODE: "development" | "production"` - type of Vite integration to use
  - `VITE_PORT: int` - port where Vite is expected to respond
  - `VITE_OUTDIR: str` - file system path where the Vite output is expected
- removed Vite template helpers from snapcraft.io's helpers
- removed `vite_dev_tools` helper, dev tools are now injected automatically when running in dev mode
- improved `vite_import`
  - support for more file extensions
  - clearer code
  - support for unknown file extensions (they just log an error but otherwise fail silently)
    - (maybe it should raise an exception, at least in development mode?)
- updated tests

### Minimal usage example
```python
# app.py
app = FlaskBase()

app.config["VITE_MODE"] = "development"
app.config["VITE_PORT"] = 5173
app.config["VITE_OUTDIR"] = "path/to/vite/dist"

vite = FlaskVite()
vite.init_app(app)
```
```html
<head>
  { vite_import("path/to/file.tsx") }
</head>
```

## How to QA
Vite dev mode
- checkout this branch
- `dotrun -p 5004:5004`
- check that the site loads correctly
Vite prod mode
- check that the demo loads correctly
- check that tests are green

## Testing
- [x] This PR has tests: updated previous tests to match the new interface

## Issue / Card
Fixes WD-30416

### DIFF
--- a/tests/tests_vite_integration.py
+++ b/tests/tests_vite_integration.py
@@ -4,6 +4,9 @@ from shutil import rmtree
 from typing import cast
 from urllib.parse import urlparse
 from pathlib import Path
+from unittest.mock import patch
+
+from webapp.app import create_app
 from webapp.vite_integration.impl import (
     ProdViteIntegration,
     DevViteIntegration,
@@ -12,7 +15,8 @@ import webapp.vite_integration.exceptions as vite_exceptions
 
 
 MOCK_CONFIG = {
-    "port": 5173,
+    "mode": "development",
+    "port": 9999,
     "outdir": "/tmp/python_vite_test",
 }
 MOCK_ASSET_PATH = "test/path/for/asset.ts"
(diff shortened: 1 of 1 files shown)


bartaz:
Out of curiosity, what do you mean by "upstreaming", for our own internal use cross project, or to Flask/Vite OSS?

webapp/vite_integration/impl/__init__.py
```diff
@@ -1,5 +1,7 @@
-from webapp.vite_integration.impl.dev import DevViteIntegration
-from webapp.vite_integration.impl.prod import ProdViteIntegration
+from .dev import DevViteIntegration
```
bartaz, line 1: We seem to be using `webapp.` absolute package imports elsewhere, why switching to relative ones here?

edisile:
@bartaz I plan to publish it on PyPI as a separate "canonicalwebteam.flask_vite" package that other projects can use.
The idea initially was to help charmhub.io and our other internal Flask apps adopt this Vite setup, but I'm trying to make it as generic as possible so that it won't be limited to just our internal projects. In the long term, I'll probably submit a PR to add the plugin to [vitejs/awesome-vite](https://github.com/vitejs/awesome-vite?tab=readme-ov-file#flask) as one of the recommended Flask integrations, since the only one that is currently listed there is a bit lacking IMO.

Maybe "upstreaming" isn't the proper term for this, but it sounded good to me 😂 I'll try to come up with something better

webapp/vite_integration/impl/__init__.py
edisile: That's just because of the way I'm working on this, I'll go back to absolute imports once I publish the package. I moved the whole directory into a separate project outside of the snapcraft.io repo and `mount --bind` that path in the current directory, so I couldn't use the `webapp` imports when working in the other directory

webapp/vite_integration/impl/prod.py
```diff
@@ -108,13 +113,13 @@ def get_imported_css(self, asset_name: str) -> List[str]:
         files = []
         for chunk in chunks:
             for file in chunk.get("css", []):
-                files.append(path.join(ProdViteIntegration.OUT_DIR, file))
+                files.append(path.join(self.outdir, file))
 
         for f in files:
```
alvaromateo: ### Nitpick

There is some repetition of this `for` block.
Also of this one:
```
entry = ProdViteIntegration.manifest.get(asset_name)
        if not entry:
...
```
You could extract them into 2 functions to avoid duplication.

PR #5437, opened 2025-10-24
WD-30045 - Implement sliding navigation in snapcraft.io

alvaromateo:
## Done

- Move changes from `global-nav` to snapcraft.io.

More specifically these changes include:
- Changes to the header markup to make the sliding animations of dropdowns work in all screen sizes.
- Necessary styles for the sliding animations.
- JS listeners for handling all the clicks in the navigation links and menus.
- Some patches to adapt the `global-nav` to the sliding style (mainly

## How to QA

- Open Demos and check that the navigation works the same as it does in Production.

## Issue / Card

Fixes [WD-30045](http://warthogs.atlassian.net/browse/WD-30045)

[WD-30045]: https://warthogs.atlassian.net/browse/WD-30045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/static/sass/_snapcraft_p-navigation.scss
+++ b/static/sass/_snapcraft_p-navigation.scss
@@ -20,9 +20,13 @@
         // starts on the right outside of the screen
         --right: 110%;
 
-        &--wrapper {
+        .p-navigation__items-wrapper {
           position: relative;
         }
+      }
+
+      .l-docs__main .p-navigation__nav {
+        margin-top: 0;
       }
 
       .p-navigation__items {
@@ -72,7 +76,7 @@
         syntax: "<percentage>";
       }
 
-      .p-navigation__nav--wrapper {
+      .p-navigation__items-wrapper {
         display: flex;
         flex-direction: row;
         justify-content: space-between;
--- a/templates/_header.html
+++ b/templates/_header.html
@@ -29,7 +29,7 @@
           <div class="p-navigation__row u-fixed-width">
         {% endif %}
             <nav class="p-navigation__nav">
-              <div class="p-navigation__nav--wrapper">
+              <div class="p-navigation__items-wrapper">
                 <ul class="p-navigation__items" role="menu">
                   <li
                     class="p-navigation__item {% if request.path == '/store' %}is-selected{% endif %}"


entrypoint
```diff
@@ -6,6 +6,7 @@ RUN_COMMAND="gunicorn webapp.app:create_app() --bind $1 --worker-class gevent --
 
 if [ "${FLASK_DEBUG}" = true ] || [ "${FLASK_DEBUG}" = 1 ]; then
     RUN_COMMAND="${RUN_COMMAND} --reload --log-level debug --timeout 9999"
+    RUN_COMMAND="${RUN_COMMAND} --logger-class canonicalwebteam.flask_base.log_utils.GunicornDevLogger"
```
steverydz, line 9: Out of curiosity, why is this change needed?

static/js/base/navigation/dropdownUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (parentLevelDropdown) {
+    parentLevelDropdown.classList.toggle("is-active", isActive);
+  }
+
+  // set active state of the top navigation list under p-navigation__nav
+  // to set the position of the sliding panel properly
+  const topLevelNavigation = dropdownToggleButton.closest(".p-navigation__nav");
+  if (topLevelNavigation) {
+    const topLevelItems = topLevelNavigation.querySelectorAll(
+      ":scope > .p-navigation__items",
```
steverydz, line 35: Nice!

entrypoint
alvaromateo: It's optional, but given that you are already in the latest flask-base version, with this change you will get the pretty printed logs in DEV mode, which is a nice addition :smile:

static/js/base/navigation/listeners.ts
```diff
@@ -0,0 +1,222 @@
...
+          collapseAllDropdowns(target);
+        }
+
+        if (target.getAttribute("aria-hidden") === "true") {
+          unfocusAllLinks();
+          console.log(toggle.parentElement?.classList);
+          expandDropdown(toggle, target);
+          navigation.classList.add("has-menu-open");
+        } else {
+          console.log(toggle.parentElement?.classList);
```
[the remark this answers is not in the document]
alvaromateo: Done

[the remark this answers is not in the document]
alvaromateo: Done

static/js/base/navigation/globalNav.ts
```diff
@@ -0,0 +1,35 @@
+export function patchAllCanonicalMobileMarkup() {
+  const allCanonicalMobile = document.getElementById("all-canonical-mobile");
+  const topMobileSections = allCanonicalMobile?.querySelectorAll(
+    ".global-nav__dropdown-toggle",
+  );
+
+  topMobileSections?.forEach((section: Element) => {
+    const sectionButton = section.querySelector("button");
+    const sectionHref = sectionButton?.getAttribute("href");
```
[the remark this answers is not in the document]
alvaromateo: Done

static/js/base/navigation/globalNav.ts
```diff
@@ -0,0 +1,35 @@
...
+  return div.childNodes[0];
+}
+
+function createBackButtonItem(href: string) {
+  // remove the # from the href
+  const ariaControls = href.slice(1);
+  return createFromHTML(`<li class="p-navigation__item--dropdown-close">
+      <button href=${href} aria-controls=${ariaControls} class="p-navigation__link js-back-button">
+        Back
+      </button>
```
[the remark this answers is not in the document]
alvaromateo: Done

static/sass/_snapcraft_p-navigation.scss
```diff
@@ -52,14 +72,11 @@
         syntax: "<percentage>";
       }
 
-      .p-navigation__items {
+      .p-navigation__nav--wrapper{
```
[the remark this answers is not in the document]
alvaromateo: Done

static/sass/_snapcraft_p-navigation.scss
```diff
@@ -19,6 +19,15 @@
 
         // starts on the right outside of the screen
         --right: 110%;
+
+        &--wrapper {
```
bartaz: There are also `.p-navigation__nav--wrapper` styles in the line 75. Do we need them in 2 places?

Also, if we keep them here, it would be good to spell out the whole class name `p-navigation__nav--wrapper` as the `&` shorthand makes it harder to understand what the classname really is (and search for the related styles in code).

templates/_header.html
```diff
@@ -29,8 +29,8 @@
           <div class="p-navigation__row u-fixed-width">
         {% endif %}
             <nav class="p-navigation__nav">
-              <div class="p-navigation__items">
-                <ul class="p-navigation__items--section" role="menu">
+              <div class="p-navigation__nav--wrapper">
```
bartaz: This is not exactly a correct BEM class name for this element. `--modifier` is meant to create a variant of existing class name, and be used alongside (or instead of) it. So `p-navigation__nav--wrapper` would be used on `nav` element that is parent to this one.

If we need another kind of child element it should be a `__child` of `p-navigation`, so maybe `p-navigation__wrapper`, or if we want to be more specific `p-navigation__items-wrapper`.

static/js/base/navigation/globalNav.ts
```diff
@@ -0,0 +1,35 @@
+export function patchAllCanonicalMobileMarkup() {
```
bartaz, line 1: Why not put the necessary elements in HTML template (that we control) instead of "patching" in JS?

bartaz:
image

On the /docs page (not available on demos) on mobile, there is excessive margin top between top nav and expanded navigation. On standard layout this margin accommodates for top nav height, but in docs layout it seems it's not needed. We may need to apply it only on standard layout.

static/sass/_snapcraft_p-navigation.scss
alvaromateo: We need them in 2 places cause one sets styles for mobile and the second one different styles for desktop.
I did the change to spell out the whole class name. Thanks for the suggestion!

static/js/base/navigation/globalNav.ts
alvaromateo: Cause that HTML template is in global-nav, and given that this is just for the sliding navigation I thought we didn't want to mess with it to add any kind of changes related to sliding navigation. That's why I patch it with a bit of JS instead of adding again things to the global-nav.

templates/_header.html
alvaromateo: Done

alvaromateo:
> image
> On the /docs page (not available on demos) on mobile, there is excessive margin top between top nav and expanded navigation. On standard layout this margin accommodates for top nav height, but in docs layout it seems it's not needed. We may need to apply it only on standard layout.

Good catch @bartaz! I just saw this bug is present also in PROD.
I've pushed changes that fix it and addressed all your other comments too.
Could you check these latest changes and approve if there's nothing else? Thanks!

static/js/base/navigation/globalNav.ts
bartaz: As discussed separately in chat, we need to do it to make "All Canonical" mobile view compatible with sliding navigation (adding Back buttons).

This is handled similarly in ubuntu.com: https://github.com/canonical/ubuntu.com/blob/main/static/js/src/navigation.js#L746-L782

static/sass/_snapcraft_p-navigation.scss
```diff
@@ -20,11 +20,15 @@
         // starts on the right outside of the screen
         --right: 110%;
 
-        &--wrapper {
+        .p-navigation__items-wrapper {
```
bartaz, line 23: Does this style need to be here (nested under `p-navigation__nav`)?

We already have styles for `p-navigation__items-wrapper` separately in line 79. Could the `position: relative` be moved there to keep all styles together?

bartaz: I just noticed you answered this in other comment.

static/sass/_snapcraft_p-navigation.scss
bartaz: Ah, OK. Didn't notice that the top part is wrapped in mobile media query.

PR #5438, opened 2025-10-27
feat: Add popular snaps to SSR explore page

steverydz:
## Done
Adds the popular snaps section to the server-side rendered explore page

## How to QA
- Go to https://snapcraft-io-5438.demos.haus/explore
- Check that there is a "Most popular snaps" section and that it is the same as it is on the [live Explore page](https://snapcraft.io/explore)

## Testing
- [x] This PR has tests
- [ ] No testing required (explain why): No behavioural changes

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-28154

### DIFF
--- a/templates/explore/_popular-snaps.html
+++ b/templates/explore/_popular-snaps.html
@@ -1,0 +1,37 @@
+<section class="p-strip is-shallow u-no-padding--bottom">
+  <div class="u-fixed-width">
+    <h2>Most popular snaps</h2>
+  </div>
+</section>
+<section class="p-strip is-shallow u-no-padding--bottom">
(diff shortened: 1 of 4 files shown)


codeEmpress1 (approved):

edisile:
Publisher names are missing the verified badge (none of the snaps have a star dev badge, but I would guess those would also be missing)

Demo:
immagine

Live:
immagine

steverydz:
@edisile Good catch. I've fixed that and for the stars too. There are actually two snaps that should have had stars so there's clearly a bug in the live version too, but it's fixed here.

PR #5462, opened 2025-11-19
Add cache to /endpoints/publisher/listing

codeEmpress1:
## Done
- add cache utility
- add cache to endpoints/publisher/listing.py
## How to QA
- Go to a snap listing page and check that the page renders as expected
- update the snap metadata and the details should be updated as expected
## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #

### DIFF
--- a/webapp/endpoints/publisher/listing.py
+++ b/webapp/endpoints/publisher/listing.py
@@ -25,8 +25,7 @@ device_gateway = DeviceGW("snap", api_session)
 
 
 def get_snap_info_key(snap_name):
-    user_id = flask.session.get("publisher", {}).get("identity_url", "")
-    cache_key = f"snapcraft:dashboard:snap_info:{snap_name}:{user_id}"
+    cache_key = f"snap_info:{snap_name}"
     return cache_key
 
 


cache/README.md
```diff
@@ -0,0 +1,23 @@
+# Local Redis for Development
+
+This setup provides a local Redis instance for development and testing purposes.
+
+## Getting Started
+
+1. Ensure Docker is installed.
```
ilayda-cp, line 7: every time we run snapcraft locally do we need to do this? or is it just for testing redis?

codeEmpress1: This is for testing redis locally only.

alvaromateo (approved):
LGTM

The only thing that I would like to add if possible are a bit of logging for when we set/delete info in Redis, to be able to track the flow of our program properly. If I remember correctly there were some logs already in the RedisCache utility? Otherwise we should add them.

codeEmpress1:
> LGTM
> 
> The only thing that I would like to add if possible are a bit of logging for when we set/delete info in Redis, to be able to track the flow of our program properly. If I remember correctly there were some logs already in the RedisCache utility? Otherwise we should add them.

@alvaromateo we have logs already in the[ RedisCache utility in storeAPI module](https://github.com/canonical/canonicalwebteam.store-api/blob/main/canonicalwebteam/stores_web_redis/utility.py)

codeEmpress1:
> > LGTM
> > The only thing that I would like to add if possible are a bit of logging for when we set/delete info in Redis, to be able to track the flow of our program properly. If I remember correctly there were some logs already in the RedisCache utility? Otherwise we should add them.
> 
> @alvaromateo we have logs already in the[ RedisCache utility in storeAPI module](https://github.com/canonical/canonicalwebteam.store-api/blob/main/canonicalwebteam/stores_web_redis/utility.py)

I would update the failing tests and ping you for re-approval

webapp/endpoints/publisher/listing.py
```diff
@@ -18,14 +18,26 @@
...
 
 dashboard = Dashboard(api_session)
 device_gateway = DeviceGW("snap", api_session)
 
+def get_snap_info_key(snap_name):
+    user_id = flask.session.get(
+        "publisher", {}
+    ).get("identity_url", "").split("+id")[1]
+
+    return f"{user_id}:snap_info:{snap_name}"
```
edisile: Not sure if `user_id` should be a part of the key, since snaps can have multiple contributors. This would mean that if one contributor updates the listing info, the others might not see the new data if they have an old cache entry in their name

codeEmpress1: Good catch!

edisile (approved):

PR #5463, opened 2025-11-19
fix: Fix show more functionality in the snap usage chart

bartaz:
## Done

Fixes "Show more" functionality of the OS usage chart on snap details page.

### Before:

When OS usage chart contained more than 20 OSes on the list, it gets fully collapsed, not showing anything, not even hinting what is collapsed (just "Show more" link).
image

Only when expanded it shows everything:
image

Heading size is also inconsistent with snaps that have less than 20 OSes on the list:

image

### After:

image

## How to QA

- Go to a snap with many OSes in usage graph, make sure the list is truncated (but visible) https://snapcraft-io-5463.demos.haus/josm
- Go to a snap with few OSes in usage graph, make sure the list still renders fine https://snapcraft.io/bomber
- In both cases heading should look the same

## Testing
- [ ] This PR has tests

### DIFF
--- a/templates/store/snap-details.html
+++ b/templates/store/snap-details.html
@@ -234,9 +234,7 @@
                     </div>
                     <div class="p-show-more__fade">
                       <div class="p-show-more__link-container">
-                        <small>
-                          <button class="p-button--link p-show-more__link" aria-expanded="false">Show more</button>
-                        </small>
+                        <button class="p-button--link p-show-more__link is-small u-no-padding" aria-expanded="false">Show more</button>
                       </div>
                     </div>
                   </div>
--- a/static/sass/_snapcraft_distro-chart.scss
+++ b/static/sass/_snapcraft_distro-chart.scss
@@ -26,7 +26,6 @@
       line-height: 1rem;
       margin-top: 1px;
       overflow: hidden;
-      text-align: right;
       text-overflow: ellipsis;
       white-space: nowrap;
     }


templates/store/snap-details.html
```diff
@@ -197,38 +197,56 @@ <h4 class="p-heading--2">Where people are using {{ snap_title }}</h4>
...
+                            class="snapcraft-distro-chart__bar"
+                            style="width: {{ distro.value * 100 }}%;"
+                          ></div>
+                        {% endfor %}
+                      </div>
+                    </div>
+                    <div class="p-show-more__fade">
+                      <div class="p-show-more__link-container">
+                        <small>
+                          <button class="p-button--link p-show-more__link" aria-expanded="false">Show more</button>
```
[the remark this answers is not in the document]
bartaz: Not needed, as button is hidden (removed from accessibility tree) after expanding.

templates/store/snap-details.html
```diff
@@ -197,38 +197,56 @@ <h4 class="p-heading--2">Where people are using {{ snap_title }}</h4>
...
           {% endif %}
           {% if normalized_os %}
-            {% if normalized_os|length > 21 %}
-            <div class="col-4 p-show-more is-collapsed is-deep" data-js="js-show-more">
+            {% set list_limit = 20 %}
+            {% if normalized_os|length > list_limit + 1 %}
+            <div class="col-4">
               <div class="js-sschart-holder" data-live="weekly_installed_base_by_operating_system_normalized">
-                <h4>Users by distribution (log)</h4>
+                <h5 class="p-muted-heading">Users by distribution (log)</h5>
```
ilayda-cp, line 204: the only comment that i have is the"Where people are using josm" is heaing two and the right after the other heading its "Users by distribution (log)" and its heading 5. Its not good for a11y https://www.w3.org/WAI/tutorials/page-structure/headings/

bartaz: Yes, I plan on addressing this when refreshing the page layout.

PR #5465, opened 2025-11-19
Publisher pages lazy import, pt3

edisile:
fixes the ugly redraw issue mentioned in #5461

## Done
- new layout for snap management pages
- rebuilt breadcrumbs and tab navigation in layout
- removed old breadcrumbs and tab navigation from each of the pages
- wrapped all snap management pages in the new layout
- simple shared Loader component
- some cleanup
- test for new breadcrumbs and tab navigation

## How to QA
- visit the demo
- log in
- open the network tab in dev tools and filter by JS
- visit a snap listing page and then navigate between tabs in any order
  - the first time you will see a loading spinner
  - every page you visit will fetch at least one JS chunk
  - the second time the JS will be cached and there will be no loader
  - the breadcrumbs and the tabs won't disappear when changing route

## Testing
- [x] This PR has tests

## Issue / Card
Fixes #

### DIFF
--- a/static/js/publisher/layouts/SnapsManagementLayout/Breadcrumbs.tsx
+++ b/static/js/publisher/layouts/SnapsManagementLayout/Breadcrumbs.tsx
@@ -35,6 +35,8 @@ function Breadcrumbs(): React.JSX.Element {
         ))}
       </h1>
 
+      {/* TODO: this is the proper accessible implementation of the breadcrumbs pattern,
+          we should change the layout to use this rather than the h1 above */}
       {/* <nav className="p-breadcrumbs" aria-label="Breadcrumb" aria-live="polite">
         <ol className="p-breadcrumbs__items">
           <li className="p-breadcrumbs__item">
--- a/static/js/publisher/index.tsx
+++ b/static/js/publisher/index.tsx
@@ -65,6 +65,7 @@ root.render(
                 if any of the children routes changes, make sure to update
                 static/js/publisher/layouts/SnapsManagementLayout/routes.ts
               */}
+              {/* TODO: move to data routing with RouterProvider */}
               <Route path="publicise" element={<Publicise />} />
               <Route
                 path="publicise/badges"


static/js/publisher/layouts/SnapsManagementLayout/Breadcrumbs.tsx
```diff
@@ -0,0 +1,58 @@
...
+                <Link to={to}>{label}</Link>&nbsp;/&nbsp;
+              </>
+            ) : (
+              label
+            )}
+          </Fragment>
+        ))}
+      </h1>
+
+      {/* <nav className="p-breadcrumbs" aria-label="Breadcrumb" aria-live="polite">
```
alvaromateo, line 40: This commented out code is to be removed or do we want to keep it for some reason?

static/js/publisher/layouts/SnapsManagementLayout/routes.ts
```diff
@@ -0,0 +1,55 @@
+type Route = {
+  path: string;
+  label?: string;
+  children?: Route[];
+};
+
+// TODO: this should be tied to the ACTUAL routes... maybe move to data routing?
```
alvaromateo, line 7: Maybe for a future task, but what about changing the BrowserRouter that we currently have by the [RouterProvider](https://reactrouter.com/start/data/installation)?
This way we could move this object closer to the router creation and use it both for the router creation and for the SnapsManagementLayout, thus making sure that we have everything in one place.

If you decide to leave this for a future task maybe we could add a cross reference comment here, like the one you added at the BrowserRouter code - to make sure that if someone needs to modify one, they realize that they have to modify the other too.

alvaromateo (approved):
LGTM!

I left a minor comment and a suggestion.

Regarding the QA, I noticed that in the "Builds" and "Releases" pages I always get the loading spinner, although I imagine that has to do with the data being fetched because in the Network tab I don't get anymore JS requests. I consider the QA passed, I just mentioned this in case I'm wrong in my assumption.

Good job 👍

static/js/publisher/layouts/SnapsManagementLayout/Breadcrumbs.tsx
edisile: The code below is the "correct" implementation of breadcrumbs following the Vanilla pattern. The current implementation where the breadcrumb sits within a heading isn't accessible (or at least I think so), so we should replace it with the commented code at some point in the future; I'll add a TODO comment above to explain this issue and keep track of it

static/js/publisher/layouts/SnapsManagementLayout/routes.ts
edisile: Yeah, this is the intention in the long run, but it's a bit outside the scope of this PR. I'll expand this comment to point to the BrowserRouter in case of changes

PR #5472, opened 2025-11-21
JetBrains publisher page: replacing legacy featured PyCharm snaps with new IntelliJ IDEA and PyCharm snaps

DmPanov:
Hi! 
We would like to update our featured snap list in https://snapcraft.io/publisher/jetbrains.
More details about that change can be found in the following threads:
1. https://forum.snapcraft.io/t/classic-confinement-for-intellij-idea-and-pycharm/49152
2. https://forum.snapcraft.io/t/redirecting-users-from-a-legacy-snap-to-a-new-snap/48339

Could anyone please take a look? Thank you!

### DIFF
--- a/webapp/store/content/publishers/jetbrains.yaml
+++ b/webapp/store/content/publishers/jetbrains.yaml
@@ -16,24 +16,24 @@ publishers: [
 featured_snaps: [
     {
         "developer_validation": "verified",
-        "icon_url": "https://dashboard.snapcraft.io/site_media/appmedia/2017/11/PyCharmCore256.png",
+        "icon_url": "https://dashboard.snapcraft.io/site_media/appmedia/2024/06/Avatar-5.png",
         "origin": "jetbrains",
-        "package_name": "pycharm-community",
+        "package_name": "intellij-idea",
         "publisher": "JetBrains",
-        "summary": "Python IDE for Professional Developers",
-        "description": "PyCharm Community Edition is a free and open-source IDE which is perfect for pure Python coding. For professional Web and Scientific development see PyCharm Professional Edition.",
-        "title": "PyCharm CE",
-        "background": "#3d3532"
+        "summary": "The Leading IDE for Professional Development in Java and Kotlin",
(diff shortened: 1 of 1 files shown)


bartaz:
Thanks @DmPanov. Are the new snaps ready to go? Pycharm page looks fine, but `intellij-idea` is private and doesn't have too much of metadata set yet as far as I see. 

You want this change to go live as is, or you wanna wait until `intelij-idea` snap work is finished and public?

webapp/store/content/publishers/jetbrains.yaml
```diff
@@ -16,24 +16,24 @@ publishers: [
 featured_snaps: [
     {
         "developer_validation": "verified",
-        "icon_url": "https://dashboard.snapcraft.io/site_media/appmedia/2017/11/PyCharmCore256.png",
+        "icon_url": "https://dashboard.snapcraft.io/site_media/appmedia/2024/06/Avatar-5.png",
```
bartaz, line 19: A bit of nitpicking, but seems that both icons use a bit of different format, with intellij one having black background and smaller logotype.

This makes them look a bit off side by side:

Image

Also, you sure you want the black background? Currently it makes the border in heading disappear, but we can fix it on our side with styling, if you want to keep the black.

DmPanov:
@bartaz I'd like to wait for the `intellij-idea` snap to go public, thanks for a heads-up

webapp/store/content/publishers/jetbrains.yaml
DmPanov: The black background is intentional, ~~but the IDEA's icon size is not, thanks, looking into that~~

DmPanov: Actually, we already have a similar issue right below the featured snap list, so let's keep it as is

bartaz:
This looks good. Once the IntelliJ snap is live we can give this a final quick look and get it live.

DmPanov:
@bartaz https://snapcraft.io/intellij-idea is live, we're ready to merge

DmPanov:
@bartaz thank you! May I ask when this change is expected to be deployed to production?

bartaz:
Should be automatically published now that's merged. May take a bit for it to propagate through caches.

PR #5479, opened 2025-11-24
fix: fill the missing data in the metrics response

ilayda-cp:
## Done
Sometimes the metrics information that is provided from store API includes None values. This pr fillsout those missing data as follows:
- take avg of before non-null and after non-null value
- set the prev non-null val if its the last value
- set the after non-null val if its the first value
- set 0 if every thing is None

## How to QA
- go to https://snapcraft-io-5479.demos.haus/test-flight-tracker/metrics
- verify there is no sudden drop in the middle of the chart
- go to https://snapcraft-io-5479.demos.haus/snaps
- verify there is no sudden drop in the middle of the chart

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-15452

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -12,7 +12,7 @@ jobs:
       contents: read
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - uses: actions/setup-node@v6
         with:
@@ -59,7 +59,7 @@ jobs:
       contents: write
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           ref: gh-pages
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


webapp/metrics/helper.py
```diff
@@ -152,18 +152,48 @@ def build_snap_installs_metrics_query(snaps, get_filter=get_filter):
...
+        return [0] * n
+
+    for i in range(n):
+        if filled[i] is None:
+            # Find previous non-null value
+            prev_value = None
+            prev_idx = i - 1
+            while prev_idx >= 0 and prev_value is None:
+                prev_value = filled[prev_idx]
+                prev_idx -= 1
```
edisile, line 178: Is there a case where the `while` is actually necessary? Since we're iterating items in order I think we can assume that the previous spot in the list is not None because it was either a valid number from the start or it has already been filled if it was None...

edisile: On second thought not sure if changing it is worth it, I find it's nice that it's symmetrical to the `next_value` loop

steverydz (approved):
LGTM 👍

edisile (commented):
Can't QA because I don't have access to that snap, but code looks good

PR #5490, opened 2025-11-27
TS types for releases API

edisile:
## Done
- added types for releases API payloads and responses
- removed useless `getErrorMessage` helper function (all errors were swallowed in the API helpers no the function always returned the same message)
- fixed a couple of tests that were effectively broken (they used to return responses with "sucess" rather than "success", apparently that didn't count as a failure ¯\\_(ツ)_/¯)

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): just types

## Issue / Card
Fixes WD-31604


steverydz (approved):
LGTM 👍

PR #5513, opened 2025-12-04
Cross platform vite_import plugin

edisile:
## Done
- replaced grep command with ripgrep (pulled using `@lvce-editor/ripgrep`) for cross-platform support
- improved vite_import detection plugin
- moved the plugin to a separate file
  - at some point it could also be moved to a separate package

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): if tests pass it's good

### DIFF
--- a/viteDetectImportPlugin.js
+++ b/viteDetectImportPlugin.js
@@ -1,5 +1,6 @@
 import { defineConfig } from "vite";
 import { execSync } from "node:child_process";
+import { EOL } from "node:os";
 import { rgPath as rg } from "@lvce-editor/ripgrep";
 
 // TODO: this could be generalized to support other regex patterns and moved into a separate NPM package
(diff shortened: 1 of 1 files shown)


viteDetectImportPlugin.js
```diff
@@ -0,0 +1,68 @@
...
+        : `\\$1`; // any other platform does though
+
+    const viteImports =
+      execSync(
+        `${rg} -oNI -t "html" -e "${pattern}" ./templates -r "${captureString}"`,
+      ).toString() || "";
+
+    // the string has multiple lines, some of which might be blank, so split
+    // on newlines and filter out the empty strings
+    const imports = viteImports.split("\n").filter(Boolean);
```
alvaromateo: In the captureString you consider support for Windows platform (brave 😂).
I believe Windows uses a different new line characters. Won't that be a problem here (and later in the `.join("\n")`)?

edisile: Good point, splitting this way leaves a bad "\r" in the filename... Vite doesn't seem to mind but I'll add a check for this edge case. No need for "\r\n" in the .join call because that only gets printed in the console

alvaromateo (approved):

edisile:
There are some connection time out issues when downloading ripgrep from github that can lead to builds failing. Adding "don't merge" label until I find some time to investigate

PR #5517, opened 2025-12-08
feat: Populate channel map security tab

steverydz:
## Done
- Populated the channel map security tab to display either SBOM link or the text "Not available"
- Updated the SBOM proxy URL so the downloaded JSON file is named correctly

## How to QA
- Run locally
- Go to http://localhost:8004/staging-muchohan-chisel-snapcraft
- Open the channel map
- Check that there is a "Security" tab
- Check that clicking on the "Security" tab displays a table that contains a link to a SPDX file
- Click the link to the SPDX file and confirm that a JSON file is downloaded
- Go to another snap, e.g. http://localhost:8004/telegram-sergiusens
- Open the channel map
- Check that there is **not** a "Security" tab (in the case of this snap there should be no tabs)

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-31840
Fixes https://warthogs.atlassian.net/browse/WD-31166
Fixes https://warthogs.atlassian.net/browse/WD-20986

### DIFF
--- a/static/js/public/snap-details/channelMap.ts
+++ b/static/js/public/snap-details/channelMap.ts
@@ -613,42 +613,40 @@ class ChannelMap {
             trackInfo["revision"],
           ]);
         }
-
-        if (this.hasSboms && securityRows.length > 0) {
-          Promise.all(
-            securityRows.map(async (row) => {
-              const revision = row[3];
-              const sbomUrl = getSbomUrl(revision);
-              const downloadLink = `<a href="${sbomUrl}" download>SPDX file&nbsp;<i class="p-icon--begin-downloading"></i></a>`;
-              const res = await fetch(sbomUrl, { method: "HEAD" });
-
-              if (res.status === 200) {
-                row.push(downloadLink);
-              } else {
-                row.push("Not available");
-              }
-
-              return row;
-            }),
-          ).then(() => {
-            this.writeTable(
-              tbodySecurityEl,
-              this.sortRows(securityRows),
-              "security",
-            );
-
-            // Enable "Security" tab only when SBOM requests
-            // are complete to avoid a race condition causing
-            // the table to have not rendered
-            const securityTab = document.querySelector(
-              "#channel-map-security-tab",
-            );
-            securityTab?.classList.remove("is-disabled");
-            securityTab?.setAttribute("aria-disabled", "false");
-          });
-        }
       });
     });
+
+    if (this.hasSboms && securityRows.length > 0) {
+      Promise.all(
+        securityRows.map(async (row) => {
+          const revision = row[3];
+          const sbomUrl = getSbomUrl(revision);
+          const downloadLink = `<a href="${sbomUrl}" download>SPDX file&nbsp;<i class="p-icon--begin-downloading"></i></a>`;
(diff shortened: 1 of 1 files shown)


static/js/public/snap-details/channelMap.ts
```diff
@@ -140,6 +148,14 @@ class ChannelMap {
         "channel-map-row-template",
       );
     }
+    let channelSecurityRowTemplateEl = document.querySelector(
+      '[data-js="channel-map-security-table-row"]',
+    );
+    if (!channelSecurityRowTemplateEl) {
```
alvaromateo: Given that we control the Jinja templates, is this fallback necessary?
Instead of the fallback we could throw an error to warn the developer that the 'data-js' attribute is needed (or the id, whichever option you prefer).

static/js/public/snap-details/channelMap.ts
```diff
@@ -577,6 +610,49 @@ class ChannelMap {
...
+        if (this.hasSboms) {
+          securityRows.push([
+            trackName,
+            trackInfo["risk"],
+            trackInfo["version"],
+            trackInfo["revision"],
+          ]);
+        }
+
+        if (this.hasSboms && securityRows.length > 0) {
```
alvaromateo: I may be wrong here - first time looking at this source file of snapcraft.io - but it looks like if there are more than one row in the channel map, then we will repeat queries because this `Promise.all` will run for each row (in the second row it will do the fetch for the first and second rows, in the third row it will do it for the 1st to the 3rd, etc.). The same goes for the extra rows being added to the table.

The example in QA only has one row so I could not test this out, but it looks to me that this Promise should be run outside of the `Object.keys` loop to avoid repetition.

templates/store/snap-details.html
```diff
@@ -374,7 +374,7 @@ <h5 class="p-notification__title">Error</h5>
 
       {% if channel_map %}
         try {
-            snapcraft.public.storeDetails.channelMap('#js-channel-map', {{ package_name|tojson }}, {{ channel_map | tojson }}, "{{ default_track }}");
+snapcraft.public.storeDetails.channelMap('#js-channel-map', {{ package_name|tojson }}, {{ snap_id|tojson }}, {{ channel_map | tojson }}, "{{ default_track }}", {{ has_sboms|tojson }});
```
edisile: Was this indentation change on purpose?

alvaromateo (commented):
Quite a complex piece of code. There's a part that concerns me regarding the `Promise.all` call.
For the rest it looks good to me (although maybe we could start thinking of doing some refactoring to reduce the complexity of the code 😂).

templates/store/snap-details.html
steverydz: @edisile No 😆 - I'll fix it

static/js/public/snap-details/channelMap.ts
steverydz: Good point. I'll remove them

static/js/public/snap-details/channelMap.ts
steverydz: Good catch - fixed

alvaromateo (approved):
Thanks for the changes!

PR #5519, opened 2025-12-09
Fix cache access for private snaps

codeEmpress1:
## Done
- Fix cache leakages for private snaps
- Remove cache from store snaps

## How to QA
- You need two accounts to QA this. One of the accounts should have a private snap.

**Locally**
- `cd cache` run `docker compose run redis-cli`
- In another terminal window, run `dotrun`
- With the account that owns the private snap, go to `localhost:8004/api/{private-snap-name}/listing`, `localhost:8004/api/{private-snap-name}/releases`, `localhost:8004/api/{private-snap-name}/builds`
- Then go to the redis CLI, run `KEYS *`, confirm that the key for snap-info for the private snap is suffixed with your username
- In another browser, with your second account, go to `localhost:8004/api/{private-snap-name}/listing`, `localhost:8004/api/{private-snap-name}/releases`, `localhost:8004/api/{private-snap-name}/builds` these should return 404.
- To confirm that this works as expected for public snaps, go to the listing page or any other publisher page for the snaps, confirm that both accounts can access the pages.

**In Demos**
- With the same accounts above or any other acccounts (one should have a private snap) go to `/api/{private-snap-name}/listing` and other publisher routes, Confirm that only the private snap owner is able to acess the pages, the other account should 404.
## Testing
- [x] This PR has tests

## Issue / Card
Fixes #

### DIFF
--- a/tests/endpoints/publisher/tests_cache_leak.py
+++ b/tests/endpoints/publisher/tests_cache_leak.py
@@ -77,7 +77,6 @@ class TestCacheLeakage(TestCase):
             "publisher": {"username": "ownerA"},
         }
 
-        # Prepare other mocks used by the listing view
         mock_device_gateway.get_categories.return_value = []
         mock_get_categories.return_value = []
         mock_categorise_media.return_value = ([], [], [])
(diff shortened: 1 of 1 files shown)


steverydz:
It works for all of the snaps I've tried. There is one strange thing when running the `redis-cli`. When logging out the keys I get:
`(empty array)`

I've tried:
`KEYS *`
`KEYS "*"`
`keys *`
`keys "*"`

The changes still work as expected so not sure why the keys wouldn't be showing. Perhaps that's an issue on my machine, but it doesn't seem to effect the functionality. I'm using MacBook Pro M1 (macOS Tahoe).

codeEmpress1:
Closing as changes has been reverted in other PRs

PR #5541, opened 2026-01-05
Ratings in snap details page

M7mdisk:
## Done
- added ratings to the snap details page
> [!NOTE]
> Dear reviewer: you can completely ignore everything under `webapp/ratings/generated` when reviewing this PR.
## How to QA
- go to https://snapcraft-io-5541.demos.haus/code or any other snap
- go to a less popular snap, see no ratings 
- see ratings in the header (after the category)
## Testing
- [x] This PR has tests

## Security

  - [ ] Examples:
  - [ ] ...
- [x] This PR has no security considerations (explain why): this simply exposes read only information that is already available in the desktop store, we don't support submitting ratings

## Issue / Card
Fixes WD-31734

### DIFF
--- a/konf/site.yaml
+++ b/konf/site.yaml
@@ -74,11 +74,6 @@ env: &env
       key: lp-api-token-secret
       name: snapcraft-io
 
-  - name: YOUTUBE_API_KEY
-    secretKeyRef:
-      key: youtube-api-key
-      name: snapcraft-io
-
   - name: DISCOURSE_API_KEY
     secretKeyRef:
       key: discourse-api-key
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # App dependencies
 canonicalwebteam.flask-base==3.1.1
-canonicalwebteam.flask-vite==0.3.1
+canonicalwebteam.flask-vite==0.4.0
 canonicalwebteam.candid==0.9.0
 canonicalwebteam.discourse==7.0.0
 canonicalwebteam.blog==6.6.0
--- a/templates/_base-layout.html
+++ b/templates/_base-layout.html
@@ -14,10 +14,11 @@
     <link rel="preload" href="https://assets.ubuntu.com/v1/9689339a-snapcraft-hero-background--light.png" as="image">
 
     <!-- Google Tag Manager -->
-    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    <script nonce="{{ CSP_NONCE }}">(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
     new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
     j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    'https://www.googletagmanager.com/gtm.js?id='+i+dl;var n=d.querySelector('[nonce]');
+    n&&j.setAttribute('nonce',n.nonce||n.getAttribute('nonce'));f.parentNode.insertBefore(j,f);
     })(window,document,'script','dataLayer','GTM-KCGXHQS');</script>
     <!-- End Google Tag Manager -->
 
@@ -25,7 +26,7 @@
 
     {{ vite_import('static/js/base/base.ts') }}
     {% block scripts_includes %}{% endblock %}
-    <script src="https://assets.ubuntu.com/v1/703e23c9-lazysizes+noscript+native-loading.5.1.2.min.js" defer></script>
+    <script src="https://assets.ubuntu.com/v1/703e23c9-lazysizes+noscript+native-loading.5.1.2.min.js" defer nonce="{{ CSP_NONCE }}"></script>
 
     <meta name="description" content="{% block meta_description %}Snaps are containerised software packages that are simple to create and install. They auto-update and are safe to run. And because they bundle their dependencies, they work on all major Linux systems without modification.{% endblock %}">
     <meta name="copydoc" content="{% block meta_copydoc %}{% endblock %}">
@@ -80,13 +81,13 @@
 
     {% block scripts %}{% endblock %}
 
-    <script>
+    <script nonce="{{ CSP_NONCE }}">
       window.COMMIT_ID = "{{ COMMIT_ID }}";
       window.ENVIRONMENT = "{{ ENVIRONMENT }}";
       window.SENTRY_DSN = "{{ SENTRY_DSN }}";
     </script>
 
-    <script type="application/ld+json">
+  <script type="application/ld+json" nonce="{{ CSP_NONCE }}">
       {
         "@context": "http://schema.org/",
         "@id": "https://snapcraft.io/#organization",
@@ -105,7 +106,7 @@
       }
     </script>
 
-    <script type="application/ld+json">
+    <script type="application/ld+json" nonce="{{ CSP_NONCE }}">
       {
         "@context": "http://schema.org",
         "@id": "https://snapcraft.io/#website",
--- a/templates/_layout-embedded.html
+++ b/templates/_layout-embedded.html
@@ -45,7 +45,7 @@
 
     {% block scripts %}{% endblock %}
 
-    <script type="application/ld+json">
+    <script type="application/ld+json" nonce="{{ CSP_NONCE }}">
       {
         "@context": "http://schema.org/",
         "@id": "https://snapcraft.io/#organization",
--- a/templates/about/publish.html
+++ b/templates/about/publish.html
@@ -196,7 +196,7 @@
 {% endblock %}
 
 {% block scripts %}
-<script>
+<script nonce="{{ CSP_NONCE }}">
   window.addEventListener("DOMContentLoaded", function () {
     try {
       snapcraft.about.initFSFLanguageSelect(document.querySelector('.js-fsf-language-select'));
--- a/templates/admin/admin.html
+++ b/templates/admin/admin.html
@@ -6,7 +6,7 @@
 <div id="root">
 </div>
 
-<script>
+<script nonce="{{ CSP_NONCE }}">
   window.CSRF_TOKEN = "{{ csrf_token() }}";
   window.SENTRY_DSN = "{{ SENTRY_DSN }}";
   window.API_URL = "{{ api_url }}";
--- a/templates/blog/article.html
+++ b/templates/blog/article.html
@@ -9,7 +9,7 @@
 {% endif %}
 
 {% block meta_schema %}
-<script type="application/ld+json">
+<script type="application/ld+json" nonce="{{ CSP_NONCE }}">
     {
       "@context": "http://schema.org",
       "@id": "https://snapcraft.io/#article",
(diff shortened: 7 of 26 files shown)


webapp/config.py
```diff
@@ -48,4 +48,8 @@ class ConfigurationError(Exception):
 SEARCH_API_URL = "https://www.googleapis.com/customsearch/v1"
 SEARCH_CUSTOM_ID = "009048213575199080868:i3zoqdwqk8o"
 
+# Ratings service configuration
+# Set RATINGS_SERVICE_URL to enable ratings (e.g., "ratings.example.com:443")
+RATINGS_SERVICE_URL = os.getenv("RATINGS_SERVICE_URL")
```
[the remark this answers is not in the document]
M7mdisk: this would be relevant if there were multiple rating services, but I don't see much value in it given we know that the deployed service has SSL

webapp/ratings/ratings_client.py
```diff
@@ -0,0 +1,102 @@
...
+            return None
+        finally:
+            channel.close()
+
+    def get_snap_rating(self, snap_id: str) -> Optional[Dict[str, Any]]:
+        if not self.token:
+            self.token = self._authenticate()
+            if not self.token:
+                logger.error("Failed to authenticate with ratings service")
+                return None
```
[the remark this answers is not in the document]
M7mdisk: fixed

steverydz:
I've noticed that the rating seems a bit inconsistent, such as [JOSM](https://snapcraft-io-5541.demos.haus/josm) has 94 votes and shows as "Good", whereas [HEY Mail](https://snapcraft-io-5541.demos.haus/hey-mail) has 134 votes and shows as "Poor"

Screenshot 2026-01-06 at 09 21 37

Screenshot 2026-01-06 at 09 21 43

M7mdisk:
@steverydz this is the number of total votes (including both negative and positive), so 134 votes doesn't necessarily mean they're all good. I think of it as a confidence indicator (more votes means the rating is probably more accurate)

Although you raise a good point that this is confusing now that I think about it. maybe we should change it to "(134 total votes)" instead

steverydz:
@M7mdisk I wonder if it might be better to have thumbs up and down icons with the number of positive/negative votes next to them? Or maybe something like this:

Screenshot 2026-01-06 at 09 52 01

Of course, this would be a design decision though

templates/store/snap-details/_package_heading_data.html
```diff
@@ -59,6 +62,9 @@
           <a href="/search?categories={{ category.slug }}">{{ category.name }}</a>
         </li>
       {% endfor %}
+      {% if ratings and ratings.total_votes > 0 %}
```
ilayda-cp, line 65: is rendering the ratings twice intentional?

webapp/store/snap_details_views.py
```diff
@@ -26,6 +27,17 @@
 
 logger = logging.getLogger(__name__)
 
+ratings_client = None
+
+
+def get_ratings_client():
+    global ratings_client
+    if ratings_client is None:
```
ilayda-cp, line 35: this might be an invalid case or rarely happen but will there be a case where multiple clients run this function at the same time and causing a race condition?

templates/store/snap-details/_package_heading_data.html
M7mdisk: yes one for desktop and one for mobile

webapp/store/snap_details_views.py
M7mdisk: hmm, it could happen, although even if it does, I don't see it causing any problems (worst case, we just have another client created)

bartaz:
> @steverydz this is the number of total votes (including both negative and positive), so 134 votes doesn't necessarily mean they're all good. I think of it as a confidence indicator (more votes means the rating is probably more accurate)
> 
> Although you raise a good point that this is confusing now that I think about it. maybe we should change it to "(134 total votes)" instead

I don't find the rating system too confusing as is. It seems quite similar to what Steam shows (rating + number of votes). Although they do add % in a tooltip. Maybe that would help?

image

It would be good to get design (UX and visual) review as well. I'm not 100% sure about using coloured text, it may be against our guidelines and it harder to made accessible across light/dark themes.

bartaz:
@M7mdisk As for security considerations: are we sure that the data we show is public? It is available via public API, doesn't need any authorisation to access, user doesn't need to be signed in to see it?

M7mdisk:
> > @steverydz this is the number of total votes (including both negative and positive), so 134 votes doesn't necessarily mean they're all good. I think of it as a confidence indicator (more votes means the rating is probably more accurate)
> > Although you raise a good point that this is confusing now that I think about it. maybe we should change it to "(134 total votes)" instead
> 
> I don't find the rating system too confusing as is. It seems quite similar to what Steam shows (rating + number of votes). Although they do add % in a tooltip. Maybe that would help?
> image
> 
> It would be good to get design (UX and visual) review as well. I'm not 100% sure about using coloured text, it may be against our guidelines and it harder to made accessible across light/dark themes.

I've asked @leiaru-c for a review, but you're right maybe its not as confusing (just checked the [microsoft store](https://apps.microsoft.com/detail/9nksqgp7f2nh?hl=en-US&gl=KW) for example, they do the same thing too).

good point about the colored text, hadn't thought of that :+1: 


---

> @M7mdisk As for security considerations: are we sure that the data we show is public? It is available via public API, doesn't need any authorisation to access, user doesn't need to be signed in to see it?


No this is all public, if you have the [app-center](https://github.com/ubuntu/app-center) installed you can see it unauthenticated (or more accurately, authenticated with a machine id, but even that is only really needed for making votes, not reading them). 

the API is public under `ratings.ubuntu.com` (uses gRPC, and the repo for the service is [public](https://github.com/ubuntu/app-center-ratings/))

static/sass/_snapcraft_p-ratings.scss
```diff
@@ -0,0 +1,19 @@
+@mixin snapcraft-p-ratings {
+  .p-snap-ratings {
+    &__band {
+      &--very-good,
+      &--good {
+        color: $color-positive;
```
bartaz: As mentioned in separate comment it would be good to confirm how do we want to display it. Coloured text doesn't seem to be something that is recommended in Vanilla.

Ideally we should use themed colour variables, but there are not coloured variables for text in themes. We have tinted backgrounds and borders. So border is likely the closest what we would need here in terms of contrast.

https://github.com/canonical/vanilla-framework/blob/fdf62cd229e50960de7e1aebfaf24c548c9de0ba/scss/_settings_colors.scss#L261-L265

Although, it seems a bit of a hack to use border colour for text, so would be good to confirm the direction with designers.

Simplest workaround would be not to use color at all, and just use default text. This should be fine for start.

If we want to highlight the score maybe we can use Chip component? https://github.com/canonical/vanilla-framework/blob/fdf62cd229e50960de7e1aebfaf24c548c9de0ba/scss/_settings_colors.scss#L261-L265

templates/store/_ratings_info.jinja
```diff
@@ -0,0 +1,18 @@
...
+  {%- elif ratings.ratings_band == "good" -%}
+    <span class="p-snap-ratings__band p-snap-ratings__band--good" title="Good rating">Good</span>
+  {%- elif ratings.ratings_band == "neutral" -%}
+    <span class="p-snap-ratings__band p-snap-ratings__band--neutral" title="Neutral rating">Neutral</span>
+  {%- elif ratings.ratings_band == "poor" -%}
+    <span class="p-snap-ratings__band p-snap-ratings__band--poor" title="Poor rating">Poor</span>
+  {%- elif ratings.ratings_band == "very-poor" -%}
+    <span class="p-snap-ratings__band p-snap-ratings__band--very-poor" title="Very poor rating">Very poor</span>
+  {%- endif -%}
+  <span class="p-snap-ratings__votes"> ({{ ratings.total_votes }} votes)</span>
```
bartaz: If we have any additional info (for example % bands of how the ratings are calculated) maybe we can show them in a tooltip?

Something like "Rated Poor based on 80% of 123 votes".

webapp/ratings/proto/ratings_features_app.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
```
bartaz, line 1: Where do these proto files come from? They don't say they are genearted, so are they copied over from rating service or something? Do we need to keep them up to date with the service part?

webapp/ratings/generate_proto_bindings.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
bartaz, line 1: What's the process of using this script? When do we need to run it?

If it generates the code in /generated subfolder, do we need to keep those files in git? Or can we generate them on build time?

M7mdisk: this needs to be run if/when the BE (the desktop ratings service) changes the API definition.
we have had a similar discussion on where this code should live in the original [ratings spec](https://docs.google.com/document/d/1A27O7Hz6rNay1Hnma2vnNPKIV5m8v0aYQhvX1c-75Kc/edit?tab=t.0#heading=h.1czsjbke1ky7).

I don't see much benefit in introducing build time overhead for this given that it will very rarely change, but happy to discuss further. we could also just manually add the only proto file we need to reduce the amount of generated code. but I tried to follow exactly what we did in snaprecommend

webapp/ratings/proto/ratings_features_app.proto
M7mdisk: the proto files are cloned from the app-center-ratings repo using the  `generate_proto_bindings.sh` script. and yes we do need to keep them up to date if the service changes, again the spec I linked in the other comments has more details.

templates/store/_ratings_info.jinja
M7mdisk: the ratings service uses the [wilson score](https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Wilson_score_interval)  to generate the band (nice [article](https://www.evanmiller.org/how-not-to-sort-by-average-rating.html) explaining why its good)

and then maps the score to a rating bands with these values

```rust
    /// The percentage of votes that denotes an upper bound between good and very good
    const GOOD_UPPER: f64 = 0.8;
    /// The percentage of votes that denotes the line between neutral and good
    const NEUTRAL_UPPER: f64 = 0.55;
    /// The percentage of votes that denotes the line between poor and neutral
    const POOR_UPPER: f64 = 0.45;
    /// The percentage of votes that denotes a line between poor and very poor
    const VERY_POOR_UPPER: f64 = 0.2; 
```
if something is `VERY_GOOD`, what we can say is that "the proportion of positive votes is at least 80%."

static/sass/_snapcraft_p-ratings.scss
M7mdisk: I like colors because they provide quick visual feedback, but if its not recommended I'm happy to change it to normal text. 

I'll wait on Leia's input before proceeding with this, but I'm ok with either choice.

templates/store/_ratings_info.jinja
bartaz: Sounds good. And looks that it is symmetrical. So for very poor it would be at least 80% negative votes (less than 20% positive).

static/sass/_snapcraft_p-ratings.scss
M7mdisk: I had an offline discussion with Leia on this:

the tldr is that yes colors like this aren't great. and that we might instead keep the normal text color but add icons. to keep the quick visual input.

Leia will ask designers what they think is a good icon choice. I'll keep the PR up for now @bartaz

PR #5550, opened 2026-01-12
Charm for snapcraft.io

edisile:
## Done
- added charm snapcraft-io using PaaS App Charmer
- added rock for the application 
- added GitHub Actions workflows for deployment across all environments
- added a `pack-rock` test to CI workflow during PR
- removed unused `SEARCH_API_KEY` env var from application code and k8s `konf` configs

## How to QA
- visit https://staging.snapcraft.io/ which is now being served in its charmed version
- log in
- try to edit snap details, trigger builds, release, etc.
- visit https://staging-api.snapcraft.io/ which is now being served in its charmed version
- log in, **this time with a staging account**
- production env QA is tracked [here](https://docs.google.com/spreadsheets/d/1dYy5HCiixgoyxEIHdcNlvvCxbytnY8IeAuUW9Chg_Bk/edit?gid=701107956#gid=701107956)

## Testing
- [x] This PR has tests

## Security
- [x] Security considerations for review:
  - [x] Sensitive data: secret variables necessary for deployment are kept in GitHub secrets
- [ ] This PR has no security considerations (explain why):

## Issue / Card
Fixes WD-27717

## UX Approval


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,64 @@
...
+          echo "charm_name=${{ vars.CHARM_NAME }}" >> $GITHUB_OUTPUT
+          echo "channel=${{ vars.CHANNEL }}" >> $GITHUB_OUTPUT
+          echo "juju_controller_name=${{ vars.JUJU_CONTROLLER_NAME }}" >> $GITHUB_OUTPUT
+          echo "juju_model_name=${{ vars.JUJU_MODEL_NAME }}" >> $GITHUB_OUTPUT
+          echo "environment=${{ github.event.inputs.environment != '' && github.event.inputs.environment || (github.ref == 'refs/heads/main' && 'Production' || 'Staging') }}" >> $GITHUB_OUTPUT
+
+  deploy:
+    needs: setup
+    name: Deploy
+    uses: canonical/webteam-devops/.github/workflows/deploy.yaml@fix-charmcraft-upload
```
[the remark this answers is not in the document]
edisile: Will commit this when https://github.com/canonical/webteam-devops/pull/13 is merged

rockcraft.yaml
```diff
@@ -0,0 +1,44 @@
...
+  flask-framework/install-app:
+    after:
+      - build-ui
+    prime:
+      - flask/app/.env
+      - flask/app/app.py
+      - flask/app/cache
+      - flask/app/webapp
+      - flask/app/templates
+      # - flask/app/static # it already gets copied in the build-ui step
```
alvaromateo: I wonder if they have changed something in rockcraft cause as far as I understood copying a directory to the "install" directory was not the same as copying/including it in the "prime" directory 🤔

charm/charmcraft.yaml
```diff
@@ -0,0 +1,125 @@
...
+      type: string
+
+    environment:
+      description: "Environment in which the application is running, mostly useless for the charmed version of the app"
+      default: "production"
+      type: string
+
+    marketo-client-id:
+      description: "Marketo API client ID"
+      type: string
```
alvaromateo: I wonder if some of the configuration options should be secrets.
At least in the PR I opened for [Demos in PS7](https://github.com/canonical/snapcraft.io/pull/5642/changes#diff-85be363d1e2ea1dc453fac472e447583db6c546459d3bf39c98bb8c24b6fe867) I changed some of them to secrets.

I'm not sure of the added security of having them as secrets in Juju; for example, I know that k8s handles secrets by just base64 encoding the string...

alvaromateo (commented):
Just one discussion point regarding configuration with plain strings vs secrets that I think we should have.

charm/charmcraft.yaml
edisile: good point, I'll try to cherry pick your changes and add them to this PR before merging it. afaik juju stores secrets in either k8s or a hashicorp vault instance, with the latter being the safer option and what we use on PS6

edisile: I tried but a clean cherry pick isn't possible, also I'm afraid the secrets wouldn't work correctly on the PS7 demo branch. the names I used for the settings are derived from the env vars the app looks for and based on how `juju config` works (`marketo-client-id` becomes `FLASK_MARKETO_CLIENT_ID`). on the other hand, juju's secrets are maps of key-value pairs and can't be simple strings like config values are. secrets are injected as env vars with names of the form `FLASK_<secret name>_<key>` so the names wouldn't match the env vars the app is looking for...

edisile: I think the best way to solve this is group env vars that share a prefix under the same secret, e.g. `github-*` options that map to `GITHUB_*` env vars are replaced by a `github` secret where all the `*` are keys in the secret so the injected env var name matches. I'll make sure to add a description that lists all the keys you'd have to provide in each secret to properly configure the app

edisile: I made the change and tested it locally, everything seems to work as expected. I'll deploy it to staging and then to the prod env as well next week, won't do it now because it's a bit of a hassle to configure all these secrets

charm/charmcraft.yaml
```diff
@@ -0,0 +1,109 @@
...
+    github:
+      type: secret
+      description: |
+        Needed for various functionality. Must contain following keys:
+          client-id: OAuth application ID for prompting users for access to their repositories
+          client-secret: OAuth application client secret for prompting users for access to their repositories
+          snapcraft-bot-user-token: GitHub application token for CVE data
+          snapcraft-user-token: application token for automated builds
+          webhook-host-url: URL of the automated build webhooks' host
+          webhook-secret: Secret salt used for signing automated build webhooks
```
M7mdisk, line 53: very nice 👍

charm/charmcraft.yaml
```diff
@@ -0,0 +1,109 @@
...
+extensions:
+  - flask-framework
+
+requires:
+  redis:
+    interface: redis
+    optional: true
+    limit: 1
+
+config:
```
M7mdisk, line 26: just given the sheer amount of secrets and config options, I think we should encode this in terraform somewhere (as a future item ofc). redploying on another model with this setup is going to be a lot of work

M7mdisk, line 26: the  `ANALYTICS_ENDPOINT` env var is missing

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,64 @@
...
+          - Production
+          - Staging
+          - Staging API
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    environment: ${{ github.event.inputs.environment != '' && github.event.inputs.environment || (github.ref == 'refs/heads/main' && 'Production' || 'Staging') }}
```
M7mdisk: I think, when triggering this with workflow_call, this check might have issues? since that needs `inputs.environment` not `github.event.inputs.environment` which might be undefined . I asked gemini and it agrees, but I am not sure if its a real problem or not

https://github.blog/changelog/2022-06-09-github-actions-inputs-unified-across-manual-and-reusable-workflows/

rockcraft.yaml
edisile: good point, going back and reading the docs I don't even understand why it works... for some reason I used to get errors when copying the directory in the "prime" step, but now I uncommented the line and everything seems to work

charm/charmcraft.yaml
edisile: opened [WD-37187](https://warthogs.atlassian.net/browse/WD-37187) to keep track of this work

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,64 @@
...
+          echo "charm_name=${{ vars.CHARM_NAME }}" >> $GITHUB_OUTPUT
+          echo "channel=${{ vars.CHANNEL }}" >> $GITHUB_OUTPUT
+          echo "juju_controller_name=${{ vars.JUJU_CONTROLLER_NAME }}" >> $GITHUB_OUTPUT
+          echo "juju_model_name=${{ vars.JUJU_MODEL_NAME }}" >> $GITHUB_OUTPUT
+          echo "environment=${{ inputs.environment != '' && inputs.environment || (github.ref == 'refs/heads/main' && 'Production' || 'Staging') }}" >> $GITHUB_OUTPUT
+
+  deploy:
+    needs: setup
+    name: Deploy
+    uses: canonical/webteam-devops/.github/workflows/deploy.yaml@main
```
M7mdisk, line 50: I was a bit confused/hesitant about triggering the flow from itself like this, when I asked copilot its saying:

```
 setup job uses github.event.inputs.environment to determine the target environment. This is null for workflow_call triggers (only workflow_dispatch populates it). When deploy-staging-api.yaml calls this workflow, the fallback logic
 evaluates github.ref == 'refs/heads/main' → true → resolves to 'Production'. So a push to main triggers two Production deployments instead of Production + Staging API.                                                                                                                                                  
```

I'm not sure if its onto something, if not, please ignore.

edisile: I think our boy is a bit confused, there's no way for an input to be null... the documentation explicitly states that `on.workflow_call.inputs` entries can only be strings, numbers or booleans, and none of them can be null not even by default.

in this case the default for `environment` on a workflow_call is an empty string, but deploy-staging-api.yaml passes "Staging API" as an input

M7mdisk (approved):
lgtm

.github/workflows/deploy.yaml
M7mdisk: stupid clanker

PR #5563, opened 2026-01-20
feature: Node-based entry point detection plugin for Vite

edisile:
## Done
- moved vite plugin to a separate file
  - will later move to a separate NPM package
- added `tinyglobby` as dependency
  - we already install it because Vite depends on it
  - dependency will be removed when plugin moves to separate package
- reworked vite plugin
  - no implicit dependency on `grep`
  - runs entirely in node
  - template path matching is done via `tinyglobby`
  - uses any user-provided RegExp pattern to detect entry points
  - added pretty printing for the entry point list

## How to QA
- check out the main branch locally
- run `dotrun build` and take note of the entry points array in the console
- check out this branch locally
- run `dotrun build`, the entry points list should match
- check that the GitHub actions are green and demo works

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): if actions are green all is fine

## Security

  - [ ] Examples:
  - [ ] ...
- [x] This PR has no security considerations (explain why): it's just tooling

## Issue / Card

Fixes #

## UX Approval


steverydz (approved):
LGTM 👍 

The change to `yarn.lock` is missing so would be good to commit that. Otherwise all good.

PR #5569, opened 2026-01-27
feat: Replace YouTube thumbnails with embeds on listing page

steverydz:
## Done
Replaces the YouTube thumbnail with a YouTube embed. This is because the thumbnail isn't always available, or the resolution is too low etc. This PR also updates the CSP headers so that YouTube video URLs without the `www` subdomain are also allowed.

## How to QA
- Go to a snap listing page where the video URL has a YouTube video embed without the `www` subdomain, e.g. https://snapcraft-io-5569.demos.haus/0ad
- Check that there is a YouTube video embed and that it is playable
- Go to a snap listing page where the video URL has a YouTube video embed with the `www` subdomain, e.g. https://snapcraft-io-5569.demos.haus/code
- Check that there is a YouTube video embed and that it is playable

## Testing

- [ ] This PR has tests
- [x] No testing required (explain why): No logic changes

## Security

  - [ ] Examples:
  - [ ] ...
- [x] This PR has no security considerations (explain why): XSS is already handled by escaping user input and the particular input will only accept URLs which must meet a specified format

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-33185

## Screenshots
n/a

## UX Approval

### DIFF
--- a/webapp/handlers.py
+++ b/webapp/handlers.py
@@ -95,7 +95,8 @@ CSP = {
     "frame-src": [
         "'self'",
         "td.doubleclick.net",
-        "www.youtube.com/",
+        "www.youtube.com",
+        "youtube.com",
         "asciinema.org",
         "player.vimeo.com",
         "snapcraft.io",
--- a/konf/site.yaml
+++ b/konf/site.yaml
@@ -71,11 +71,6 @@ env: &env
       key: lp-api-token-secret
       name: snapcraft-io
 
-  - name: YOUTUBE_API_KEY
-    secretKeyRef:
-      key: youtube-api-key
-      name: snapcraft-io
-
   - name: DISCOURSE_API_KEY
     secretKeyRef:
       key: discourse-api-key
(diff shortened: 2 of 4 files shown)


bartaz:
My memories are vague, but I recall that one of the reasons for going for thumbnails instead of straight embeds was performance: to avoid loading whole YT iframe until user actively requests it.

But potentially things have changed on the YT side in last couple years, so maybe it's not too much of a consern. Would be interesting to check if there is any significant impact.

From code perspective, it does seem to simplify things quite a lot.

steverydz:
@bartaz Yes, the initial impetus was an attempt at performance gains, but given we were making a `POST` request to check thumbnails, the gains were negligible. Given the embed loads asynchronously and the bulk of the media isn't downloaded until the video is interacted with should lessen the impact.

The URL format for thumbnails seems to be inconsistent now also, which means we're often stretching out a low res thumbnail or showing the default missing thumbnail image in places.

webapp/handlers.py
```diff
@@ -95,7 +95,8 @@
     "frame-src": [
         "'self'",
         "td.doubleclick.net",
-        "www.youtube.com/",
+        "www.youtube.com",
+        "youtube.com",
```
edisile, line 99: do we also support www.youtube-nocookie.com embeds? if we do, you should add that as well

(for context, the default youtube domains set tracking cookies as soon as the iframe loads, the nocookie domain does it only after the user interacts with the video)

edisile: I checked the logic and we don't, so it's fine

PR #5573, opened 2026-01-30
Ratings

M7mdisk:
## Done
- add ratings to snap details page
## How to QA
- go to https://snapcraft-io-5573.demos.haus/store
- open as many snaps as you can, see if they all load quickly, see if you get any 5xx errors
## Testing

- [x] This PR has tests

## Security

  - [ ] Examples:
  - [ ] ...
- [x] This PR has no security considerations (explain why): all public data accessible through an open source and documented API

## UX approval:
already approved on  #5541


static/sass/styles.scss
```diff
@@ -181,6 +181,8 @@ $color-social-icon-foreground: $color-light;
 @include vf-p-icon-toggle-side-nav;
 @include vf-p-icon-log-out;
 @include vf-p-icon-restart;
+@include vf-p-icon-thumbs-up;
+@include vf-p-icon-thumbs-down;
```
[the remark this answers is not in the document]
M7mdisk: this ships with base vanilla icons

templates/store/snap-details/_package_heading_data.html
```diff
@@ -59,6 +62,9 @@ <h1 class="u-sv-1" data-live="title">{{ package_title }}</h1>
           <a href="/search?categories={{ category.slug }}">{{ category.name }}</a>
         </li>
       {% endfor %}
+      {% if ratings and ratings.total_votes > 0 %}
+        <li class="p-inline-list__item u-hide--small">{{ ratings_info(ratings) }}</li>
+      {% endif %}
```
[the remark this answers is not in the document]
M7mdisk: this would cause an empty `<li>` element

edisile (commented):
code looks good, and QA looked ok initially. the ratings only showed up for a bit though, there's something wrong now because the ratings aren't showing anymore and the demo logs say it's because the gRPC channel is closed or not initialized
```
{"level": "ERROR", "message": "Channel error fetching rating for snap 6gjU50q0lEDaMTtSUxYoVDSmjR5fERpB: Channel is closed or not initialized", "timestamp": "2026-02-02T11:53:24.060060+00:00"}
```

static/js/public/snap-details/ratings.ts
```diff
@@ -0,0 +1,103 @@
...
+    const li = document.createElement("li");
+    li.className = "p-inline-list__item";
+    li.innerHTML = html;
+    list.appendChild(li);
+  }
+}
+
+export default function initSnapRatings(
+  snapId: string,
+  containerSelectors: string[],
```
bartaz, line 90: Do we need multiple selectors?

Even if we do want to render it in exactly the same way in multiple places, it should be possible using a single selector string (that allows to select multiple elements `#js-snap-details-list, #js-some-other-place` and then `document.querySelectorAll()` on it.

webapp/endpoints/views.py
```diff
@@ -81,3 +98,39 @@ def get_brand_store(store_id: str):
...
 
     return response
+
+
+@endpoints.route("/api/snap/<snap_id>/ratings")
+def get_snap_ratings(snap_id):
+    """
+    Get ratings for a snap by snap_id.
+    """
+    if not snap_id:
```
ilayda-cp, line 108: do we need this check? because if snap_id doesnt exist it will throw 404 for /api/snap//ratings?

static/js/public/snap-details/ratings.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  const title = text.charAt(0).toUpperCase() + text.slice(1) + " rating";
+  const html = `
+    <span class="p-snap-ratings">
+      <span title="${title}"><i class="${icon}"></i> ${text}</span>
+      <span> (${ratingsData.total_votes} votes)</span>
+    </span>
+  `;
+
+  const list = document.querySelector(containerSelector);
```
bartaz, line 79: This makes it very tied in to a specific place in our template. Which (in addition to my other comment from line 90) suggests even more that we only mean to append it in one, very specific place (as a list item).

Overall it's fine, we don't need to make it more generic at this point.

static/js/public/snap-details/ratings.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function renderRatingsInfo(
+  ratingsData: RatingsData | null,
+  containerSelector: string,
+): void {
+  const container = document.querySelector(containerSelector) as HTMLElement;
+
+  if (!container) {
+    return;
```
bartaz, line 35: Throwing or logging an error in such case could help notice and debug such unexpected cases - unless we think it's valid use case to call this function without those params?

webapp/endpoints/views.py
```diff
@@ -81,3 +98,39 @@ def get_brand_store(store_id: str):
...
+        return jsonify({"error": "snap_id is required"}), 400
+
+    cache_key = f"snap_ratings:{snap_id}"
+    cached_ratings = redis_cache.get(cache_key)
```
ilayda-cp, line 112: we might need to add expected_type=dict in there. because it returns raw string. redis_cache.get(key) defaults to expected_type=str. So on a cache hit cached_ratings is the raw json string not a dict.

templates/store/package_header/_package_header_data.html
```diff
@@ -1,5 +1,6 @@
 {%- from 'store/package_header/_developer_info.html' import developer_info -%}
 {%- from 'store/package_header/_publisher_info.html' import publisher_info -%}
+{%- from 'store/package_header/_ratings_info.html' import ratings_info -%}
```
ilayda-cp, line 3: it looks like its not called anywhere?

templates/store/package_header/_ratings_info.html
```diff
@@ -0,0 +1,18 @@
+{%- macro ratings_info(ratings) -%}
```
ilayda-cp, line 1: its only imported in the _package_header_data.html and not used anywhere? if thats the case this macro isnt used anywhere we can delete it.

ilayda-cp:
When the detail page is opened it takes some time for it to be loaded like 10secs initially. should we add something like a spinner to indicate "there will be something but it will take some time to be loaded"?

webapp/endpoints/views.py
```diff
@@ -81,3 +98,39 @@ def get_brand_store(store_id: str):
...
+    if cached_ratings is not None:
```
edisile, line 113: this check is correct but has some unintended consequences, see my next comment

webapp/endpoints/views.py
```diff
@@ -81,3 +98,39 @@ def get_brand_store(store_id: str):
...
+        if not ratings_client:
+            return jsonify(None), 200
+
+        ratings_data = ratings_client.get_snap_rating(snap_id)
+
+        if (
+            ratings_data
+            and ratings_data.get("ratings_band") == "insufficient-votes"
+        ):
+            redis_cache.set(cache_key, None, ttl=86400)
```
edisile, line 127: this means that snaps with insufficient votes always bypass the cache check and cause a call to the ratings service even when we already know the result. I think that caching the result as is and handling the "insufficient-votes" band on the frontend would be a better solution

webapp/endpoints/views.py
M7mdisk: You're right but I'd rather be explicit here, the 404 thats returned is with an HTML body which might cause issues if not parsed correctly/if called in other places

PR #5578, opened 2026-02-05
refactor snap reports

M7mdisk:
## Done
- fix reporting and send information to a google sheet
## How to QA
- go to any snap
- click "Report this snap"
- fill the form, go to https://docs.google.com/spreadsheets/d/1BdNwlccAFdX0Y6eKvex-I0i4uQBogBWE_7PViVg5cT4/edit?usp=sharing and see it populated 
## Testing

- [ ] This PR has tests

## Security

- [x] Security considerations for review (list them):
The previous implementation did the request on the client side to the google form directly, which meant we had no control over how many requests, now, at least the submissions have to go through the app, which means we can add any rate limiting and so on if needed. 
We don't sanitize user input, but I do not think that would be a big problem (given that content is just stored in a sheet)
- [ ] This PR has no security considerations (explain why):

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -74,39 +74,68 @@ jobs:
           docker run --network="host" -v .:/app cypress/base:22.18.0 \
             bash "-c" "cd /app && npx cypress install && yarn run test-e2e"
 
+  changes:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      py: ${{ steps.filter.outputs.py }}
+      js: ${{ steps.filter.outputs.js }}
+      scss: ${{ steps.filter.outputs.scss }}
+    steps:
+    - uses: dorny/paths-filter@v3
+      id: filter
+      with:
+        filters: |
+          py:
+            - '**/*.py'
+            - 'requirements.txt'
+            - 'templates/**/*'
+          js:
+            - '**/*.js'
+            - '**/*.ts'
+            - '**/*.jsx'
+            - '**/*.tsx'
+            - 'package.json'
+            - 'yarn.lock'
+            - 'eslint.config.mjs'
+            - 'vitest.config.js'
+            - 'vite.config.js'
+            - 'tsconfig.json'
+            - 'babel.config.json'
+          scss:
+            - '**/*.scss'
+            - '.stylelintrc'
+            - 'package.json'
+            - 'yarn.lock'
+
   lint-python:
+    needs: changes
+    if: ${{ needs.changes.outputs.py == 'true' }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
     steps:
       - name: Checkout
         uses: actions/checkout@v6
-      - uses: dorny/paths-filter@v3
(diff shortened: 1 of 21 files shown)


templates/store/snap-details/_report_snap_modal.html
```diff
@@ -41,18 +40,16 @@ <h5 class="p-notification__title">
...
           <option value="Copyright or trademark violation">Copyright or trademark violation</option>
           <option value="Snap Store terms of service violation">Snap Store terms of service violation</option>
         </select>
         <label for="report-snap-comment">Please provide more detailed reason to your report</label>
-        <textarea id="report-snap-comment" type="text" name="entry.1974584359" placeholder="Comment..." rows="5" maxlength="1000" required></textarea>
+        <textarea id="report-snap-comment" type="text" name="comment" placeholder="Comment..." rows="5" maxlength="1000" required></textarea>
 
         <label for="report-snap-email">Your email (optional)</label>
-        <input id="report-snap-email" type="email" name="entry.1624813972" placeholder="email@example.com" />
-        <label for="report-snap-confirm" style="position: absolute; top: -9999999px;">I agree</label>
```
edisile: why are we removing the honeypot fields?

konf/site.yaml
```diff
@@ -86,6 +86,11 @@ env: &env
       key: dns-verification-salt
       name: snapcraft-io
 
+  - name: REPORT_SHEET_URL
```
bartaz, line 89: Is this set on our prod env already? @M7mdisk

static/js/public/snap-details/reportSnap.ts
```diff
@@ -48,13 +48,6 @@ export default function initReportSnap(
   const modal = document.querySelector(modalSelector) as HTMLElement;
   const reportForm = modal.querySelector("form") as HTMLFormElement;
 
-  const honeypotField = reportForm.querySelector(
```
bartaz, line 51: Wasn't that a bit of spam protection? Don't we need this anymore?

webapp/store/snap_details_views.py
```diff
@@ -535,24 +535,30 @@ def snap_distro_install(snap_name, distro):
...
-            ]
-        )
+        form_url = flask.current_app.config.get("REPORT_SHEET_URL")
+        if not form_url:
+            logger.warning("REPORT_SHEET_URL is not configured")
+            return flask.jsonify({"error": "report_url_missing"}), 503
 
         fields = flask.request.form
 
-        # If the honeypot is activated or a URL is included in the message,
```
bartaz, line 551: Seems we used to have a "honeypot" spam prevention in place (a checkbox that bots were "lured" to click). And any comments with http links were ignored as well. Could be worth keeping them at least to filter out some of the most rudimentary bot spam.

bartaz (commented):
Seems to be working fine, and looks good as a quick fix, but I wonder if we could keep some of the spam protection mechanisms that were there before.

bartaz (approved):
Let's merge it as is for now, to start getting reports and if we start getting spam we can mitigate that later.

PR #5579, opened 2026-02-05
feat: Add endpoint to post new allowlist

steverydz:
## Done
Adds an endpoint to create a new remodel allowlist

## How to QA
Run locally and use [Postman](https://www.postman.com/) to test the responses

### URL
http://localhost:8004/api/store/njwQYXFnS7ppo7LaGxoh7aqVZc1CPi26/models/remodel-allowlist

### Headers
- Content-Type: application/json
- Cookie: YOUR COOKIE
- X-Csrftoken: YOUR TOKEN

### Body
`[{"from-model":"testmodelzzzz","to-model":"my-name-is-luke","from-serial":"GN4R66MulBPDMooy8Ch_VNzkJBW56zFwEupqSodfKhFbfVz_5hSUgSWztUwbvNPB-Hlkj"}]`

### Requests
- Make a request - you should get a conflict error
- Change the serial to a random string and make a request - it should be successful
- Change one of the model names to a random string - you should get an error
- Change the body to `[]` - you should get an error

## Testing

- [x] This PR has tests

## Security

  - [ ] Examples:
  - [ ] ...
- [x] This PR has no security considerations (explain why): Follows existing patterns for posting data

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-33419

## Screenshots
n/a

## UX Approval

### DIFF
--- a/tests/endpoints/tests_models.py
+++ b/tests/endpoints/tests_models.py
@@ -261,3 +261,84 @@ class TestGetRemodelAllowlist(TestModelServiceEndpoints):
         self.assertEqual(response.status_code, 500)
         self.assertFalse(data["success"])
         self.assertEqual(data["message"], "Internal server error")
+
+
+class TestCreateRemodelAllowlist(TestModelServiceEndpoints):
+    @patch(
+        "canonicalwebteam.store_api.publishergw.PublisherGW"
+        + ".create_remodel_allowlist"
+    )
+    def test_create_remodel_allowlist_success(
+        self, mock_create_remodel_allowlist
+    ):
+        mock_create_remodel_allowlist.return_value = None
+
+        payload = {
+            "description": "Test remodel allowlist",
+            "from-model": "test-from-model",
+            "from-serial": "test-from-serial",
+            "to-model": "test-to-model",
+        }
+
+        response = self.client.post(
+            "/api/store/1/models/remodel-allowlist", json=payload
+        )
+        data = response.json
+
+        self.assertEqual(response.status_code, 201)
+        self.assertTrue(data["success"])
+        mock_create_remodel_allowlist.assert_called_once()
+
+    @patch(
+        "canonicalwebteam.store_api.publishergw.PublisherGW"
+        + ".create_remodel_allowlist"
+    )
+    def test_create_remodel_allowlist_store_not_found(
+        self, mock_create_remodel_allowlist
+    ):
+        mock_create_remodel_allowlist.side_effect = StoreApiResponseErrorList(
+            "Store not found", 404, [{"message": "Store not found"}]
+        )
+
+        payload = {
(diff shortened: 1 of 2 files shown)


webapp/endpoints/models.py
```diff
@@ -301,3 +301,44 @@ def get_remodel_allowlist(store_id: str):
...
+        store_id (str): The ID of the store.
+
+    Returns:
+        dict: A dictionary containing the response message and success
+        status.
+    """
+    res = {}
+
+    try:
+        allowlist = flask.request.json
```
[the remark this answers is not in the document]
steverydz: The validation is handled on the store side

webapp/endpoints/models.py
```diff
@@ -301,3 +301,45 @@ def get_remodel_allowlist(store_id: str):
...
+        publisher_gateway.create_remodel_allowlist(
+            flask.session, store_id, allowlist
+        )
+
+        res["success"] = True
+        return make_response(res, 201)
+    except StoreApiResponseErrorList as error_list:
+        res["success"] = False
```
bartaz, line 332: Not sure if we do for other similar endpoints, but should we log something here when error happens to have some kind of trace of it?

tests/endpoints/tests_models.py
```diff
@@ -262,3 +262,84 @@ def test_get_remodel_allowlist_general_error(
...
+            "from-model": "test-from-model",
+            "to-model": "test-to-model",
+        }
+
+        response = self.client.post(
+            "/api/store/999/models/remodel-allowlist", json=payload
+        )
+        data = response.json
+
+        self.assertEqual(response.status_code, 500)
```
bartaz: Doesn't matter too much, but would it be possible to pass the HTTP code from the store? So when it's 404 on store side, we return the same?

webapp/endpoints/models.py
```diff
@@ -300,3 +300,45 @@ def get_remodel_allowlist(store_id: str):
...
+        messages = [
+            f"{error.get('message', 'An error occurred')}"
+            for error in error_list.errors
+        ]
+        res["message"] = " ".join(messages)
+        return make_response(res, 500)
+
```
edisile, line 344: it would make sense to add a dedicated case for `StoreApiResourceNotFound`, right now sending an incorrect serial value means we swallow the 404 from the store backend and the user doesn't know why the request failed

webapp/endpoints/models.py
steverydz: We don't do that for the other endpoints at the moment, but we should probably have a standardised way of doing that

steverydz:
@bartaz @edisile I've updated this so that we will use the status code from the store api when there is an error, and also have an `StoreApiResourceNotFound` to handle that case

edisile (approved):

PR #5586, opened 2026-02-09
Fix Python/JS/styles tests/linting on CI

edisile:
## Done
- reverted `responses` package to `0.22.0`
- added check to run `test-python` and `lint-python` jobs when `requirements.txt` changes
- added check to run `test-js` and `lint-js` jobs when `package.json` or `yarn.lock` change
- fixed `lint-scss` issues
- ran `black` to fix some formatting issues

## How to QA
- check that the `test-python` job is green
- check that the `test-python` job ran *all* steps

## Testing

- [ ] This PR has tests
- [x] No testing required (explain why): it's a simple CI fix

## Security

  - [ ] Examples:
  - [ ] ...
- [x] This PR has no security considerations (explain why): it's a simple CI fix

## Issue / Card

Fixes #

## UX Approval

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -90,6 +90,7 @@ jobs:
           py:
             - '**/*.py'
             - 'requirements.txt'
+            - 'templates/**/*'
           js:
             - '**/*.js'
             - '**/*.ts'
@@ -99,6 +100,9 @@ jobs:
             - 'yarn.lock'
             - 'eslint.config.mjs'
             - 'vitest.config.js'
+            - 'vite.config.js'
+            - 'tsconfig.json'
+            - 'babel.config.json'
           scss:
             - '**/*.scss'
             - '.stylelintrc'


.github/workflows/pr.yml
```diff
@@ -173,6 +173,7 @@ jobs:
           filters: |
             py:
               - '**/*.py'
+              - 'requirements.txt'
 
       - name: Install dotrun
```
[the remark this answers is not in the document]
edisile: good catch

.stylelintrc
```diff
@@ -16,6 +16,7 @@
     "scss/no-global-function-names": null,
     "declaration-block-no-redundant-longhand-properties": null,
     "selector-not-notation": null,
+    "color-function-alias-notation": null,
```
bartaz, line 19: I was wondering if we should ignore this rule, or just fix the couple `rgba` aliases we use… Although, I guess we don't really care about which notation we use, so it's fair go ignore.

.github/workflows/pr.yml
bartaz: We are missing package.json/yarn.lock checks here.

.github/workflows/pr.yml
```diff
@@ -149,6 +150,9 @@ jobs:
               - '**/*.jsx'
             tsx:
               - '**/*.tsx'
+            package.json:
+              - 'package.json'
+              - 'yarn.lock'
 
       - name: Install JS dependencies
         if: ${{ steps.filter.outputs.js == 'true' || steps.filter.outputs.ts == 'true' || steps.filter.outputs.jsx == 'true' || steps.filter.outputs.tsx == 'true' }}
```
bartaz: If we are adding new filter it probably needs to be added here as well. For that reason, maybe better to avoid naming it `package.json` as it may mess up the dot notation here. Maybe `dependencies`?

.github/workflows/pr.yml
bartaz: When stylelintrc changes, we should run this as well. Now, do we have any other linter configs?

More and more I think we should ditch those filters altogether. Or at least make them quite explicit if they skipped any part, so it's visible in the checks list.

bartaz: When stylelintrc changes, we should run this as well. Now, do we have any other linter configs?

More and more I think we should ditch those filters altogether. Or at least make them quite explicit if they skipped any part, so it's visible in the checks list.

.github/workflows/pr.yml
edisile: it doesn't have to be in a separate property, but I agree the name isn't that great. I'll change it

edisile: it doesn't have to be in a separate property, but I agree the name isn't that great. I'll change it

.github/workflows/pr.yml
```diff
@@ -173,6 +181,8 @@ jobs:
           filters: |
             py:
               - '**/*.py'
+            py_dependencies:
```
bartaz: I think it was fine in one filter. Otherwise we need to check them both in every `if` statement in the job. Currently we are just checking the `py` filter (line 188).

edisile: good catch, I completely checked out when I added these checks

.github/workflows/pr.yml
edisile: apparently there's no way to quit a job early and set the state to something other than a failure... ([source](https://github.com/orgs/community/discussions/82744)) there is a way to move the checks at a job level and skip the job altogether which should show it as a skip in the review summary, maybe I can try that

.stylelintrc
edisile: I think it's fine to ignore it, from what I could gather from Sass docs the syntax we use is still fully supported

.github/workflows/pr.yml
edisile: moving the check upfront looks and works better, the status is much clearer now

.github/workflows/pr.yml
```diff
@@ -74,91 +74,96 @@ jobs:
...
+            - 'vitest.config.js'
+          scss:
+            - '**/*.scss'
+            - '.stylelintrc'
+            - 'package.json'
+            - 'yarn.lock'
+
   lint-python:
+    needs: changes
+    if: ${{ needs.changes.outputs.py == 'true' }}
```
bartaz, line 114: I wonder if that's gonna work well with them being "required" checks.

Because I assume when they are skipped, they will not show up as completed (which is good on it's own), but may be harder to enforce them passing.

Because if we remove them from required checks, it may allow to merge with them failing.

I wonder if we can make given check required to pass, but still allow it to be skipped in settings somehow. But that we would have to QA in practice on another PR I guess.

bartaz:
yarn.lock doesn't seem up to date, it get's recreated when I run `dotrun install` locally.

.github/workflows/pr.yml
```diff
@@ -74,91 +74,96 @@ jobs:
...
+            - 'requirements.txt'
+          js:
+            - '**/*.js'
+            - '**/*.ts'
+            - '**/*.jsx'
+            - '**/*.tsx'
+            - 'package.json'
+            - 'yarn.lock'
+            - 'eslint.config.mjs'
+            - 'vitest.config.js'
```
bartaz, line 102: We also seem to have babel.config.json, tsconfig.json … not sure if anything else may be needed here?

.github/workflows/pr.yml
```diff
@@ -74,91 +74,96 @@ jobs:
...
+      py: ${{ steps.filter.outputs.py }}
+      js: ${{ steps.filter.outputs.js }}
+      scss: ${{ steps.filter.outputs.scss }}
+    steps:
+    - uses: dorny/paths-filter@v3
+      id: filter
+      with:
+        filters: |
+          py:
+            - '**/*.py'
```
bartaz, line 91: Interesting consideration (pointed out by copilot locally) is that we seem to have some python tests that render HTML templates, so changes to HTML templates could require running python tests as well.

.github/workflows/pr.yml
edisile: according to the docs `skipped` doesn't block merging:
> Required status checks must have a successful, skipped, or neutral status before collaborators can make changes to a protected branch
[source](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging)

PR #5602, opened 2026-02-25
Package name in sidebar (and also near the title, if necessary)

edisile:
## Done
- added package name in details sidebar
- added package name near package title if it differs too much from the name
  - title might be misleading if `package_title.replace(" ", "-").lower() != package_name`

## How to QA
- visit https://snapcraft-io-5602.demos.haus/lxd
  - package title is "LXD"
  - the package name does not appear near the title
- visit https://snapcraft-io-5602.demos.haus/gedit
  - package title is "Text Editor (gedit)"
  - hovering over "(gedit)" displays a tooltip
- visit https://snapcraft-io-5602.demos.haus/test-name-that-is-a-test
  - package title is "looooong package title != package name < line break > (test-name-that-is-a-test)"
  - resize the window, the package name size shrinks

## Testing

- [ ] This PR has tests

## Security

  - [ ] Examples:
  - [ ] ...
- [x] This PR has no security considerations (explain why): this PR doesn't have any impact on the application from a security point of view

## Issue / Card

Fixes WD-33034

## UX Approval

- [x] This PR does require UX approval (add context):
The idea behind this change is to mitigate the risk of fraudulent packages: by making it clear when a snap's title and its name don't match, users will hopefully be able to avoid installing malicious snaps.
To achieve this, adding the package name to the details sidebar is a good first step, but this sidebar isn't always immediately visible on first scroll (e.g. snaps with screenshots/videos), meaning this measure isn't very effective IMO. Adding the package name near the title at the top of the page, on the other hand, makes the mismatch between name and title easier to spot.

Other ideas to consider/explore:
- moving the package name down, putting it on the same line as the publisher info and snap categories
- using some form of string distance to quantify the difference between the package name and package title, adding a warning if it's above a certain threshold
- only showing applying this logic for non-verified publishers

### DIFF
--- a/static/sass/_snapcraft_snap-details.scss
+++ b/static/sass/_snapcraft_snap-details.scss
@@ -106,4 +106,39 @@
       opacity: 0.3;
     }
   }
+
+  .p-package-header {
+    container: pkgname / inline-size;
+
+    &__name {
+      display: inline-block;
+      white-space: nowrap;
+
+      .p-tooltip__message {
+        font-weight: 400;
+      }
+
+      --length: 40; // default value, it should be overridden via inline styles on the element
+      --fs-ratio: 1.765; // height/width (1rem/1ch) ratio measured in the browser
+      // we take the full width of the container and divide it by the number of characters in the
+      // package name (+2 to account for the parentheses): this is the horizontal space that each
+      // character has on average; we multiply this by the fs-ratio to get the font size
+      --target-font-size: calc(
+        (100cqw / (2 + var(--length))) * var(--fs-ratio)
+      );
+      // we limit the font size to the smallest and largest available sizes
+      --max-font-size: #{map-get($settings-text-h1-mobile, font-size)}rem; // size of the title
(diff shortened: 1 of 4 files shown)


leiaru-c:
I agree with the rationale of being more explicit with package names that differ too much from their titles, and that it needs to be more visible to help users find clues on whether or not they want to trust that application. 
I would recommend making the parentheses content the same size as the main title for a more homogeneous look. There will be issues with headers that are close to the character limits, but I believe that combining a reduced size above 20-ish characters, similar to what we do for smaller screen sizes, and careful line breaks to avoid cutting the content inside the parentheses, it should not create too many legibility issues.

edisile:
I added some styles to implement an adaptive font size based on the length of the package name; for most snaps the font will be as close as possible to the rest of the title and won't wrap awkwardly anymore, but for those snaps with long names the font will shrink in order to fit the whole name on one line

abbiesims (approved):
LGTM! 👍

PR #5653, opened 2026-04-07
WD-34911 - RTK migration and tests

alvaromateo:
## Done

- Migration of releases UI to RTK.


static/js/publisher/pages/Releases/selectors/index.ts
```diff
@@ -196,7 +203,7 @@ export function getFilteredAvailableRevisionsForArch(
 // get list of architectures of uploaded revisions
 export function getArchitectures(state: ReleasesReduxState) {
   return state.architectures && state.architectures.length > 0
-    ? state.architectures.sort()
+    ? [...state.architectures].sort()
```
edisile: can't we remove the `.sort` from here and do it when setting the architectures array? no idea why we didn't do it before...
it wasn't really an issue in the old implementation because that would mutate the underlying array and then make the next `.sort()` a no-op, whereas this would copy and sort the array every time which is quite wasteful

static/js/publisher/pages/Releases/slices/architectures.ts
```diff
@@ -0,0 +1,26 @@
...
+  name: "architectures",
+  initialState: [] as ArchitecturesState,
+  reducers: {
+    updateArchitectures: {
+      prepare(revisions: Revision[]) {
+        let archs: string[] = [];
+        revisions.forEach((revision) => {
+          archs = archs.concat(revision.architectures);
+        });
+        archs = archs.filter((item, i, ar) => ar.indexOf(item) === i);
```
edisile: from the odd `indexOf` duplicates check I guess this is a carryover from the old implementation? using a `Set` would be a more modern way to do it

static/js/publisher/types/releaseTypes.ts
```diff
@@ -45,8 +46,8 @@ export type Series = "16" | (string & {}); // series is and will always be 16, b
 
 export type Progressive = {
   "current-percentage": number | null;
-  paused: boolean | null;
```
edisile, line 48: should we keep this in the type since it's necessary for the API? even though we don't actually have a way to use it?

steverydz (approved):
LGTM 👍 

Everything is working fine as far as I can see. The code looks good. I looked at the code overall rather than the diff. The code is definitely a lot cleaner and easier to follow than it was. I couldn't find any issues that stood out.

edisile (approved):
great job!

static/js/publisher/types/releaseTypes.ts
alvaromateo: I added it to a separate type:
```
export type ProgressivePayload = {
  percentage: number | null;
  paused: false;
};
```

This way we know that it is necessary for the API but when working with the type for our UI we don't get it in the middle or distracting anyone. If in the future the Stores Team decide to add that functionality then we can move it.

static/js/publisher/pages/Releases/selectors/index.ts
alvaromateo: Done

static/js/publisher/pages/Releases/slices/architectures.ts
alvaromateo: Done

PR #5670, opened 2026-04-16
Restore new combobox + fixes

edisile:
## Done
Reapplied reverted combo box PR (#5617) along with a few fixes. The reverted code has already been reviewed, so you can skip it and review only the new changes by comparing the tip of this branch to the reapply commit: b25b1b5d219e185074ae7fc7c9e3a35d4a480f50

The reverted version had a problem where navigating to a nested brand store route and refreshing the page would bring you back to `/admin/< store id >/snaps` rather than the page you were viewing before. Fixes:
- compare current store id with selected store id in StoreSelector component, call `navigate(...)` only if they're different
- improved ComboBox behavior during init (onChange doesn't fire until there are some options to choose from)

As a drive-by I added a primary navigation fix for highlighting the "Models" section when visiting remodel pages

## How to QA
- visit https://snapcraft-io-5670.demos.haus/admin/ahnuP3quahti9vis8aiw/models/test-superalpaca-00/remodel
  - the "Models" link in the side nav should be highlighted
- refresh the page
  - the URL doesn't change
- select a different store
  - the URL changes to /< store id >/snaps

## Testing

- [x] This PR has tests (restores previously merged tests)

## Security

  - [ ] Examples:
  - [ ] ...

## Issue / Card

Fixes #

## UX Approval


static/js/publisher/components/PrimaryNav/PrimaryNav.tsx
```diff
@@ -104,8 +104,9 @@ function PrimaryNav(): React.JSX.Element {
                         component: NavLink,
                         to: `/admin/${storeId}/models`,
                         icon: "models",
-                        "aria-current":
-                          location.pathname === `/admin/${storeId}/models`,
+                        "aria-current": location.pathname.includes(
+                          `/admin/${storeId}/models`,
```
ilayda-cp: changing to "includes" makes it too loose. It might work now but in the future if we add something like "/admin/${storeId}/models-archive" it would cause a bug.

ilayda-cp (approved):

PR #5708, opened 2026-05-20
Remove major updates grouping from Renovate config

abbiesims:
## Done
- Removes major updates grouping so that they are not omitted


steverydz (approved):
LGTM 👍

PR #5714, opened 2026-05-22
fix: block list-parser rule bypass

ilayda-cp:
## Done
- Stopped the mistune 3 upgrade from re-enabling h1, blockquote, and pre code rendering when a list is followed by # heading, > quote, or a triple-backtick fence (``` some code ```). mistune 3's list parser was calling restricted parse methods directly, bypassing our reduced rule set.
- Added four regression tests covering the leak scenarios and one for the email autolink.

## How to QA
- go to https://snapcraft-io-5714.demos.haus/ilayda-test-hello
- Find the line that reads # 
  - This should NOT be a heading.
  - It should be the same font size and weight as the surrounding paragraph text
- Find > This should NOT be a blockquote.
  - It should sit flush-left like normal text, not indented and not in a grey/bordered quote box.
- Find the this should not be a code block line.
  - It should be plain paragraph text with literal backticks visible around it. It should not be inside a grey monospace box.
- On the line with "bold and italic and code and strike"
  - Verify bold is in bold, italic is in italics, code is in a monospace inline span, strike is ~~struck through~~.
- The "one / two / three" block should appear as a bulleted list with visible bullet points, not as three separate lines of plain text.
- https://snapcraft.io in the "Visit ... for more." sentence should be a clickable blue link.
- The three-space indented code line should appear inside a grey monospace  box.

## Testing

- [x] This PR has tests

## Security

  - [ ] Output tags allowlist: the fix narrows what HTML the parser can emit. Anything outside the allowlist is now coerced into a paragraph whose contents flow through HTML escaping, so injected <script>, <style>, etc. stay escaped.
  - [ ]  mailto autolink: auto_email was re-enabled, but the regex only matches email-shaped input (no : in the local part), so alternate schemes like javascript: cannot be smuggled in. The URL goes through mistune's escape_url and the renderer's safe_url (which already blocks harmful protocols).
  - [ ] No new user-input sinks, no new outbound calls, no auth/permission changes.

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-36737

### DIFF
--- a/tests/tests_markdown_parser.py
+++ b/tests/tests_markdown_parser.py
@@ -175,3 +175,34 @@ class TestMarkdownParser(unittest.TestCase):
         expected_result = "<p>" + markdown + "</p>\n"
 
         self.assertEqual(result, expected_result)
+
+    def test_list_followed_by_heading_does_not_render_heading(self):
+        markdown = "- a\n# Title"
+        expected = "<ul>\n<li>a</li>\n</ul>\n<p># Title</p>\n"
+        self.assertEqual(parse_markdown_description(markdown), expected)
+
+    def test_list_followed_by_blockquote_does_not_render_blockquote(self):
+        markdown = "- a\n> quoted"
+        expected = "<ul>\n<li>a</li>\n</ul>\n<p>&gt; quoted</p>\n"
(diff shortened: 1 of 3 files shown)


steverydz (approved):
LGTM 👍

PR #5727, opened 2026-06-02
feat: Add date selectors to the serial log UI

steverydz:
## Done
Adds date filters to the serial logs view as currently it only shows the past 30 days of logs

## How to QA
- Go to https://snapcraft-io-5727.demos.haus/admin/ahnuP3quahti9vis8aiw/models/test-superalpaca-00/serial-log
- Click "Select custom date range"
- Choose a 30 day time period which includes 23rd March (these are the only serial logs we have)
- Check that the serial log table is populated
- Reload the page
- Check that the custom range panel is open and that the start and end values are still populated
- Close the panel
- Check that the start and end have reverted to the current 30 day period
- Try expanding the date range to more than 30 days
- Check that there is a notification that warns you about this

## Testing

- [x] This PR has tests

## Security

  - [ ] Examples:
  - [ ] ...
- [x] This PR has no security considerations (explain why): User input is escaped

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-35106

## Screenshots
n/a

## UX Approval

- [x] This PR does require UX approval (add context):

The serial log API only shows results for up to a 30 day period, and by default it returns data for the previous 30 days. Currently the user has no way to select a time period of logs to view. This PR implements a panel where a user can specify a time period to view.

### DIFF
--- a/static/js/publisher/pages/SerialLog/SerialLogDateSelectors.tsx
+++ b/static/js/publisher/pages/SerialLog/SerialLogDateSelectors.tsx
@@ -1,0 +1,257 @@
+import { useEffect, useState } from "react";
+import { useLocation, useNavigate, useSearchParams } from "react-router-dom";
+import { Button, Notification, Row, Col } from "@canonical/react-components";
+import { differenceInCalendarDays, format, parseISO, subDays } from "date-fns";
+
+type Props = {
+  onApplyDateRange: () => void;
+};
+
+const DEFAULT_START_TIME = "00:00:00";
+const DEFAULT_END_TIME = "23:59:59";
+
+function formatDateInputValue(date: Date): string {
+  return format(date, "yyyy-MM-dd");
+}
+
+function formatTimeInputValue(date: Date): string {
+  return format(date, "HH:mm:ss");
+}
+
+function getDefaultDateRange(): { startDate: string; endDate: string } {
+  const today = new Date();
+
+  return {
+    startDate: formatDateInputValue(subDays(today, 29)),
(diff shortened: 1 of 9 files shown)


static/js/publisher/pages/SerialLog/SerialLogDateSelectors.tsx
```diff
@@ -0,0 +1,257 @@
...
+}
+
+function getDefaultDateRange(): { startDate: string; endDate: string } {
+  const today = new Date();
+
+  return {
+    startDate: formatDateInputValue(subDays(today, 30)),
+    endDate: formatDateInputValue(today),
+  };
+}
```
edisile, line 28: submitting the form with the default date range leads to this API error (I guess the backend counts the dates differently compared to date-fns? or is it the start/end time?)
image
if I shift the start date forward by 1 it works as expected (it shows there are no logs in the last 30 days)

edisile: also not sure if this is intentional, but once this error appears there's no way to go back and change the dates (outside of going to another page and back to this one or changing the URL manually)

steverydz: @edisile Good catch. Standard off-by-one problem. I've fixed the logic so it's 30 days inclusive and doesn't hide the rest of the UI when the error shows. I've also disabled the button when the date range exceeds 30 days.

edisile (approved):

PR #5742, opened 2026-06-11
feat: Add AGENTS.md and CLAUDE.md for usage with agents

steverydz:
## Done

Adds a `AGENTS.md` file to the project for when developers are using agents for local development. Added `CLAUDE.md` also which reads from `AGENTS.md`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,79 @@
+# AGENTS.md
+
+This file provides guidance to agents when working with code in this repository.
+
+snapcraft.io is the web frontend for the Snap Store: a stateless Flask (Python 3.10+) app serving Jinja2 templates, with React 19 / TypeScript islands bundled by Vite 7 and styled with Vanilla Framework (SCSS).
+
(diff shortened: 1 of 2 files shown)


alvaromateo (commented):
I have a couple of suggestions:
1. I think we should include links to documentation for specific tools like dotrun and Vanilla Framework that may be a bit more unknown to LLMs. This might be a bit tricky cause if you just put a URL I think the agent is not smart enough to browse all the site. I wonder if it is better to give it the Git Repo URL and tell the agent to clone the repo and inspect the code when it needs to work with the given tool.
2. I think that given that we have good test coverage on snapcraft we are missing a section on "workflow" to tell the agent how to work. For example something simple like:
    - Analyse and read about any tool that you may need to use (point 1 of this comment).
    - Make your changes.
    - Write tests for your changes.
    - Run all the tests with `dotrun test-all`
    - Run linting with `dotrun lint-all`
    - (If we still have e2e cypress/playwright tests) Run all e2e tests

In the future we can keep working on this section by creating a separate agent for the testing and another for the reviewing (which we kinda already have thanks to the copilot instructions - we just need to be able to run it locally). And include a review in the above workflow to further improve the results.

alvaromateo (approved):
Looks good! Thanks for the initiative @steverydz

PR #5771, opened 2026-06-22
chore: refactor login and macaroon auth

M7mdisk:
## Done
- Fix the login approach to exchange the raw root + discharge macaroons for a single  store token at login time, then discard both originals. (instead of storing the exchange and discharge one, we only keep the store one, reducing cookie size)

## How to QA
- Login with demo link
- Compare the `session` cookie size with what you get in snapcraft.io
- hopefully its less
- Make sure all auth routes work fine, no 401s, weird errors etc

## Issue / Card

Fixes WD-37297


edisile:
code looks good, login and logout work. before approving, I'll wait for the token to expire to see if the refresh flow works correctly as well

edisile (approved):
refresh flow worked correctly as well

PR #5799, opened 2026-07-17
Wd 37736 auditable snaps

ilayda-cp:
## Done
- Under the Install button, snaps built on Launchpad now show the exact source commit they came from with a direct link.
- A new Security tab lists recent versions with their build and source-commit links, per architecture.
- When a snap has no public source (uploaded manually or private), nothing misleading is shown.
- Added analytics to measure how often provenance is available and whether people use it.

## How to QA
- go to https://snapcraft-io-5799.demos.haus/openshell
- verify there is no public commit /build for this version
Screenshot 2026-07-20 at 11 20 55 AM

- go to https://snapcraft-io-5799.demos.haus/mumble
- Verify the commit and build links are shown
- Click on the commit 
- verify it navigates to the correct github commit 
- Click on the build link
- Verify it navigates to the correct Launchpad build
Screenshot 2026-07-20 at 11 27 07 AM

- go to https://snapcraft-io-5799.demos.haus/audacity
- Verify the commit and build are not available for the latest but available for other channels
Screenshot 2026-07-20 at 12 22 07 PM

- verify the build and commit information are fetch async and doesnt block the detail page to be loaded.
- Verify while loading a loading state is shown both on the header part and the table

## Testing

- [x] This PR has tests

## Security

  - [ ] Examples:
  - [ ] ...

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-37736

## UX Approval


webapp/endpoints/snaps.py
```diff
@@ -86,6 +95,160 @@ def dns_verified_status(snap_name):
     return response
 
 
+def _get_provenance_map(snap_name):
+    """Return the (cached) Launchpad provenance map for a snap.
+
+    The map is expensive to build (paginated Launchpad calls), so it is cached
+    for an hour and shared by both auditable endpoints.
+    """
+    cache_key = get_auditable_map_cache_key(snap_name)
```
edisile, line 114: maybe when we release a snap we should remove its auditable data from the cache? still it's not that big of a deal if the data is stale for 1 hour I think

static/js/public/snap-details/escape.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/**
+ * Return an href-safe, HTML-escaped URL. Only http(s) and relative URLs are
+ * allowed through; anything else collapses to "#" so publisher/recipe-controlled
+ * URLs can't smuggle a script scheme into a link.
+ */
+export function safeUrl(value: unknown): string {
+  const url = String(value ?? "").trim();
+  if (/^https?:\/\//i.test(url) || url.startsWith("/")) {
```
edisile: I think this covers most cases, but using the builtin `URL.parse` method and checking the `protocol` value might be a bit safer.

Also, can a valid URL we use in this feature start with "/"? Or is this supposed to a generic helper we can use in other places across the app?

static/js/public/snap-details/detailsTabs.ts
```diff
@@ -0,0 +1,40 @@
+export default function initDetailsTabs(): void {
```
edisile, line 1: super-mega-ultra nitpick: the label for the tab is "Description", maybe we should call it "description" rather than "detail" in the filename and HTML attributes for consistency?

edisile (approved):
great work! left a few comments but they're mostly nitpicks

webapp/endpoints/snaps.py
ilayda-cp: yea we can add it

static/js/public/snap-details/escape.ts
ilayda-cp: its not a generic helper YET. nothing else needs it and a generic version might want different rules. better to promote it when a real second use case shows up. in this case the url never start with '/'

static/js/public/snap-details/detailsTabs.ts
ilayda-cp: details-tab is on both the Description and Security tabs "details" here means the snap details page, not the Description tab specifically. so i would keep it.

leiaru-c:
Works for me :)

PR #5805, opened 2026-07-22
feat: Setup project for Pragma adoption

canonical-steverydz:
## Done

Sets up the project for Pragma migration. After this, React components that exist in Pragma should be able to be migrated. Available components in Storybook: https://ds-react-global.canonical.com/

### Icons

According to the [implementation spec](https://docs.google.com/document/d/1WhO8JYCEFpa6v-rXthe19qPmXvAAm1PU8zkKJoVnRIw/edit?tab=t.iww1koxwfh0#heading=h.agssxzylu7ma), icons need to be served from /icons at the root of the project. On snapcraft.io, with the exception of a few pages, any path from the root relates to a specific snap so whilst there doesn't appear to be a snap called "Icons", this approach could cause problems in the future. 

The solution I have proposed is to put the icons in the `/static` directory and redirect any requests to `/icons/<ICON_NAME>.svg` to `/static/icons/<ICON_NAME>.svg`. This targets only SVG files so that if there were to be a snap called "Icons" then /icons would show the snap, and the publisher pages, e.g. /icons/listing etc. would also load correctly.

### Fonts

According to the [implementation spec](https://docs.google.com/document/d/1WhO8JYCEFpa6v-rXthe19qPmXvAAm1PU8zkKJoVnRIw/edit?tab=t.iww1koxwfh0#heading=h.8unqzxutbu9v), we shouldn't have to include fonts from Pragma if they are already loaded by Vanilla. In practice this doesn't seem to work, as there is a conflict with the variables. The solution was to redefine the font variables after importing the pragma styles. The alternative is to include the Pragma fonts, but the original Vanilla font still loads, therefore we'd be downloading unused fonts.

See [adoption spec](https://docs.google.com/document/d/1WhO8JYCEFpa6v-rXthe19qPmXvAAm1PU8zkKJoVnRIw/edit?tab=t.0)

## Issue
Fixes https://warthogs.atlassian.net/browse/WD-37851


scripts/link-ds-icons.mjs
```diff
@@ -0,0 +1,103 @@
...
+      "Cannot link DS icons: node_modules/@canonical/ds-assets/icons " +
+        "does not exist. Run yarn install first.",
+    );
+    process.exit(1);
+  }
+
+  const sourceRealPath = await realpath(sourcePath);
+  await mkdir(path.dirname(targetPath), { recursive: true });
+
+  if (await pathExists(legacyTargetPath)) {
```
alvaromateo: What do we need this legacy part for? I don't see any "icons" folder in our code.
Is there any corner case that I'm missing?

scripts/link-ds-icons.mjs
```diff
@@ -0,0 +1,103 @@
...
+      if (legacyLinkRealPath === sourceRealPath) {
+        await unlink(legacyTargetPath);
+      }
+    }
+  }
+
+  if (await pathExists(targetPath)) {
+    const targetStat = await lstat(targetPath);
+
+    if (!targetStat.isSymbolicLink()) {
```
alvaromateo: I'm having issues understanding this part.

Just by looking at the code there is the creation of a directory `static/icons` in line 49 with `mkdir`. Then we come into this if because the path exists given that we just created it (or I see it could also be created in the Dockerfile by the COPY instruction). Then if the target is not a symbolic link we log an error and exit. It looks to me like this is what is going to happen all the time unless we go and manually create a soft link with `ln -s <source> <target>`.

alvaromateo (commented):
I like the approach you did regarding the `redirect` to make sure the icons don't mess up with any snap.
On the other hand, I have a couple doubts regarding the script that sets up the linking with the ds-assets repo.
Finally, now that snapcraft is charmed maybe we'll need to add also something to the rockcraft.yaml file; same way you added it to Dockerfile.

scripts/link-ds-icons.mjs
canonical-steverydz: true, we don't need this - I'll remove it

canonical-steverydz:
@alvaromateo I've made some changes based on your comments. I added copying the icons to `rockcraft.yaml` but I'm not sure how to test it works without deploying the site.

scripts/link-ds-icons.mjs
alvaromateo: Okay, I see with your changes that I had mistakenly thought you were creating `static/icons` but in fact you were creating just `static/`. My bad 😅 Now it makes sense what the script is doing, thanks!

alvaromateo:
> @alvaromateo I've made some changes based on your comments. I added copying the icons to `rockcraft.yaml` but I'm not sure how to test it works without deploying the site.

Thanks for the changes @steverydz! All is clear now.

I think it's a good approach given the constraints that we have been imposed by pragma's requirements. I was thinking if to detect if the user is requesting an icon or a snap called 'icons' we could use maybe the MIME type in one of the headers of the request; but I don't know if it would be as simple as your solution, which is a good thing IMO. Let's see what opinions do other people of the team have :)

The only thing I would recommend is maybe adding a comment on top of `add_url_rule` and maybe the linking script to explain that we are doing this because of pragma's requirements so that new people (and us in the future) can understand why we have this. But I'm not sure what's the standard approach regarding that in snapcraft, so I let you decide if and where to put it.

Good job :+1:

alvaromateo (approved):

canonical-steverydz:
@alvaromateo Thanks - I added a comment

scripts/link-ds-icons.mjs
```diff
@@ -0,0 +1,85 @@
...
+  } catch (error) {
+    if (error.code === "ENOENT") {
+      return false;
+    }
+
+    throw error;
+  }
+}
+
+async function main() {
```
bartaz-canonical: Why do we need such an elaborate script to link the icons?
Could it work just to add a symlink in /static and add it to git? (it would of course depend on the package being installed for the link to work, but seem simpler than recreating the link every time).

Unless I don't see something missing there?

Dockerfile
```diff
@@ -15,6 +15,7 @@ FROM node:24 AS yarn-dependencies
 WORKDIR /srv
 ADD package.json .
 ADD yarn.lock .
+ADD scripts/link-ds-icons.mjs scripts/link-ds-icons.mjs
```
bartaz-canonical: Do we need this script copied? Later in line 33 we are copying the icons anyway to the right path.

webapp/app.py
```diff
@@ -53,6 +54,23 @@ def create_app(testing=False):
...
+    # the URL pattern we use for snap names. To resolve this, we are
+    # redirecting any requests to `/icons/<ICON_NAME>.svg` to
+    # `/static/icons/<ICON_NAME>.svg`, which leaves the possibility for a
+    # snap called "icons" and its subsequent publisher pages.
+    #
+    # The script that handles this is `scripts/link-ds-icons.mjs`
+    app.add_url_rule(
+        "/icons/<icon_name>.svg",
+        "ds-icons",
+        lambda icon_name: redirect(
```
bartaz-canonical: Probably not an issue at the moment, but this may have a bit of a performance downside of serving all the icons via redirects. If we can add specific route handling just for `/icons/<icon_name>.svg` maybe we could serve the icons directly from there, without redirect?

But, if it's easier as is, let's keep it for now.

.dockerignore
```diff
@@ -21,3 +21,5 @@ run
 
 # Node is only needed for building, not running
 node_modules
+icons
```
bartaz-canonical: Is there an `icons` folder in root that needs to be ignored?

canonical-steverydz: No, this is a hangover from an earlier attempt. Good catch!

Dockerfile
canonical-steverydz: Yes, because `yarn install --production` will fail otherwise when building the Docker image because it needs access to that script.

webapp/app.py
```diff
@@ -53,6 +54,23 @@ def create_app(testing=False):
...
+    # which leaves the possibility for a snap called "icons" and its
+    # subsequent publisher pages.
+    #
+    # The script that handles this is `scripts/link-ds-icons.mjs`
+    app.add_url_rule(
+        "/icons/<icon_name>.svg",
+        "ds-icons",
+        lambda icon_name: send_from_directory(
+            f"{app.static_folder}/icons",
+            f"{icon_name}.svg",
```
bartaz-canonical: Just realised this may conflict with the badges that we serve from s.io/snap-name/badge.svg

Image

bartaz-canonical: As for now, there is no "badge" icon, but we are kind of sitting edge case here, where is someone registers a snap named "icons", and there is a "badge" icon, they will have a name conflict.

Not sure if there is a way around it other than restricting the usage of "icons" snap name and/or "badge" icon name…

canonical-steverydz: Good point. This is something we will definitely need to be careful of!

scripts/link-ds-icons.mjs
canonical-steverydz: Good idea. I've updated to reflect this

canonical-steverydz:
@copilot resolve the merge conflicts in this pull request

rockcraft.yaml
```diff
@@ -26,18 +26,24 @@ parts:
       yarn install --immutable
       # build the UI
       yarn run build
+      # The final image excludes node_modules, so copy icon files instead of the dev symlink.
```
edisile, line 29: priming from the `build-ui` step would be simpler and avoids those annoying permission issues (tbf I should have done this from the start) 
```diff
diff --git a/rockcraft.yaml b/rockcraft.yaml
index a62b53319e..66fde337c6 100644
--- a/rockcraft.yaml
+++ b/rockcraft.yaml
@@ -31,19 +31,17 @@ parts:
       cp -r node_modules/@canonical/ds-assets/icons static/icons
       mkdir -p "$CRAFT_PART_INSTALL/flask/app"
       cp -r static "$CRAFT_PART_INSTALL/flask/app/"
+    prime:
+      - flask/app/static
   flask-framework/install-app:
     after:
       - build-ui
-    stage:
-      - -flask/app/static/icons
     prime:
       - flask/app/.env
       - flask/app/app.py
       - flask/app/cache
       - flask/app/webapp
       - flask/app/templates
-      - flask/app/static
-      - -flask/app/static/icons
       - flask/app/deleted.yaml
       - flask/app/redirects.yaml
       - flask/app/security.md

```

canonical-steverydz: Thanks - I've made that change

edisile (approved):
deployed this branch to staging and it looks good.
the only obvious changes seem to be:
- a slight baseline shift and additional `margin-bottom` on all headings
- a change in the `h5` style to use the `small-caps` variant of the font

I think neither of these is significant enough to worry about

PR #5806, opened 2026-07-23
Open Desktop button fixes

ilayda-cp:
## Done
Adds a "View in Desktop store" button on the snap details / channel map that launches the snap in the user's desktop store via a snap:// handoff.

- Platform-gated: the button is hidden by default and only revealed on platforms that can register a snap:// handler (Linux desktop; excludes Android/ChromeOS), so it never shows where it would silently fail.
- Feedback states: shows a loading spinner while launching, and a caution notification with the CLI install fallback if no handler responds within 1.5s.
- Launch detection: treats the window losing focus/visibility as a successful handoff, and the timeout as "no handler registered".
- Shared button markup extracted into a Jinja macro; adds analytics for impressions and launch outcomes; unit tests for platform detection, reveal, and launch flow.

Scenarios

  - Non-Linux (macOS/Windows) or Android/ChromeOS -> button stays hidden, nothing tracked.
  - Linux desktop, page load -> button revealed, *_shown impression fired.
  - Click with snapd installed-> spinner, window blurs/tab hides, returns to idle, *_success fired.
  - Click with no snap:// handler -> spinner, 1.5s timeout, caution notification + CLI fallback shown, *_error fired.
  - Click on a specific channel (channel map) -> same flow with the channel's snap://…?channel= and details_channel_* events.
  - Re-click after an error → previous error cleared, launch retried.
  - Rapid double-click / click while loading -> ignored, single launch and single event.
  - Channel install screen re-rendered -> newly injected block re-checked and revealed on supported platforms.

## How to QA
- go to https://snapcraft-io-5806.demos.haus/microcloud
- Click on the install button
    - if your computer is Linux verify the "view in desktop store" is visible
    - if you are using non-Luinux verify the 'view in desktop store" is hidden
- click  "view in desktop store" if visible
    - if your Os is Ubuntu it should open the store
    - if not an error message should be shown 
- Click on the channel button near the install button
- hover on a channel
- click on the install button:
    - if your computer is Linux verify the "view in desktop store" is visible
    - if you are using non-Luinux verify the 'view in desktop store" is hidden
- click  "view in desktop store" if visible
    - if your Os is Ubuntu it should open the store
    - if not an error message should be shown 

## Testing

- [x] This PR has tests

## Security

  - [ ] Examples:
  - [ ] ...

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-36645

### DIFF
--- a/static/js/public/snap-details/__tests__/openDesktop.test.ts
+++ b/static/js/public/snap-details/__tests__/openDesktop.test.ts
@@ -73,9 +73,10 @@ describe("isDesktopStoreSupported", () => {
       setUserAgent(MAC_UA);
       expect(isDesktopStoreSupported()).toBe(true);
     } finally {
-      delete (navigator as any).userAgentData;
+      delete (navigator as { userAgentData?: unknown }).userAgentData;
     }
   });
+});
 
 describe("applyDesktopStoreSupport", () => {
   beforeEach(() => {
--- a/static/js/public/snap-details/__tests__/openDesktop.test.ts
+++ b/static/js/public/snap-details/__tests__/openDesktop.test.ts
@@ -1,0 +1,213 @@
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import "@testing-library/jest-dom";
+
+import openDesktop, {
+  applyDesktopStoreSupport,
+  isDesktopStoreSupported,
+} from "../openDesktop";
+
+const { trackEvent } = vi.hoisted(() => ({ trackEvent: vi.fn() }));
+
+vi.mock("@canonical/analytics-events", () => ({ trackEvent }));
+
+const LINUX_UA =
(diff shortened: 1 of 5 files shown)


abbiesims:
When I click on the "View in Desktop store" for any revision from the channel dropdown, it opens with an error in the app centre:

image

However, the default release one works, but only from the main green "Install" button in the header

Edit: I've just realised this also seems to be a bug on prod so perhaps is out of scope for this

canonical-steverydz (commented):
I can verify that the button doesn't show up on macOS

ilayda-cp:
@abbiesims does this happen on the production as well? or did i break something

abbiesims:
> @abbiesims does this happen on the production as well? or did i break something

ah, just edited my previous comment - it's a bug on prod! I'll make a maintenance ticket but it's out of scope for this task :)

abbiesims (approved):
LGTM! 👍

PR #5850, opened 2026-08-25
Add Pragma buttons to publisher save and preview bar

canonical-steverydz:
## Done
Adds [Pragma buttons](https://ds-react-global.canonical.com/?path=/docs/components-button--docs) to the "Save and preview" section of the publisher pages.

## How to QA
- Go to a snaps listing page, e.g. https://snapcraft-io-5850.demos.haus/steve-test-snap/listing
- Check that the "Preview" button has a tooltip on hover
- Check that the "Preview" button works as expected
- Check that the "Revert" button works as expected
- Check that the "Save" button works as expected
- Go to a snaps settings page, e.g. https://snapcraft-io-5850.demos.haus/steve-test-snap/settings
- Check that the "Revert" button works as expected
- Check that the "Save" button works as expected

Some things to note:
- The "Save and preview" bar is not sticky which is [an existing issue not introduced by this PR](https://warthogs.atlassian.net/browse/WD-38482)
- The `js-sticky-bar` class was removed from the sticky bar as it isn't actually used
- Using the [withTooltip HOC](https://ds-react-global.canonical.com/?path=/docs/components-tooltip-withtooltip--docs) showed an error in the console caused by two Reacts loaded on the page, fixed by the changes to `package.json` and `vite.config.js`
- `/tests/setup.ts` is needed to mock `ResizeObserver` in `jsdom` as it is used by the Pragma Tooltip component

## Testing

- [x] This PR has tests

## Security

  - [ ] Examples:
  - [ ] ...

## Issue / Card

Fixes #

## Screenshots
n/a

## UX Approval

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,10 +27,10 @@
     "@babel/preset-typescript": "7.28.5",
     "@canonical/analytics-events": "1.1.2",
     "@canonical/cookie-policy": "3.10.0",
-    "@canonical/ds-assets": "0.32.0",
+    "@canonical/ds-assets": "0.34.0",
     "@canonical/global-nav": "3.8.0",
     "@canonical/react-components": "4.5.2",
-    "@canonical/react-ds-global": "0.32.0",
+    "@canonical/react-ds-global": "0.34.0",
     "@canonical/store-components": "0.55.0",
     "@canonical/styles": "0.32.0",
     "@dnd-kit/core": "6.3.1",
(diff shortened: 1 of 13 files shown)


edisile (commented):
lgtm for the most part, code is good. only issue I noticed is the low contrast on the disabled buttons. the Vanilla styles add an `opacity: .33;` for `button:disabled` on top of Pragma's styles.
current:
image
correct:
image

not sure what the best approach would be here... an override like `.ds.button:disabled {opacity: 1}`? something more/less specific? if we add this kind of compatibility patches we should probably make a new stylesheet for this kind of ad-hoc rules and compare with what other teams found.

another unrelated thing I noticed (but is very out of scope for this PR): Vanilla and Pragma select light/dark theme in different ways, which can lead to "interesting" results when the theme selected for snapcraft doesn't match the `prefers-color-scheme` value

Vanilla dark theme, but `prefers-color-scheme: light`:
Image

viceversa:
Image

edisile (approved):

PR #5857, opened 2026-08-31
Update robots.txt to improve crawl efficiency

ilayda-cp:
## Done
robots.txt has been empty since the flask-base migration (#2565, 2020) which replaced a hardcoded empty response with an empty file. Everything was crawlable including paths require login.
Added rules that matches with [ubuntu.com](https://ubuntu.com/robots.txt) and  [canonical.com](https://canonical.com/robots.txt):
- Block account, admin, login, register-snap and validation-sets
- Block the per-snap publisher dashboard (/<snap>/listing, /metrics, /releases etc) which redirects to SSO and sits under the public snap namespace
- Block /api/, .json endpoints, search and contact forms
- Declare 4 sitemaps

There are rules with $ or a trailing slash because snap pages live at the root. Just adding  Disallow: /account would also block snaps whose name starts with "account". /about/listing, /about/release and /about/publicise are reallowed explicitly since they collide with the dashboard patterns.

## How to QA
n/a

## Testing

- [x] This PR has tests
- [] No testing required (explain why):

## Security

  - [ ] Examples:
  - [ ] ...

## Issue / Card

Fixes #

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -35,6 +35,7 @@ Disallow: /login/
 Disallow: /login-beta$
 Disallow: /logout$
 Disallow: /github/auth
+Disallow: /publisher/github/
 Disallow: /snaps$
 Disallow: /snaps/
 Disallow: /register-snap$
@@ -111,6 +112,7 @@ Disallow: /login/
 Disallow: /login-beta$
 Disallow: /logout$
 Disallow: /github/auth
+Disallow: /publisher/github/
 Disallow: /snaps$
 Disallow: /snaps/
 Disallow: /register-snap$
--- a/robots.txt
+++ b/robots.txt
@@ -1,9 +1,8 @@
 # ===========================================
 # snapcraft.io robots.txt
 # Strategy:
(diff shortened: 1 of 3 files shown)


robots.txt
```diff
@@ -0,0 +1,171 @@
...
+Disallow: /search$
+Disallow: /search?
+Disallow: /docs/search
+Disallow: /about/contact-us$
+Disallow: /about/thank-you$
+Disallow: /api/
+Disallow: /*.json$
+Disallow: /*/embedded
+Disallow: /*.svg$
+Disallow: /install/
```
edisile: why block /install/* pages? they're a good resource (e.g. users might search "how to install < snap > on < distro >"), I think we should allow them

edisile (approved):
lgtm, just a small suggestion

robots.txt
ilayda-cp: done! also removed the agent specififc configuration as discussed

PR #5861, opened 2026-09-02
add llms.txt

ilayda-cp:
## Done
This PR adds an llms.txt file to the snapcraft.io repository, a machine-readable document that describes the site's purpose, structure, and navigation links in a format optimized for LLM consumption

llms.txt will be created during deployment so that it is served static (faster) and automatic (no maintenance required when a new public page is added)

## How to QA
- go to https://snapcraft-io-5861.demos.haus/llms.txt
- verify if opens md page (most of the links are md versions of pages and directed to snapcraft.io so some might not work)
- Go to some of the md pages verify they work:
     - https://snapcraft-io-5861.demos.haus/about/publish.md
     - https://snapcraft-io-5861.demos.haus/about/listing.md
     - https://snapcraft-io-5861.demos.haus/iot.md
     - https://snapcraft-io-5861.demos.haus/lxd.md
Note: this is valid for public pages only. documentation pages doesnt have md files yet as we dont handle them. 

## Testing

- [x] This PR has tests

## Security

  - [ ] Examples:
  - [ ] ...

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-38673


webapp/site_pages.py
```diff
@@ -0,0 +1,245 @@
...
+                    "exhaustive crawling rather than reading."
+                ),
+                "sitemap": False,
+            },
+        ],
+    },
+]
+
+# Public pages that are not added to llms.txt.
+# Anything new must be listed above or excluded here on purpose.
```
bartaz-canonical: Why "anything new must be listed above or excluded"? What happens if we add a new route but don't list it here?

Having such list may be tricky to maintain (remember to add routes here).

What's the use of `EXCLUDED_PAGES`? If we have a list of `PAGES` do we need to specifically exclude others?

webapp/markdown_suffix.py
```diff
@@ -0,0 +1,63 @@
+"""
+Serve /page.md as the Markdown rendering of /page.
+
+canonicalwebteam.markdown-response is worked with a query parameter but
```
bartaz-canonical, line 4: Would it be possible to add .md support to the markdown-response library itself?

webapp/markdown_suffix.py
```diff
@@ -0,0 +1,63 @@
...
+def cache_markdown(response):
+    if response.mimetype == "text/markdown":
+        response.headers["Cache-Control"] = CACHE_CONTROL
+
+    return response
+
+
+def add_suffix(path):
+    if path.endswith("/"):
+        path += "index"
```
bartaz-canonical: Do we need support for /index.md? If we have llms.txt for home page equivalent, we probably don't need this.

Also, looking again and the proposed standard instead of `path/index.md` they use `path/llms.txt`. So if anything, maybe we could support that as alrearnative?

webapp/site_pages.py
ilayda-cp: PAGES is a subset of pagers that we show in the llms.txt. i added here because i want those subset of all routes are sync with sitemap. the comment is misleading. in that case  exclusion list is knida useless. my inital idea was adding everything to llms.txt but that would be not useful so i added exclusion list. But i'll check a way to collect ALLL routes except snap specific and login pages

webapp/markdown_suffix.py
ilayda-cp: yes. i can add it once this llms.txt thing is finished.

webapp/markdown_suffix.py
ilayda-cp: that would be nice. but it think thats ok for now. because this will require creating llms.txt for each index file. BUTTTT we dont have it so i will remove this as well

bartaz-canonical: It's not really about adding new llms.txt for other routes. It's more to serve the existing md versions from there.

So: snapcraft.io/spotify.md and snapcraft.io/spotify/llms.txt would serve the same md version of the details page.

But it's just a nice to have suggestion, we don't need it now.

ilayda-cp: hmm yes it is doable but why do we want to do that?

templates/store/snap-details/_channel_map.html
```diff
@@ -131,7 +131,7 @@
           </nav>
         </div>
         <div class="u-fixed-width p-channel-map__version-table" role="tabpanel" id="release-versions" aria-hidden="false">
-          <table aria-label="Release versions">
+          <table aria-label="Release versions" data-md-strip>
```
bartaz-canonical, line 134: What is `data-md-strip` doing? Hiding it from md version?

ilayda-cp: yes its because channel map is filled with js after the rendering is completed so its always empty.

ilayda-cp: also demo is working you can test it from there

templates/_base-layout.html
```diff
@@ -50,6 +50,7 @@
 
     {% block extra_meta %}{% endblock %}
     <link rel="canonical" href="https://snapcraft.io{{ request.path }}" />
+    <link rel="alternate" type="text/markdown" href="https://snapcraft.io{{ request.path }}.md" />
```
bartaz-canonical: This is not correct, at least not always.

On home page I guess it should link to llms.txt (?)

Image

Also authenticated pages show it:

Image

Authenticated views seem to also kinda work, which shouldn't be the case I think. We don't want to give LLMs access there or pretend we do.

Image

I guess we would have to be much more specific about where we actually show this link alternate link, and which routes support the md format.

rockcraft.yaml
```diff
@@ -17,15 +17,22 @@ parts:
...
     override-build: |
       set -eux
       # install dependencies
       npm install -g yarn
       yarn install --immutable
       # build the UI
       yarn run build
+      # write static/llms.txt from the staged Python dependencies
+      export PYTHONPATH="$(ls -d "$CRAFT_STAGE"/lib/python3*/site-packages)"
+      yarn run build-llms-txt
```
bartaz-canonical, line 35: Will the pre-built llms.txt be part of the rock? I'm not sure if I see it being explicitly mentioned to be included?

Dockerfile
```diff
@@ -51,6 +51,9 @@ RUN rm -rf package.json yarn.lock .babelrc requirements.txt
 COPY --from=build /srv/static/js static/js
 COPY --from=build /srv/static/icons static/icons
 
+# Write static/llms.txt from the pages found in the routing table
+RUN python3 scripts/generate_llms_txt.py
```
bartaz-canonical, line 56: Doesn't matter too much maybe, but should it be ran earlier (in one of previous layers?). So that this one remains to be purely prod image without any unnecessary build elements?

templates/_base-layout.html
ilayda-cp: great catch yes it should be llms.txt. for authenticated views we might have md versions of it as its create alllllll pages md versions but i think we can block it. for all public pages we will have md versions i will prevent private ones to have md versions. for authenticated version i will hide the alternate one and fix the pubic alternate url

rockcraft.yaml
ilayda-cp: yes it will be in the rock. The rock build copies the entire static folder as a whole.

Dockerfile
ilayda-cp: its just for staging and local. also the script requires web app's code and python libs so its the most appropriate place. One thing that worries me is becausde we are adding categories to llms.txt it requires a store API call. so if it fails the build will be blocked. so im thinking about having a static list of categories.

bartaz-canonical: Ah, yes. If we do it on the build step, we should not rely on API calls. I think it's OK to make the list static. Categories change very rarely.

templates/_base-layout.html
ilayda-cp: @bartaz-canonical about the the md version of the home page: I checked out the proposal https://llmstxt.org/ they say "We furthermore propose that pages with information that agents might need provide a clean markdown version of those pages at the same URL as the original page, either with .md appended (page.html.md) or with the extension replaced by .md (page.md). (URLs without file names should append index.html.md or index.md instead.)" and 
""rel="alternate" type="text/markdown" points to the markdown version of a page, and rel="describedby" points  to the llms.txt file"'

so index.md would be a better approach.

canonical-steverydz (approved):
LGTM 👍

bartaz-canonical (approved):

PR #5889, opened 2026-09-21
feat: Add vanilla styles adaptor

canonical-steverydz:
## Done
Implemented the [Pragma Vanilla adapter](https://github.com/canonical/pragma/tree/main/packages/styles/vanilla-adapter).

Things of note:
- Fonts need to be symlinked in the same way that icons are to comply with the way they are loaded by the framework
- The `style.scss` changes look like a lot however they are for the most part just moving the existing styles into a layer
- The Pragma buttons don't have even vertical spacing in the app layout - this is expected, and in a subsequent release will be evened out
- The forms aren't included in this PR as they will be addressed in another task
- There was also an issue with the settings icon in the side navigation in the application layout, where it was conflicting with custom styles from a legacy icon implementation

## How to QA
Theres no functional changes. There will be some minor visual changes due to the new design systems spacing etc. The page most effected would be [the listing page](https://snapcraft-io-5889.demos.haus/steve-test-snap/listing) as it has the most buttons and icons so check things look ok there, and the tour also.

## Testing

- [ ] This PR has tests
- [x] No testing required (explain why): Only replacing frontend components

## Security

  - [ ] Examples:
  - [ ] ...
- [x] This PR has no security considerations (explain why): Only replacing frontend components

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-38940

## Screenshots
n/a

## UX Approval


static/sass/styles.scss
```diff
@@ -1,586 +1,608 @@
 @use "sass:map";
+@use "@canonical/styles-vanilla-adapter/layers.css";
```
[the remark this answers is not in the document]
canonical-steverydz: @copilot This is incorrect

templates/admin/admin.html
```diff
@@ -1,6 +1,8 @@
 {% set page_slug = "admin" %}
 {% set show_header = False %}
 {% extends "_base-layout.html" %}
+{% block html_class %}app comfortable light{% endblock %}
```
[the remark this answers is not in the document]
canonical-steverydz: @copilot This is incorrect

templates/store/publisher.html
```diff
@@ -1,5 +1,7 @@
 {% set show_header = False %}
 {% extends "_base-layout.html" %}
+{% block html_class %}app comfortable light{% endblock %}
```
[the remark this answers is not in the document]
canonical-steverydz: @copilot This is incorrect

static/js/publisher/components/Tour/TourStep.tsx
```diff
@@ -18,18 +15,14 @@ function TourStep(props: TooltipRenderProps): React.JSX.Element {
   } = props;
 
   return (
-    <Card style={{ border: "none" }}>
+    <Card style={{ border: "none", backgroundColor: "#fff" }}>
```
edisile: tour steps break in dark mode:

Image

something like `var(--color-background-layer3)` would probably be a better choice (not sure if it's the correct variable, but considering it's an element that sits on top of the page it makes sense to me)

edisile: currently this is also an issue in production, so it's not this additional `backgroundColor` attribute that broke it

static/js/publisher/pages/Model/PoliciesTable.tsx
```diff
@@ -200,20 +197,18 @@ function PoliciesTable({
               </Button>
               <Button
                 className="u-no-margin--bottom u-no-margin--right"
-                appearance="positive"
+                importance="primary"
+                anticipation="constructive"
```
edisile: considering we're deleting, shouldn't this be `destructive`?

edisile (changes requested):

static/js/publisher/pages/Model/PoliciesTable.tsx
canonical-steverydz: @edisile Good catch. I think I'd reached the point of button blindness!

canonical-steverydz:
@edisile I've address the changes

edisile (approved):
@canonical-steverydz thanks! lgtm

